PR #4, opened 2022-01-24
feat(contracts): Add initial LP functionality

chrismaree:


nicholaspai (approved):

package.json
```diff
@@ -8,7 +8,11 @@
     "type": "git",
     "url": "git+https://github.com/across-protocol/across-smart-contracts-v2.git"
   },
-  "dependencies": {},
+  "dependencies": {
+    "@openzeppelin/contracts": "^4.4.2",
```
nicholaspai, line 12: why shouldn't these be dev dependencies?

nicholaspai (approved):

package.json
chrismaree: I made them normal dependencies because that's what we do in the V1 repo: https://github.com/across-protocol/across-smart-contracts/blob/master/package.json

PR #5, opened 2022-01-24
feat: Implement deposit functionality

nicholaspai:
Signed-off-by: Nick Pai <npai.nyc@gmail.com>


contracts/SpokePool.sol
```diff
@@ -1,16 +1,169 @@
...
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import "@uma/core/contracts/common/implementation/Testable.sol";
+import "@uma/core/contracts/common/implementation/Lockable.sol";
+import "@uma/core/contracts/common/implementation/MultiCaller.sol";
 
+/**
+ * @title SpokePool
+ * @notice Contract deployed on source and destination chains enabling depositors to transfer assets from source to 
+ * destination. Deposit orders are fulfilled by off-chain relayers who also interact with this contract. Deposited
+ * tokens are burned on the source chain and relayers send the recipient the desired token currency and amount
```
chrismaree, line 14: nothing is burnt. they are locked and later sent over the canonical bridge

contracts/SpokePool.sol
```diff
@@ -1,16 +1,169 @@
...
+ * on the destination chain.
+ */
+contract SpokePool is Testable, Lockable, MultiCaller {
+    using SafeERC20 for IERC20;
+
+    // Chain ID for this contract.
+    uint256 public chainId;
```
chrismaree, line 21: do we need a state variable for this? could we have a view function that returns `chainid()` (a built in solidity function that returns the evm chainID where the function is called)

contracts/SpokePool.sol
```diff
@@ -1,16 +1,169 @@
...
+
+    // Timestamp when contract was constructed. Relays cannot have a quote time before this.
+    uint64 public deploymentTime;
```
chrismaree, line 24: very good

contracts/SpokePool.sol
```diff
@@ -1,16 +1,169 @@
...
+
+    // Track the total number of deposits. Used as a unique identifier for relays.
```
chrismaree, line 26:
```suggestion
    // Track the total number of deposits. Used as a unique identifier for deposits.
```

contracts/SpokePool.sol
```diff
@@ -1,16 +1,169 @@
...
+    uint256 public numberOfDeposits;
+
+    // // Address of WETH on L1. If the deposited token maps to this L1 token then wrap ETH to WETH on the users behalf.
+    // address public l1Weth;
+
+    struct DestinationToken {
+        address token;
```
chrismaree, line 33: should this be `destinationToken`?

contracts/SpokePool.sol
nicholaspai: Amended language

contracts/SpokePool.sol
nicholaspai: Sure

contracts/SpokePool.sol
```diff
@@ -1,16 +1,169 @@
...
+    // Chain ID for this contract.
+    uint256 public chainId;
+
+    // Timestamp when contract was constructed. Relays cannot have a quote time before this.
+    uint64 public deploymentTime;
+
+    // Track the total number of deposits. Used as a unique identifier for relays.
+    uint256 public numberOfDeposits;
+
+    // // Address of WETH on L1. If the deposited token maps to this L1 token then wrap ETH to WETH on the users behalf.
```
chrismaree, line 29: would it be better to bake this information into `DestinationToken` with a bool? that way we dont need to store `l1WethAddress`

nicholaspai: Good call

contracts/SpokePool.sol
```diff
@@ -1,16 +1,169 @@
...
+    // address public l1Weth;
+
+    struct DestinationToken {
+        address token;
+        address spokePool;
+        uint256 chainId;
```
chrismaree, line 32: having this in here has a big implication: each `SpokePool` needs to know about the waitlists of all other tokens, on all other chains. for example boba needs to know for each origin token what destination is whitelisted

chrismaree: would the right data structure for `whitelistedDestinationTokens` not be:
```
mapping(address => uint256 => DestinationToken)
```

so the extra step is for the target chainID?

contracts/SpokePool.sol
```diff
@@ -1,16 +1,164 @@
...
+        _;
+    }
+
+    /**************************************
+     *          ADMIN FUNCTIONS           *
+     **************************************/
+
+    /**
+     * @notice Whitelist an origin token <-> destination oken pair for bridging.
+     */
```
chrismaree, line 79: I think we should ignore the NatSpec for now OR do it correctly. half NatSpec is almost worse than nothing. what do you think?

contracts/SpokePool.sol
```diff
@@ -1,16 +1,164 @@
...
+    function _whitelistToken(
```
chrismaree, line 80: as I mentioned above I dont think we need to just whitelist a token, we actually need to whitelist a _path_ a token can traverse. i.e Arbitrum->Optimism, with both the originToken and destinationToken for that given _path_

contracts/SpokePool.sol
nicholaspai: the name Destination is in the struct so no need to repeat

contracts/SpokePool.sol
nicholaspai: I mildly disagree i find it useful to know what the methods are meant to do high level so i'm just orienting myself

contracts/SpokePool.sol
```diff
@@ -1,16 +1,161 @@
...
+        uint256 destinationChainId,
         uint256 amount,
         address recipient,
-        uint256 destinationChainId,
-        uint256 relayerFee
-    ) public {}
+        uint64 relayerFeePct,
+        uint64 quoteTimestamp
+    ) public onlyIfDepositsEnabled(originToken, destinationChainId) {
+        require(isWhitelistedRoute(originToken, destinationChainId), "deposit token not whitelisted");
```
chrismaree, line 118: can this be rolled into the modifier above? or should it be it's own modifier?

contracts/SpokePool.sol
```diff
@@ -1,16 +1,161 @@
...
+    mapping(address => mapping(uint256 => DestinationToken)) public whitelistedDestinationRoutes;
+
+
+    /****************************************
+     *                EVENTS                *
+     ****************************************/
+    event WhitelistRoute(address originToken, uint256 destinationChainId, address destinationToken, address spokePool, address wethContract);
+    event DepositsEnabled(address originToken, uint256 destinationChainId, bool depositsEnabled);
+    event FundsDeposited(
+        uint256 nonce,
```
chrismaree, line 43: should we call this something different? should it be `depositId`?

contracts/SpokePool.sol
nicholaspai: On second thought, these kind of check the same thing. I'm going to remove  the method `isWhitelistedRoute`, which i copied from Across V1.

contracts/SpokePool.sol
nicholaspai: what do you think? I'm good with either.

nicholaspai:
@chrismaree let's reach consensus on contract implementation and then I'll add tests

contracts/SpokePool.sol
```diff
@@ -1,16 +1,164 @@
...
+        address originToken,
+        uint256 destinationChainId,
+        address recipient,
+        address sender,
+        address destinationToken,
+        uint256 amount,
+        uint64 relayerFeePct,
+        uint64 quoteTimestamp
+    );
```
chrismaree, line 56: should we try pack these types closer? put all the addresses,uints256 and uint64 close to each other?

chrismaree, line 56: should this include `originChainId`?

nicholaspai: done

chrismaree: mainly just useful in the context of the event. does not really add much cost anywhere.

contracts/SpokePool.sol
```diff
@@ -1,16 +1,180 @@
...
+
+    // Timestamp when contract was constructed. Relays cannot have a quote time before this.
+    uint64 public deploymentTime;
+
+    // Track the total number of deposits. Used as a unique identifier for deposits.
+    uint256 public numberOfDeposits;
+
+    struct DestinationToken {
+        address token;
+        address spokePool;
```
chrismaree, line 32: why do we need `spokePool`?

contracts/SpokePool.sol
```diff
@@ -1,16 +1,180 @@
...
+    function _whitelistRoute(
+        address originToken,
+        address destinationToken,
+        address spokePool,
+        bool isWethToken,
+        uint256 destinationChainId
+    ) internal {
+        // We use a correctly set spoke pool address as the determinant of whether a token path is whitelisted or not.
+        // So, explicitly prevent the caller from unintentionally setting the spoke pool to the zero address.
+        require(spokePool != address(0), "Invalid spoke pool");
```
chrismaree, line 96: I dont think we need this prop 🤔

contracts/SpokePool.sol
nicholaspai: after talking offline we agree this can be removed. I incuded bc i thought relayer might need it but relayer can assume that latest spoke pool is the one to use

contracts/SpokePool.sol
nicholaspai: removed

contracts/SpokePool.sol
```diff
@@ -1,16 +1,176 @@
...
+        // Note We assume that L2 timing cannot be compared accurately and consistently to L1 timing. Therefore, 
+        // `block.timestamp` is different from the L1 EVM's. Therefore, the quoteTimestamp must be within 10
+        // mins of the current time to allow for this variance.
+        // Note also that `quoteTimestamp` cannot be less than 10 minutes otherwise the following arithmetic can result
+        // in underflow. This isn't a problem as the deposit will revert, but the error might be unexpected for clients.
+        // Consider requiring `quoteTimestamp >= 10 minutes`.
+        require(
+            getCurrentTime() >= quoteTimestamp - 10 minutes && 
+            getCurrentTime() <= quoteTimestamp + 10 minutes &&
+            quoteTimestamp >= deploymentTime,
```
mrice32, line 143: Do we really need to check this? Wouldn't this check only be possible to violate in the first 10 minutes of the SpokePool's existence? Otherwise, you would always hit `getCurrentTime() <= quoteTimestamp + 10 minutes`, right? Also, what would happen if the quote timestamp were a few minutes before this contract is deployed (i.e. what is this trying to prevent?)? That wouldn't prevent the issue of the quote timestamp being before the HubPool whitelists a particular token, right?

contracts/SpokePool.sol
```diff
@@ -1,16 +1,176 @@
...
+            "invalid quote time"
+        );
+        DestinationToken memory tokenPath = whitelistedDestinationRoutes[originToken][destinationChainId];
+        // If the address of the destination token is a WETH contract and there is a msg.value with the transaction 
+        // then the user is sending ETH. In this case, the ETH should be deposited to WETH.
+        if (tokenPath.isWethToken && msg.value > 0) {
```
mrice32, line 149: As mentioned above, I think it's a bit odd that we're using the token path to determine whether the _deposited token_ is WETH, since that seems to be something that only depends on the origin chain.

contracts/SpokePool.sol
```diff
@@ -38,4 +198,12 @@ contract SpokePool {
         uint256[] memory relayerRefundAmounts,
         bytes32[] memory inclusionProof
     ) public {}
+
+    /**************************************
+     *           VIEW FUNCTIONS           *
+     **************************************/
+
+    function chainId() public view returns (uint256) {
+        return block.chainid;
```
mrice32, line 207: +1. Hopefully this is cheaper than reading a variable from storage (seems likely!),

mrice32 (commented):
Looks great! A few comments

contracts/SpokePool.sol
chrismaree: ye I think there is a good meta point here: where posible (and even when not 100% posible) we should try and stick with the same interface/variable name as in v1 to simplify bot/front end compatibility.

contracts/SpokePool.sol
```diff
@@ -1,16 +1,176 @@
...
+        uint256 nonce,
+        uint256 destinationChainId,
+        uint256 originChainId,
+        uint256 amount,
+        uint64 relayerFeePct,
+        uint64 quoteTimestamp,
+        address originToken,
+        address recipient,
+        address sender,
+        address destinationToken
```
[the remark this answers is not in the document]
chrismaree: ye, this is a good point. we dont actually need to know the destination token address. really, we just need to know that the `originToken` is whitelisted to send to a particular destination chainID. 

At the most extreme we simply have `whitelistedDestinationRoutes` being :
```
// mapping of origin token to destination chainID to bool signalling if it enabled (or not).
mapping(address => mapping(uint256 => bool))
```

we then store the weth address that the spokePool is deployed on in a separate variable and using this we check for any weth related logic.

thoughts?

contracts/SpokePool.sol
chrismaree: ye, that's true. not needed. let's keep it lean.

contracts/SpokePool.sol
```diff
@@ -1,16 +1,176 @@
...
+        address destinationToken,
+        bool isWethToken,
+        uint256 destinationChainId
+    ) internal {
+        // We use a correctly set destination token address as the determinant of whether a token path is whitelisted 
+        // or not. So, explicitly prevent the caller from unintentionally setting this to the zero address.
+        require(destinationToken != address(0), "Invalid spoke pool");
+        whitelistedDestinationRoutes[originToken][destinationChainId] = DestinationToken({
+            token: destinationToken,
+            isWethToken: isWethToken,
```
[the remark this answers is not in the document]
chrismaree: ye, this links into [this](https://github.com/across-protocol/across-smart-contracts-v2/pull/5/files#r792209503) recommendation.

contracts/SpokePool.sol
chrismaree: good point! would save a small bit of gas.

test/HubPool.LiquidityProvision.ts
```diff
@@ -31,7 +31,7 @@ describe("HubPool Liquidity Provision", function () {
     await dai.connect(liquidityProvider).approve(hubPool.address, amountToLp);
     await hubPool.connect(liquidityProvider).addLiquidity(dai.address, amountToLp);
 
-    //The balance of the collateral should be equal to the original amount minus the LPed amount. The balance of LP
```
chrismaree, line 34: <3

test/HubPool.TokenWhitelisting.ts
```diff
@@ -6,32 +6,32 @@ import { getContractFactory, SignerWithAddress } from "./utils";
...
 
 describe("HubPool Token Whitelisting", function () {
   before(async function () {
-    [owner, liquidityProvider, other] = await ethers.getSigners();
+    [owner, other] = await ethers.getSigners();
     ({ weth, hubPool } = await deployHubPoolTestHelperContracts(owner));
   });
 
   it("Can add L1 token to whitelisted lpTokens mapping", async function () {
-    await expect((await hubPool.callStatic.lpTokens(weth.address)).lpToken).to.equal(ZERO_ADDRESS);
```
chrismaree, line 18: thanks for cleaning these for me

test/SpokePool.Admin.ts
```diff
@@ -0,0 +1,80 @@
+import { expect } from "chai";
+import { Contract } from "ethers";
+import { ethers } from "hardhat";
+import { SignerWithAddress } from "./utils";
+import { ZERO_ADDRESS } from "@uma/common";
+import { deploySpokePoolTestHelperContracts } from "./SpokePool.Fixture";
+import { depositDestinationChainId } from "./constants";
+
+let spokePool: Contract, erc20: Contract;
+let admin: SignerWithAddress;
```
chrismaree, line 10: nit: in the other tests we call this the `owner`. we should be consistant. either we should change this files test or the other one to reflect `admin` over `owner.

test/SpokePool.Admin.ts
```diff
@@ -0,0 +1,80 @@
...
+        .whitelistRoute(
+          erc20.address,
+          ZERO_ADDRESS,
+          true,
+          depositDestinationChainId
+        )).to.be.reverted;
+
+    await expect(spokePool
+      .connect(admin)
+      .whitelistRoute(
```
chrismaree, line 30: is this listed? I feel that this can be done on fewer lines, no? do you have the prettier vs code plug in installed: https://github.com/prettier/prettier-vscode?

this will let you lint without needing to run prettier, using the config that I set when creating the repo.

test/SpokePool.Admin.ts
```diff
@@ -0,0 +1,80 @@
...
+        depositDestinationChainId
+      )).to.emit(spokePool, "WhitelistRoute").withArgs(
+        erc20.address,
+        depositDestinationChainId,
+        erc20.address,
+        true
+      );
+
+    // Whitelisted path should be saved in contract.
+    const whitelistedDestinationRoutes = await spokePool.whitelistedDestinationRoutes(erc20.address, depositDestinationChainId)
```
chrismaree, line 43: this looks too long. please try the linter

test/SpokePool.Fixture.ts
```diff
@@ -0,0 +1,40 @@
...
+  return { timer, weth, erc20, destWeth, destErc20, spokePool };
+}
+
+export interface DepositRoute {
+    originToken: string;
+    destinationToken: string;
+    isWethToken: boolean;
+    destinationChainId?: number;
+}
+export async function whitelistRoutes(spokePool: Contract, routes: DepositRoute[]) {
```
chrismaree, line 30: nice structure.

contracts/SpokePool.sol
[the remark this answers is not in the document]
nicholaspai: We'd need to map chain Id to weth address right? I think this is possible not to include destinationToken. 

For a given deposit with origin token and destination chain ID, how would the relayer know which currency to relay in?

contracts/SpokePool.sol
nicholaspai: Sure I can remove this check. 

Do you agree though that we should keep this deployment time stamp somewhere in the contract? We probably will need it in initiateRelay to block any deposit quote times before the deployment time

contracts/SpokePool.sol
nicholaspai: cool will change

contracts/SpokePool.sol
nicholaspai: removed

contracts/SpokePool.sol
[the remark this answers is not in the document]
nicholaspai: After talking offline i've decided to change the mapping to `originTokenAddress => destChainId => destChainToken`. 
Also, `wethAddress` is stored in contract state

contracts/SpokePool.sol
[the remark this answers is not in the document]
nicholaspai: agreed

test/SpokePool.Admin.ts
nicholaspai: ok i'm good with owner

test/SpokePool.Admin.ts
nicholaspai: I think the main issue is that I was erroneously relying on CI to lint like it does in `protocol` repo. So I will manually lint and commit

nicholaspai:
@chrismaree @mrice32 I responded to your comments. Main changes were:
- refactoring `whitelistedDestinationPaths` struct (is there a better name for this)?
- adding a `lint-fix` script and running it

contracts/SpokePool.sol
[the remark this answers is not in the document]
mrice32: So I still think this is a bit confusing. Is there no central source of truth for mainnet token address 0x1234 <=> optimism token address 0x5678?

If you consider the deposits _alone_, you only need to know the mapping between a token on the source and destination. However, the data worker, DVM voters, OO disputers, etc must have some central understanding of what mainnet token in the HubPool maps to which tokens on each destination chain. Otherwise, there's no way to do the rebalancing actions. If we have each destination chain defining these source destination mappings locally, we could technically end up with conflicting sources of truth with no concrete way to resolve.

I also don't quite see why destination chains need this info at all? It seems like this mapping could be stored on L1 rather than duplicated on each of these chains, but I could be missing the use case.

test/utils.ts
```diff
@@ -25,3 +24,14 @@ export const toBN = (num: string | number | BigNumber) => {
 export interface SignerWithAddress extends Signer {
   address: string;
 }
+
+export async function seedWallet(
+  walletToFund: any,
```
mrice32, line 29: Why do these need to be any? Doesn't Ethers have a type that you can pull in for signers/wallets?

mrice32 (commented):
Just this thread remaining: https://github.com/across-protocol/across-smart-contracts-v2/pull/5#discussion_r792341304.

contracts/SpokePool.sol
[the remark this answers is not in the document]
nicholaspai: yeah bringing the data worker requirements into context makes this more apparent. I agree we should store this exact mapping (`originTokenAddress => destChainId => destChainToken`) on L1 and then change the last part of the mapping to a `bool` on L2.

test/utils.ts
nicholaspai: Good point will fix

chrismaree (approved):
this looks awesome to me! good work @nicholaspai

PR #6, opened 2022-01-24
feat(contracts): Add LP logic

chrismaree:
This PR adds liquidity addition/removal logic.


nicholaspai (approved):

test/HubPool.LiquidityProvision.ts
```diff
@@ -1,20 +1,124 @@
...
+    await expect(() =>
+      hubPool.connect(liquidityProvider).removeLiquidity(weth.address, amountToLp.div(2), false)
+    ).to.changeEtherBalance(liquidityProvider, 0);
+
+    expect(await weth.balanceOf(liquidityProvider.address)).to.equal(amountToSeedWallets.sub(amountToLp.div(2))); // WETH balance should increase by the amount removed.
+
+    // Next, remove half the liquidity as ETH (set sendETH = true). This should modify the eth bal but not the weth bal.
+    await expect(() =>
+      hubPool.connect(liquidityProvider).removeLiquidity(weth.address, amountToLp.div(2), true)
+    ).to.changeEtherBalance(liquidityProvider, amountToLp.div(2)); // There should be ETH transferred, not WETH.
```
chrismaree, line 100: this syntax is great! unfortunately it cant be chained along with a `.to.changeTokenBalance` so I did this the old fashion way.

test/HubPool.TokenWhitelisting.ts
```diff
@@ -0,0 +1,38 @@
+import { expect } from "chai";
```
chrismaree, line 1: notice that I've separated sets of tests into files. goal is to not have one massive file and rather just run some fixtures and have fewer shorter test files.

hardhat.config.ts
```diff
@@ -23,12 +23,11 @@ task("accounts", "Prints the list of accounts", async (taskArgs, hre) => {
 // Go to https://hardhat.org/config/ to learn more
 
 const config: HardhatUserConfig = {
-  solidity: "0.8.11",
+  solidity: { compilers: [{ version: "0.8.11", settings: { optimizer: { enabled: true, runs: 200 } } }] },
```
nicholaspai, line 26: Set runs to 1_000_000?

test/HubPool.Fixture.ts
```diff
@@ -0,0 +1,45 @@
+import { TokenRolesEnum, ZERO_ADDRESS } from "@uma/common";
+import { getContractFactory, toWei } from "./utils";
+import { ethers } from "hardhat";
+import { Contract, BigNumber } from "ethers";
+
+export async function deployHubPoolTestHelperContracts(deployerWallet: any) {
```
nicholaspai, line 6: +1

nicholaspai (approved):
Awesome first PR on the HubPool lgtm!

test/HubPool.Fixture.ts
```diff
@@ -0,0 +1,45 @@
...
+  await dai.addMember(TokenRolesEnum.MINTER, deployerWallet.address);
+
+  // Deploy the hubPool
+  const hubPool = await (await getContractFactory("HubPool", deployerWallet)).deploy(weth.address, timer.address);
+
+  return { timer, weth, usdc, dai, hubPool };
+}
+
+export async function seedWallet(
+  walletToFund: any,
```
mrice32, line 24: Same here.

hardhat.config.ts
mrice32: +1

hardhat.config.ts
```diff
@@ -23,12 +23,11 @@ task("accounts", "Prints the list of accounts", async (taskArgs, hre) => {
...
   networks: {
     ropsten: {
       url: process.env.ROPSTEN_URL || "",
-      accounts:
-        process.env.PRIVATE_KEY !== undefined ? [process.env.PRIVATE_KEY] : [],
+      accounts: process.env.PRIVATE_KEY !== undefined ? [process.env.PRIVATE_KEY] : [],
```
mrice32, line 30: Why not use a mnemonic? Isn't that the preferred way of generating accounts?

contracts/HubPool.sol
```diff
@@ -51,9 +64,49 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
      *          LIQUIDITY PROVIDER FUNCTIONS         *
      *************************************************/
 
-    function addLiquidity(address token, uint256 amount) public {}
+    function addLiquidity(address l1Token, uint256 l1TokenAmount) public payable {
```
mrice32, line 67: Is this all new code? Or is it all copied? Or is it a mix? If a mix, can you comment on the parts that you have changed?

contracts/HubPool.sol
```diff
@@ -15,29 +19,38 @@ interface WETH9Like {
...
-        string memory lpTokenSymbol
-    ) public onlyOwner {
-        ERC20 lpToken = new ERC20(lpTokenName, lpTokenSymbol);
-        lpTokens[l1Token] = LPToken({ lpToken: address(lpToken), isWeth: isWeth, isEnabled: true });
+    function enableL1TokenForLiquidityProvision(address l1Token) public onlyOwner {
+        ExpandedERC20 lpToken = new ExpandedERC20(
+            append("Across ", IERC20Metadata(l1Token).name(), " LP Token"), // LP Token Name
+            append("Av2-", IERC20Metadata(l1Token).symbol(), "-LP"), // LP Token Symbol
+            IERC20Metadata(l1Token).decimals() // LP Token Decimals
+        );
```
mrice32, line 50: Minor note: we could use a contract like the TokenFactory here down the line so the ERC20 bytecode doesn't need to live in this contract (and thus, we give ourselves more room in the bytecode limit).

contracts/HubPool.sol
```diff
@@ -4,9 +4,13 @@ pragma solidity ^0.8.0;
 import "@uma/core/contracts/common/implementation/Testable.sol";
 import "@uma/core/contracts/common/implementation/Lockable.sol";
 import "@uma/core/contracts/common/implementation/MultiCaller.sol";
+import "@uma/core/contracts/common/implementation/ExpandedERC20.sol";
```
mrice32, line 7: nit: something we might consider for a follow-up. We could swap this out for an ERC20 contract that's optimized for this use case (immutable and singular minter/burner).

contracts/HubPool.sol
```diff
@@ -15,29 +19,38 @@ interface WETH9Like {
...
+        ExpandedERC20 lpToken;
         bool isEnabled;
     }
 
+    WETH9Like public l1Weth;
+
     mapping(address => LPToken) public lpTokens; // Mapping of L1TokenAddress to the associated LPToken.
 
-    constructor(address timerAddress) Testable(timerAddress) {}
+    event LiquidityAdded(address l1Token, uint256 amount, uint256 lpTokensMinted, address indexed liquidityProvider);
```
mrice32, line 33: Should we index the l1Token?

contracts/HubPool.sol
```diff
@@ -15,29 +19,38 @@ interface WETH9Like {
...
+    event LiquidityRemoved(uint256 amount, uint256 lpTokensBurnt, address indexed liquidityProvider);
```
mrice32, line 34: Should this include the l1Token?

contracts/HubPool.sol
```diff
@@ -72,4 +125,29 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
...
+    function append(
+        string memory a,
+        string memory b,
+        string memory c
+    ) internal pure returns (string memory) {
+        return string(abi.encodePacked(a, b, c));
+    }
+
+    // Added to enable the BridgePool to receive ETH. used when unwrapping Weth.
+    receive() external payable {}
```
mrice32, line 152: Should we revert if this comes from an address that is not the WETH contract?

hardhat.config.ts
chrismaree: I did not change this, this was just a listing change. I think we should ignore this until we get there but ye I agree a mnemonic would be better

test/HubPool.TokenWhitelisting.ts
```diff
@@ -0,0 +1,38 @@
...
+import { Contract } from "ethers";
+import { ethers } from "hardhat";
+import { ZERO_ADDRESS } from "@uma/common";
+import { getContractFactory } from "./utils";
+import { deployHubPoolTestHelperContracts } from "./HubPool.Fixture";
+
+let hubPool: Contract, weth: Contract, owner: any, liquidityProvider: any, other: any;
```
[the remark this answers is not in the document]
chrismaree: I could not easily find. will try again

contracts/HubPool.sol
chrismaree: it's copied but changed to include the new mapping structure. the v1 version of this did not specify the `l1Token` (one token per pool). the changes therefore are simply to define the l1Token within the mapping. the other small change is the mint logic is now called on the external contract, rather than on `this` as the LPtoken is external.

contracts/HubPool.sol
chrismaree: this is a good point. we can make this change if we run out of space. I'll add a comment.

contracts/HubPool.sol
chrismaree: good catch, it should. done and indexed

contracts/HubPool.sol
chrismaree: that's an interesting idea but its somewhat inconsistant. we dont prevent users from sending tokens so why should we prevent them from sending eth?

mrice32: I see what you're saying. ETH transfers work differently, so I would argue it's more likely to send ETH accidentally. However, this would incur a slight gas cost, so it probably isn't worth the cost of the check.

PR #8, opened 2022-01-25
feat: add a first draft of the proof library

mrice32:
Note: the proof verification functions all return true for now to make interacting with them simpler until we have some js or ts that can be used to construct them.

### DIFF
--- a/contracts/MerkleLib.sol
+++ b/contracts/MerkleLib.sol
@@ -19,9 +19,9 @@ library MerkleLib {
         // This only needs to be emitted in an event so future proposals know the most recent block number.
         uint256 endBlock;
 
-        // The following arrays are required to be the same length.
+        // The following arrays are required to be the same length. They are parallel arrays for the given chainId and should be ordered by the `tokenAddresses` field.
         address[] tokenAddresses; // All whitelisted tokens for this chain in the order of whitelisting.
-        uint256[] bundleLpFees; // LP fee amount per token in this bundle.
+        uint256[] bundleLpFees; // Total LP fee amount per token in this bundle, encompassing all associated bundled relays.
 
         // This array is grouped with the two above, and it represents the amount to send or request back from the
         // SpokePool. If positive, the pool needs to pay the SpokePool. If negative the SpokePool needs to pay the
@@ -44,7 +44,7 @@ library MerkleLib {
         uint256 amountToReturn;
         // The associated L2TokenAddress that these claims apply to.
         address l2TokenAddress;
-        // These two arrays must be the same length.
+        // These two arrays must be the same length and are parallel arrays. They should be order by refundAddresses.
         // This array designates each address that must be refunded.
         address[] refundAddresses;
         // This array designates how much each of those addresses should be refunded.
--- a/contracts/MerkleLib.sol
+++ b/contracts/MerkleLib.sol
@@ -16,21 +16,25 @@ library MerkleLib {
         uint256 leafId;
         // This is used to know which chain to send cross-chain transactions to (and which SpokePool to sent to).
         uint256 chainId;
-        // This only needs to be emitted in an event so future proposals know the most recent block number.
-        uint256 endBlock;
 
         // The following arrays are required to be the same length. They are parallel arrays for the given chainId and should be ordered by the `tokenAddresses` field.
-        address[] tokenAddresses; // All whitelisted tokens for this chain in the order of whitelisting.
+        // All whitelisted tokens with nonzero relays on this chain in this bundle in the order of whitelisting.
+        address[] tokenAddresses;
         uint256[] bundleLpFees; // Total LP fee amount per token in this bundle, encompassing all associated bundled relays.
 
         // This array is grouped with the two above, and it represents the amount to send or request back from the
-        // SpokePool. If positive, the pool needs to pay the SpokePool. If negative the SpokePool needs to pay the
-        // HubPool. There can be arbitrarily complex rebalancing rules defined offchain. This number is only nonzero
-        // when the rules indicate that a rebalancing action should occur. This means that the DestinationDistribution
-        // associated with this PoolRebalance can indicate a balance change, but that the HubPool does not rebalance.
-        // When a rebalance does occur, the netSendAmount is the sum of all netBalanceChange values in the
-        // DestinationDistribution structs since the last rebalance action.
+        // SpokePool. If positive, the pool will pay the SpokePool. If negative the SpokePool will pay the HubPool.
+        // There can be arbitrarily complex rebalancing rules defined offchain. This number is only nonzero
+        // when the rules indicate that a rebalancing action should occur. When a rebalance does not occur,
+        // runningBalance for this token should change by the total relays - deposits in this bundle. When a rebalance
+        // does occur, runningBalance should be set to zero for this token and the netSendAmount should be set to the
+        // previous runningBalance + relays - deposits in this bundle.
         int256[] netSendAmount;
+
+        // This is only here to be emitted in an event to track a running unpaid balance between the L2 pool and the L1 pool.
+        // A positive number indicates that the HubPool owes the SpokePool funds. A negative number indicates that the
+        // SpokePool owes the HubPool funds. See the comment above for the dynamics of this and netSendAmount
+        int256[] runningBalance;
     }
 
     // This leaf is meant to be decoded in the SpokePool in order to pay out individual relayers for this bundle.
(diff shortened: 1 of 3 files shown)


contracts/MerkleLib.sol
```diff
@@ -0,0 +1,101 @@
...
+        // This is the amount to return to the HubPool. This occurs when there is a PoolRepayment netSendAmount that is
+        // negative. This is just that value inverted. 
+        uint256 amountToReturn;
+        // The associated L2TokenAddress that these claims apply to.
+        address l2TokenAddress;
+        // These two arrays must be the same length.
+        // This array designates each address that must be refunded.
+        address[] refundAddresses;
+        // This array designates how much each of those addresses should be refunded.
+        uint256[] refundAmounts;
```
nicholaspai, line 51: Does the ordering of this struct matter in terms of packing types together?

contracts/MerkleLib.sol
```diff
@@ -0,0 +1,101 @@
...
+
+
+/**
+ * @notice Library to help with merkle roots, proofs, and claims.
+ */
+library MerkleLib {
+    // TODO: some of these data structures can be moved out if convenient.
+    // This data structure is used in the settlement process on L1.
+    // Each PoolRepayment structure is responsible for balancing a single chain's SpokePool across all tokens.
+    struct PoolRepayment {
```
nicholaspai, line 14: nit: repayment sounds like paying back the relayer. Why not use another more clear word like "rebalance"?

nicholaspai (approved):
Good start +1.

contracts/MerkleLib.sol
mrice32: Good point. Will update.

mrice32: Fixed.

contracts/MerkleLib.sol
mrice32: Maybe, but probably not:
1. The only type that's currently <32 bytes is address, which is in an array.
2. Array elements are padded to 32 bytes even if using abi.encodePacked.

contracts/MerkleLib.sol
```diff
@@ -0,0 +1,101 @@
...
+        // Used as the index in the bitmap to track whether this leaf has been executed or not.
+        uint256 leafId;
+        // This is used to know which chain to send cross-chain transactions to (and which SpokePool to sent to).
+        uint256 chainId;
+        // This only needs to be emitted in an event so future proposals know the most recent block number.
+        uint256 endBlock;
+
+        // The following arrays are required to be the same length.
+        address[] tokenAddresses; // All whitelisted tokens for this chain in the order of whitelisting.
+        uint256[] bundleLpFees; // LP fee amount per token in this bundle.
```
chrismaree, line 24:
```suggestion
        uint256[] bundleLpFees; // total LP fee amount per token in this bundle, encompassing all associated bundled relays.
```

contracts/MerkleLib.sol
```diff
@@ -0,0 +1,101 @@
...
+
+        // This array is grouped with the two above, and it represents the amount to send or request back from the
+        // SpokePool. If positive, the pool needs to pay the SpokePool. If negative the SpokePool needs to pay the
+        // HubPool. There can be arbitrarily complex rebalancing rules defined offchain. This number is only nonzero
+        // when the rules indicate that a rebalancing action should occur. This means that the DestinationDistribution
+        // associated with this PoolRebalance can indicate a balance change, but that the HubPool does not rebalance.
+        // When a rebalance does occur, the netSendAmount is the sum of all netBalanceChange values in the
```
chrismaree, line 31: this does not match the variables in the `DestinationDistribution`. you refer to `netBalanceChange` in the comment but there is no such variable.

contracts/MerkleLib.sol
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    // This leaf is meant to be decoded in the SpokePool in order to pay out individual relayers for this bundle.
+    struct DestinationDistribution {
+        // Used as the index in the bitmap to track whether this leaf has been executed or not.
+        uint256 leafId;
+        // Used to verify that this is being decoded on the correct chainId.
+        uint256 chainId;
+        // This is the amount to return to the HubPool. This occurs when there is a PoolRebalance netSendAmount that is
+        // negative. This is just that value inverted.
```
chrismaree, line 43: I'm not sure if we want the inverted amount. it might be easier if this is always a int256. a positive number always means L1->L2 flow and a negative number always means L2->L1 flow. dont bake any polarity changes into the variable.

contracts/MerkleLib.sol
```diff
@@ -0,0 +1,101 @@
...
+        uint256 leafId;
+        // Used to verify that this is being decoded on the correct chainId.
+        uint256 chainId;
+        // This is the amount to return to the HubPool. This occurs when there is a PoolRebalance netSendAmount that is
+        // negative. This is just that value inverted. 
+        uint256 amountToReturn;
+        // The associated L2TokenAddress that these claims apply to.
+        address l2TokenAddress;
+        // These two arrays must be the same length.
+        // This array designates each address that must be refunded.
```
chrismaree, line 48:
```suggestion
        // This array designates each address that must be refunded, ordered by address.
```

contracts/MerkleLib.sol
```diff
@@ -0,0 +1,101 @@
...
+    // Each PoolRebalance structure is responsible for balancing a single chain's SpokePool across all tokens.
+    struct PoolRebalance {
+        // Used as the index in the bitmap to track whether this leaf has been executed or not.
+        uint256 leafId;
+        // This is used to know which chain to send cross-chain transactions to (and which SpokePool to sent to).
+        uint256 chainId;
+        // This only needs to be emitted in an event so future proposals know the most recent block number.
+        uint256 endBlock;
+
+        // The following arrays are required to be the same length.
```
chrismaree, line 22:
```suggestion
        // The following arrays are required to be the same length. They are parallel arrays for the given chainId and should be ordered by the `tokenAddresses` field.
```

contracts/MerkleLib.sol
```diff
@@ -0,0 +1,101 @@
...
+        // Used as the index in the bitmap to track whether this leaf has been executed or not.
+        uint256 leafId;
+        // Used to verify that this is being decoded on the correct chainId.
+        uint256 chainId;
+        // This is the amount to return to the HubPool. This occurs when there is a PoolRebalance netSendAmount that is
+        // negative. This is just that value inverted. 
+        uint256 amountToReturn;
+        // The associated L2TokenAddress that these claims apply to.
+        address l2TokenAddress;
+        // These two arrays must be the same length.
```
chrismaree, line 47:
```suggestion
        // These two arrays must be the same length and are parallel arrays. They should be order by refundAddresses.
```

contracts/MerkleLib.sol
```diff
@@ -0,0 +1,101 @@
...
+        // HubPool. There can be arbitrarily complex rebalancing rules defined offchain. This number is only nonzero
+        // when the rules indicate that a rebalancing action should occur. This means that the DestinationDistribution
+        // associated with this PoolRebalance can indicate a balance change, but that the HubPool does not rebalance.
+        // When a rebalance does occur, the netSendAmount is the sum of all netBalanceChange values in the
+        // DestinationDistribution structs since the last rebalance action.
+        int256[] netSendAmount;
+    }
+
+    // This leaf is meant to be decoded in the SpokePool in order to pay out individual relayers for this bundle.
+    struct DestinationDistribution {
```
chrismaree, line 37: there was a bool within this struct at one point to indicate if `amountToReturn` should be sent. what happened to this?

contracts/MerkleLib.sol
mrice32: As discussed today, we would only set this number if we wanted to tell the receiving chain to send tokens back to mainnet. So I made this positive since there is no need to communicate when tokens are being sent _to_ the L2 (aka the opposite sign).

contracts/MerkleLib.sol
mrice32: As discussed today, this is no longer needed. We don't need to communicate when the rebalances are happening in this data structure. The netSendAmount in the struct above is the only place where we denote whether money is being sent to the chain or not.

contracts/MerkleLib.sol
mrice32: Agreed. Will fix these up.

chrismaree (approved):

PR #9, opened 2022-01-26
feat: relay logic

nicholaspai:
Signed-off-by: Nick Pai <npai.nyc@gmail.com>


contracts/SpokePool.sol
```diff
@@ -147,19 +179,88 @@ abstract contract SpokePool is Testable, Lockable, MultiCaller {
...
+    ) public {
+        // We limit the relay fees to prevent the user spending all their funds on fees.
+        require(relayerFeePct <= 0.5e18 && realizedLpFeePct <= 0.5e18, "invalid fees");
+
+        // Use relay count as unique ID for new relay.
+        require(relays[numberOfRelays].relayAmount == 0, "Relay exists");
+        relays[numberOfRelays] =
+            RelayData(
+                recipient,
+                relayToken,
```
chrismaree, line 199: can we rename this variable to `relayDestinationToken` to make it clear it's the token on the destination?

contracts/SpokePool.sol
```diff
@@ -147,19 +179,88 @@ abstract contract SpokePool is Testable, Lockable, MultiCaller {
...
+        address relayToken,
+        address sender,
+        address recipient
+    ) public {
+        // We limit the relay fees to prevent the user spending all their funds on fees.
+        require(relayerFeePct <= 0.5e18 && realizedLpFeePct <= 0.5e18, "invalid fees");
+
+        // Use relay count as unique ID for new relay.
+        require(relays[numberOfRelays].relayAmount == 0, "Relay exists");
+        relays[numberOfRelays] =
```
chrismaree, line 196: is it better to assign to an index within in an array (like you have here) or better to push into the array then define a method for `numberOfRelays` to be equal to the `relays.length`?

contracts/SpokePool.sol
```diff
@@ -147,19 +179,88 @@ abstract contract SpokePool is Testable, Lockable, MultiCaller {
...
+            relayerFeePct, 
+            realizedLpFeePct, 
+            relayToken, 
+            sender, 
+            recipient
+        );
+
+        // Questions:
+        // - do we need to store relayer (i.e. msg.sender) anywhere on-chain? Or can we assume data worker
+        //   can grab it from event data.
```
chrismaree, line 220: I think you should emit it. just add the prop the the even then emit `msg.sender`.

contracts/SpokePool.sol
```diff
@@ -147,19 +179,88 @@ abstract contract SpokePool is Testable, Lockable, MultiCaller {
...
+            amount > 0 &&
+            relay.relayAmount >= relay.filledAmount + amount, 
+            "Invalid remaining relay amount"
+        );
+
+        // Update total filled amount with this new fill. Each fill can be uniquely identified by the
+        // total filled amount including itself, the relay ID, and the chain ID of this contract.
+        relays[relayId].filledAmount += amount;
+
+        // Pull fill amount net fees from caller, which is the amount owed to the recipient.  The relayer will receive
```
chrismaree, line 244:
```suggestion
        // Pull fill amount net fees from caller, which is the amount owed to the recipient. The relayer will receive 
```

contracts/SpokePool.sol
```diff
@@ -147,19 +179,88 @@ abstract contract SpokePool is Testable, Lockable, MultiCaller {
...
+        IERC20(relay.relayToken).safeTransferFrom(msg.sender, relay.recipient, amountNetFees);
+
+        // If relay token is weth then unwrap and send eth.
+        if (relay.relayToken == wethAddress) {
+            _unwrapWETHTo(payable(relay.recipient), amountNetFees);
+            // Else, this is a normal ERC20 token. Send to recipient.
+        } else IERC20(relay.relayToken).safeTransfer(relay.recipient, amountNetFees);
+
+        // Questions:
+        // - same question as above: do we need to store relayer (i.e. msg.sender) anywhere on-chain? Or can we assume data worker
```
chrismaree, line 259: I think including it in the event is best. something like `filler` or `relayFiller`

contracts/SpokePool.sol
```diff
@@ -147,19 +179,88 @@ abstract contract SpokePool is Testable, Lockable, MultiCaller {
...
+        //   can grab it from event data.
+
+        emit FilledRelay(relayId, relays[relayId].filledAmount, repaymentChain);
```
chrismaree, line 262: I think you should also emit `amountNetFees` as this is the actual transfer amount.

contracts/SpokePool.sol
```diff
@@ -147,19 +179,88 @@ abstract contract SpokePool is Testable, Lockable, MultiCaller {
...
+        );
+
+        // Questions:
+        // - do we need to store relayer (i.e. msg.sender) anywhere on-chain? Or can we assume data worker
+        //   can grab it from event data.
+
+        numberOfRelays += 1;
+    }
 
     function fillRelay(
```
chrismaree, line 225: I like the structure you have here and how you defined `amount` as including fees and then send amount without fees.

chrismaree (commented):
This looks great to me! I think you should add a bit more to events and perhaps consider changing how you are appending to arrays but other than this it looks great!

contracts/SpokePool.sol
nicholaspai: sure

contracts/SpokePool.sol
nicholaspai: done, easy

contracts/SpokePool.sol
nicholaspai: ah true

contracts/SpokePool.sol
nicholaspai: good point, i removed `numberOfRelays`, changed `relays` to an array of `RelayData` and use the length as the nonce

contracts/SpokePool.sol
```diff
@@ -147,19 +182,73 @@ abstract contract SpokePool is Testable, Lockable, MultiCaller {
...
+            RelayData(
+                recipient,
+                destinationToken,
+                relayerFeePct,
+                realizedLpFeePct,
+                amount, // total relay amount
+                0 // total amount filled
+            )
+        );
+        uint64 relayId = uint64(relays.length - 1);
```
chrismaree, line 208: do you think it will use less gas to not define this variable at all and rather just put `uint64(relays.length - 1)` on line 214?

contracts/SpokePool.sol
```diff
@@ -147,19 +182,73 @@ abstract contract SpokePool is Testable, Lockable, MultiCaller {
...
+        // Update total filled amount with this new fill. Each fill can be uniquely identified by the
+        // total filled amount including itself, the relay ID, and the chain ID of this contract.
+        relays[relayId].filledAmount += amount;
+
+        // Pull fill amount net fees from caller, which is the amount owed to the recipient. The relayer will receive
+        // this amount plus the relayer fee after the relayer refund is processed.
+        uint256 amountNetFees = amount - _getAmountFromPct(relay.realizedLpFeePct + relay.relayerFeePct, amount);
+        IERC20(relay.destinationToken).safeTransferFrom(msg.sender, relay.recipient, amountNetFees);
+
+        // If relay token is weth then unwrap and send eth.
```
chrismaree, line 244: do you think that we should try let the user choose to receive WETH or ETH? it's kinda strange that a user sending WETH or ETH always gets ETH? (obs not something for now but I'm curious what you think).

chrismaree (approved):
This is excellent! nice and simple, well commented and good code. good work

contracts/SpokePool.sol
mrice32, line 208: nit: you might be able to avoid the `- 1` if you move this above the push call.

mrice32 (commented):
Looks great! A few high-level comments

contracts/SpokePool.sol
nicholaspai: Yeah it should ill change

contracts/SpokePool.sol
```diff
@@ -147,19 +182,73 @@ abstract contract SpokePool is Testable, Lockable, MultiCaller {
...
+            sender,
+            recipient,
+            msg.sender
+        );
+    }
 
     function fillRelay(
-        uint256 relayId,
-        uint256 fillAmount,
+        uint64 relayId,
```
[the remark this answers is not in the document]
nicholaspai: hmm this is a good point. Using a state variable counter `numberOfRelays` also wouldn't work. We could make the relay ID determined solely by unique deposit data: `depositId` + `originChainId`?

contracts/SpokePool.sol
nicholaspai: See above but I think I'm going to remove `relayId` in favor of a `relayHash` which is a hash of the relay data + deposit data

contracts/SpokePool.sol
```diff
@@ -39,21 +43,57 @@ abstract contract SpokePool is Testable, Lockable, MultiCaller {
     // Origin token to destination token routings can be turned on or off.
     mapping(address => mapping(uint256 => bool)) public enabledDepositRoutes;
 
+    struct RelayData {
```
mrice32, line 46: This is really interesting and raises a tradeoff.

In v1, we store a hash of the full relay data and require the caller to pass it in on each call. In v2, it looks like the current design is to store the relay data. I'm guessing this is because a) it's simpler and b) storage is cheap on L2s and calldata is expensive.

However, this amount of storage might make our L2 -> L1 transfers more expensive. Also, I think the common path through this code is that there will be a single relayer who fills the entire thing on one go, so there will be usually be no need to call into it multiple times to read this data.

Another, related thought is that since calldata is expensive on L2, it will be an added cost to have to multicall initiate and fill relay on a typical relay. This requires more calldata than just having a combined method that can do both things.

Going back to these structs, the _only_ piece of dynamic data is filledAmount. Everything else is static after initialization I think.

Putting this all together, what do you think of a setup where _all_ the static data (everything in both these structs except for filledAmount) is hashed to look up a single number, which is filledAmount?

In this setup you could have a single method called `fill` that takes all this data + the amount you want to fill. If the relay exists, just adds to the filledAmount. If the relay doesn't exist, it initializes it.

Thoughts? Note: I could totally be missing something, I was just trying to think of ways to keep this cheap on both L1 and L2.

mrice32 (commented):
This looks awesome! Thanks for adding the hashing logic. +1. I have one high-level comment

contracts/SpokePool.sol
nicholaspai: I didn't go into nearly as much thought about cost when designing this interface. I was motivated mostly by simplicity. 

I think generating the relay hash as it's stored here is simple enough and I also agree we should key fillAmount by relayHash.

contracts/SpokePool.sol
```diff
@@ -147,19 +187,73 @@ abstract contract SpokePool is Testable, Lockable, MultiCaller {
...
+        uint64 realizedLpFeePct,
+        address destinationToken,
+        address sender,
+        address recipient
+    ) public {
+        // We limit the relay fees to prevent the user spending all their funds on fees.
+        require(relayerFeePct <= 0.5e18 && realizedLpFeePct <= 0.5e18, "invalid fees");
+
+        // Relay will be keyed to unique deposit plus relay data hash.
+        RelayData memory relayData = RelayData(
```
chrismaree, line 203: is it cheaper to get the relayer to pass a `memory relayData` as a prop in this function and not construct this in the method like this?

chrismaree: main downside is this makes calling this method a bit more complex and we would require the caller to set the `totalFillAmount` to 0.

chrismaree: if you choose to leave it like this, can you use Key-value assignment syntax that we use here: https://github.com/across-protocol/across-smart-contracts-v2/blob/977a9a48f8b3993c95d2a4d724462355823c2458/contracts/HubPool.sol?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L82

contracts/SpokePool.sol
chrismaree: the depositor would need an extra bool, which is kind of a pain 🤔 . might not be worth it tbh

contracts/SpokePool.sol
```diff
@@ -146,20 +174,65 @@ abstract contract SpokePool is Testable, Lockable, MultiCaller {
...
-    ) public {}
-
-    function fillRelay(
-        uint256 relayId,
+        address destinationToken,
+        uint64 realizedLpFeePct,
+        uint64 relayerFeePct,
+        uint64 depositId,
+        uint256 originChainId,
+        uint256 relayAmount,
```
mrice32, line 185: When we say "fillAmount", is this the amount of tokens that are transferred from the relayers wallet? Or something else? What do you think the right UX for this is?

We could make this the max amount they're willing to have pulled from their wallet?

Thoughts?

contracts/SpokePool.sol
```diff
@@ -39,21 +43,45 @@ abstract contract SpokePool is Testable, Lockable, MultiCaller {
...
         uint256 amount,
+        uint64 depositId,
         uint64 relayerFeePct,
         uint64 quoteTimestamp,
         address originToken,
         address recipient,
         address sender
     );
+    event FilledRelay(
+        bytes32 relayHash,
```
mrice32, line 78: nit: do you think this should be indexed?

contracts/SpokePool.sol
```diff
@@ -39,21 +43,45 @@ abstract contract SpokePool is Testable, Lockable, MultiCaller {
...
+        uint256 newFilledAmount,
+        uint256 repaymentChain,
+        uint256 fillAmountNetFees,
+        address relayer,
```
mrice32, line 82: indexed maybe?

mrice32 (approved):
LGTM!

contracts/SpokePool.sol
chrismaree: it's not the amount pulled from the Wallet. it's the amount pulled plus the realised LP Fee, plus the relayer fee. put another way, the amount sent from the relayer(and thus pulled from their wallet) is `fillAmount * (1 - relayerFeePct - realizedLpFeePct`

contracts/SpokePool.sol
```diff
@@ -39,21 +43,45 @@ abstract contract SpokePool is Testable, Lockable, MultiCaller {
...
+        uint64 relayerFeePct;
+        uint64 depositId;
+        uint256 originChainId;
+        uint256 relayAmount;
+    }
+
+    // Each relay is associated with the hash of parameters that uniquely identify the original deposit and a relay
+    // attempt for that deposit. The relay itself is just represented as the amount filled so far. The total amount to 
+    // relay, the fees, and the agents are all parameters included in the hash key.
+    mapping(bytes32 => uint256) relays;
```
chrismaree, line 60: is `relays` the best name for this variable? what it really is is a mapping of hashes, to their fills, like your comment describes. would a better variable name be `relayFillAmounts` or `relayFills`? curious what you think.

contracts/SpokePool.sol
nicholaspai: Yeah chris is correct

contracts/SpokePool.sol
nicholaspai: yeah

contracts/SpokePool.sol
nicholaspai: This leads to a follow up question. For the events we have so far, which params should we index?

contracts/SpokePool.sol
nicholaspai: relayFills could be good, but it also tracks relays generally right? I think I'll change to `relayFills` unless there's disagreement

contracts/SpokePool.sol
nicholaspai: ended up using key-value syntax but not going to make caller pass in `relayData` because it marginally makes the interface harder to use for EOA's on etherscan

nicholaspai: it adds a little calldata too

contracts/SpokePool.sol
mrice32: So a few points on this:
1. I think it's a bit unclear what fillAmount is, although we might be able to resolve this via comments.
2. If a relayer is capital limited (which is the primary motivation for partial fills), they would likely want to use this method to send exactly as many tokens as they have to the contract. Instead of requiring them to compute fillAmount from that, why can't we just let them specify how many tokens they want to send (or the max amount they are willing to send) as that's what they would care about I would think?

contracts/SpokePool.sol
mrice32: SGTM!

contracts/SpokePool.sol
mrice32: I think the ones you have indexed look good to me!

mrice32 (commented):
Only remaining comment: https://github.com/across-protocol/across-smart-contracts-v2/pull/9#discussion_r796100335.

contracts/SpokePool.sol
chrismaree: ye, I agree with that 2nd point. it's confusing. what should we call it though @mrice32 ?

nicholaspai: what if we just renamed them to `relayAmount => totalAmountToRelay` and `fillAmount => amountToRelay`...?

nicholaspai: I don't quite understand point 2 @mrice32

mrice32: Sorry, I think I'm making this too confusing. That's my bad.

If I have 1000 USDC, it's difficult to tell this method that I only want to send 1000 USDC since you have to do some math to get there. Instead, you could have a field called `maxTokensToSend` or similar, that allows the caller to specify the max number of tokens they're willing to contribute.

Not only is the easier for the caller, but it's nice from a frontrunning perspective too. If I try to fill 1000 USDC, but someone frontruns my txn and there's only room for 500 USDC, my transaction still goes through, it just pulls less from my wallet.

chrismaree: the point is that `fillAmount`, the thing that the relayer is passing into the prop, is not the actual amount of funds that gets pulled from their wallet. I.e if the relay is 100 weth but I only have 5 weth I cant put `fillAmount` to 5 weth as this won't use _all_ my funds, it'll pull `5*(1-relayerFeePct-realizedLpFeePct)` thereby leaving over some in my wallet. If I want to spend the _full_ 5 WETH I will need some compute `5*(1+relayerFeePct+realizedLpFeePct)` thereby pulling exactly 5 WETH from my wallet.

The point here is the prop name `fillAmount` is a bit confusing as ideally the amount the relayer chooses for this prop is what is pulled from their wallet so they can choose exactly how much to send.

contracts/SpokePool.sol
mrice32: +1 @nicholaspai

contracts/SpokePool.sol
nicholaspai: ah i see so is it naive to just change the name to `maxAmountToFill`? or `fillAmountPreFees`? This way the caller has more clarity that they won't send the full amount that they pass in?

chrismaree: I think that's ok for now @nicholaspai . we might want to re-consider this to be the amount that's actually pulled from their wallet but that's a bit more complex to reason about. a good name could be `fillAmountSansFees` or like you have it with `PreFees`.

mrice32: Hmmm, I don't really think the name is the core issue. I think it's the fact that a caller has no way to say "I want to send this amount of funds". Is it easy to compute fillAmount from maxTokensToSend in this function?

contracts/SpokePool.sol
```diff
@@ -29,30 +32,54 @@ abstract contract SpokePool is Testable, Lockable, MultiCaller {
...
 
-    // Track the total number of deposits. Used as a unique identifier for deposits.
-    uint256 public numberOfDeposits;
+    // Use count of deposits as unique deposit identifier.
+    uint64 public numberOfDeposits;
 
     // Address of WETH contract for this network. If an origin token matches this, then the caller can optionally
     // instruct this contract to wrap ETH when depositing.
-    address public wethAddress;
+    WETH9Like public weth;
```
chrismaree, line 40: nice. I prefer this pattern.

contracts/SpokePool.sol
```diff
@@ -62,7 +89,7 @@ abstract contract SpokePool is Testable, Lockable, MultiCaller {
     ) Testable(timerAddress) {
```
chrismaree, line 89: in the constructor params might it be better to put the timer last? we always do this in all the other UMA contracts. wdyt?

test/SpokePool.Relay.ts
```diff
@@ -0,0 +1,200 @@
...
+        0, // deposit ID
+        666, // origin chain ID
+        destErc20.address
+    )
+
+    await expect(
+      spokePool
+        .connect(relayer)
+        .fillRelay(
+            relayData[0],
```
chrismaree, line 49: can you not just use the spread operator here? `...relayData`?

chrismaree: I see in a few of these cases you are overriding the data within `relayData`. this could be solved by returning an object from `getRelayhash` then overriding the named props using `{...relayData, namedPropYouWantToOverride}` or something like this so we dont need to define these long lists of indexes.

wdyt, my goal is fewer lines here

contracts/SpokePool.sol
chrismaree: it's trivial to compute this I think. if we shifted it to this you can modify some of the logic within this function to deal with this. if the prop was called `relayerFillAmount` or `relayerAmountSpend` indicating how much funds are actually pulled from the relayers wallet, then you would fill the amount of the original relay by `relayFillAmount*(1+relayerFeePct+realizedLpFeePct)`. i.e the amount _filled_ includes these fees but the amount _sent_ is before the fees. 

it's really just a framing question and both are equally simple to implement IMHO.

nicholaspai: I think I understand what you're saying. In this [commit](https://github.com/across-protocol/across-smart-contracts-v2/pull/9/commits/977ffa9d7db0e6f590c44d15f8a963e80d0737e3) I allow the user to pass in `maxTokensToSend`, and the contract now computes internal accounting of `fillAmountPreFees`

contracts/SpokePool.sol
nicholaspai: sure

test/SpokePool.Relay.ts
nicholaspai: done

contracts/SpokePool.sol
```diff
@@ -146,20 +173,67 @@ abstract contract SpokePool is Testable, Lockable, MultiCaller {
...
+            originChainId: originChainId,
+            relayAmount: totalRelayAmount
+        });
+        bytes32 relayHash = _getRelayHash(relayData);
+
+        // Compute the equivalent amount to be sent by the relayer before fees have been taken out. This is the amount
+        // that we'll add to the `relayFills` counter, and we do this math here in the contract for the user's
+        // convenience so that they don't have to do this math before calling this function. The user can simply
+        // pass in `maxTokensToSend` and assume that the contract will pull exactly that amount of tokens (or revert).
+        uint256 fillAmountPreFees = (maxTokensToSend / (1e18 - (realizedLpFeePct + relayerFeePct))) * 1e18;
```
mrice32, line 214: Would swapping the order of the multiplication and division mean less rounding?

contracts/SpokePool.sol
```diff
@@ -146,20 +174,66 @@ abstract contract SpokePool is Testable, Lockable, MultiCaller {
...
+
+        // Check that the caller is filling a non zero amount of the relay and that the filled amount will not
+        // send to the recipient more than they are expected to receive. Note that the `relays` mapping will point
+        // to the amount filled so far for a particular `relayHash`, so this will start at 0 and increment with each
+        // fill.
+        require(
+            maxTokensToSend > 0 && totalRelayAmount >= relayFills[relayHash] + fillAmountPreFees,
```
mrice32, line 221: I think this would fail if the user passes in a number _larger_ than the amount needed to fill the remainder of the relay. Should we do some something like `min(totalRelayAmount - relayFills[relayHash], fillAmountPreFees)` to determine how much they will actually fill?

mrice32 (approved):
LGTM! Two minor comments on the fill math, but this looks really good!

contracts/SpokePool.sol
nicholaspai: Ah yes i forgot to add this logic, will do

contracts/SpokePool.sol
nicholaspai: hm yeah

contracts/SpokePool.sol
nicholaspai: Do you think this logic is too convoluted and are there gas savings? https://github.com/across-protocol/across-smart-contracts-v2/pull/9/commits/c6970ae1bd291221333415d20d570cee21dda1c6

contracts/SpokePool.sol
```diff
@@ -146,20 +174,73 @@ abstract contract SpokePool is Testable, Lockable, MultiCaller {
...
+        // that we'll add to the `relayFills` counter, and we do this math here in the contract for the user's
+        // convenience so that they don't have to do this math before calling this function. The user can simply
+        // pass in `maxTokensToSend` and assume that the contract will pull exactly that amount of tokens (or revert).
+        uint256 fillAmountPreFees = 1e18 * maxTokensToSend / (1e18 - (realizedLpFeePct + relayerFeePct));
+
+        // If user's specified max amount to send is greater than the amount of the relay remaining pre-fees, 
+        // we'll pull exactly enough tokens to complete the relay.
+        uint256 amountToSend;
+        if (totalRelayAmount - relayFills[relayHash] < fillAmountPreFees) {
+            amountToSend = totalRelayAmount - relayFills[relayHash];
```
mrice32, line 228: Won't this cause them to send the amount _including fees_. Don't we need to multiply this by `(1e18 - (realizedLpFeePct + relayerFeePct)) / 1e18`?

mrice32 (approved):

contracts/SpokePool.sol
nicholaspai: Ah yes good call, fixed

contracts/SpokePool.sol
```diff
@@ -180,4 +261,34 @@ abstract contract SpokePool is Testable, Lockable, MultiCaller {
...
+    /**************************************
+     *         INTERNAL FUNCTIONS         *
+     **************************************/
+
+    function _computeAmountPreFees(uint256 amount, uint256 feesPct) private pure returns (uint256) {
+        return (1e18 * amount) / (1e18 - feesPct);
+    }
+
+    function _computeAmountPostFees(uint256 amount, uint256 feesPct) private pure returns (uint256) {
+        return (amount * feesPct) / 1e18;
```
mrice32, line 274: nit: should this also be `(1e18 - feesPct)` rather than just `feesPct`?

mrice32 (approved):
Everything looks good, just that one nit: https://github.com/across-protocol/across-smart-contracts-v2/pull/9#discussion_r796923511

contracts/SpokePool.sol
nicholaspai: shoot yeah, the test case where the fees are equally set to 0.25 and 0.25 is confusing because 1e18 - totalFeesPct == totalFeesPct. I changed the test values to 0.1 and 0.2 and changed this solidity code to `1e18 - feesPct`

mrice32 (approved):
LGTM!

PR #10, opened 2022-01-27
feat(hubpool): Add Initial initiateRelayerRefund implementation 

chrismaree:
Adds initial `initiateRelayerRefund` PR.


contracts/MerkleLib.sol
```diff
@@ -3,7 +3,6 @@ pragma solidity ^0.8.0;
 
 import "@openzeppelin/contracts/utils/cryptography/MerkleProof.sol";
```
chrismaree, line 4: changes in this file were due to running the linter config.

contracts/BridgeAdmin.sol
```diff
@@ -0,0 +1,14 @@
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity ^0.8.0;
+
+import "./interfaces/BridgeAdminInterface.sol";
+
+
+contract BridgeAdmin is BridgeAdminInterface {
```
nicholaspai, line 7: why not just import this from uma/core?

contracts/HubPool.sol
```diff
@@ -26,13 +31,40 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
...
+
+    RelayerRefundRequest[] public relayerRefundRequests;
 
     // Whitelist of origin token to destination token routings to be used by off-chain agents.
     mapping(address => mapping(uint256 => address)) public whitelistedRoutes;
 
     mapping(address => LPToken) public lpTokens; // Mapping of L1TokenAddress to the associated LPToken.
 
+    // Address of L1Weth. Enable LPs to deposit/receive ETH, if they choose, when adding/removing liquidity.
+    WETH9Like public l1Weth;
```
nicholaspai, line 54: in general do we prefer storing contracts as addresses or as interfaces? I stored `weth` as an address in the `SpokePool` for example but we should just standardize

contracts/BridgeAdmin.sol
nicholaspai: Ah I see that this would be a brand new contract

contracts/HubPool.sol
```diff
@@ -26,13 +31,40 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
...
+
+    // Address of the bridgeAdmin that is used to manage the hub.
+    BridgeAdminInterface public bridgeAdmin;
+
+    // Token used to bond the data worker for proposing relayer refund bundles.
+    IERC20 public bondToken;
+
+    // The bondToken's final fee from the UMA Store is scaled by this number to increase the bonding amount.
+    uint256 public bondTokenFinalFeeMultiplier;
```
nicholaspai, line 63: Can we make this a uint64?

nicholaspai: for context, I use uint64 for all percentage values in spokepool

contracts/HubPool.sol
```diff
@@ -47,14 +79,39 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
...
+        bytes32 indexed destinationDistributionProof,
+        address proposer
+    );
+
+    constructor(
+        address _l1Weth,
+        address _bridgeAdmin,
+        address _bondToken,
+        uint256 _bondTokenFinalFeeMultiplier,
+        address _timerAddress
```
nicholaspai, line 95: nit: group similar types together

contracts/HubPool.sol
```diff
@@ -47,14 +79,39 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
...
+        bondToken = IERC20(_bondToken);
+        bondTokenFinalFeeMultiplier = _bondTokenFinalFeeMultiplier;
     }
 
     /*************************************************
      *                ADMIN FUNCTIONS                *
      *************************************************/
 
+    function setBondToken(address _bondToken) public onlyOwner {
+        bondToken = IERC20(_bondToken);
```
nicholaspai, line 108: These should emit events since they change state

contracts/HubPool.sol
```diff
@@ -139,11 +196,35 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
...
+        bytes32 destinationDistributionProof
+    ) public {
+        relayerRefundRequests.push(
+            RelayerRefundRequest({
+                poolRebalanceProof: poolRebalanceProof,
+                destinationDistributionProof: destinationDistributionProof,
+                status: RefundRequestStatus.Pending
+            })
+        );
+        uint256 relayerRefundId = relayerRefundRequests.length - 1;
```
nicholaspai, line 215: int: should the ID be a smaller type like uint64?

nicholaspai (approved):
LGTM i agree with the structure but have nits on types

contracts/HubPool.sol
chrismaree: ye, good point. my main question which should inform this decision is is this more expensive to store? is this more expensive to use?

my main reason for doing it this way is it means you dont need to do any type casting when using the interface which makes things a bit easier.

contracts/HubPool.sol
chrismaree: but ye we can. I'll change the type to uint64

contracts/HubPool.sol
```diff
@@ -47,14 +83,41 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
...
     /*************************************************
      *                ADMIN FUNCTIONS                *
      *************************************************/
 
+    function setBondToken(address newBondToken) public onlyOwner {
+        bondToken = IERC20(newBondToken);
+        emit BondTokenSet(newBondToken);
+    }
+
+    function setBondTokenFinalFeeMultiplier(uint64 newBondTokenFinalFeeMultiplier) public onlyOwner {
```
mrice32, line 116: This is interesting. I totally get the rationale: it's easier to just set the final fees and have this move around as a result since they're kinda a valuation anchor, but do you think this is weird in that it relies on internal UMA system variables for setting its own security bounds? Like if the UMA voting system changed and spam were less of a concern, so final fees were dropped by 95%, we probably wouldn't want this value to change as a result, right?

It's also not _super_ hard to manage this since it's only one bond and one token for the entire Across v2 system.

As a side note: this also means that we need to be sure that the bond token is an approved UMA collateral currency.

contracts/HubPool.sol
```diff
@@ -139,11 +202,34 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
...
+        relayerRefundRequests.push(
+            RelayerRefundRequest({
+                poolRebalanceProof: poolRebalanceProof,
+                destinationDistributionProof: destinationDistributionProof,
+                status: RefundRequestStatus.Pending
+            })
+        );
+
+        // Pull bonds from from the caller.
+        bondToken.safeTransferFrom(msg.sender, address(this), bondAmount);
```
mrice32, line 223: This is annoying, but how do we deal with a changing bond amount mid-refund? I know this code is to pull the bond, but do you think we need to store the bond amount for each refund request?

mrice32 (approved):
LGTM! A few minor comments

contracts/HubPool.sol
chrismaree: this is all true! would you prefer it to simply have one number that is set within the contract and we don't pull anything from the store? my logic was that the store is a _reasnoble_ bond size that can be scaled. we can easily just define one number in this contract that sets the bond for everything

contracts/HubPool.sol
chrismaree: just dont change it if there is a pending refund I guess.

contracts/HubPool.sol
chrismaree: OK I've changed this. we now have a `bondAmount` that the owner can set. much cleaner

nicholaspai: agreed with this, its cleaner and in the same spirit of the UMA final fee

PR #11, opened 2022-01-27
feat(github-actions): Add hardhat test to CI

chrismaree:

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,14 @@
+name: CI
+on: [push]
+jobs:
+  build:
+    name: Test
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: borales/actions-yarn@v2.3.0
+        with:
+          cmd: install
+      - uses: borales/actions-yarn@v2.3.0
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -0,0 +1,14 @@
...
+jobs:
+  build:
+    name: Test
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: borales/actions-yarn@v2.3.0
+        with:
+          cmd: install
+      - uses: borales/actions-yarn@v2.3.0
```
mrice32, line 12: What is this image? Is it official?

mrice32 (approved):
LGTM!

.github/workflows/test.yml
chrismaree: it's a wrapper of the official package (this: https://github.com/actions/setup-node) to work with yarn. we can easily use the default one like this:
```
steps:
- uses: actions/checkout@v2
- uses: actions/setup-node@v2
  with:
    node-version: '14'
- run: yarn install
- run: yarn test
```

but from my testing this one runs faster

PR #12, opened 2022-01-27
feat(github-actions): Add Linter to CI

chrismaree:


.github/workflows/lint.yml
```diff
@@ -0,0 +1,14 @@
+name: CI
```
mrice32, line 1: Is this file a dup from the other PR?

mrice32 (approved):
LGTM!

.github/workflows/lint.yml
chrismaree: nope. it has a different name and runs a diffrent command.

the reason they are separate files it shows up as separate jobs. 

I think you can merge them to achieve the same thing but it just seemed a bit cleaner to have separate files.

PR #13, opened 2022-01-28
feat(HubPool): Second iteration on init, dispute and execute relayer refund bundles

chrismaree:
This PR adds the second iteration of `initiateRelayerRefund`, `disputeRelayerRefund` and `executeRelayerRefund`.

### DIFF
--- a/contracts/HubPool.sol
+++ b/contracts/HubPool.sol
@@ -79,7 +79,7 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
     );
     event WhitelistRoute(address originToken, uint256 destinationChainId, address destinationToken);
 
-    event RelayerRefundRequested(
+    event InitiateRefundRequested(
         uint64 indexed relayerRefundId,
         uint64 requestExpirationTimestamp,
         uint64 poolRebalanceLeafCount,
@@ -88,12 +88,7 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
         bytes32 destinationDistributionRoot,
         address indexed proposer
     );
-    event RelayerRefundExecuted(
-        uint256 relayerRefundId,
-        MerkleLib.PoolRebalance poolRebalance,
-        bytes32[] proof,
-        address caller
-    );
+    event RelayerRefundExecuted(uint256 relayerRefundId, MerkleLib.PoolRebalance poolRebalance, address caller);
 
     event RelayerRefundDisputed(uint256 relayerRefundId, address disputer);
 
@@ -226,19 +221,19 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
         );
 
         uint64 requestExpirationTimestamp = uint64(getCurrentTime() + refundProposalLiveness);
-        uint256 relayerRefundId = relayerRefundRequests.length;
-        relayerRefundRequests.push();
-        relayerRefundRequests[relayerRefundId].requestExpirationTimestamp = requestExpirationTimestamp;
-        relayerRefundRequests[relayerRefundId].unclaimedPoolRebalanceLeafs = poolRebalanceLeafCount;
-        relayerRefundRequests[relayerRefundId].poolRebalanceRoot = poolRebalanceRoot;
-        relayerRefundRequests[relayerRefundId].destinationDistributionRoot = destinationDistributionRoot;
-        relayerRefundRequests[relayerRefundId].proposer = msg.sender;
+
+        RelayerRefundRequest storage relayerRefundRequest = relayerRefundRequests.push();
+        relayerRefundRequest.requestExpirationTimestamp = requestExpirationTimestamp;
+        relayerRefundRequest.unclaimedPoolRebalanceLeafs = poolRebalanceLeafCount;
+        relayerRefundRequest.poolRebalanceRoot = poolRebalanceRoot;
+        relayerRefundRequest.destinationDistributionRoot = destinationDistributionRoot;
+        relayerRefundRequest.proposer = msg.sender;
 
         // Pull bondAmount of bondToken from the caller.
         bondToken.safeTransferFrom(msg.sender, address(this), bondAmount);
 
-        emit RelayerRefundRequested(
-            uint64(relayerRefundId),
+        emit InitiateRefundRequested(
+            uint64(relayerRefundRequests.length - 1),
             requestExpirationTimestamp,
             poolRebalanceLeafCount,
             bundleEvaluationBlockNumbers,
@@ -280,7 +275,7 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
         // in a later PR.
         // TODO: modify the associated utilized and pending reserves for each token sent.
 
-        emit RelayerRefundExecuted(relayerRefundRequestId, poolRebalance, proof, msg.sender);
+        emit RelayerRefundExecuted(relayerRefundRequestId, poolRebalance, msg.sender);
     }
 
     function disputeRelayerRefund(uint256 relayerRefundRequestId) public {
--- a/contracts/HubPool.sol
+++ b/contracts/HubPool.sol
@@ -57,7 +57,7 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
     // The computed bond amount as the UMA Store's final fee multiplied by the bondTokenFinalFeeMultiplier.
     uint256 public bondAmount;
 
-    // Each refund proposal must stay in liveness for at least this long until it can be considered finalized. It can
+    // Each refund proposal must stay in liveness for this period of time before it can be considered finalized. It can
     // be disputed only during this period of time.
     uint64 public refundProposalLiveness;
 
--- a/contracts/HubPool.sol
+++ b/contracts/HubPool.sol
@@ -31,12 +31,6 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
         bool isEnabled;
     }
 
-    enum RefundRequestStatus {
-        Pending, // Request is in liveness and waiting
-        Finalized,
-        Disputed
-    }
-
     struct RelayerRefundRequest {
         uint64 requestExpirationTimestamp;
         uint64 unclaimedPoolRebalanceLeafs;
--- a/contracts/HubPool.sol
+++ b/contracts/HubPool.sol
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: GPL-3.0-only
 pragma solidity ^0.8.0;
 
+import "./MerkleLib.sol";
+
 import "@uma/core/contracts/common/implementation/Testable.sol";
 import "@uma/core/contracts/common/implementation/Lockable.sol";
 import "@uma/core/contracts/common/implementation/MultiCaller.sol";
@@ -30,15 +32,19 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
     }
 
     enum RefundRequestStatus {
-        Pending,
+        Pending, // Request is in liveness and waiting
         Finalized,
         Disputed
     }
 
     struct RelayerRefundRequest {
-        bytes32 poolRebalanceProof;
-        bytes32 destinationDistributionProof;
-        RefundRequestStatus status;
+        uint64 requestExpirationTimestamp;
+        uint64 unclaimedPoolRebalanceLeafs;
+        bytes32 poolRebalanceRoot;
+        bytes32 destinationDistributionRoot;
+        mapping(uint256 => uint256) claimedBitMap;
+        address proposer;
+        bool proposerBondRepaid;
     }
 
     RelayerRefundRequest[] public relayerRefundRequests;
@@ -57,8 +63,9 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
     // The computed bond amount as the UMA Store's final fee multiplied by the bondTokenFinalFeeMultiplier.
     uint256 public bondAmount;
 
-    // There should only ever be one refund proposal pending at any point in time. This bool toggles accordingly.
-    bool public currentPendingRefundProposal = false;
+    // Each refund proposal must stay in liveness for at least this long until it can be considered finalized. It can
+    // be disputed only during this period of time.
+    uint64 public refundProposalLiveness;
 
     event BondAmountSet(uint64 newBondMultiplier);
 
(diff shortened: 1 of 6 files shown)


contracts/HubPool.sol
```diff
@@ -29,16 +31,14 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
         bool isEnabled;
     }
 
-    enum RefundRequestStatus {
-        Pending,
-        Finalized,
-        Disputed
-    }
-
     struct RelayerRefundRequest {
```
chrismaree, line 34: this structure changed somewhat.

contracts/HubPool.sol
```diff
@@ -201,48 +214,101 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
...
 
+        uint64 requestExpirationTimestamp = uint64(getCurrentTime() + refundProposalLiveness);
+        uint256 relayerRefundId = relayerRefundRequests.length;
+        relayerRefundRequests.push();
+        relayerRefundRequests[relayerRefundId].requestExpirationTimestamp = requestExpirationTimestamp;
+        relayerRefundRequests[relayerRefundId].unclaimedPoolRebalanceLeafs = poolRebalanceLeafCount;
+        relayerRefundRequests[relayerRefundId].poolRebalanceRoot = poolRebalanceRoot;
+        relayerRefundRequests[relayerRefundId].destinationDistributionRoot = destinationDistributionRoot;
+        relayerRefundRequests[relayerRefundId].proposer = msg.sender;
+
```
chrismaree, line 236: unfortunately this is the syntax we need to use because this struct contains a mapping for `mapping(uint256 => uint256) claimedBitMap;`. doing it like this leaves that mapping as uninitialized, giving it an empty mapping.

contracts/MerkleLib.sol
```diff
@@ -57,7 +57,7 @@ library MerkleLib {
      * @param repayment the repayment struct.
      * @param proof the merkle proof.
      */
-    function verifyRepayment(
+    function verifyPoolRebalance(
```
chrismaree, line 60: renamed these to be more consistant.

test/utils.ts
```diff
@@ -1,18 +1,24 @@
...
 export interface SignerWithAddress extends Signer {
   address: string;
 }
 
-export async function getContractFactory(name: string, signer: SignerWithAddress) {
+export async function getContractFactory(name: string, signer: SignerWithAddress): Promise<ContractFactory> {
+  console.log("getting", name);
   try {
     // Try fetch from the local ethers factory from HRE. If this exists then the contract is in this package.
+    if (name == "HubPool") {
```
chrismaree, line 12: if we are deploying a `HubPool` then add the library linking

contracts/HubPool.sol
```diff
@@ -201,48 +214,101 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
...
         );
 
+        uint64 requestExpirationTimestamp = uint64(getCurrentTime() + refundProposalLiveness);
+        uint256 relayerRefundId = relayerRefundRequests.length;
+        relayerRefundRequests.push();
+        relayerRefundRequests[relayerRefundId].requestExpirationTimestamp = requestExpirationTimestamp;
+        relayerRefundRequests[relayerRefundId].unclaimedPoolRebalanceLeafs = poolRebalanceLeafCount;
+        relayerRefundRequests[relayerRefundId].poolRebalanceRoot = poolRebalanceRoot;
+        relayerRefundRequests[relayerRefundId].destinationDistributionRoot = destinationDistributionRoot;
+        relayerRefundRequests[relayerRefundId].proposer = msg.sender;
```
mrice32, line 235: nit: push() returns the storage reference, so you can do:
```
MyStruct storage myStruct = array.push();
myStruct.someField = 1;
myStruct.otherField = 2;
```

This makes the syntax a bit cleaner and will save us from having to index into the array multiple times.

contracts/HubPool.sol
```diff
@@ -29,16 +31,14 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
...
-        bytes32 poolRebalanceProof;
-        bytes32 destinationDistributionProof;
-        RefundRequestStatus status;
+        uint64 requestExpirationTimestamp;
+        uint64 unclaimedPoolRebalanceLeafs;
+        bytes32 poolRebalanceRoot;
+        bytes32 destinationDistributionRoot;
+        mapping(uint256 => uint256) claimedBitMap;
+        address proposer;
+        bool proposerBondRepaid;
```
mrice32, line 41: Might not be something we want to do in this initial PR, but have you considered doing something similar to what we did in v1, where we hash all of this and just store the hash to verify the calldata that's passed in on later calls to reduce storage costs?

contracts/HubPool.sol
```diff
@@ -29,16 +31,14 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
...
     }
 
     RelayerRefundRequest[] public relayerRefundRequests;
```
mrice32, line 44: Do you think it's possible to somehow avoid storing all of these since only one is active at a time? We could reuse the same storage slot each time and always know if the prev one has been closed out when starting the next one because the storage slot isn't zero'd out. Note: only problem I see with this approach is that it might not work well with the mapping, since you can't really close out/delete a mapping.

mrice32 (commented):
This is awesome! I have some high level comments, but this is super cool so far.

contracts/HubPool.sol
```diff
@@ -79,20 +80,32 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
...
-        bytes32 indexed destinationDistributionProof,
-        address proposer
+        bytes32 poolRebalanceRoot,
+        bytes32 destinationDistributionRoot,
+        address indexed proposer
     );
+    event RelayerRefundExecuted(
+        uint256 relayerRefundId,
+        MerkleLib.PoolRebalance poolRebalance,
+        bytes32[] proof,
```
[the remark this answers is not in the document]
chrismaree: ye, that's a good point. we dont need it off-chain. I was just emitting everything I could and not really considering this. as a more meta point, we should generally consider the impact of emitting more complex structures vs less complex structures in events. is this largely impactful? I did not think it was.

contracts/HubPool.sol
```diff
@@ -201,48 +214,101 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
...
+        relayerRefundRequests[relayerRefundId].poolRebalanceRoot = poolRebalanceRoot;
+        relayerRefundRequests[relayerRefundId].destinationDistributionRoot = destinationDistributionRoot;
+        relayerRefundRequests[relayerRefundId].proposer = msg.sender;
+
         // Pull bondAmount of bondToken from the caller.
         bondToken.safeTransferFrom(msg.sender, address(this), bondAmount);
 
         emit RelayerRefundRequested(
-            uint64(relayerRefundRequests.length - 1), // Index of the relayerRefundRequest within the array.
+            uint64(relayerRefundId),
```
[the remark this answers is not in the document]
chrismaree: yup yup. good point. will rename.

contracts/HubPool.sol
```diff
@@ -201,48 +214,101 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
...
+        // Set the leafId in the claimed bitmap.
+        MerkleLib.setClaimed(relayerRefund.claimedBitMap, poolRebalance.leafId);
+
+        // Decrement the unclaimedPoolRebalanceLeafs.
+        relayerRefund.unclaimedPoolRebalanceLeafs--;
+
+        // Transfer the bondAmount to back to the proposer, if this was not done before for this refund bundle.
+        if (!relayerRefund.proposerBondRepaid) {
+            relayerRefund.proposerBondRepaid = true;
+            bondToken.safeTransfer(relayerRefund.proposer, bondAmount);
```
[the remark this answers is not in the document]
chrismaree: we assume it does not. I think that's fair assumption. we should pause the bridge if changing this number.

contracts/HubPool.sol
```diff
@@ -201,48 +214,101 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
...
-        relayerRefundRequests.push(
-            RelayerRefundRequest({
-                poolRebalanceProof: poolRebalanceProof,
-                destinationDistributionProof: destinationDistributionProof,
-                status: RefundRequestStatus.Pending
-            })
+        // The most recent refund proposal must be fully claimed before the next relayer refund bundle is initiated.
+        require(
+            relayerRefundRequests.length == 0 || // If this is the first initiated relay refund.
+                relayerRefundRequests[getNumberOfRelayRefunds()].unclaimedPoolRebalanceLeafs == 0,
```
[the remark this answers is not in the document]
chrismaree: well we are not storing the number of relayer refunds, no? we are simply setting this (as the number of leaves) at the beginning. in my mind the big advantage with doing it this way is we know with certainty that there is no pending unclaimed leafs before initialising the next bundle of leafs. this is really critical as you can have a situation with inconsistant state on L1 vs what the bot sees if we do this optimistically (i.e I have not seen the final executed leafs for a set of repayments and therefore dispute the next bundle that comes in as I'm not up too date with master).

obviously, this is not that important if we design the bots well to avoid this but it just seems consistant to store this, with a marginally small gas cost.

put another way, storing this data and having this extra computation is likely less than 10k extra gas per relay refund in total. sending to one L2 costs at least 100k gas so this cost is small when compared to the other costs.

contracts/HubPool.sol
chrismaree: we can do that but we can only do it with the static data, right?

contracts/HubPool.sol
chrismaree: we can do that. your goal is we'll get a refund if we re-use the same storage slots between refunds, right?

chrismaree: one option would to simply store one instance of `RelayerRefundRequest` within the hubPool. it tracks the exact variables we have in it now, but it's not within an array. When the `unclaimedPoolRebalanceLeafs` hits 0 and a data worker calls `initiateRelayerRefund` again we simply delete this instance of this struct and re-set it. this way we only ever have the storage of _one_ instance of this data and will net many of the refunds. I'll give this a go and see how it looks.

chrismaree: I tried this out and it does not really work as you cant delete/reset the mapping in the way we'd need to to work. I tried creating this basic test:
```
pragma solidity ^0.8.0;

contract StructTest {
    struct RelayerRefundRequest {
        uint64 requestExpirationTimestamp;
        uint64 unclaimedPoolRebalanceLeafs;
        bytes32 poolRebalanceRoot;
        mapping(uint256 => uint256) claimedBitMap;
        address proposer;
        bool proposerBondRepaid;
    }

    RelayerRefundRequest public activeRelayerRefundRequest;

    function setValue() public {
        activeRelayerRefundRequest.proposer = msg.sender;
        activeRelayerRefundRequest.claimedBitMap[0] = 10;
    }

    function getClaim(uint256 index) public view returns (uint256) {
        return activeRelayerRefundRequest.claimedBitMap[index];
    }

    function resetStruct() public {
        delete activeRelayerRefundRequest;
    }
}
```

The delete call worked fine at resetting the `proposer` but after deleting the instance of `RelayerRefundRequest` the `claimedBitMap` was still initialised on the value set in the `setValue` call.

chrismaree: long story short you cant use `delete` on complex data types and so cant reset it. this means we either need:
1) to create a 3d mapping where we index some information to spectate out `claimedBitMap` for each proposal 2) do an array like I originally proposed 3) move away from a claimedBitMap to some other structure for tracking claims. I personally think leaving it like it is now is probs the cleanest and cheapest option.

mrice32: @chrismaree as discussed IRL, I think we landed on making the bitmap a single integer and limit us to supporting 256 chains.

contracts/HubPool.sol
mrice32: Agreed. As discussed IRL, the savings may not make this worthwhile. It would save us around 15-20k gas, but require each subsequent proposal to pass in the previous proposal to prove that it was completed.

contracts/HubPool.sol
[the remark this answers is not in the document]
mrice32: Would pause-unpause work well within a governance process? Can we game this out a little bit? Could we make the setBond method detect whether there is an active, unresolved proposal to guarantee this doesn't result in broken accounting/lost funds?

contracts/HubPool.sol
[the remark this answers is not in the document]
mrice32: I agree with your broader point around emitting structs. Etherscan and other offchain parsers (like HAL) don't seem to be able to read this. I think we should generally avoid putting important information in structs within events.

contracts/HubPool.sol
chrismaree: That's right. I've implemented this in a separate PR to try minimise the changes here: https://github.com/across-protocol/across-smart-contracts-v2/pull/14

contracts/HubPool.sol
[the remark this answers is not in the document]
chrismaree: ye, that can work! I can set that restriction on this method. I'll leave this for a separate PR.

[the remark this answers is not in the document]
chrismaree: I've added this in this PR: https://github.com/across-protocol/across-smart-contracts-v2/pull/14

contracts/HubPool.sol
```diff
@@ -29,16 +31,14 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
...
-        Finalized,
-        Disputed
-    }
-
     struct RelayerRefundRequest {
-        bytes32 poolRebalanceProof;
-        bytes32 destinationDistributionProof;
-        RefundRequestStatus status;
+        uint64 requestExpirationTimestamp;
+        uint64 unclaimedPoolRebalanceLeafs;
```
nicholaspai, line 36: nit: change this name to end with `LeafCounts` instead of `Leafs`. I was a bit confused because without "counts" i expected this to be an array or a bitmap

contracts/HubPool.sol
nicholaspai: Yeah too much complexity for the gas saved

contracts/HubPool.sol
```diff
@@ -201,48 +209,101 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
...
-        address[] memory l1TokenAddress,
-        uint256[] memory accumulatedLpFees,
-        uint256[] memory netSendAmounts,
-        bytes32[] memory inclusionProof
-    ) public {}
+        MerkleLib.PoolRebalance memory poolRebalance,
+        bytes32[] memory proof
+    ) public {
+        RelayerRefundRequest storage relayerRefund = relayerRefundRequests[relayerRefundRequestId];
+        require(getCurrentTime() > relayerRefund.requestExpirationTimestamp, "Not passed liveness");
```
nicholaspai, line 252: nit: should this be >= ?

contracts/HubPool.sol
```diff
@@ -201,48 +209,101 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
...
+        // in a later PR.
+        // TODO: modify the associated utilized and pending reserves for each token sent.
+
+        emit RelayerRefundExecuted(relayerRefundRequestId, poolRebalance, msg.sender);
+    }
+
+    function disputeRelayerRefund(uint256 relayerRefundRequestId) public {
+        RelayerRefundRequest storage relayerRefund = relayerRefundRequests[relayerRefundRequestId];
+        require(
+            getCurrentTime() > relayerRefundRequests[relayerRefundRequestId].requestExpirationTimestamp,
```
nicholaspai, line 284: nit: same as as above, should this be >=

contracts/HubPool.sol
```diff
@@ -201,48 +209,101 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
...
+        // Delete the last element in the relayerRefundRequests array. This acts to throw out the request.
+        emit RelayerRefundDisputed(relayerRefundRequestId, msg.sender);
+
+        relayerRefundRequests.pop();
+
+        // TODO: pull bonds. request price from OO.
+    }
+
+    function getNumberOfRelayRefunds() public view returns (uint256) {
+        if (relayerRefundRequests.length == 0) return 0;
```
nicholaspai, line 297: pretty sure you can just `return relayerRefundRequests.length` here and the current implementation is incorrect

contracts/HubPool.sol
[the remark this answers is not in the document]
mrice32: That seems reasonable, and I think we should probably just not worry about this until we've nailed down how we're storing everything, as that probably changes a lot of the gas math here.

contracts/HubPool.sol
chrismaree: sure.

mrice32 (approved):
LGTM once @nicholaspai's comments are resolved.

PR #14, opened 2022-01-31
feat: hubpool third iteration. Add dispute logic and refine bitmap structure.

chrismaree:


contracts/BridgeAdmin.sol
```diff
@@ -1,13 +0,0 @@
-// SPDX-License-Identifier: GPL-3.0-only
```
chrismaree, line 1: I removed this for now...we dont have a definition of what this requires just yet and I thought it cleaner to directly add the finder to the hub.

contracts/HubPool.sol
```diff
@@ -31,32 +34,36 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
...
     }
 
-    struct RelayerRefundRequest {
+    struct RefundRequest {
         uint64 requestExpirationTimestamp;
-        uint64 unclaimedPoolRebalanceLeafs;
+        uint64 unclaimedPoolRebalanceLeafCount;
         bytes32 poolRebalanceRoot;
         bytes32 destinationDistributionRoot;
-        mapping(uint256 => uint256) claimedBitMap;
```
chrismaree, line 39: this is an important change: we use a 1D bitMap after this PR, thereby reducing storage requirements and enabling us to re-use state between each proposal. tradeoff is we can only have up to 256 different chains supported, but this is likely fine :)

contracts/HubPool.sol
```diff
@@ -80,41 +87,50 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
     event WhitelistRoute(address originToken, uint256 destinationChainId, address destinationToken);
 
     event InitiateRefundRequested(
-        uint64 indexed relayerRefundId,
         uint64 requestExpirationTimestamp,
-        uint64 poolRebalanceLeafCount,
+        uint64 unclaimedPoolRebalanceLeafCount,
```
chrismaree, line 91: variable name changed for @nicholaspai

contracts/HubPool.sol
```diff
@@ -248,25 +260,24 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
         MerkleLib.PoolRebalance memory poolRebalance,
         bytes32[] memory proof
     ) public {
-        RelayerRefundRequest storage relayerRefund = relayerRefundRequests[relayerRefundRequestId];
-        require(getCurrentTime() > relayerRefund.requestExpirationTimestamp, "Not passed liveness");
+        require(getCurrentTime() >= refundRequest.requestExpirationTimestamp, "Not passed liveness");
 
         // Verify the leafId in the poolRebalance has not yet been claimed.
-        require(!MerkleLib.isClaimed(relayerRefund.claimedBitMap, poolRebalance.leafId), "Already claimed");
```
chrismaree, line 255: we use a separate `MerkleLib` method that I added to deal with 1D bit maps.

chrismaree: I will add unit tests for this once @mrice32 's MerkleLib PR has been merged.

contracts/HubPool.sol
```diff
@@ -278,24 +289,85 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
...
 
-        // TODO: pull bonds. request price from OO.
+        // Finally, delete the state pertaining to the active refundRequest.
+        delete refundRequest;
     }
 
-    function getNumberOfRelayRefunds() public view returns (uint256) {
-        if (relayerRefundRequests.length == 0) return 0;
-        return relayerRefundRequests.length - 1;
+    function _getRefundProposalAncillaryData() public view returns (bytes memory ancillaryData) {
```
chrismaree, line 347: this was my initial pass at constructing the ancillary data. curious what you guys think & if we need more/less in this.

contracts/MerkleLib.sol
```diff
@@ -105,4 +106,23 @@ library MerkleLib {
...
         claimedBitMap[claimedWordIndex] = claimedBitMap[claimedWordIndex] | (1 << claimedBitIndex);
     }
+
+    /**
+     * @notice Tests whether a claim is contained within a 1D claimedBitMap mapping.
+     * @param claimedBitMap a simple uint256 value, encoding a 1D bitmap.
+     * @param index the index to check in the bitmap.
+     \* @return bool indicating if the index within the claimedBitMap has been marked as claimed.
+     */
+    function isClaimed1D(uint256 claimedBitMap, uint256 index) public pure returns (bool) {
```
chrismaree, line 116: these are the new merkleLib methods I added. v simple and adapted from the original ones to take in a `uint256 claimedBitMap` rather than `mapping(uint256 => uint256) storage claimedBitMap`

test/Uma.Fixture.ts
```diff
@@ -0,0 +1,33 @@
+import { getContractFactory, utf8ToHex } from "./utils";
```
chrismaree, line 1: ~30 lines to deploy the whole UMA ecosystem <3. I love clean repos.

contracts/HubPool.sol
nicholaspai: yeah this seems reasonable until v3 :D

contracts/HubPool.sol
```diff
@@ -31,32 +34,36 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
...
+    RefundRequest public refundRequest;
 
     // Whitelist of origin token to destination token routings to be used by off-chain agents.
     mapping(address => mapping(uint256 => address)) public whitelistedRoutes;
 
     mapping(address => LPToken) public lpTokens; // Mapping of L1TokenAddress to the associated LPToken.
 
-    // Address of L1Weth. Enable LPs to deposit/receive ETH, if they choose, when adding/removing liquidity.
-    WETH9Like public l1Weth;
+    FinderInterface finder;
```
nicholaspai, line 54: nit: make `public`

contracts/HubPool.sol
```diff
@@ -80,41 +87,50 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
...
 
     function setBondToken(address newBondToken) public onlyOwner {
+        require(refundRequest.unclaimedPoolRebalanceLeafCount == 0, "Cant set during active request");
         bondToken = IERC20(newBondToken);
         emit BondTokenSet(newBondToken);
     }
 
-    function setBondTokenFinalFeeMultiplier(uint64 newBondAmount) public onlyOwner {
+    function setBondAmount(uint64 newBondAmount) public onlyOwner {
+        require(refundRequest.unclaimedPoolRebalanceLeafCount == 0, "Cant set during active request");
```
nicholaspai, line 133: might want to make this a modifier if you re-use it

contracts/HubPool.sol
```diff
@@ -214,26 +230,22 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
         bytes32 destinationDistributionRoot
     ) public {
         // The most recent refund proposal must be fully claimed before the next relayer refund bundle is initiated.
-        require(
-            relayerRefundRequests.length == 0 || // If this is the first initiated relay refund.
-                relayerRefundRequests[getNumberOfRelayRefunds()].unclaimedPoolRebalanceLeafs == 0,
-            "Last bundle has unclaimed leafs"
-        );
+        require(refundRequest.unclaimedPoolRebalanceLeafCount == 0, "Last bundle has unclaimed leafs");
```
nicholaspai, line 233: could reuse the modifier suggestion from previous comment

contracts/HubPool.sol
nicholaspai: these seem reasonable. I'm thinking about ways we can reduce the amount of calls to `appendKeyValue...` because these functions are expensive I believe

contracts/HubPool.sol
nicholaspai: also, is there any reason not to combine `setBondToken` and `setBondAmount` into `setBond(address token, uint256 amount)`?

contracts/HubPool.sol
```diff
@@ -214,26 +230,22 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
...
-        relayerRefundRequest.poolRebalanceRoot = poolRebalanceRoot;
-        relayerRefundRequest.destinationDistributionRoot = destinationDistributionRoot;
-        relayerRefundRequest.proposer = msg.sender;
+        delete refundRequest; // Remove the existing information relating to the previous relayer refund request.
+
+        refundRequest.requestExpirationTimestamp = requestExpirationTimestamp;
+        refundRequest.unclaimedPoolRebalanceLeafCount = poolRebalanceLeafCount;
+        refundRequest.poolRebalanceRoot = poolRebalanceRoot;
+        refundRequest.destinationDistributionRoot = destinationDistributionRoot;
+        refundRequest.proposer = msg.sender;
```
nicholaspai, line 243: All things being equal I prefer using key-value syntax to set `refundRequest` here because it means that the developer is less likely to forget to set a parameter. Are there gas savings to setting each param individually and leaving default values as-is for keys like `proposerBondRepaid`?

contracts/HubPool.sol
```diff
@@ -214,26 +230,22 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
...
-        RelayerRefundRequest storage relayerRefundRequest = relayerRefundRequests.push();
-        relayerRefundRequest.requestExpirationTimestamp = requestExpirationTimestamp;
-        relayerRefundRequest.unclaimedPoolRebalanceLeafs = poolRebalanceLeafCount;
-        relayerRefundRequest.poolRebalanceRoot = poolRebalanceRoot;
-        relayerRefundRequest.destinationDistributionRoot = destinationDistributionRoot;
-        relayerRefundRequest.proposer = msg.sender;
+        delete refundRequest; // Remove the existing information relating to the previous relayer refund request.
+
+        refundRequest.requestExpirationTimestamp = requestExpirationTimestamp;
+        refundRequest.unclaimedPoolRebalanceLeafCount = poolRebalanceLeafCount;
```
nicholaspai, line 240: Do we need to check that `unclaimedPoolRebalanceLeafCount > 0`? I only ask because the condition `unclaimedPoolRebalanceLeafCount == 0` is an important one in this contract but I also don't think its exploitable to set it to 0.

nicholaspai (commented):
I haven't checked all the tests out carefully yet but first pass the contract logic makes sense!

contracts/HubPool.sol
chrismaree: done. changed `identifier` as well.

contracts/HubPool.sol
chrismaree: so that would replace the `delete` call plus this assigmeneet with this:

```solidity
// Remove the existing information relating to the previous relayer refund request and set to the new info.
        refundRequest = RefundRequest({
            requestExpirationTimestamp: requestExpirationTimestamp,
            unclaimedPoolRebalanceLeafCount: poolRebalanceLeafCount,
            poolRebalanceRoot: poolRebalanceRoot,
            destinationDistributionRoot: destinationDistributionRoot,
            claimedBitMap: 0,
            proposer: msg.sender,
            proposerBondRepaid: false
        });
```

I tried out the gas costs:
using the old assignment + delete that I had cost: 155863
using the syntax above cost: 156041

I think the reason the original syntax is cheaper is you are deleting storage and then not re-asigining some things, such as the `claimedBitMap` and the `proposerBondRepaid`. for these variables the default (post delete values) are the correct in this context.

As a result, I'm going leave it with the original implementation, unless you have anything else I can try that is syntactically better but still gas efficient.

contracts/HubPool.sol
chrismaree: ye, this is a good point and is actually something we should check. if the relayer provided a set of proofs with `poolRebalanceLeafCount` set to 0 then there is actually no L1->L2 actions that this bundle can do. they could be providing valid roots that do _nothing_. I agree that we should only process a bundle if it has _at least_ one valid L2 action. I'll add this require.

contracts/HubPool.sol
chrismaree: The only argument I could see against having this as a modifier would be that we can have more descriptive revert messages. I think we can find a more common one that covers both though. I'll revert for that.

Regarding the `setBond` that's a good point. no reason to not have just one method seeing both are restricted by the `onlyOwner`. I'll change this as well.

contracts/HubPool.sol
```diff
@@ -80,43 +85,51 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
...
         uint256[] bundleEvaluationBlockNumbers,
         bytes32 poolRebalanceRoot,
         bytes32 destinationDistributionRoot,
         address indexed proposer
     );
     event RelayerRefundExecuted(uint256 relayerRefundId, MerkleLib.PoolRebalance poolRebalance, address caller);
 
-    event RelayerRefundDisputed(uint256 relayerRefundId, address disputer);
+    event RelayerRefundDisputed(
+        address disputer,
```
mrice32, line 98: indexed?

contracts/HubPool.sol
```diff
@@ -80,43 +85,51 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
...
+        SkinnyOptimisticOracleInterface.Request ooPriceRequest,
+        RefundRequest refundRequest
```
mrice32, line 100: One minor comment on these: we might want to select out parts of these structs we want to emit to 1) reduce the data and 2) make etherscan better able to encode/decode the events rather than just showing bytes.

contracts/HubPool.sol
```diff
@@ -80,43 +85,51 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
...
      *                ADMIN FUNCTIONS                *
      *************************************************/
 
-    function setBondToken(address newBondToken) public onlyOwner {
-        bondToken = IERC20(newBondToken);
-        emit BondTokenSet(newBondToken);
-    }
-
-    function setBondTokenFinalFeeMultiplier(uint64 newBondAmount) public onlyOwner {
+    function setBond(IERC20 newBondToken, uint256 newBondAmount) public onlyOwner onlyIfNoActiveRequest {
```
mrice32, line 129: +1 for this modifier

contracts/HubPool.sol
```diff
@@ -212,28 +225,23 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
...
         bytes32 destinationDistributionRoot
-    ) public {
-        // The most recent refund proposal must be fully claimed before the next relayer refund bundle is initiated.
-        require(
-            relayerRefundRequests.length == 0 || // If this is the first initiated relay refund.
-                relayerRefundRequests[getNumberOfRelayRefunds()].unclaimedPoolRebalanceLeafs == 0,
-            "Last bundle has unclaimed leafs"
-        );
+    ) public onlyIfNoActiveRequest {
+        require(poolRebalanceLeafCount > 0, "Bundle must have at least 1 leaf");
```
mrice32, line 229: +1 good check. This will make us more spam resistant.

contracts/HubPool.sol
mrice32: I also think there are some annoying solidity inefficiencies with key value syntax. I think they create the entire struct in mem and then assign it rather than just assigning each value directly to storage. This increased mem usage might be why the gas costs are higher.

contracts/HubPool.sol
mrice32: Do we call this in the happy path? Or only on disputes? If only on disputes, we might not need to worry about gas quite as much (although it's still worth thinking about!!).

mrice32 (approved):
This looks awesome!! Only minor comments and responses to threads.

contracts/HubPool.sol
```diff
@@ -278,24 +285,85 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
...
-        if (relayerRefundRequests.length == 0) return 0;
-        return relayerRefundRequests.length - 1;
+    function _getRefundProposalAncillaryData() public view returns (bytes memory ancillaryData) {
+        ancillaryData = AncillaryData.appendKeyValueUint(
+            "",
+            "requestExpirationTimestamp",
+            refundRequest.requestExpirationTimestamp
+        );
+
+        ancillaryData = AncillaryData.appendKeyValueUint(
```
mrice32, line 350: If we do end up caring about hyper-optimizing gas usage here, we probably need to take a second look at the Uint method, since I don't think we optimized that last time around.

chrismaree: it only is called in the dispute case so I don't think it matters too much

contracts/HubPool.sol
chrismaree: it only happens in the dispute case so the gas does not matter too much I don't think.

contracts/HubPool.sol
nicholaspai: sounds good to me

contracts/HubPool.sol
nicholaspai: this is a good point and i'll change the same for SpokePool

nicholaspai (approved):
Thanks for responding to my comments, lgtm

contracts/MerkleLib.sol
```diff
@@ -105,4 +106,23 @@ library MerkleLib {
...
+        return claimedBitMap & (1 << index) > 0;
```
mrice32, line 117: nit: can you explain the differences in logic from the uniswap one?

I think their version of this code is:
```
        uint256 mask = (1 << index);
        return claimedBitMap & mask == mask;
```

Not trying to be nit-picky, this logic is just very dense, and I want to make sure we're super confident in the correctness.

contracts/HubPool.sol
```diff
@@ -31,39 +34,41 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
...
 
-    struct RelayerRefundRequest {
+    struct RefundRequest {
         uint64 requestExpirationTimestamp;
-        uint64 unclaimedPoolRebalanceLeafs;
+        uint64 unclaimedPoolRebalanceLeafCount;
         bytes32 poolRebalanceRoot;
         bytes32 destinationDistributionRoot;
-        mapping(uint256 => uint256) claimedBitMap;
+        uint256 claimedBitMap;
```
mrice32, line 42: I think this is a good change, but we've started making some assumptions like "there will never be more than X tokens or Y chains". I want to write these down somewhere. Should we go back to the interface doc and write these down? Or start a known issues and constraints doc to outline this sort of stuff? Thoughts?

I will go ahead and start such a doc, but let me know if you disagree.

contracts/HubPool.sol
```diff
@@ -31,39 +34,41 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
...
+    WETH9Like public l1Weth;
+
     // Each refund proposal must stay in liveness for this period of time before it can be considered finalized. It can
     // be disputed only during this period of time.
     uint64 public refundProposalLiveness;
 
-    event BondAmountSet(uint64 newBondMultiplier);
-
-    event BondTokenSet(address newBondMultiplier);
+    event BondSet(address newBondToken, uint256 newBondAmount);
```
mrice32, line 71: Should we index newBondToken?

contracts/HubPool.sol
chrismaree: I'm going to leave this for now and will try and see how much of a gas difference this makes.

contracts/MerkleLib.sol
chrismaree: going to add unit tests now on this and will verify the behaviour. can test out this other code example as well (note I created my implementation from modifying the uniswap code to be 1 dimensional :) )

contracts/HubPool.sol
chrismaree: ok great. ye, you right, we need to write this down! the interface doc is now somewhat out of date though in some regards. our options are to update the interface doc or start a new one.

in either case, I'm going to add some documentation on this.

contracts/MerkleLib.sol
chrismaree: both work equivalently. I reverted to your recommendation as it's easier to read.

contracts/HubPool.sol
```diff
@@ -80,43 +85,51 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
...
-        uint64 indexed relayerRefundId,
         uint64 requestExpirationTimestamp,
-        uint64 poolRebalanceLeafCount,
+        uint64 unclaimedPoolRebalanceLeafCount,
         uint256[] bundleEvaluationBlockNumbers,
         bytes32 poolRebalanceRoot,
         bytes32 destinationDistributionRoot,
         address indexed proposer
     );
     event RelayerRefundExecuted(uint256 relayerRefundId, MerkleLib.PoolRebalance poolRebalance, address caller);
```
mrice32, line 95: Can we enumerate the fields in poolRebalance? Also, do you think there are fields we can remove? Do we think these emissions are important for bot verification or anything?

contracts/HubPool.sol
```diff
@@ -278,24 +285,85 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
...
+            identifier,
+            uint32(getCurrentTime()),
+            _getRefundProposalAncillaryData(),
+            ooPriceRequest,
+            msg.sender,
+            address(this)
+        );
 
-        relayerRefundRequests.pop();
+        emit RelayerRefundDisputed(msg.sender, ooPriceRequest, refundRequest);
```
mrice32, line 337: Can we enumerate these structs for better parseability? Also might be nice to do so so we can drop fields that are unnecessary.

mrice32 (approved):
LGTM! Minor comments on events!

contracts/HubPool.sol
nicholaspai: i don't think it should change anything

contracts/HubPool.sol
chrismaree: will do this on a follow up.

PR #15, opened 2022-02-01
feat: add MerkleLib tests

mrice32:
This PR adds MerkleLib tests. To do this, it needed:
- A library to construct merkle trees from custom leaf types.
- A test contract to proxy calls to the lib and provide storage.

Note: the smart contract changes are only cosmetic, as there was a minor naming inconsistency.


hardhat.config.ts
```diff
@@ -6,6 +6,7 @@ import "@nomiclabs/hardhat-waffle";
 import "@typechain/hardhat";
 import "hardhat-gas-reporter";
 import "solidity-coverage";
+import "hardhat-deploy";
```
mrice32, line 9: To use fixtures, hardhat deploy was needed.

contracts/test/MerkleLibTest.sol
```diff
@@ -0,0 +1,35 @@
+// SPDX-License-Identifier: GPL-3.0-only
```
mrice32, line 1: Test contract

contracts/MerkleLib.sol
```diff
@@ -54,15 +54,15 @@ library MerkleLib {
     /**
      * @notice Verifies that a repayment is contained within a merkle root.
      * @param root the merkle root.
-     * @param repayment the repayment struct.
+     * @param rebalance the rebalance struct.
```
mrice32, line 57: Fixed a naming inconsistency

utils/MerkleTree.ts
```diff
@@ -0,0 +1,141 @@
+// This script provides some useful methods for building MerkleTrees. It is essentially the uniswap implementation
+// https://github.com/Uniswap/merkle-distributor/blob/master/src/merkle-tree.ts with some added convenience methods
```
mrice32, line 2: See this comment for the differences between this file and the one in protocol.

test/utils.ts
```diff
@@ -6,7 +6,7 @@ export interface SignerWithAddress extends Signer {
   address: string;
 }
 
-export async function getContractFactory(name: string, signer: SignerWithAddress) {
+export async function getContractFactory(name: string, signer: Signer) {
```
mrice32, line 9: I made these normal signers because it seemed like there was nothing in these methods that explicitly _required_ them to be `SignerWithAddress`.

test/MerkleLib.Fixture.ts
```diff
@@ -0,0 +1,17 @@
+import { getContractFactory } from "./utils";
+import hre from "hardhat";
+
+export const merkleLibFixture = hre.deployments.createFixture(async ({ deployments }) => {
```
mrice32, line 4: I wasn't sure if this was the type of fixtures we were wanting to use, but this is the type I'm familiar with (where hardhat keeps a snapshot of the fixture so it doesn't have to redeploy).

contracts/MerkleLib.sol
```diff
@@ -54,15 +54,15 @@ library MerkleLib {
...
      * @param proof the merkle proof.
      */
-    function verifyRepayment(
+    function verifyRebalance(
         bytes32 root,
-        PoolRebalance memory repayment,
+        PoolRebalance memory rebalance,
         bytes32[] memory proof
     ) public pure returns (bool) {
-        return MerkleProof.verify(proof, root, keccak256(abi.encode(repayment))) || true; // Run code but set to true.
```
mrice32, line 65: Note: I temporarily removed the `|| true` to prove that these tests were working. I can put it back before merging if helpful, but now that we have the MerkleTree ts library, we should be able to construct test data with relative ease. Let me know what y'all prefer!

nicholaspai: whatever is easiest!

hardhat.config.ts
nicholaspai: welcome back my old friend

nicholaspai (approved):

contracts/MerkleLib.sol
mrice32: I'll go ahead and remove. Feel free to add back if needed!

chrismaree: ye, let's avoid this part to force us to use correct trees, generated from this lib. the only place I touch these methods has not yet had any meaningful implementation added so this won't impact anything in the hubPool.

test/MerkleLib.Claims.ts
```diff
@@ -0,0 +1,29 @@
+import { expect } from "chai";
+import { merkleLibFixture } from "./MerkleLib.Fixture";
+import { Contract, BigNumber } from "ethers";
+
+describe("MerkleLib Claims", async function () {
+  let merkleLibTest: Contract;
```
chrismaree, line 6: nit: in the other test files we've done the declaration of these types of variables outside of the `describe` block.

test/MerkleLib.Fixture.ts
chrismaree: ye, so this is why you needed the `hardhat-deploy` package but @nicholaspai and my "fixtures" thus far have not. we simply created functions that did the deployment, which is not great. we should converge on one standard that does the same thing throughout the repo.

I think what you have here is better, though, as it will actually do the snapshotting and no re-deployment. I can refactor the other "fixtures" once this PR is in.

test/MerkleLib.Fixture.ts
```diff
@@ -0,0 +1,17 @@
...
+  await deployments.fixture();
+  const [signer] = await hre.ethers.getSigners();
+  const merkleLib = await (await getContractFactory("MerkleLib", signer)).deploy();
+  const merkleLibTest = await (
+    await hre.ethers.getContractFactory("MerkleLibTest", {
+      signer,
+      libraries: {
```
chrismaree, line 11: can this be linted to fewer lines?

test/MerkleLib.Fixture.ts
```diff
@@ -0,0 +1,17 @@
+import { getContractFactory } from "./utils";
+import hre from "hardhat";
+
+export const merkleLibFixture = hre.deployments.createFixture(async ({ deployments }) => {
+  await deployments.fixture();
+  const [signer] = await hre.ethers.getSigners();
+  const merkleLib = await (await getContractFactory("MerkleLib", signer)).deploy();
+  const merkleLibTest = await (
+    await hre.ethers.getContractFactory("MerkleLibTest", {
```
chrismaree, line 9: there is an argument to put this linking deployment within the `getContractFactory` from `utils` so you dont need to create your own ethers factory with `getContractFactory` and then this would be hidden from this fixture. what I describe is more similar to what was done in the hub pool tests. downside with that is the `utils` needs branching logic based on the contract name to do the associated library linking. curious what pattern you think is best and if it warrants changing what's in the `HubPool`'s fixture, [here](https://github.com/across-protocol/across-smart-contracts-v2/blob/a44eeb28839345fab140f80b2de7ad00cfd7968c/test/utils.ts?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L12)

test/MerkleLib.Claims.ts
```diff
@@ -0,0 +1,29 @@
...
+    await merkleLibTest.setClaimed(1499);
+    await merkleLibTest.setClaimed(1500);
+    await merkleLibTest.setClaimed(1501);
+    expect(await merkleLibTest.isClaimed(1499)).to.equal(true);
+    expect(await merkleLibTest.isClaimed(1500)).to.equal(true);
+    expect(await merkleLibTest.isClaimed(1501)).to.equal(true);
+    const claim1499 = BigNumber.from(2).pow(219);
+    const claim1500 = BigNumber.from(2).pow(220);
+    const claim1501 = BigNumber.from(2).pow(221);
+    expect(await merkleLibTest.claimedBitMap(5)).to.equal(claim1499.add(claim1500).add(claim1501));
```
chrismaree, line 27: nice

test/MerkleLib.Proofs.ts
```diff
@@ -0,0 +1,109 @@
...
+  refundAmounts: BigNumber[];
+}
+
+function randomBigNumber() {
+  return ethers.BigNumber.from(ethers.utils.randomBytes(31));
+}
+
+function randomAddress() {
+  return ethers.utils.hexlify(ethers.utils.randomBytes(20));
+}
```
chrismaree, line 31: should we put these in the `utils` file? we've been doing something similar to this for other tests with common utils.

test/MerkleLib.Proofs.ts
```diff
@@ -0,0 +1,109 @@
...
+
+describe("MerkleLib Proofs", async function () {
+  let merkleLibTest: Contract;
```
chrismaree, line 34: same nit about where this should be defined.

test/MerkleLib.Proofs.ts
```diff
@@ -0,0 +1,109 @@
...
+    const fragment = merkleLibTest.interface.fragments.find((fragment) => fragment.name === "verifyPoolRebalance");
+    const param = fragment!.inputs.find((input) => input.name === "rebalance");
+
+    const hashFn = (input: PoolRebalance) =>
+      ethers.utils.keccak256(ethers.utils.defaultAbiCoder.encode([param!], [input]));
+    const merkleTree = new MerkleTree<PoolRebalance>(poolRebalances, hashFn);
+
+    const root = merkleTree.getHexRoot();
+    const proof = merkleTree.getHexProof(poolRebalances[34]);
+    expect(await merkleLibTest.verifyPoolRebalance(root, poolRebalances[34], proof)).to.equal(true);
```
chrismaree, line 73: nice

chrismaree: please test the negative case as well. the test you have here only tests that a valid proof passes but not that an invalid proof returns `false`.

test/utils.ts
```diff
@@ -31,12 +31,12 @@ export const toBN = (num: string | number | BigNumber) => {
...
 
 export async function seedWallet(
-  walletToFund: SignerWithAddress,
+  walletToFund: Signer,
   tokens: Contract[],
   weth: Contract | undefined,
   amountToSeedWith: number | BigNumber
 ) {
-  for (const token of tokens) await token.mint(walletToFund.address, amountToSeedWith);
+  for (const token of tokens) await token.mint(await walletToFund.getAddress(), amountToSeedWith);
```
chrismaree, line 39: why this change?

test/MerkleLib.Fixture.ts
mrice32: Yeah, I think I might've left a comment on that PR (can't remember if I deleted it before sending). I don't love branching on individual contract names.

One compromise between the two approaches is have the second argument mirror the structure of the ethers method, `Signer | Options` where options could contain the _signer_ and the libraries. Then the user can still use the function, but they can pick the library deployment they want to link.

test/MerkleLib.Fixture.ts
mrice32: Are you suggesting changing the prettier settings?

test/MerkleLib.Claims.ts
mrice32: Okay, sounds good.

test/MerkleLib.Proofs.ts
mrice32: We can!

test/MerkleLib.Proofs.ts
mrice32: Ah, good point. Can add.

test/utils.ts
mrice32: SignerWithAddress is more restrictive than Signer. It's pretty easy to get the address from the Signer and the Signer type is much easier to extract from the ethers package, so I think we should prefer using it over the more specific and cumbersome SignerWithAddress when we're building generic library functions.

test/MerkleLib.Proofs.ts
mrice32: Will fix.

test/MerkleLib.Fixture.ts
mrice32: Sounds good. I think that's the right path to avoid this PR growing too large. I'm also happy to make this change as a follow-up.

test/MerkleLib.Fixture.ts
mrice32: So I made the change I mention above. PTAL and let me know what you think.

test/MerkleLib.Proofs.ts
mrice32: Done.

test/MerkleLib.Proofs.ts
mrice32: Done.

test/MerkleLib.Proofs.ts
mrice32: Done.

test/MerkleLib.Fixture.ts
chrismaree: just seems like it can be rolled into 
```javascript
...
signer,
libraries: { MerkleLib: merkleLib.address }
```

one one line.

chrismaree (approved):
LGTM! thanks for adding this

PR #17, opened 2022-02-02
improve: move all tests to hardhat fixtures

mrice32:
This PR is just a simple refactor that moves the existing test deployment logic to using hardhat fixtures, which is the recommended way to "snapshot" deployments in tests.

I ran an a/b test on these (and ensured that all conditions were the same, like contracts were already compiled, etc). We saw a pretty decent test speedup from 16s to 12s (25%-ish). This is pretty small, and as long as the repo and tests stay simple and small this will be insignificant, but still good to know that it didn't slow things down :).

Note: this also includes one or two `any`s that I saw an opportunity to replace with something more specific.


nicholaspai (approved):
This is awesome thanks so much for doing this

PR #18, opened 2022-02-02
improve: Destructure RelayData params in FilledRelay event

nicholaspai:
In the past, emitting a complex object in an event causes Etherscan difficulty in decoding the params. We should emit all neccessary params individually.

### DIFF
--- a/contracts/SpokePool.sol
+++ b/contracts/SpokePool.sol
@@ -77,8 +77,8 @@ abstract contract SpokePool is Testable, Lockable, MultiCaller {
     event FilledRelay(
         bytes32 indexed relayHash,
         uint256 totalRelayAmount,
-        uint256 newFilledAmount,
-        uint256 amountSentToRecipient,
+        uint256 totalFilledAmount,
+        uint256 fillAmount,
         uint256 indexed repaymentChain,
         uint256 originChainId,
         uint64 depositId,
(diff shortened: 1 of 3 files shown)


contracts/SpokePool.sol
```diff
@@ -44,7 +44,7 @@ abstract contract SpokePool is Testable, Lockable, MultiCaller {
     mapping(address => mapping(uint256 => bool)) public enabledDepositRoutes;
 
     struct RelayData {
-        address sender;
+        address depositor;
```
nicholaspai, line 47: Unrelated: naming this to `depositor` is more clear and better delineates between "sender" and "relayer"

mrice32: +1

chrismaree (approved):
does this change the gas cost of the emitting?

contracts/SpokePool.sol
```diff
@@ -72,15 +72,22 @@ abstract contract SpokePool is Testable, Lockable, MultiCaller {
...
         address recipient,
-        address indexed sender
+        address indexed depositor
     );
     event FilledRelay(
         bytes32 indexed relayHash,
+        uint256 totalRelayAmount,
         uint256 newFilledAmount,
-        uint256 indexed repaymentChain,
         uint256 amountSentToRecipient,
```
mrice32, line 81: Instead of reporting how much they sent to the recipient, do you think we should report how much was added to the filled amount? We have the newFilledAmount, but not the diff. Seems like it might be easier (from a bot perspective) to compute how much the relayer is owed from that amount rather than the amountSent. Thoughts?

mrice32 (approved):
Looks good!

contracts/SpokePool.sol
nicholaspai: Hmm I think you might be right. @chrismaree do you have instincts on this? The relayer would be able to compute `(1 + relayerFeePct) * fillIncreaseAmount`. Question though is whether they even would care about this new fill amount. Also, should we simply append the fill delta or replace the `amountSentToRecipient`?

nicholaspai:
> does this change the gas cost of the emitting?

Empirically it doesn't look like it changes anything @chrismaree

contracts/SpokePool.sol
mrice32: > The relayer would be able to compute `(1 + relayerFeePct) * fillIncreaseAmount`.

I may be wrong, but I think the formula for the repayment amount would be `(1 - lpFeePct) * fillIncreaseAmount`. Does that seem right to you?

mrice32 (approved):
LGTM once we resolve this thread: https://github.com/across-protocol/across-smart-contracts-v2/pull/18#discussion_r798090752.

contracts/SpokePool.sol
nicholaspai: I guess you could compute `amountSentToRecipient` by calculating `(1 - relayerFeePct - lpFeePct) * fillIncreaseAmount`

nicholaspai: I replaced `amountToSend` with `filllAmountPreFees` and renamed the param to `fillAmount`

mrice32 (approved):
LGTM!

PR #19, opened 2022-02-03
feat(HubPool): Add initial L1->L2 call plumbing

chrismaree:
- nit
- nit
- nit
- Update contracts/HubPool.sol
- review nit
- review nit
- feat(hubpool): Refactor hub pool to use 1D bitmap integer
- nit
- WIP
- review nit
- review nit
- nit
- nit
- nit
- nit
- nit
- nit
- WIP
- nit
- nit


contracts/HubPool.sol
```diff
@@ -270,19 +287,17 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
...
+        if (refundRequest.unclaimedPoolRebalanceLeafCount == 0)
             bondToken.safeTransfer(refundRequest.proposer, bondAmount);
-        }
 
-        // TODO call into canonical bridge to send PoolRebalance.netSendAmount for the associated
-        // PoolRebalance.tokenAddresses, to the target PoolRebalance.chainId. this will likely happen within a
-        // x_Messenger contract for each chain. these messengers will be registered in a separate process that will follow
-        // in a later PR.
+        _sendTokensToTargetChain(poolRebalance, refundRequestId);
+        _executeRelayerRefundOnTargetChain(poolRebalance, refundRequestId);
```
chrismaree, line 296: all the logic of sending tokens to the chains and sending the L1->L2 calls is wrapped into these two functions :)

contracts/HubPool.sol
```diff
@@ -270,19 +287,17 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
...
         refundRequest.unclaimedPoolRebalanceLeafCount--;
 
-        // Transfer the bondAmount to back to the proposer, if this was not done before for this refund bundle.
-        if (!refundRequest.proposerBondRepaid) {
-            refundRequest.proposerBondRepaid = true;
+        // Transfer the bondAmount to back to the proposer, if this the last executed leaf. Only sending this once all
+        // leafs have been executed acts to force the data worker to execute all bundles or they wont receive their bond.
+        if (refundRequest.unclaimedPoolRebalanceLeafCount == 0)
             bondToken.safeTransfer(refundRequest.proposer, bondAmount);
-        }
```
chrismaree, line 277: it's better to only pay the relayer back their bond one all bundles are executed.

contracts/chain-adapters/Optimism_Adapter.sol
```diff
@@ -0,0 +1,42 @@
+// SPDX-License-Identifier: AGPL-3.0-only
```
chrismaree, line 1: I tried to make this as simple as posible. I think it's pretty clean

contracts/HubPool.sol
```diff
@@ -126,12 +134,21 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
...
         bondAmount = newBondAmount;
         emit BondSet(address(newBondToken), newBondAmount);
     }
 
+    function setCrossChainContracts(
+        uint256 chainId,
+        AdapterInterface adapter,
+        address spokePool
+    ) public onlyOwner noActiveRequests {
+        require(address(crossChainContracts[chainId].adapter) == address(0), "Contract already set");
```
nicholaspai, line 148: Should we not allow the owner to reset addresses for upgradeability?

contracts/HubPool.sol
nicholaspai: On L2, what happens if the relayer refund method on the spoke pool is executed before tokens arrive there?

contracts/HubPool.sol
```diff
@@ -405,6 +421,37 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
...
+        for (uint32 i = 0; i < poolRebalance.tokenAddresses.length; i++) {
+            int256 amount = poolRebalance.netSendAmount[i];
+
+            // TODO: Checking the amount is greater than 0 is not sufficient. we need to build an external library that
+            // makes the decision on if there should be an L1->L2 token transfer. this should come in a later PR.
+            if (amount > 0) {
+                // Send the adapter all the tokens it needs to bridge. This should be refined later to remove the extra
+                // token transfer through the use of delegate call.
+                IERC20(poolRebalance.tokenAddresses[i]).safeTransfer(address(adapter), uint256(amount));
+                adapter.relayTokens(
```
nicholaspai, line 445: is there any way, or even an advantage to, potentially batching these calls?

nicholaspai (commented):
Really good start, left some comments and will re-review after you add tests and respond to comments

contracts/HubPool.sol
```diff
@@ -405,6 +421,37 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
                 .rawValue;
     }
 
+    function _executeRelayerRefundOnTargetChain(MerkleLib.PoolRebalance memory poolRebalance, uint256 refundRequestId)
```
mrice32, line 424: nit: would it be clearer just to take in the chainId if that's all we're using in the poolRebalance struct?

contracts/chain-adapters/Optimism_Adapter.sol
```diff
@@ -0,0 +1,44 @@
...
+    ) CrossDomainEnabled(_crossDomainMessenger) {
+        gasLimit = _gasLimit;
+        l1ERC20Bridge = IL1ERC20Bridge(_IL1ERC20Bridge);
+    }
+
+    function relayMessage(address target, bytes memory message) external payable override onlyOwner {
+        sendCrossDomainMessage(target, uint32(gasLimit), message);
+    }
+
+    //TODO: we should look into using delegate call as this current implementation assumes the caller
```
mrice32, line 34:
```suggestion
    // TODO: we should look into using delegate call as this current implementation assumes the caller
```

contracts/HubPool.sol
```diff
@@ -270,19 +287,18 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
...
-        // PoolRebalance.tokenAddresses, to the target PoolRebalance.chainId. this will likely happen within a
-        // x_Messenger contract for each chain. these messengers will be registered in a separate process that will follow
-        // in a later PR.
+        _sendTokensToTargetChain(poolRebalance, refundRequestId);
+        _executeRelayerRefundOnTargetChain(poolRebalance, refundRequestId);
+
         // TODO: modify the associated utilized and pending reserves for each token sent.
 
-        emit RelayerRefundExecuted(relayerRefundRequestId, poolRebalance, msg.sender);
+        emit RelayerRefundExecuted(refundRequestId, poolRebalance, msg.sender);
```
mrice32, line 301: 1. Can we enumerate the values in this struct so they're decodable on etherscan?
2. Do we need to emit all of this data? With the size of these structs and the number of times this method gets called, I could see these getting expensive. Do you think we can drop some or most of this struct leaving only the most important info?

contracts/chain-adapters/Optimism_Adapter.sol
```diff
@@ -0,0 +1,44 @@
...
+    // transfers the tokens first to this contract.
+    function relayTokens(
+        address l1Token,
+        address l2Token,
+        uint256 amount,
+        address to
+    ) external payable override onlyOwner {
+        l1ERC20Bridge.depositERC20To(l1Token, l2Token, to, amount, gasLimit, "0x");
```
mrice32, line 42: Is there any ETH/WETH weirdness sending tokens _to_ optimism? Do they take WETH?

contracts/HubPool.sol
```diff
@@ -405,6 +421,37 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
                 .rawValue;
     }
 
+    function _executeRelayerRefundOnTargetChain(MerkleLib.PoolRebalance memory poolRebalance, uint256 refundRequestId)
+        internal
+    {
+        AdapterInterface adapter = crossChainContracts[poolRebalance.chainId].adapter;
+        adapter.relayMessage(
+            address(adapter),
+            abi.encodeWithSignature("initializeRelayerRefund(address)", refundRequest.destinationDistributionRoot)
```
nicholaspai, line 430: This should be "bytes32" not "address" right?

contracts/chain-adapters/Optimism_Adapter.sol
chrismaree: yup! this implementation right now will only work with ERC20s. I've added a todo to add extra WETH related logic.

contracts/HubPool.sol
```diff
@@ -405,6 +421,37 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
...
+        );
+    }
+
+    function _sendTokensToTargetChain(MerkleLib.PoolRebalance memory poolRebalance, uint256 refundRequestId) internal {
+        AdapterInterface adapter = crossChainContracts[poolRebalance.chainId].adapter;
```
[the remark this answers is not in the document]
chrismaree: yup, good call. will revert if this is 0x0.

contracts/HubPool.sol
chrismaree: it should! fixed.

contracts/HubPool.sol
chrismaree: we also using the `refundRequest.destinationDistributionRoot` so we could pass in the `chainId` and the `destinationDistributionRoot`. I'll see if there is a gas difference, else I think it's cleaner to pass around the whole object as it makes the interfaces the same between the two methods.

contracts/HubPool.sol
chrismaree: good point & I agree! I will enumerate this to the minimum required data.

contracts/HubPool.sol
chrismaree: it'll revert as there is not enough money to pay everyone.

contracts/HubPool.sol
```diff
@@ -225,7 +242,7 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
         uint64 poolRebalanceLeafCount,
         bytes32 poolRebalanceRoot,
         bytes32 destinationDistributionRoot
-    ) public onlyIfNoActiveRequest {
+    ) public noActiveRequests {
         require(poolRebalanceLeafCount > 0, "Bundle must have at least 1 leaf");
```
nicholaspai, line 246: Is there a way to verify that `poolRebalanceRoot` actually contains `poolRebalanceLeafCount` leafs?

contracts/HubPool.sol
mrice32: Let me be sure I have the right understanding here:
1. This will just send over a verified _root_ that will be saved on the L2. No funds get automatically distributed.
2. If funds haven't arrived _and someone tried to distribute on the L2_, their transaction will revert.

Is my understanding correct? In no situation should the L2 side of the `_executeRelayerRefundOnTargetChain` transaction revert, right?

contracts/HubPool.sol
mrice32: @chrismaree doesn't destinationDistributionRoot come from somewhere else (not poolRebalance)?

contracts/chain-adapters/Optimism_Adapter.sol
mrice32: Makes sense!

contracts/HubPool.sol
chrismaree: I dont think you can do this by just looking at the root 🤔 . you will need to know the data that went into constructing the leafs. if this is wrong the whole bundle is disputable.

contracts/HubPool.sol
chrismaree: I dont think we can batch these easily as this is the actual action that sends tokens from L1->L2 so we'd need a way to batch these across the canonical bridge.

contracts/SpokePool.sol
```diff
@@ -242,7 +242,9 @@ abstract contract SpokePool is Testable, Lockable, MultiCaller {
         emit FilledRelay(relayHash, relayFills[relayHash], repaymentChain, amountToSend, msg.sender, relayData);
     }
 
-    function initializeRelayerRefund(bytes32 relayerRepaymentDistributionProof) public {}
```
chrismaree, line 245: nit to give me the required interface. @nicholaspai 's [PR](https://github.com/across-protocol/across-smart-contracts-v2/pull/20) implements this actual logic.

test/HubPool.RefundExecution.ts
```diff
@@ -0,0 +1,81 @@
+import { expect } from "chai";
```
chrismaree, line 1: I have split the HubPool rebalance unit tests into 3 separate files: `Initialization``Execution` and `Dispute`.

test/HubPool.RefundExecution.ts
```diff
@@ -0,0 +1,81 @@
...
+    await seedWallet(liquidityProvider, [dai], weth, consts.amountToLp.mul(10));
+
+    await enableTokensForLiquidityProvision(owner, hubPool, [weth, dai]);
+    await weth.connect(liquidityProvider).approve(hubPool.address, consts.amountToLp);
+    await hubPool.connect(liquidityProvider).addLiquidity(weth.address, consts.amountToLp);
+    await dai.connect(liquidityProvider).approve(hubPool.address, consts.amountToLp.mul(10)); // LP with 10000 DAI.
+    await hubPool.connect(liquidityProvider).addLiquidity(dai.address, consts.amountToLp.mul(10));
+  });
+
+  it("Execute relayer refund correctly produces the refund bundle call and sends cross-chain repayment actions", async function () {
```
chrismaree, line 28: this test covers the main logic added in this PR. please read through it to see how the leafs are created, tree constructed, refund initiated and finally executed.

test/MerkleLib.utils.ts
```diff
@@ -0,0 +1,60 @@
+import { expect } from "chai";
+import { getParamType } from "./utils";
+import { merkleLibFixture } from "./MerkleLib.Fixture";
+import { MerkleTree } from "../utils/MerkleTree";
+import { ethers } from "hardhat";
+const { defaultAbiCoder, keccak256 } = ethers.utils;
+import { BigNumber, Signer, Contract } from "ethers";
+
+export interface PoolRebalance {
```
chrismaree, line 9: these interfaces were moved to here to from `test/MerkleLib.Proofs.ts`.

test/utils.ts
```diff
@@ -58,5 +60,11 @@ export function randomBigNumber() {
 }
 
 export function randomAddress() {
-  return ethers.utils.hexlify(ethers.utils.randomBytes(20));
+  return ethers.utils.getAddress(ethers.utils.hexlify(ethers.utils.randomBytes(20)));
+}
+
+export async function getParamType(contractName: string, functionName: string, paramName: string) {
```
chrismaree, line 66: new function to return the types of a function's parameters. used when building Merkle trees.

test/utils.ts
```diff
@@ -58,5 +60,11 @@ export function randomBigNumber() {
 }
 
 export function randomAddress() {
-  return ethers.utils.hexlify(ethers.utils.randomBytes(20));
+  return ethers.utils.getAddress(ethers.utils.hexlify(ethers.utils.randomBytes(20)));
```
chrismaree, line 63: using the `.getAddress` call checksum cases the address.

test/HubPool.RefundInitilization.ts
```diff
@@ -0,0 +1,66 @@
+import { expect } from "chai";
```
chrismaree, line 1: nothing changed here, just moved and cut down on redundant imports/method calls.

test/HubPool.RelayerDispute.ts
```diff
@@ -1,91 +1,35 @@
 import { expect } from "chai";
```
chrismaree, line 1: nothing changed here, just moved and cut down on redundant imports/method calls.

mrice32 (approved):
There are a few commend threads that still need to be resolved, but otherwise, LGTM!

contracts/HubPool.sol
```diff
@@ -92,15 +100,20 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
         bytes32 destinationDistributionRoot,
         address indexed proposer
     );
-    event RelayerRefundExecuted(uint256 relayerRefundId, MerkleLib.PoolRebalance poolRebalance, address caller);
-
-    event RelayerRefundDisputed(
-        address indexed disputer,
-        SkinnyOptimisticOracleInterface.Request ooPriceRequest,
-        RefundRequest refundRequest
+    event RelayerRefundExecuted(
```
nicholaspai, line 103: are any of these events indexable?

test/HubPool.Fixture.ts
```diff
@@ -44,7 +44,23 @@ export const hubPoolFixture = hre.deployments.createFixture(async ({ ethers }) =
...
-  return { weth, usdc, dai, hubPool, ...parentFixtureOutput };
+  // Deploy a mock chain adapter and add it as the chainAdapter for the test chainId. Set the SpokePool to address 0.
+  const mockAdapter = await (await getContractFactory("Mock_Adapter", signer)).deploy();
+  await mockAdapter.transferOwnership(hubPool.address);
+  const mockSpoke = await (
+    await getContractFactory("MockSpokePool", signer)
+  ).deploy(weth.address, 0, parentFixtureOutput.timer.address);
+  await hubPool.setCrossChainContracts(repaymentChainId, mockAdapter.address, mockSpoke.address);
+
+  // Deploy mock l2 tokens for each token created before and whitelist the routes.
```
nicholaspai, line 55: nit: is there any benefit to running all these async funcs in parallel? I know the test network is already super fast but we could possibly get even faster?

test/HubPool.RefundExecution.ts
```diff
@@ -0,0 +1,81 @@
...
+    await weth.connect(dataWorker).approve(hubPool.address, consts.bondAmount.mul(10));
+
+    // Construct the leafs that will go into the merkle tree. For this test create a simple set of leafs that will repay
+    // two token to one chain Id with simple lpFee, netSend and running balance amounts.
+    const wethToSend = toBNWei(100);
+    const daiToSend = toBNWei(1000);
+    const leafs = buildPoolRebalanceLeafs(
```
nicholaspai, line 35: Useful function +1

contracts/HubPool.sol
```diff
@@ -252,7 +274,7 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
     }
 
     function executeRelayerRefund(
-        uint256 relayerRefundRequestId,
+        uint256 refundRequestId,
```
nicholaspai, line 277: As discussed offline, i think we can remove `refundRequestId`, rename `poolRebalance` to `poolRebalanceLeaf` and rename `proof` to `inclusionProof`

contracts/HubPool.sol
```diff
@@ -225,7 +247,7 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
         uint64 poolRebalanceLeafCount,
         bytes32 poolRebalanceRoot,
         bytes32 destinationDistributionRoot
-    ) public onlyIfNoActiveRequest {
+    ) public noActiveRequests {
```
nicholaspai, line 250: As discussed IRL:
- consider adding a comment that after `initiateRelayerRefund` is called, if the `poolRebalanceRoot`, `destinationDistributionRoot`, `poolRebalanceLeafCount`, and  `bundleEvaluationBlockNumbers` set is disputable then this can be challenged. Once the challenge period passes, then the roots are no longer disputable, and only `executeRelayerRefund` can be called and `initiateRelayerRefund` can't be called again until all leafs are executed. 
- Examples of some characteristics of these roots that could be disputed: `poolRebalanceRoot.netSendAmount != destinationDistributionRoot.amountToReturn` for example

nicholaspai: Motivation is updated comments + interface for the spoke pool `distributeRelayerRefund` in my PR: https://github.com/across-protocol/across-smart-contracts-v2/pull/20

chrismaree: will do!

contracts/HubPool.sol
chrismaree: yup!

contracts/HubPool.sol
chrismaree: demos. will index appropriate ones.

contracts/MerkleLib.sol
```diff
@@ -15,9 +15,9 @@ library MerkleLib {
         uint256 leafId;
         // This is used to know which chain to send cross-chain transactions to (and which SpokePool to sent to).
         uint256 chainId;
-        // The following arrays are required to be the same length. They are parallel arrays for the given chainId and should be ordered by the `tokenAddresses` field.
+        // The following arrays are required to be the same length. They are parallel arrays for the given chainId and should be ordered by the `l1Token` field.
         // All whitelisted tokens with nonzero relays on this chain in this bundle in the order of whitelisting.
-        address[] tokenAddresses;
+        address[] l1Token;
```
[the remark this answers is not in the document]
chrismaree: if we make this change should we make all the arrays plural? `netSendAmounts` `runningBalances`?

test/HubPool.Fixture.ts
chrismaree: I dont think really. I think optimising for readability is the most important thing in fixtures IMHO.

contracts/HubPool.sol
chrismaree: this is exactly right @mrice32

contracts/HubPool.sol
chrismaree: yes, it comes directly from the `refundRequest` object on this contract via `refundRequest.destinationDistributionRoot`. The only thing from `MerkleLib.PoolRebalance memory poolRebalance` that is used is in this method is the `poolRebalance.chainId`. For this method, it would be cleaner to just pass in the chainId but the downside is that it then has a diffrent interface to `_sendTokensToTargetChain` which needs: `chainId`,`netSendAmount[]`,`l1Token[]`.

what do you prefer? is there a gas difference here do you think? IMHO from a readability perspective passing the whole root is cleanest but curious what you think!

contracts/MerkleLib.sol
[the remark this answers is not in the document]
nicholaspai: I think arrays should always be plural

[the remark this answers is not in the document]
mrice32: Yeah, I agree. If we didn't do so before in our structs (which could be my fault), I think we should update.

contracts/HubPool.sol
mrice32: Sorry, yeah, I don't know how this is handled gas-wise. My stylistic preference is always to pass in the _simplest_ data possible, so the reader can clearly see what's being used. But this is totally up to you if it makes little or no gas difference (copying a large struct in memory could be expensive though so we might want to check on that).

contracts/MerkleLib.sol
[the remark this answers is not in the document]
chrismaree: ok will update accordingly on all variables in all places.

contracts/HubPool.sol
chrismaree: I'm going to refactor this to pass in as little data as posible. I think you right @mrice32 that this, as a general rule, is always the right way to go.

PR #20, opened 2022-02-03
feat: Implement _initiateRelayerRefund and begin Optimism-specific SpokePool

nicholaspai:


contracts/SpokePool.sol
```diff
@@ -43,6 +43,16 @@ abstract contract SpokePool is Testable, Lockable, MultiCaller {
...
     mapping(address => mapping(uint256 => bool)) public enabledDepositRoutes;
 
+    // Associates relayer refund distribution proofs to unique IDs.
+    mapping(uint256 => bytes32) public relayerRefundRoots;
+
+    // Count of additions to `relayerRefundRoots`.
+    uint256 public relayerRefundRootCount;
+
+    // This is a 2D bitmap tracking which leafs in the relayer refund roots have been claimed, with max size of
+    // 256 elements, limiting us to 256 relayer refund leafs per distribution proof.
```
chrismaree, line 53: this comment is wrong. we have no such limitation.

contracts/SpokePool.sol
```diff
@@ -257,7 +267,15 @@ abstract contract SpokePool is Testable, Lockable, MultiCaller {
         );
     }
 
-    function initializeRelayerRefund(bytes32 relayerRepaymentDistributionProof) public {}
+    // This internal method should be called by an external "initializeRelayerRefund" function that validates the
+    // cross domain sender is the HubPool. This validation step differs for each L2, which is why the implementation
+    // specifics are left to the implementor of this abstract contract.
+    function _initializeRelayerRefund(bytes32 relayerRepaymentDistributionProof) internal {
+        relayerRefunds.push();
+        uint256 relayerRefundId = relayerRefunds.length - 1;
```
chrismaree, line 275: small gas optitimization: get the `uint256 relayerRefundId = relayerRefunds.length;` before doing the push operation so you dont need to do the extra subtraction operation.

contracts/SpokePool.sol
nicholaspai: fixed

nicholaspai:
@mrice32 @chrismaree not sure why but I the `yarn.lock` wasn't regenerating after I added something to the package.json, so i removed and then re-added and it caused a big diff

yarn.lock
```diff
@@ -279,20 +279,20 @@
     minimatch "^3.0.4"
     strip-json-comments "^3.1.1"
 
-"@eth-optimism/contracts@^0.5.5":
-  version "0.5.8"
-  resolved "https://registry.yarnpkg.com/@eth-optimism/contracts/-/contracts-0.5.8.tgz#975313a219b5537bad7f4f98f8841c98592882d0"
-  integrity sha512-INvIMiE72Y6tjxm+BZyN9f3+LiMKJgmNpR0qEimrvpg49hK8gOXQTnI+0C7/qIIGvgaQgDyejG9AfkxZizjLVQ==
+"@eth-optimism/contracts@^0.5.11", "@eth-optimism/contracts@^0.5.5":
```
mrice32, line 282: @nicholaspai the reason you didn't see changes in the yarn.lock (I think) was that this package was _already in the dependency tree_ with a permissible version, most likely due to the fact that we import uma/core. You could probably drop these yarn.lock changes if you wanted and I think everything would work (but nothing wrong with regenerating things at this early stage).

nicholaspai: interesting, what do you think we should do--keep the new yarn.lock or drop? Also should we update the eth-optimism version if its available?

mrice32: I think keeping the updates is totally fine! I just wanted to figure out what happened to make sure something wasn't broken in our setup or something :)

contracts/Optimism_SpokePool.sol
```diff
@@ -0,0 +1,65 @@
...
+        SpokePool(_wethAddress, _depositQuoteTimeBuffer, timerAddress)
+    {
+        _setCrossDomainAdmin(_crossDomainAdmin);
+    }
+
+    /**************************************
+     *          ADMIN FUNCTIONS           *
+     **************************************/
+
+    // TODO:
```
mrice32, line 35: What's TODO on these? The natspec or something in the implementation?

contracts/Optimism_SpokePool.sol
```diff
@@ -0,0 +1,65 @@
...
+        onlyFromCrossDomainAccount(crossDomainAdmin)
+    {
+        _initializeRelayerRefund(relayerRepaymentDistributionProof);
+    }
+
+    /**************************************
+     *         INTERNAL FUNCTIONS         *
+     **************************************/
+
+    function _setCrossDomainAdmin(address newCrossDomainAdmin) internal {
```
mrice32, line 60: Is there any way to call this function after construction? Is this something we'd like to be able to change? (genuine question, not suggesting a change)

contracts/Optimism_SpokePool.sol
```diff
@@ -0,0 +1,65 @@
...
+    function setEnableRoute(
+        address originToken,
+        uint256 destinationChainId,
+        bool enable
+    ) public onlyFromCrossDomainAccount(crossDomainAdmin) {
+        _setEnableRoute(originToken, destinationChainId, enable);
+    }
+
+    // TODO:
+    function setDepositQuoteTimeBuffer(uint64 buffer) public onlyFromCrossDomainAccount(crossDomainAdmin) {
```
mrice32, line 45: Is there any interface or anything that ensures that these functions match across all SpokePool implementations?

mrice32 (approved):
Looks good! Only minor questions/comments

contracts/Optimism_SpokePool.sol
chrismaree: there is a `setCrossDomainAdmin` that can call this after construction enabling us to change it. I think we'd want to change it in the event something broke within the L1<->L2 logic. one problem, though, that I can see with this pattern is it assumes that the cross domain owner implements sufficient methods to call `setCrossDomainAdmin` over the bridge. this should be doable as this is going to be sitting within the "Adapter" logic that I've created for the L1->L2 calls. We can create a base contract (`CrossDomainAdminBase`) or something equivalent that these adapters will inherit from. can think more on this in the next few PRs.

test/constants.ts
```diff
@@ -39,3 +39,7 @@ export const zeroBytes32 = "0x00000000000000000000000000000000000000000000000000
 export const identifier = utf8ToHex("IS_ACROSS_V2_RELAY_VALID");
 
 export const zeroRawValue = { rawValue: "0" };
+
+export const spokePoolRelayerRefundRoot = createRandomBytes32();
```
chrismaree, line 43: I have a const for this in my other PR called `mockDestinationDistributionRoot`.

contracts/Optimism_SpokePool.sol
nicholaspai: yeah i guess these are fully implemented actually

contracts/Optimism_SpokePool.sol
nicholaspai: Yeah i added in this function without telling you. We should allow this to be changed from L1

contracts/Optimism_SpokePool.sol
nicholaspai: Great point I'll add an interface

nicholaspai: do you think all L2 spoke pools will immplement the notion of a `crossDomainAdmin`? @chrismaree

test/constants.ts
nicholaspai: Will merge that in

contracts/Optimism_SpokePool.sol
```diff
@@ -0,0 +1,80 @@
...
+import "./SpokePoolInterface.sol";
+
+/**
+ * @notice OVM specific SpokePool.
+ * @dev Uses OVM cross-domain-enabled logic for access control.
+ */
+
+contract Optimism_SpokePool is CrossDomainEnabled, SpokePoolInterface, SpokePool {
+    // Address of the L1 contract that acts as the owner of this SpokePool.
+    address public override crossDomainAdmin;
```
nicholaspai, line 16: @chrismaree @mrice32 should `crossDomainAdmin` be in the `SpokePool` or do you think this notion of a cross domain admin will be different for each. L2?

contracts/Optimism_SpokePool.sol
mrice32: Sorry, I'm a little confused. Are there two functions that allow you to `setCrossDomainAdmin`? If so, can we consolidate them into a single function?

mrice32 (approved):
LGTM!

contracts/SpokePool.sol
```diff
@@ -43,6 +45,15 @@ abstract contract SpokePool is Testable, Lockable, MultiCaller {
     // Origin token to destination token routings can be turned on or off.
     mapping(address => mapping(uint256 => bool)) public enabledDepositRoutes;
 
+    struct RelayerRefund {
+        // Merkle root of relayer refunds.
+        bytes32 distributionRoot;
```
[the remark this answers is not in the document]
nicholaspai: Hm so you think this could work by the `HubPool` relaying all leafs to the `SpokePool` here and storing them on-chain, as opposed to just relaying the root. @chrismaree wdyt?

contracts/Optimism_SpokePool.sol
nicholaspai: No there's only one, implemented in this contract. `setCrossDomainAdmin` is included in the `SpokePoolInterface`

contracts/Optimism_SpokePool.sol
nicholaspai: Its the contract that calls `relayMessage` on the canonical bridge, so its either the `BridgeAdmin` or the `HubPool`

chrismaree: for now `HubPool` me thinks. we might want to change this later but I think lets leave it there optimistically at the moment.

contracts/SpokePool.sol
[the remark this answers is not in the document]
chrismaree: I dont think @mrice32  is proposing the HubPool passing the leafs; it cant do that in a efficient way. I _think_ @mrice32 is proposing a technique where you can execute multiple leafs at once in a compressed structure and prove they are within the provided root. @mrice32 what you describe would fit into the execution stage of the process, right?

test/SpokePool.RelayerRefund.ts
```diff
@@ -0,0 +1,22 @@
+import { expect } from "chai";
```
chrismaree, line 1: In your next PR it might be good to re-name this file to `SpokePool.RefundInitialization.ts` and then you can create `SpokePool.RefundExecution.ts` for the execution step, following the pattern I did in the hub.

chrismaree (approved):
this looks great to me! keen to get the next bit of functionality in with the proofs.

contracts/SpokePool.sol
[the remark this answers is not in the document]
nicholaspai: How do you imagine we improve on just making the contract multicall and assuming `distributeRelayerRefund` is called multiple times in one batched txn?

test/SpokePool.RelayerRefund.ts
nicholaspai: done!

nicholaspai:
Just giving @mrice32 a chance to comment [here](https://github.com/across-protocol/across-smart-contracts-v2/pull/20#discussion_r800108266) before merging

contracts/SpokePool.sol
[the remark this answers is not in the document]
mrice32: I wanna be super clear that I don't think we should do this now, and I think before we add this complexity, we should try to better understand the amount of calldata/gas savings we'd see from something like this with different tree sizes.

Agree @chrismaree. I'm not suggesting the HubPool do anything different since any added data anywhere on mainnet is going to be expensive.

If you did multicall, you would effectively be passing:

[{distributionStruct1, proofArray1}, {distributionStruct2, proofArray2}...]

What I'm suggesting is an additional method that allows you to pass _all_ leaves this way:

[{distriubtionStruct1}, {distributionStruct2}, ...]

If we get the logic right in the contract, the leaf data _could be used_ to generate the proofs so you wouldn't need to pass them in, thereby reducing calldata.

mrice32 (approved):

PR #22, opened 2022-02-04
feat: add release process via github actions

mrice32:
Just adds a simple config using github's recommended process whereby it publishes when you create a release on the github UI.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,18 @@
+name: Publish Package to npmjs
+on:
+  release:
+    types: [created]
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      # Setup .npmrc file to publish to npm
+      - uses: actions/setup-node@v2
+        with:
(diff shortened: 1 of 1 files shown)


chrismaree (approved):
nice!

.github/workflows/publish.yml
```diff
@@ -0,0 +1,18 @@
...
+      - uses: actions/checkout@v2
+      # Setup .npmrc file to publish to npm
+      - uses: actions/setup-node@v2
+        with:
+          node-version: "16.x"
+          registry-url: "https://registry.npmjs.org"
+      - run: yarn
+      - run: yarn publish
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
```
chrismaree, line 18: how is this set?

mrice32: I set one in the across github org. It can be set at the org or repo level.

mrice32: One thing I want to better understand is who has access to these secrets via PRs.

PR #24, opened 2022-02-07
feat(spoke-pool): Distribute relayer refund

nicholaspai:


contracts/Optimism_SpokePool.sol
```diff
@@ -44,37 +39,68 @@ contract Optimism_SpokePool is CrossDomainEnabled, SpokePoolInterface, SpokePool
...
+        // that sends tokens from this contract to relayers on the same network?
+        if (distributionLeaf.amountToReturn > 0) {
+            // Do we need to perform any check about the last time that funds were bridged from L2 to L1?
+            IL2ERC20Bridge(Lib_PredeployAddresses.L2_STANDARD_BRIDGE).withdrawTo(
+                distributionLeaf.l2TokenAddress, // _l2Token. Address of the L2 token to bridge over.
+                hubPool, // _to. Withdraw, over the bridge, to the l1 pool contract.
+                distributionLeaf.amountToReturn, // _amount. Send the full balance of the deposit box to bridge.
+                6_000_000, // _l1Gas. Unused, but included for potential forward compatibility considerations
+                "" // _data. We don't need to send any data for the bridging action.
+            );
```
mrice32, line 95: nit: should we just put this bit of special logic in a virtual internal method that's called by the base class? That way the more generic logic could all live in the base class and this dispatch bit would be the only part that's custom per chain?

contracts/SpokePool.sol
```diff
@@ -282,11 +323,59 @@ abstract contract SpokePool is Testable, Lockable, MultiCaller {
...
+        );
+
+        // Verify the leafId in the leaf has not yet been claimed.
+        require(!MerkleLib.isClaimed(refund.claimedBitmap, distributionLeaf.leafId), "Already claimed");
+
+        // Set leaf as claimed in bitmap.
+        MerkleLib.setClaimed(refund.claimedBitmap, distributionLeaf.leafId);
+
+        // TODO: Should we be checking that distributionLeaf.refundAddresses.length == distributionLeaf.refundAmounts.length
+        // or can we assume that the `initiateRelayerRefund` on the HubPool would have been disputed on L1?
```
mrice32, line 352: As long as it's fairly cheap to check (which I think it is), I think this check makes sense to avoid any unexpected behavior below.

contracts/SpokePool.sol
```diff
@@ -282,11 +323,59 @@ abstract contract SpokePool is Testable, Lockable, MultiCaller {
...
+        require(distributionLeaf.refundAddresses.length == distributionLeaf.refundAmounts.length, "invalid leaf");
+
+        // TODO: Similar question to above: should we assume that the L2 token is not address(0) or any other address
+        // that would have been disputed on L1?
+        require(distributionLeaf.l2TokenAddress != address(0), "invalid leaf");
```
mrice32, line 357: I wouldn't worry about this unless there's something _particularly bad_ that would happen if it were address 0.

contracts/Optimism_SpokePool.sol
```diff
@@ -44,37 +39,68 @@ contract Optimism_SpokePool is CrossDomainEnabled, SpokePoolInterface, SpokePool
...
+        }
 
-    function _setCrossDomainAdmin(address newCrossDomainAdmin) internal {
-        require(newCrossDomainAdmin != address(0), "Bad bridge router address");
-        crossDomainAdmin = newCrossDomainAdmin;
-        emit SetXDomainAdmin(crossDomainAdmin);
+        emit TokensBridged(
```
mrice32, line 98: Should we only emit this if we actually bridged tokens?

mrice32 (approved):
Looks great! A few minor comments.

contracts/Optimism_SpokePool.sol
nicholaspai: Great idea, done: https://github.com/across-protocol/across-smart-contracts-v2/pull/24/commits/833d32056b8c31cf6cc33f144c4d8f77a4af903f

contracts/SpokePool.sol
nicholaspai: Agreed

contracts/SpokePool.sol
nicholaspai: Yeah this would revert anyways if l2 token was address 0, and also should be disputed on L1

contracts/Optimism_SpokePool.sol
nicholaspai: yup! This is now implemented here https://github.com/across-protocol/across-smart-contracts-v2/pull/24/commits/833d32056b8c31cf6cc33f144c4d8f77a4af903f

`_bridgeTokensToHubPool` is only called if `amountToReturn > 0`, and that method should emit an event for each L2-specific implementation, like it is in the Optimism_SpokePool demo code

contracts/Optimism_SpokePool.sol
```diff
@@ -44,37 +47,55 @@ contract Optimism_SpokePool is CrossDomainEnabled, SpokePoolInterface, SpokePool
...
 
-    /**************************************
-     *         INTERNAL FUNCTIONS         *
-     **************************************/
-
-    function _setCrossDomainAdmin(address newCrossDomainAdmin) internal {
-        require(newCrossDomainAdmin != address(0), "Bad bridge router address");
-        crossDomainAdmin = newCrossDomainAdmin;
-        emit SetXDomainAdmin(crossDomainAdmin);
+    function _bridgeTokensToHubPool(MerkleLib.DestinationDistribution memory distributionLeaf) internal override {
```
chrismaree, line 85: this will need branching logic for WETH. can do it in a future PR.

contracts/Optimism_SpokePool.sol
```diff
@@ -44,37 +47,55 @@ contract Optimism_SpokePool is CrossDomainEnabled, SpokePoolInterface, SpokePool
...
+            hubPool, // _to. Withdraw, over the bridge, to the l1 pool contract.
+            distributionLeaf.amountToReturn, // _amount. Send the full balance of the deposit box to bridge.
+            6_000_000, // _l1Gas. Unused, but included for potential forward compatibility considerations
+            "" // _data. We don't need to send any data for the bridging action.
+        );
+        emit TokensBridged(
+            distributionLeaf.l2TokenAddress,
+            hubPool,
+            distributionLeaf.amountToReturn,
+            6_000_000,
```
chrismaree, line 97: I think we should set this in a variable and then let the owner change it. does not need to happen in this PR though but should probs do it.

contracts/SpokePool.sol
```diff
@@ -286,7 +321,54 @@ abstract contract SpokePool is Testable, Lockable, MultiCaller {
         uint256 relayerRefundId,
         MerkleLib.DestinationDistribution memory distributionLeaf,
         bytes32[] memory inclusionProof
-    ) public {}
+    ) public override {
+        // Check integrity of leaf structure:
+        require(distributionLeaf.chainId == chainId(), "Invalid chainId");
+        require(distributionLeaf.refundAddresses.length == distributionLeaf.refundAmounts.length, "invalid leaf");
```
chrismaree, line 327: I dont think any of these checks are needed. If either of these fail then the proof will fail. it's somewhat redundant, no?

contracts/SpokePool.sol
```diff
@@ -286,7 +321,54 @@ abstract contract SpokePool is Testable, Lockable, MultiCaller {
...
+
+        // Grab distribution root stored at `relayerRefundId`.
+        RelayerRefund storage refund = relayerRefunds[relayerRefundId];
+
+        // Check that `inclusionProof` proves that `distributionLeaf` is contained within the distribution root.
+        // Note: This should revert if the `distributionRoot` is uninitialized.
+        require(
+            MerkleLib.verifyRelayerDistribution(refund.distributionRoot, distributionLeaf, inclusionProof),
```
chrismaree, line 335: nit: if you re-name the `inclusionProof` to `proof` t his might fit on one line and then it's consistant with the `HubPool`.

contracts/Optimism_SpokePool.sol
```diff
@@ -12,22 +13,24 @@ import "./SpokePoolInterface.sol";
  */
 
 contract Optimism_SpokePool is CrossDomainEnabled, SpokePoolInterface, SpokePool {
-    // Address of the L1 contract that acts as the owner of this SpokePool.
-    address public override crossDomainAdmin;
-
-    event SetXDomainAdmin(address indexed newAdmin);
+    event TokensBridged(
```
mrice32, line 16: Do we need this event to be different per chain? Or could we standardize and put in the base class?

mrice32 (approved):

contracts/Optimism_SpokePool.sol
nicholaspai: I can change in this PR. what's annoying is that this is potentially Optimism specific, but HubPool will need to implement a method that can trigger the cross chain admin function to make this call. Shold we just implement `setL2GasLimit` in the `SpokePool` and have it be useless for some L2s?

contracts/SpokePool.sol
nicholaspai: I think its better to be safe in case the bots miss something, we won't get into weird state

nicholaspai: merging this comment: https://github.com/across-protocol/across-smart-contracts-v2/pull/24#discussion_r801314310

contracts/Optimism_SpokePool.sol
nicholaspai: I think it should be diff per chain since each chain will use different params to bridge tokens. For example, Arbitrum uses something called a `maxSubmissionCost` whereas optimsim does not

mrice32: Gotcha, okay. Could make sense to have a standard event that we use for all contracts, if the specific parameters aren't relevant. Just something to think about later. Not something I think we should change now.

mrice32 (approved):

contracts/Optimism_SpokePool.sol
nicholaspai: added a TODO

contracts/Optimism_SpokePool.sol
nicholaspai: Hmmm ill add a TODO

contracts/SpokePool.sol
nicholaspai: done

contracts/Optimism_SpokePool.sol
nicholaspai: And consider other constants such as `maxSubmissionCost` and `l2GasPrice` (and `refundAddress`) that will be Arbitrum specific?

nicholaspai: Perhaps in the HubPool we can allow it to call any arbitrary encoded method to the SpokePool so it can call L2-specific admin functions like `setL2GasLimit` for Optimism and `setL2GasPrice` for Arbitrum

mrice32 (approved):
LGTM!

contracts/Optimism_SpokePool.sol
chrismaree: ye, so I think that what makes the most sence is to have this be a chain specific event, like you mention. then, each SpokePool specific implementation for a given chain has these extra methods on them for those chain specific things.

contracts/Optimism_SpokePool.sol
```diff
@@ -44,37 +47,56 @@ contract Optimism_SpokePool is CrossDomainEnabled, SpokePoolInterface, SpokePool
...
-    /**************************************
-     *         INTERNAL FUNCTIONS         *
-     **************************************/
-
-    function _setCrossDomainAdmin(address newCrossDomainAdmin) internal {
-        require(newCrossDomainAdmin != address(0), "Bad bridge router address");
-        crossDomainAdmin = newCrossDomainAdmin;
-        emit SetXDomainAdmin(crossDomainAdmin);
+    function _bridgeTokensToHubPool(MerkleLib.DestinationDistribution memory distributionLeaf) internal override {
+        // TODO: Handle WETH token unwrapping
```
chrismaree, line 86: yup! this part you can copy some of the work done in the other contracts, fortunately.

contracts/SpokePool.sol
```diff
@@ -285,8 +320,54 @@ abstract contract SpokePool is Testable, Lockable, MultiCaller {
...
+        // Set leaf as claimed in bitmap.
+        MerkleLib.setClaimed(refund.claimedBitmap, distributionLeaf.leafId);
+
+        // For each relayerRefundAddress in relayerRefundAddresses, send the associated refundAmount for the L2 token address.
+        // Note: Even if the L2 token is not enabled on this spoke pool, we should still refund relayers.
+        for (uint32 i = 0; i < distributionLeaf.refundAmounts.length; i++) {
+            uint256 amount = distributionLeaf.refundAmounts[i];
+            if (amount > 0) {
+                IERC20(distributionLeaf.l2TokenAddress).safeTransfer(distributionLeaf.refundAddresses[i], amount);
+            }
```
chrismaree, line 348:
```suggestion
            if (amount > 0) IERC20(distributionLeaf.l2TokenAddress).safeTransfer(distributionLeaf.refundAddresses[i], amount);            
```

chrismaree (approved):

contracts/Optimism_SpokePool.sol
nicholaspai: Cool added a universal event emitted by `SpokePool` and added an Optimism specific `OptimismTokenBridged` event. Also added a `setL1Gas onlyOwner` method on the OptimismSpokePool

PR #25, opened 2022-02-07
improve: Allow filled amount to be 0 to empower slow relayers

nicholaspai:


contracts/SpokePool.sol
```diff
@@ -225,16 +225,16 @@ abstract contract SpokePool is Testable, Lockable, MultiCaller {
...
+        // the amount filled so far for a particular `relayHash`, so this will start at 0 and increment with each fill.
+        require(relayFills[relayHash] < totalRelayAmount, "relay filled");
 
         // Stores the equivalent amount to be sent by the relayer before fees have been taken out.
-        uint256 fillAmountPreFees = _computeAmountPreFees(maxTokensToSend, (realizedLpFeePct + relayerFeePct));
+        uint256 fillAmountPreFees = 0;
 
         // Adding brackets "stack too deep" solidity error.
-        {
+        if (maxTokensToSend > 0) {
```
mrice32, line 236: Can you think of anywhere that we use the filledAmount stored in the contract as a proxy for the relay being "initialized"? If so, this relay would still be considered uninitialized. I can't think of anywhere this would cause problems, just wanted to ask in case you could.

mrice32 (approved):
LGTM!

contracts/SpokePool.sol
nicholaspai: We don't, there isn't really a good way of checking whether a relay was initialized at the moment. For that we'd need to expand the data associated with each relay hash

chrismaree (approved):

PR #27, opened 2022-02-08
feat(spoke-pool): Allow relayer and depositor to agree to update relay fee

nicholaspai:


contracts/SpokePool.sol
```diff
@@ -268,6 +272,61 @@ abstract contract SpokePool is Testable, Lockable, MultiCaller {
...
+        // Check if relay is not yet fully filled.
+        // Note: we cannot easily check that this relay is actually pending, as the relay hash above can point to an
+        // uninitialized relay, and this function will not revert as long as the signed message and signature correspond
+        // to the depositor's address, and the updated relayer fee is contained in the message. This is not a problem
+        // because relayers who store relay data for deposits that don't exist will not get rewarded. They are instead
+        // just wasting gas to submit this transaction.
+        require(relayFills[relayHash] < totalRelayAmount, "filled relay");
+
+        // Depositor should have signed a hash of the relayer fee % to update to.
+        bytes32 expectedDepositorMessageHash = keccak256(abi.encode(newRelayerFeePct));
```
mrice32, line 313: I think we need more than this to avoid replay attacks. For instance, if I agree to a relayer fee of 23% on a small relay, then I send a large relay, someone could take my signed message from the small relay and replay it to get a large fee from me in the in the large relay.

A few additional fields I think we need: depositId, chainId. We might also want include a constant value in front, like "ACROSS-V2 FEE 1.0" or something that we could increment if we ever updated spoke pools and ended up resetting the depositId counter.

contracts/SpokePool.sol
```diff
@@ -268,6 +272,61 @@ abstract contract SpokePool is Testable, Lockable, MultiCaller {
         );
     }
 
+    function increaseRelayFee(
```
mrice32, line 275: Should the relayer be able to use this method to actually fill the relay? If not, how does it get connected with a particular fill? Or is the idea that we would swap out the relayerFeePct for every FilledRelay that was submitted afterward?

mrice32 (commented):
Looks awesome! A few minor questions/comments

contracts/SpokePool.sol
nicholaspai: The relayer in this implementation would call this method separately and it would not add to the fill. We could add an additonal method `increaseRelayFeeAndFillRelay` (bad name) that would call `fillRelay` AND modify the fee WDYT?

mrice32: > relayer in this implementation would call this method separately and it would not add to the fill

How would this work? Wouldn't the subsequent fill still be using the _old_ relayer fee (since that's the only one that would match the hash)? So in that case, too much money would be sent to the recipient, no?

We don't have to do it this way, but the way I had conceptualized this was that we'd have a fill method that's identical to the existing one with two additional args: `bytes memory signature, uint64 updatedRelayerFeePct`. The signature would be verified against the provided fee, and the updated fee fee would be emitted in the `FilledRelay` event (we could also include a second event to signify that this fee was overridden or something).

mrice32 (commented):
https://github.com/across-protocol/across-smart-contracts-v2/pull/27#discussion_r801931300

contracts/SpokePool.sol
nicholaspai: So the relay hash would stay the same, with the old relayer fee, but the event would just change? (or have an additional one emitted)

I agree wth you that `fillRelay` could be overridden with a separate interface

mrice32: Yeah, in the solution I had been thinking of, the relay hash/data would always be the same, the signature would be provided essentially as an override to the originally-set relayer fee.

nicholaspai: interesting let me try this

chrismaree: ye, what @mrice32 describes fits with my understanding of this method. you can actually just overload the `fillRelay` method with an extra param which optionally lets you change the originally provided `relayFeePct`.

contracts/SpokePool.sol
```diff
@@ -225,49 +222,130 @@ abstract contract SpokePool is Testable, Lockable, MultiCaller {
...
+    // proving that the depositor agreed to the updated fee.
+    struct UpdatedRelayerFeeData {
+        // Note: This struct is used to make sure that the following overloaded `fillRelay` does not create a stack
+        // too deep compile-time error.
+        uint64 newRelayerFeePct;
+        bytes32 depositorMessageHash;
+        bytes depositorSignature;
+    }
+
+    function fillRelayUpdatedFee(
```
mrice32, line 291: nit: since this method is public, should it be moved up next to the fillRelay?

contracts/SpokePool.sol
```diff
@@ -225,49 +222,130 @@ abstract contract SpokePool is Testable, Lockable, MultiCaller {
...
+        {
+            // Depositor should have signed a hash of the relayer fee % to update to and information uniquely identifying
+            // the deposit to relay. This ensures that this signature cannot be re-used for other deposits. The version
+            // string is included as a precaution in case this contract is upgraded.
+            // Note: we use encode instead of encodePacked because it is more secure, more in the "warning" section
+            // here: https://docs.soliditylang.org/en/v0.8.11/abi-spec.html#non-standard-packed-mode
+            bytes32 expectedDepositorMessageHash = keccak256(
+                abi.encode("ACROSS-V2-FEE-1.0", updatedRelayerFeeData.newRelayerFeePct, depositId, originChainId)
+            );
+            require(expectedDepositorMessageHash == updatedRelayerFeeData.depositorMessageHash, "incorrect new fee");
```
mrice32, line 318: Why do we need them to pass in the message hash? Aren't we already generating it from other data they've provided?

contracts/SpokePool.sol
```diff
@@ -225,49 +222,130 @@ abstract contract SpokePool is Testable, Lockable, MultiCaller {
...
+            recipient: recipient,
+            destinationToken: destinationToken,
+            realizedLpFeePct: realizedLpFeePct,
+            relayerFeePct: relayerFeePct,
+            depositId: depositId,
+            originChainId: originChainId,
+            relayAmount: totalRelayAmount
+        });
+        bytes32 relayHash = _getRelayHash(relayData);
+        _fillRelay(relayHash, relayData, updatedRelayerFeeData.newRelayerFeePct, maxTokensToSend, repaymentChain);
```
mrice32, line 346: +1!

contracts/SpokePool.sol
```diff
@@ -323,6 +401,32 @@ abstract contract SpokePool is Testable, Lockable, MultiCaller {
...
 
+    struct FillRelayEventData {
+        bytes32 relayHash;
+        RelayData relayData;
+        uint64 updatableRelayerFeePct;
+        uint256 fillAmountPreFees;
+        uint256 repaymentChain;
+    }
+
+    function _emitFillRelay(FillRelayEventData memory eventData) internal {
```
mrice32, line 412: Is this extra method here just to avoid stack too deep errors?

mrice32 (commented):
Looks good! Minor comments

contracts/SpokePool.sol
```diff
@@ -225,49 +222,130 @@ abstract contract SpokePool is Testable, Lockable, MultiCaller {
         });
         bytes32 relayHash = _getRelayHash(relayData);
 
+        _fillRelay(relayHash, relayData, relayerFeePct, maxTokensToSend, repaymentChain);
+    }
+
+    function _fillRelay(
+        bytes32 relayHash,
+        RelayData memory relayData,
```
chrismaree, line 230: is there a gas difference between this being `memory` and `calldata`?

contracts/SpokePool.sol
```diff
@@ -225,49 +222,130 @@ abstract contract SpokePool is Testable, Lockable, MultiCaller {
         });
         bytes32 relayHash = _getRelayHash(relayData);
 
+        _fillRelay(relayHash, relayData, relayerFeePct, maxTokensToSend, repaymentChain);
+    }
+
+    function _fillRelay(
```
chrismaree, line 228: do you have a section for internal methods? should probs be located there.

contracts/SpokePool.sol
```diff
@@ -225,49 +222,130 @@ abstract contract SpokePool is Testable, Lockable, MultiCaller {
...
+        bytes32 relayHash,
+        RelayData memory relayData,
+        uint64 updatableRelayerFeePct,
+        uint256 maxTokensToSend,
+        uint256 repaymentChain
+    ) internal {
+        // We limit the relay fees to prevent the user spending all their funds on fees.
+        require(
+            updatableRelayerFeePct <= 0.5e18 &&
```
chrismaree, line 237: I think one of these checks is redundant. if you validate `relayData.realizedLpFeePct` and `(updatableRelayerFeePct + relayData.realizedLpFeePct) < 1e18` then it's impossible for `updatableRelayerFeePct` to be `>0.5e18`.

contracts/SpokePool.sol
```diff
@@ -225,49 +222,130 @@ abstract contract SpokePool is Testable, Lockable, MultiCaller {
...
-            relayData.depositId,
-            relayData.relayerFeePct,
-            relayData.realizedLpFeePct,
-            relayData.destinationToken,
-            msg.sender,
-            relayData.depositor,
-            relayData.recipient
+        // Needed to resolve stack too deep compile-time error
+        _emitFillRelay(
+            FillRelayEventData(relayHash, relayData, updatableRelayerFeePct, fillAmountPreFees, repaymentChain)
```
chrismaree, line 277: dont emit `RelayData`. we dont want to emit structs. ideally what you had before was better so etherscan can parse it.

chrismaree: I take that back. your method you call does this 🙏

contracts/SpokePool.sol
nicholaspai: The only problem with overloading is that overloading doesn't play nicely with [ethers like it does with web3](https://github.com/ethers-io/ethers.js/issues/407#issuecomment-458329708). Do you have a strong opinion about overloading versus creating a new function name?

contracts/SpokePool.sol
nicholaspai: ah true!

contracts/SpokePool.sol
nicholaspai: exactly

contracts/SpokePool.sol
nicholaspai: done

contracts/SpokePool.sol
nicholaspai: what if `realizedLpFeePct = 0` but `updatableRelayerFeePct = 0.6`?

contracts/SpokePool.sol
nicholaspai: haha yep this is just unreadable to get around stack too deep

contracts/SpokePool.sol
```diff
@@ -225,59 +226,70 @@ abstract contract SpokePool is Testable, Lockable, MultiCaller {
...
+            );
+
+            // Check the hash corresponding to the https://eth.wiki/json-rpc/API#eth_sign[`eth_sign`]
+            // JSON-RPC method as part of EIP-191. We use OZ's signature checker library with adds support for
+            // EIP-1271 which can verify messages signed by smart contract wallets like Argent and Gnosis safes.
+            // If the depositor signed a message with a different updated fee (or any other param included in the
+            // above keccak156 hash), then this will revert.
+            bytes32 ethSignedMessageHash = ECDSA.toEthSignedMessageHash(expectedDepositorMessageHash);
+            require(
+                SignatureChecker.isValidSignatureNow(depositor, ethSignedMessageHash, depositorSignature),
```
mrice32, line 270: Minor thing we should be aware of: since this allows for ERC1271 "contract signatures", this can make a call out to the caller under the hood. Shouldn't be a problem, just wanted to mention it.

mrice32 (approved):
LGTM!

contracts/SpokePool.sol
nicholaspai: Hmmm thats a good point. As long as we reentrancy guard this method it should be ok though? I suppose also the calling contrct could call this contract back into `fillRelay` which I don't think could cause issues, but we could easily remedy by moving all this signature verification logic AFTER `_fillRelay`. This would also be more consistent perhaps with Check-effects-interaction pattern. WDYT?

contracts/SpokePool.sol
nicholaspai: according to [this answer](https://ethereum.stackexchange.com/questions/74442/when-should-i-use-calldata-and-when-should-i-use-memory), `calldata` is cheaper.

Also aside, when testing I noticed that declaring a variable as `memory` does not contribute to the "stack too deep error", while `calldata` does?

nicholaspai: Note that `calldata` is only avaialble for `external` methods. [source](https://solidity-by-example.org/data-locations/)

nicholaspai:
Remaining comment here: https://github.com/across-protocol/across-smart-contracts-v2/pull/27#discussion_r802105356

@chrismaree wdyt?

contracts/SpokePool.sol
mrice32: Reading more into the method, I really don't think this is a problem, and it's arguable that it's even an interaction, because it makes a staticcall (https://github.com/OpenZeppelin/openzeppelin-contracts/blob/63b466901fb015538913f811c5112a2775042177/contracts/utils/cryptography/SignatureChecker.sol#L35), meaning no state can be modified anywhere inside that call (and subcalls). It's more akin to making a balanceOf call, which I think is considered a check rather than an interaction.

I think this is pretty safe.

mrice32 (approved):
LGTM!

contracts/SpokePool.sol
chrismaree: then just check that: `updatableRelayerFeePct <= 0.5e18 && relayData.realizedLpFeePct <= 0.5e18`?

nicholaspai: true

PR #28, opened 2022-02-08
feat: add slow relay functionality

mrice32:
This doesn't include tests or some of the meat of the verification and fill logic just to allow us to agree on the interfaces before moving forward.

### DIFF
--- a/contracts/HubPool.sol
+++ b/contracts/HubPool.sol
@@ -29,7 +29,7 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
 
     struct RefundRequest {
         uint64 requestExpirationTimestamp;
-        uint64 unclaimedPoolRebalanceLeafLeafCount;
+        uint64 unclaimedPoolRebalanceLeafCount;
         bytes32 poolRebalanceRoot;
         bytes32 destinationDistributionRoot;
         bytes32 slowRelayFulfillmentRoot;
@@ -113,7 +113,7 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
 
     event InitiateRefundRequested(
         uint64 requestExpirationTimestamp,
-        uint64 unclaimedPoolRebalanceLeafLeafCount,
+        uint64 unclaimedPoolRebalanceLeafCount,
         uint256[] bundleEvaluationBlockNumbers,
         bytes32 indexed poolRebalanceRoot,
         bytes32 indexed destinationDistributionRoot,
@@ -133,7 +133,7 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
     event RelayerRefundDisputed(address indexed disputer, uint256 requestTime, bytes disputedAncillaryData);
 
     modifier noActiveRequests() {
-        require(refundRequest.unclaimedPoolRebalanceLeafLeafCount == 0, "Active request has unclaimed leafs");
+        require(refundRequest.unclaimedPoolRebalanceLeafCount == 0, "Active request has unclaimed leafs");
         _;
     }
 
@@ -277,19 +277,19 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
     // initiateRelayerRefund can't be called again until all leafs are executed.
     function initiateRelayerRefund(
         uint256[] memory bundleEvaluationBlockNumbers,
-        uint64 poolRebalanceLeafLeafCount,
+        uint64 poolRebalanceLeafCount,
         bytes32 poolRebalanceRoot,
         bytes32 destinationDistributionRoot,
         bytes32 slowRelayFulfillmentRoot
     ) public noActiveRequests {
-        require(poolRebalanceLeafLeafCount > 0, "Bundle must have at least 1 leaf");
+        require(poolRebalanceLeafCount > 0, "Bundle must have at least 1 leaf");
 
         uint64 requestExpirationTimestamp = uint64(getCurrentTime() + refundProposalLiveness);
 
         delete refundRequest; // Remove the existing information relating to the previous relayer refund request.
 
         refundRequest.requestExpirationTimestamp = requestExpirationTimestamp;
-        refundRequest.unclaimedPoolRebalanceLeafLeafCount = poolRebalanceLeafLeafCount;
+        refundRequest.unclaimedPoolRebalanceLeafCount = poolRebalanceLeafCount;
         refundRequest.poolRebalanceRoot = poolRebalanceRoot;
         refundRequest.destinationDistributionRoot = destinationDistributionRoot;
         refundRequest.slowRelayFulfillmentRoot = slowRelayFulfillmentRoot;
@@ -300,7 +300,7 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
 
         emit InitiateRefundRequested(
             requestExpirationTimestamp,
-            poolRebalanceLeafLeafCount,
+            poolRebalanceLeafCount,
             bundleEvaluationBlockNumbers,
             poolRebalanceRoot,
             destinationDistributionRoot,
@@ -309,45 +309,45 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
         );
     }
 
-    function executeRelayerRefund(PoolRebalanceLeaf memory poolRebalanceLeafLeaf, bytes32[] memory proof) public {
+    function executeRelayerRefund(PoolRebalanceLeaf memory poolRebalanceLeaf, bytes32[] memory proof) public {
         require(getCurrentTime() >= refundRequest.requestExpirationTimestamp, "Not passed liveness");
 
-        // Verify the leafId in the poolRebalanceLeafLeaf has not yet been claimed.
-        require(!MerkleLib.isClaimed1D(refundRequest.claimedBitMap, poolRebalanceLeafLeaf.leafId), "Already claimed");
+        // Verify the leafId in the poolRebalanceLeaf has not yet been claimed.
+        require(!MerkleLib.isClaimed1D(refundRequest.claimedBitMap, poolRebalanceLeaf.leafId), "Already claimed");
 
         // Verify the props provided generate a leaf that, along with the proof, are included in the merkle root.
         require(
-            MerkleLib.verifyPoolRebalanceLeaf(refundRequest.poolRebalanceRoot, poolRebalanceLeafLeaf, proof),
+            MerkleLib.verifyPoolRebalanceLeaf(refundRequest.poolRebalanceRoot, poolRebalanceLeaf, proof),
             "Bad Proof"
         );
 
         // Set the leafId in the claimed bitmap.
-        refundRequest.claimedBitMap = MerkleLib.setClaimed1D(refundRequest.claimedBitMap, poolRebalanceLeafLeaf.leafId);
+        refundRequest.claimedBitMap = MerkleLib.setClaimed1D(refundRequest.claimedBitMap, poolRebalanceLeaf.leafId);
 
-        // Decrement the unclaimedPoolRebalanceLeafLeafCount.
-        refundRequest.unclaimedPoolRebalanceLeafLeafCount--;
+        // Decrement the unclaimedPoolRebalanceLeafCount.
+        refundRequest.unclaimedPoolRebalanceLeafCount--;
 
         _sendTokensToChainAndUpdatePooledTokenTrackers(
-            poolRebalanceLeafLeaf.chainId,
-            poolRebalanceLeafLeaf.l1Tokens,
-            poolRebalanceLeafLeaf.netSendAmounts,
-            poolRebalanceLeafLeaf.bundleLpFees
+            poolRebalanceLeaf.chainId,
+            poolRebalanceLeaf.l1Tokens,
+            poolRebalanceLeaf.netSendAmounts,
+            poolRebalanceLeaf.bundleLpFees
         );
-        _executeRelayerRefundOnChain(poolRebalanceLeafLeaf.chainId);
+        _executeRelayerRefundOnChain(poolRebalanceLeaf.chainId);
 
         // Transfer the bondAmount to back to the proposer, if this the last executed leaf. Only sending this once all
         // leafs have been executed acts to force the data worker to execute all bundles or they wont receive their bond.
         //TODO: consider if we want to reward the proposer. if so, this is where we should do it.
-        if (refundRequest.unclaimedPoolRebalanceLeafLeafCount == 0)
+        if (refundRequest.unclaimedPoolRebalanceLeafCount == 0)
             bondToken.safeTransfer(refundRequest.proposer, bondAmount);
 
         emit RelayerRefundExecuted(
-            poolRebalanceLeafLeaf.leafId,
-            poolRebalanceLeafLeaf.chainId,
-            poolRebalanceLeafLeaf.l1Tokens,
-            poolRebalanceLeafLeaf.bundleLpFees,
-            poolRebalanceLeafLeaf.netSendAmounts,
-            poolRebalanceLeafLeaf.runningBalances,
+            poolRebalanceLeaf.leafId,
+            poolRebalanceLeaf.chainId,
+            poolRebalanceLeaf.l1Tokens,
+            poolRebalanceLeaf.bundleLpFees,
+            poolRebalanceLeaf.netSendAmounts,
+            poolRebalanceLeaf.runningBalances,
             msg.sender
         );
     }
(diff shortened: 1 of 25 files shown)


contracts/HubPool.sol
```diff
@@ -40,6 +40,7 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
         uint64 unclaimedPoolRebalanceLeafCount;
         bytes32 poolRebalanceRoot;
         bytes32 destinationDistributionRoot;
+        bytes32 slowRelayDataRoot;
```
chrismaree, line 43: you'll need to add this to `initiateRelayerRefund`. I can do that in a separate PR though, If you'd like as this will break a bunch of existing tests.

contracts/MerkleLib.sol
```diff
@@ -51,6 +51,18 @@ library MerkleLib {
         uint256[] refundAmounts;
     }
 
+    // TODO: should these structs be in this file?
+    struct RelayData {
```
chrismaree, line 55: would a better name be `SlowRelayFulfilment`? then use this through out, such as `verifySlowRelayFufilment`

contracts/SpokePool.sol
```diff
@@ -268,15 +259,32 @@ abstract contract SpokePool is Testable, Lockable, MultiCaller {
...
 
+    function fillRelaySlow(
+        address depositor,
+        address recipient,
+        address destinationToken,
+        uint64 realizedLpFeePct,
+        uint64 relayerFeePct,
+        uint64 depositId,
+        uint256 originChainId,
+        uint256 totalRelayAmount,
```
chrismaree, line 270: this should be called `relayAmount` to be consistant with `RelayData`, no?

contracts/HubPool.sol
chrismaree: I dont like this variable name. can it be `SlowRelayFulfilmentRoot`?

mrice32: Oh, I thought I added it everywhere, will do another pass to better understand where I missed.

On the naming, sure!

contracts/MerkleLib.sol
mrice32: Well, it's tricky here because this is _the exact_ data structure the SpokePool uses internally to represent a relay. I don't know if we want to rename that struct.

contracts/SpokePool.sol
mrice32: I just copied the parameters from the fill method, thinking they would be identical. Will double check that assumption.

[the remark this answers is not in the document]
mrice32: I was trying to mirror the fill method, as I think we will effectively have 3 fill methods that will have _mostly_ the same parameters. Would like these to be as consistent as possible, so if we use a struct here, maybe it makes sense to use a struct in the others?

contracts/MerkleLib.sol
nicholaspai: move this to SpokePool interface?

contracts/SpokePool.sol
nicholaspai: Should this be prefixed with a "distribute" instead of a "fill" to indicate that its disbursing funds after verifying that a leaf (i.e. the relay) is included in a root stored on here from an original L1 txn?

nicholaspai (approved):
This skeleton makes total sense +1

contracts/SpokePool.sol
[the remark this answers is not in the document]
chrismaree: ye, Ideally we pass around structs in all these methods methinks

contracts/MerkleLib.sol
chrismaree: hhhmm. ye I see your point. not ideal.

if we move it to the `SpokePool` interface then we'll get the MerkleLib to import from this interface, yes? equally, we can pull the Hub's Merkle `structs` to a `HubPool` interface and import them into the `MerkleLib` in the same way.

contracts/SpokePool.sol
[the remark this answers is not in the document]
mrice32: The only issue is the interpretability on etherscan :/

chrismaree: ye, @nicholaspai your framing would be good on that.

contracts/HubPool.sol
mrice32: Done on both.

contracts/MerkleLib.sol
mrice32: Renamed!

contracts/SpokePool.sol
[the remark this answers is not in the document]
chrismaree: we do this on v1 and you can decode them I think.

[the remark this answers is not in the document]
chrismaree: I take that back! looks like Etherscan has removed this for some reason :/

image


this is a settlement tx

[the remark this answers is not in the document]
chrismaree: we might need to re-evaluate the interface in the hub pool as well to spread the props, like you do it here.

chrismaree (approved):

contracts/SpokePool.sol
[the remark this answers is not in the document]
mrice32: Yeah, this definitely used to work :((

mrice32: Done.

chrismaree (approved):
this is awesome! good work on re-factoring the interfaces like this.

contracts/HubPool.sol
```diff
@@ -36,6 +37,7 @@ contract HubPool is Testable, Lockable, MultiCaller, Ownable {
         uint64 unclaimedPoolRebalanceLeafCount;
         bytes32 poolRebalanceRoot;
         bytes32 destinationDistributionRoot;
+        bytes32 slowRelayFulfilmentRoot;
```
nicholaspai, line 40: nit: I hate to be that grammar guy but fulfillment has two L's in the "fill" part

contracts/MerkleLib.sol
```diff
@@ -74,12 +31,28 @@ library MerkleLib {
...
     function verifyRelayerDistribution(
         bytes32 root,
-        DestinationDistribution memory distribution,
+        SpokePoolInterface.DestinationDistribution memory distribution,
         bytes32[] memory proof
     ) public pure returns (bool) {
         return MerkleProof.verify(proof, root, keccak256(abi.encode(distribution)));
     }
 
+    // TODO: imported RelayData for expedience. These structs should either all go in their source files or all go in this lib.
```
nicholaspai, line 40:
```suggestion
```

contracts/MerkleLib.sol
```diff
@@ -74,12 +31,28 @@ library MerkleLib {
...
+    // It seems like them all going in their source files makes sense?
```
nicholaspai, line 41:
```suggestion
```

contracts/MerkleLib.sol
```diff
@@ -2,56 +2,13 @@
...
 
 import "@openzeppelin/contracts/utils/cryptography/MerkleProof.sol";
+import "./SpokePoolInterface.sol";
+import "./HubPoolInterface.sol";
 
 /**
  * @notice Library to help with merkle roots, proofs, and claims.
  */
 library MerkleLib {
-    // TODO: some of these data structures can be moved out if convenient.
```
nicholaspai, line 10: +1 moving these to the spoke/hub pool interface i think that makes a lot of sense

contracts/SpokePool.sol
```diff
@@ -55,6 +55,8 @@ abstract contract SpokePool is SpokePoolInterface, Testable, Lockable, MultiCall
     mapping(address => mapping(uint256 => bool)) public enabledDepositRoutes;
 
     struct RelayerRefund {
+        // Merkle root of slow relays
```
nicholaspai, line 58:
```suggestion
        // Merkle root of slow relays that were not fully filled and whose recipient is still owed funds from the LP pool.
```

contracts/SpokePool.sol
```diff
@@ -480,41 +532,60 @@ abstract contract SpokePool is SpokePoolInterface, Testable, Lockable, MultiCall
             relayFills[relayHash] += fillAmountPreFees;
             // If relay token is weth then unwrap and send eth.
             if (relayData.destinationToken == address(weth)) {
-                IERC20(relayData.destinationToken).safeTransferFrom(msg.sender, address(this), amountToSend);
+                if (!isSlowRelay)
```
nicholaspai, line 535: Can you add a comment explaining that the source of funds to pay the recipient is either the caller+relayer or this contract depending on whether this is a slow relay

contracts/SpokePoolInterface.sol
```diff
@@ -1,9 +1,48 @@
 //SPDX-License-Identifier: Unlicense
 pragma solidity ^0.8.0;
 
-import "./MerkleLib.sol";
-
 interface SpokePoolInterface {
+    // This leaf is meant to be decoded in the SpokePool in order to pay out individual relayers for this bundle.
+    struct DestinationDistribution {
```
nicholaspai, line 6: nit: same as PoolRebalance, maybe add "Leaf" to end of this name

nicholaspai (approved):
Thanks for merging in my SpokePool changes! I left mostly nits on naming + comments and one suggestion for which variable to pass into the internal _fillRelay method. I also appreciate you not adding tests yet so we can re-review them in a second pass or a follow up PR. I'll be sure to follow the same commit pattern on my PR's!

contracts/HubPool.sol
mrice32: Thank you!

I don't mean to throw @chrismaree under the bus, but I just copy-pasted this from his suggestion haha

contracts/SpokePool.sol
```diff
@@ -327,19 +338,54 @@ abstract contract SpokePool is SpokePoolInterface, Testable, Lockable, MultiCall
...
+            relayAmount: totalRelayAmount
+        });
+
+        require(
+            MerkleLib.verifyRelayData(relayerRefunds[relayerRefundId].slowRelayFulfilmentRoot, relayData, proof),
+            "Invalid proof"
+        );
+
+        bytes32 relayHash = _getRelayHash(relayData);
+        uint256 fillAmountPreFees = _fillRelay(relayHash, relayData, relayerFeePct, type(uint256).max / 1e18, true);
```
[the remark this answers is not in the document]
mrice32: Interesting idea -- what you're saying is that relayAmount >= total send, so this will always result in the relay being completely filled. I like this. Will add an explanatory comment too.

contracts/SpokePool.sol
mrice32: yep!

contracts/HubPoolInterface.sol
```diff
@@ -0,0 +1,72 @@
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity ^0.8.0;
+
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import "./chain-adapters/AdapterInterface.sol";
+
+interface HubPoolInterface {
+    struct PoolRebalance {
```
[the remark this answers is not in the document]
mrice32: Agree, will try to make this change. Hopefully will just be a simple find and replace.

contracts/HubPool.sol
mrice32: Will fix.

contracts/MerkleLib.sol
```diff
@@ -2,65 +2,22 @@
...
-    }
-
     /**
      * @notice Verifies that a repayment is contained within a merkle root.
      * @param root the merkle root.
      * @param rebalance the rebalance struct.
      * @param proof the merkle proof.
      */
-    function verifyPoolRebalance(
+    function verifyPoolRebalanceLeaf(
```
nicholaspai, line 18: why rename this function to end with "Leaf"? I think its not neccessary since you take in a param called "leaf", but if you do go this route then I think you should end `verifyRelayData` and `verifyRelayerDistribution` also with "Leaf"

nicholaspai:
This LGTM but just be consistent on using "leaf" in the "verify-" function names

contracts/MerkleLib.sol
nicholaspai: Perhaps `verifyRelayData` is more appropriately named `verifyRelayDataHashLeaf` or `verifyRelayHashLeaf`

mrice32: Okay, fixed on both! I reverted the Leaf change here (it was a find-and-replace issue I think)

contracts/SpokePool.sol
```diff
@@ -16,6 +16,7 @@ import "@uma/core/contracts/common/implementation/Lockable.sol";
 import "@uma/core/contracts/common/implementation/MultiCaller.sol";
 import "./MerkleLib.sol";
 import "./SpokePoolInterface.sol";
+import "hardhat/console.sol";
```
chrismaree, line 19: @mrice32 nit

contracts/SpokePool.sol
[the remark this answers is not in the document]
chrismaree: I'll make this change in the hub :(((((

contracts/test/MockSpokePool.sol
```diff
@@ -37,9 +37,12 @@ contract MockSpokePool is SpokePoolInterface, SpokePool {
         _setDepositQuoteTimeBuffer(buffer);
     }
 
-    function initializeRelayerRefund(bytes32 relayerRepaymentDistributionProof) public override {
-        _initializeRelayerRefund(relayerRepaymentDistributionProof);
+    function initializeRelayerRefund(bytes32 relayerRepaymentDistributionRoot, bytes32 slowRelayFulfillmentRoot)
+        public
+        override
+    {
+        _initializeRelayerRefund(relayerRepaymentDistributionRoot, slowRelayFulfillmentRoot);
```
chrismaree, line 44: I know you've changed this through the play but do we think that changing these variable names would make our linting a bit easier? `relayerRepaymentDistributionRoot` is a _very_ long variable name. would `relayerRepaymentRoot` be better? and `slowRelayFulfillmentRoot` to `slowRelayRoot`? My goal is that we have a lot of functions and variables that are wrapping onto multiple lines and would perhaps not wrap if they were a bit shorter.

dont worry about this for now but just something to think about. IMHO as much as posible having code on _one_ line will make things a lot easier to read and generally makes the code look much cleaner.

package.json
```diff
@@ -58,6 +58,9 @@
     "typechain": "^5.1.2",
     "typescript": "^4.5.2"
   },
+  "engines": {
+    "node": ">=14.0.0"
+  },
```
chrismaree, line 63: why?

test/HubPool.Fixture.ts
```diff
@@ -1,7 +1,7 @@
 import { TokenRolesEnum } from "@uma/common";
 import { getContractFactory, randomAddress, toBN, fromWei } from "./utils";
 
-import { bondAmount, refundProposalLiveness, finalFee, identifier, repaymentChainId } from "./constants";
```
chrismaree, line 4: we should lint for unused variables.

test/HubPool.Fixture.ts
```diff
@@ -1,7 +1,7 @@
...
+import { bondAmount, refundProposalLiveness, finalFee, repaymentChainId } from "./constants";
 import { Contract, Signer } from "ethers";
 import hre from "hardhat";
```
chrismaree, line 6: both of these lines can be imported from `utils`. dont worry about it now, though. I'll do a clean up PR later.

test/HubPool.PooledTokenSynchronization.ts
```diff
@@ -33,7 +33,9 @@ describe("HubPool Pooled Token Synchronization", function () {
 
     // Execute a relayer refund. Check counters move accordingly.
     const { tokensSendToL2, realizedLpFees, leafs, tree } = await constructSingleChainTree(weth);
-    await hubPool.connect(dataWorker).initiateRelayerRefund([3117], 1, tree.getHexRoot(), consts.mockTreeRoot);
+    await hubPool
+      .connect(dataWorker)
+      .initiateRelayerRefund([3117], 1, tree.getHexRoot(), consts.mockTreeRoot, consts.mockSlowRelayFulfillmentRoot);
```
chrismaree, line 38: here is another example of the long ass variable names and how it makes the code ugly.

chrismaree (approved):
This is great! I left a few nits but overall this is excellent work! Feel free to merge once you've implemented the nits.

package.json
mrice32: I ran some command that failed saying it was incompatible with node version 8, and adding this fixed it. I can't remember what command it was, and I can't seem to find it in my local history, so I'll remove.

contracts/SpokePool.sol
mrice32: Thanks, will remove!

contracts/test/MockSpokePool.sol
mrice32: Yeah, I'll leave for a follow-up. We've had a lot of back and forth on naming in this PR, and just to avoid doing a bunch more find-and-replace, we can do this in one shot in a follow up if that's cool.

test/HubPool.Fixture.ts
mrice32: Fixed for this file. Haven't checked other files tho.

test/HubPool.PooledTokenSynchronization.ts
mrice32: Yeah, we can update these when we update the other names. I agree that shorter names would be better! So let's decide offline what these should be called and do a big naming PR.

test/HubPool.Fixture.ts
mrice32: Agree! We should turn on eslint, as it will catch stuff like this. I think these are all used, though.

PR #32, opened 2022-02-14
feat(spoke-pool): Reduce size of state variables

nicholaspai:
Percentages and timestamp-based variables can be smaller size to save gas costs for storing data

### DIFF
--- a/contracts/Optimism_SpokePool.sol
+++ b/contracts/Optimism_SpokePool.sol
@@ -23,11 +23,11 @@ contract Optimism_SpokePool is CrossDomainEnabled, SpokePoolInterface, SpokePool
         address _crossDomainAdmin,
         address _hubPool,
         address _wethAddress,
-        uint64 _depositQuoteTimeBuffer,
-        address timerAddress
+        address timerAddress,
+        uint32 _depositQuoteTimeBuffer
     )
         CrossDomainEnabled(Lib_PredeployAddresses.L2_CROSS_DOMAIN_MESSENGER)
-        SpokePool(_crossDomainAdmin, _hubPool, _wethAddress, _depositQuoteTimeBuffer, timerAddress)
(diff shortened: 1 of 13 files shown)


contracts/Optimism_SpokePool.sol
```diff
@@ -23,11 +23,11 @@ contract Optimism_SpokePool is CrossDomainEnabled, SpokePoolInterface, SpokePool
         address _crossDomainAdmin,
         address _hubPool,
         address _wethAddress,
-        uint64 _depositQuoteTimeBuffer,
-        address timerAddress
+        address timerAddress,
+        uint32 _depositQuoteTimeBuffer
```
chrismaree, line 27: should we default this to what we use in prod (10 mins) then remove it from the constructor? this is what I did in the hub pool: as much as posible remove the instantiation of constructor variables if we know what they will be in production. thoughts?

nicholaspai: yea lets do it

contracts/SpokePool.sol
```diff
@@ -42,21 +42,21 @@ abstract contract SpokePool is SpokePoolInterface, Testable, Lockable, MultiCall
...
 
     // Any deposit quote times greater than or less than this value to the current contract time is blocked. Forces
-    // caller to use an up to date realized fee.
-    uint64 public depositQuoteTimeBuffer;
+    // caller to use an up to date realized fee. Defaults to 10 minutes.
+    uint32 public depositQuoteTimeBuffer = 600;
 
     // Use count of deposits as unique deposit identifier.
-    uint64 public numberOfDeposits;
+    uint32 public numberOfDeposits;
```
mrice32, line 52: nit: can we move these down below weth, since I assume weth _and_ all of these will be read on each deposit, so we'll end up saving one word's worth a reads by grouping them.

mrice32 (approved):

contracts/SpokePool.sol
nicholaspai: nice done

PR #33, opened 2022-02-14
feat: Add Arbitrum Spokepool and add generic cross-chain admin relayer function on HubPool

nicholaspai:


contracts/HubPool.sol
```diff
@@ -478,7 +498,6 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
         uint256[] memory bundleLpFees
     ) internal {
         AdapterInterface adapter = crossChainContracts[chainId].adapter;
-        require(address(adapter) != address(0), "Adapter not set for target chain");
```
nicholaspai, line 481: do we need this check? elsewhere in `_executeRelayerRefundOnChain` for example, we don't check that the adapter address is non0

contracts/chain-adapters/L1_Adapter.sol
```diff
@@ -0,0 +1,50 @@
...
+import "./Base_Adapter.sol";
+import "../interfaces/AdapterInterface.sol";
+import "../interfaces/WETH9.sol";
+
+import "@uma/core/contracts/common/implementation/Lockable.sol";
+
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+
+contract L1_Adapter is Base_Adapter, Lockable {
```
nicholaspai, line 13: @mrice32 @chrismaree I originally added this because I wanted to test `HubPool.relaySpokePoolAdminFunction()` in the simplest way and check that it can delegate a call to another contract, so I figured why not just build this L1 adapter to kill two birds with one stone

test/HubPool.Admin.ts
```diff
@@ -36,6 +36,7 @@ describe("HubPool Admin functions", function () {
     await expect(hubPool.connect(other).disableL1TokenForLiquidityProvision(weth.address)).to.be.reverted;
   });
   it("Can whitelist route for deposits and rebalances", async function () {
+    await hubPool.setCrossChainContracts(destinationChainId, mockAdapter.address, mockSpoke.address);
```
nicholaspai, line 39: Neccessary now since `whitelistRoute` calls `relayMessage`

contracts/Arbitrum_SpokePool.sol
```diff
@@ -0,0 +1,118 @@
...
+ * @dev Uses AVM cross-domain-enabled logic for access control.
+ */
+
+contract Arbitrum_SpokePool is SpokePoolInterface, SpokePool {
+    // Address of the Arbitrum L2 token gateway.
+    address public l2GatewayRouter;
+
+    // Admin controlled mapping of arbitrum tokens to L1 counterpart. L1 counterpart addresses
+    // are neccessary to bridge tokens to L1.
+    mapping(address => address) public whitelistedTokens;
```
chrismaree, line 27: pain that we need this just for arbitrum but works kinda well given we have separate contracts for each chain.

contracts/HubPool.sol
```diff
@@ -184,8 +196,16 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
...
         address destinationToken
     ) public onlyOwner {
-        // TODO: Should relay message to L2 for destinationChainId and call setEnableRoute(originToken, destinationChainId, true)
         whitelistedRoutes[originToken][destinationChainId] = destinationToken;
+        relaySpokePoolAdminFunction(
+            destinationChainId,
+            abi.encodeWithSignature(
+                "setEnableRoute(address,uint32,bool)",
+                originToken,
+                uint32(destinationChainId),
```
chrismaree, line 205: unrelated to this PR but I think that the unit32 is too small.

contracts/chain-adapters/L1_Adapter.sol
```diff
@@ -0,0 +1,50 @@
...
+    using SafeERC20 for IERC20;
+
+    constructor(address _hubPool) Base_Adapter(_hubPool) {}
+
+    function relayMessage(address target, bytes memory message) external payable override nonReentrant onlyHubPool {
+        require(_executeCall(target, message), "execute call failed");
```
chrismaree, line 19: would it be cleaner to put the `require` part in the `_executeCall` function then just call this directly? we never want the `success=false` case so we might as well assume this method does not throw if it passed. i.e the `_executeCall` would not return anything and would simply have `require(success,"execute call failed")` at the end of the function.

contracts/Arbitrum_SpokePool.sol
```diff
@@ -0,0 +1,118 @@
...
+        bytes calldata _data
+    ) external payable returns (bytes memory);
+}
+
+/**
+ * @notice AVM specific SpokePool.
+ * @dev Uses AVM cross-domain-enabled logic for access control.
+ */
+
+contract Arbitrum_SpokePool is SpokePoolInterface, SpokePool {
```
chrismaree, line 21: this does not have any tests. shall we add that in this PR or a separate one? we also need to write OP tests, which can happen either at the same time, if we choose to wait on adding the tests here.

contracts/HubPool.sol
nicholaspai: Hm what should we use?

contracts/chain-adapters/L1_Adapter.sol
nicholaspai: Sure. Why not

contracts/Arbitrum_SpokePool.sol
nicholaspai: Let's do separate PR

contracts/HubPool.sol
nicholaspai: EIP 155 doesn't list the highest value for this so I guess its safe to just use 256

contracts/chain-adapters/L1_Adapter.sol
nicholaspai: done

contracts/HubPool.sol
nicholaspai: Will fix in #35

chrismaree (approved):

PR #34, opened 2022-02-15
Add weth support for Optimism Spoke pool.

chrismaree:


contracts/HubPool.sol
```diff
@@ -207,7 +207,8 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
         address originToken,
         address destinationToken
     ) public onlyOwner {
-        // TODO: Should relay message to L2 for destinationChainId and call setEnableRoute(originToken, destinationChainId, true)
+        // Note that this method makes no L1->L1 call to whitelist the route. The assumption is that the origin chain's
+        // SpokePool Owner will call setEnableRoute to enable the route. This removes the need for an L1->L2 call.
         whitelistedRoutes[originToken][destinationChainId] = destinationToken;
```
nicholaspai, line 211: This comment no longer applies as my outstanding PR #33 will make the L1-->L2 call

contracts/Optimism_SpokePool.sol
```diff
@@ -15,11 +19,17 @@ import "./SpokePoolInterface.sol";
 
 contract Optimism_SpokePool is CrossDomainEnabled, SpokePoolInterface, SpokePool, Ownable {
     // "l1Gas" parameter used in call to bridge tokens from this contract back to L1 via `IL2ERC20Bridge`.
-    uint32 l1Gas = 6_000_000;
+    uint32 public l1Gas = 1_900_000;
```
nicholaspai, line 22: This is L1 gas. Are you sure this should be set to 1.9mil? I thought only the L2 gas param in the Optimism Adapter (which calls L1--> L2 txns) should use this value

contracts/Optimism_SpokePool.sol
```diff
@@ -87,14 +97,24 @@ contract Optimism_SpokePool is CrossDomainEnabled, SpokePoolInterface, SpokePool
...
     function _bridgeTokensToHubPool(DestinationDistributionLeaf memory distributionLeaf) internal override {
-        // TODO: Handle WETH token unwrapping
+        address l2Token = distributionLeaf.l2TokenAddress;
+        address l1Destination = hubPool;
+        // If the token being bridged is WETH then we need to first unwrap it to ETH and then send ETH over the
+        // canonical bridge. On Optimism, this is address 0xDeadDeAddeAddEAddeadDEaDDEAdDeaDDeAD0000.
+        if (distributionLeaf.l2TokenAddress == address(weth)) {
+            WETH9(l2Token).withdraw(distributionLeaf.amountToReturn); // Unwrap ETH.
+            l2Token = l2Eth; // Set the bridged token to the Optimism ETH address.
+            l1Destination = l1EthWrapper; // Set the destination to the L1 ETH wrapper.
```
nicholaspai, line 107: Why is the destination the l1 eth wrapper, how does thhis wrapper contract know who the end recipient is?

nicholaspai (approved):
Looks good left some questions.

contracts/HubPool.sol
chrismaree: yup. I saw that :)

contracts/Optimism_SpokePool.sol
chrismaree: I'm not exactly sure. tbh I'm not exactly sure how it works going from L2->L1, only really know what the costing is from L1->L2.

contracts/Optimism_SpokePool.sol
chrismaree: this will be this contract: https://github.com/across-protocol/across-smart-contracts/blob/master/contracts/insured-bridge/ovm/Optimism_Wrapper.sol which simply receives eth, wraps it and sends it to the hupPool (the destination). I'll add this contract to this PR.

chrismaree: I've added a modified version of this contract to this PR.

chrismaree: @nicholaspai  I've pretty drastically refactored this.

contracts/Lockable.sol
```diff
@@ -0,0 +1,61 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity ^0.8.0;
+
+/**
+ * @title A contract that provides modifiers to prevent reentrancy to state-changing and view-only methods. This contract
+ * is inspired by https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/ReentrancyGuard.sol
+ * and https://github.com/balancer-labs/balancer-core/blob/master/contracts/BPool.sol.
+ */
+contract Lockable {
+    bool internal _notEntered;
```
chrismaree, line 10: I had to copy this to change it to be internal so the `HubPool` can read from it.

contracts/Optimism_SpokePool.sol
nicholaspai: ah i see

contracts/HubPool.sol
nicholaspai: Can remove this comment

contracts/HubPool.sol
```diff
@@ -638,6 +637,20 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
         if (protocolFeesCaptured > 0) unclaimedAccumulatedProtocolFees[l1Token] += protocolFeesCaptured;
     }
 
-    // Added to enable the SpokePool to receive ETH. used when unwrapping WETH.
-    receive() external payable {}
+    // If _notEntered is false then this method has been called by the callback function by dropping ETH onto the
+    // contract. In this case, deposit the ETH into WETH. If true then this was set as a result of unwinding LP tokens,
+    // with the intention of sending ETH to the LP. In this case, do nothing as we intend on sending the ETH to the LP.
+    function depositEthToWeth() public payable {
+        if (_notEntered) weth.deposit{ value: address(this).balance }();
```
nicholaspai, line 644: As discussed IRL, let's expose `_notEntered` via a descriptive `public` function and call that here

contracts/Optimism_SpokePool.sol
nicholaspai: I think you can leave this at 6mil, but +1 on making it `public`

nicholaspai (approved):

PR #38, opened 2022-02-16
feat(hub-pool): Reduce size of state vars

nicholaspai:


contracts/HubPoolInterface.sol
```diff
@@ -26,6 +21,11 @@ interface HubPoolInterface {
         // A positive number indicates that the HubPool owes the SpokePool funds. A negative number indicates that the
         // SpokePool owes the HubPool funds. See the comment above for the dynamics of this and netSendAmounts
         int256[] runningBalances;
+        // Used as the index in the bitmap to track whether this leaf has been executed or not.
+        uint32 leafId;
```
nicholaspai, line 25: Should this be the same size as the `poolRebalanceLeafCount` in the HubPool functions, aka uint8?

contracts/HubPool.sol
```diff
@@ -126,26 +126,26 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
     event WhitelistRoute(uint256 destinationChainId, address originToken, address destinationToken);
 
     event InitiateRefundRequested(
-        uint64 requestExpirationTimestamp,
-        uint64 unclaimedPoolRebalanceLeafCount,
         uint256[] bundleEvaluationBlockNumbers,
+        uint32 requestExpirationTimestamp,
```
mrice32, line 130: As discussed IRL, these won't get packed no matter the order, but it's totally fine to move them for style!

mrice32 (approved):
LGTM!

contracts/HubPoolInterface.sol
chrismaree: I think yes! as long as the leafId in the relayer refund remains large we are good.

chrismaree (approved):

contracts/HubPool.sol
```diff
@@ -29,30 +29,47 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
     using Address for address;
 
     struct RefundRequest {
-        uint64 requestExpirationTimestamp;
-        uint64 unclaimedPoolRebalanceLeafCount;
+        uint256 claimedBitMap; // This is a 1D bitmap, with max size of 256 elements, limiting us to 256 chainsIds.
+        uint32 requestExpirationTimestamp;
+        uint8 unclaimedPoolRebalanceLeafCount;
```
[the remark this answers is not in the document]
nicholaspai: address and bool can be packed with small uints..?

PR #42, opened 2022-02-18
feat: add neighborhood verifier

mrice32:
This proof verification can take any "neighborhood" of a merkle tree and verify it. A neighborhood is any set of leaves that contain all the information to recreate a single branch/inner node.

Tests will be added but wanted to get feedback on the algorithm and interface first.


contracts/MerkleLib.sol
```diff
@@ -99,4 +99,57 @@ library MerkleLib {
...
+            // where the end of the usable data is.
+            while (leaves[1] != bytes32(0)) {
+                // Just increments. Relies on the break statement to terminate the loop.
+                for (uint256 i = 0; ; i++) {
+                    // Store this variable since this is needed twice.
+                    bool isEven = i % 2 == 0;
+
+                    // This just detects if this index is the end of the current data. That could be the end of the
+                    // allocated array or the end of the nonzero data.
+                    if (i == leaves.length - 1 || leaves[i + 1] == bytes32(0)) {
```
nicholaspai, line 121: What if theres a leaf equal to `bytes32(0)` in the middle of the `leaves` array? Is this not possible?

PR #43, opened 2022-02-18
feat: Optimism deployment scripts

chrismaree:


nicholaspai (approved):
~I've added all deploy scripts for HubPool, Adapter, and SpokePool, and they work but etherscan-verify does not work for HubPool and SpokePool~

Got this working after i made MerkleLib functions internal and no longer needed to link the MerkleLib library when deploying.

PR #48, opened 2022-02-23
fix(HubPool): Whitelisting route on L2 uses incorrect params

nicholaspai:
The `relayWhitelistedRoute` method should be sent to the `originChainId`. For example, imagine we want to whitelist the route from network 10 to network 42161. We want to have an identical mapping of originToken => destChainId on the originChain network and the HubPool. Previously we were relaying a mapping to the destination chain, in this case that would be 42161.

### DIFF
--- a/contracts/HubPool.sol
+++ b/contracts/HubPool.sol
@@ -136,7 +136,12 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
         uint256 lpTokensBurnt,
         address indexed liquidityProvider
     );
-    event WhitelistRoute(uint256 destinationChainId, address originToken, address destinationToken);
+    event WhitelistRoute(
+        uint256 originChainId,
+        uint256 destinationChainId,
+        address originToken,
+        address destinationToken
+    );
 
     event ProposeRootBundle(
         uint64 requestExpirationTimestamp,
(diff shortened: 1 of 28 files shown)


contracts/HubPool.sol
```diff
@@ -238,16 +243,19 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
      * @notice Whitelist an origin token <-> destination token route.
      */
     function whitelistRoute(
+        uint256 originChainId,
```
nicholaspai, line 246: Do you think adding `originChainId` to the function interface is the best way to get the L2 to know about the L2Token => destinationChainId route?

contracts/HubPool.sol
```diff
@@ -238,16 +243,19 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
...
         address originToken,
         address destinationToken
     ) public onlyOwner {
         whitelistedRoutes[originToken][destinationChainId] = destinationToken;
+
+        // Whitelist the inverse route, from the L2 destinationToken to this chainID, on the L2 spoke pool.
         _relaySpokePoolAdminFunction(
             destinationChainId,
-            abi.encodeWithSignature("setEnableRoute(address,uint256,bool)", originToken, destinationChainId, true)
+            abi.encodeWithSignature("setEnableRoute(address,uint256,bool)", destinationToken, originChainId, true)
```
chrismaree, line 256: as discussed IRL this needs to change slightly.

chrismaree (approved):

mrice32 (approved):
My understanding of this fix is that the routes were being sent to the destination chain. This would effectively result in the destination chain only being able to send to itself.

I think this fix makes sense! Nice catch!

PR #49, opened 2022-02-23
fix(optimism-spokepool): Need to wrap ETH before distributing any WETH to recipients

nicholaspai:
The Optimism token bridge sends ETH (address 0xdeaddead...) to SpokePool, so SpokePool accrues an ETH balance after pool rebalances are executed on L1. We make assumptions that the SpokePool has WETH, not ETH, so we need to wrap WETH before distributing refunds or processing relays.

Originally I wanted to implement the wrapping of ETH in the `fallback` or `receive` methods but these are not triggered by ETH deposits over the token bridge because ETH is an ERC20 token on Optimism, so I opted for this approach.


contracts/SpokePool.sol
```diff
@@ -341,6 +341,9 @@ abstract contract SpokePool is SpokePoolInterface, Testable, Lockable, MultiCall
         uint32 rootBundleId,
         bytes32[] memory proof
     ) public nonReentrant {
+        // This function potentially is expected to transfer WETH to the recipient so we should wrap all ETH.
+        _depositEthToWeth();
```
chrismaree, line 345: does it make sence to only do this if the `destinationToken` is WETH?

contracts/SpokePool.sol
```diff
@@ -371,6 +374,9 @@ abstract contract SpokePool is SpokePoolInterface, Testable, Lockable, MultiCall
         SpokePoolInterface.RelayerRefundLeaf memory relayerRefundLeaf,
         bytes32[] memory proof
     ) public nonReentrant {
+        // This function potentially is expected to transfer WETH to the recipient so we should wrap all ETH.
+        _depositEthToWeth();
```
chrismaree, line 378: follow on question: does it make sence to put this in the spoke Pool? would it not make _more_ sence to put this in the Optimism spoke pool only seeing this is only relevant on optimism?

nicholaspai: As discussed IRL i'll move to Optimism_SpokePool

contracts/SpokePool.sol
nicholaspai: sure done

mrice32 (approved):
LGTM!

PR #50, opened 2022-02-23
fix: take bond into account when syncing

mrice32:
Fixes a bug where the bond can be attributed to repayments instead of reserved for the proposer.


chrismaree (approved):
great catch and clean solution!

PR #51, opened 2022-02-25
feat: update gas estimation script

mrice32:
This PR intends to make the gas estimates more realistic.

### DIFF
--- a/contracts/chain-adapters/Mock_Adapter.sol
+++ b/contracts/chain-adapters/Mock_Adapter.sol
@@ -14,18 +14,44 @@ contract Mock_Adapter is Base_Adapter {
 
     event RelayTokensCalled(address l1Token, address l2Token, uint256 amount, address to, address caller);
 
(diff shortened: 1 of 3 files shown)


contracts/chain-adapters/Mock_Adapter.sol
```diff
@@ -14,18 +14,44 @@ contract Mock_Adapter is Base_Adapter {
...
-        emit RelayMessageCalled(target, message, msg.sender);
+    Mock_Bridge public bridge;
+
+    constructor(address _hubPool) Base_Adapter(_hubPool) {
+        bridge = new Mock_Bridge();
     }
 
-    constructor(address _hubPool) Base_Adapter(_hubPool) {}
+    function relayMessage(address target, bytes memory message) external payable override {
+        bridge.bridgeMessage(target, message);
```
nicholaspai, line 24: +1

nicholaspai (approved):
I think this makes the gas estimation script a lot more realistic and also moving the `proposeRootBundle` test to its own test makes a lot of sense

PR #53, opened 2022-02-25
Refactor SpokePool inheritance structure

mrice32:
This change is mostly just simplifying. It allows the same logic to be performed but with fewer overridden methods in each SpokePool implementation and more of the implementation work done in the base class. There is also a unified modifier that all SpokePool implementations should use to gate admin functions: `onlyAdmin`.


contracts/SpokePool.sol
```diff
@@ -154,36 +154,53 @@ abstract contract SpokePool is SpokePoolInterface, Testable, Lockable, MultiCall
...
     }
 
-    function _setEnableRoute(
+    function setEnableRoute(
         address originToken,
         uint256 destinationChainId,
         bool enabled
-    ) internal {
+    ) public override onlyAdmin nonReentrant {
+        _requireAdminSender();
```
nicholaspai, line 179: Why call both `_requireAdminSender()` and use the `onlyAdmin` modifier aren't these redundant?

contracts/SpokePool.sol
```diff
@@ -154,36 +154,53 @@ abstract contract SpokePool is SpokePoolInterface, Testable, Lockable, MultiCall
...
         enabledDepositRoutes[originToken][destinationChainId] = enabled;
         emit EnabledDepositRoute(originToken, destinationChainId, enabled);
     }
 
-    function _setDepositQuoteTimeBuffer(uint32 _depositQuoteTimeBuffer) internal {
+    function setDepositQuoteTimeBuffer(uint32 _depositQuoteTimeBuffer) public override onlyAdmin nonReentrant {
+        _requireAdminSender();
```
nicholaspai, line 185: same as https://github.com/across-protocol/across-smart-contracts-v2/pull/53/files#r815169371

contracts/SpokePool.sol
```diff
@@ -154,36 +154,53 @@ abstract contract SpokePool is SpokePoolInterface, Testable, Lockable, MultiCall
...
         emit SetDepositQuoteTimeBuffer(_depositQuoteTimeBuffer);
     }
 
+    // This internal method should be called by an external "relayRootBundle" function that validates the
+    // cross domain sender is the HubPool. This validation step differs for each L2, which is why the implementation
+    // specifics are left to the implementor of this abstract contract.
+    // Once this method is executed and a distribution root is stored in this contract, then `distributeRootBundle`
+    // can be called to execute each leaf in the root.
+    function relayRootBundle(bytes32 relayerRefundRoot, bytes32 slowRelayRoot) public override onlyAdmin nonReentrant {
+        _requireAdminSender();
```
nicholaspai, line 196: same as https://github.com/across-protocol/across-smart-contracts-v2/pull/53/files#r815169371

nicholaspai (approved):
Really like this pattern but I think you have some redundant calls

contracts/SpokePool.sol
mrice32: Yes! That's a mistake.

mrice32:
> Really like this pattern but I think you have some redundant calls

Agreed. Thank you for pointing these out. I originally was just going to do a method call, then I moved to a modifier. Forgot to clean up the old calls.

contracts/SpokePool.sol
mrice32: Fixed.

contracts/SpokePool.sol
mrice32: Fixed.

contracts/SpokePool.sol
mrice32: Fixed.

nicholaspai (approved):

PR #54, opened 2022-02-25
Finish most remaining TODOs

mrice32:


contracts/HubPool.sol
```diff
@@ -359,6 +359,8 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
         bytes32 relayerRefundRoot,
         bytes32 slowRelayRoot
     ) public override nonReentrant noActiveRequests {
+        // Note: this is to prevent "empty block" style attacks where someone can make empty proposals that are
+        // technically valid but not useful. This could also potentially be enforced at the UMIP-level.
```
nicholaspai, line 363: Do you think there's a situation where we want to publish a new relayer refund root or slow relay root to the SpokePool but we don't want to execute a pool rebalance? In this case, how do we submit an "no-op" pool rebalance root?

contracts/HubPool.sol
```diff
@@ -398,6 +400,9 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
             "Bad Proof"
         );
 
+        // Before interacting with a particular chain's adapter, ensure that the adapter is set.
+        require(address(crossChainContracts[poolRebalanceLeaf.chainId].adapter) != address(0), "No adapter for chain");
```
nicholaspai, line 404: is there any advantage to prevent setting the adapter contract address to 0x0 in the `setCrossChainContracts` func?

mrice32: Actually, I think we should actually allow this to allow a chain to be de-whitelisted.

contracts/Arbitrum_SpokePool.sol
```diff
@@ -82,6 +82,8 @@ contract Arbitrum_SpokePool is SpokePoolInterface, SpokePool {
     }
 
     // l1 addresses are transformed during l1->l2 calls. See https://developer.offchainlabs.com/docs/l1_l2_messages#address-aliasing for more information.
+    // This cannot be pulled directly from Arbitrum contracts because their contracts are not 0.8.X compatible and this operation takes advantage of
+    // overflows, whose behavior changed in 0.8.0.
```
chrismaree, line 86: add a comment on the unchecked block below

chrismaree (approved):

contracts/Arbitrum_SpokePool.sol
mrice32: Done.

PR #56, opened 2022-02-27
 feat(spoke-pool): Add speedUpDeposit method

nicholaspai:
Adds convenience method that depositor can submit to signal to relayers that they can use an updated relayer fee. Reuses the same logic that is used in the fillRelayWithUpdatedFee that relayers use to submit a relay with an updated fee.

﻿Signed-off-by: nicholaspai <npai.nyc@gmail.com>

### DIFF
--- a/contracts/SpokePool.sol
+++ b/contracts/SpokePool.sol
@@ -259,6 +259,23 @@ abstract contract SpokePool is SpokePoolInterface, Testable, Lockable, MultiCall
     }
 
     // Convenience method that depositor can use to signal to relayer to use updated fee.
+    /**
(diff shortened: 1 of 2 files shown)


mrice32 (approved):
Looks great!!

contracts/SpokePool.sol
```diff
@@ -252,6 +258,20 @@ abstract contract SpokePool is SpokePoolInterface, Testable, Lockable, MultiCall
         numberOfDeposits += 1;
     }
 
+    // Convenience method that depositor can use to signal to relayer to use updated fee.
+    function speedUpDeposit(
```
chrismaree, line 262: nice. test please

chrismaree (approved):

PR #57, opened 2022-02-27
Add polygon adapters/spoke pool

mrice32:
Wanted to get feedback on this general pattern before writing tests/mocks.


contracts/PolygonTokenBridger.sol
```diff
@@ -0,0 +1,36 @@
...
+import "../Lockable.sol";
+
+interface PolygonIERC20 is IERC20 {
+    function withdraw(uint256 amount) external;
+}
+
+// Because Polygon only allows withdrawals from a particular address to go to that same address on mainnet, we need to
+// have some sort of contract that can guarantee identical addresses on Polygon and Ethereum.
+// Note: this contract is intended to be completely immutable, so it's guaranteed that the contract on each side is
+// configured identically as long as it is created via create2.
```
nicholaspai, line 15: What does create2 actually do? A comment would help here

contracts/Polygon_SpokePool.sol
```diff
@@ -0,0 +1,81 @@
...
+    bool private callValidated = false;
+
+    event PolygonTokensBridged(address indexed token, address indexed receiver, uint256 amount);
+
+    constructor(
+        PolygonTokenBridger _polygonTokenBridger,
+        address _crossDomainAdmin,
+        address _hubPool,
+        address timerAddress
+    ) SpokePool(_crossDomainAdmin, _hubPool, 0x4200000000000000000000000000000000000006, timerAddress) {
```
chrismaree, line 41: why do you have this address here? this is not the address of WETH on polygon.

contracts/Polygon_SpokePool.sol
```diff
@@ -0,0 +1,81 @@
...
+        polygonTokenBridger = _polygonTokenBridger;
+    }
+
+    function processMessageFromRoot(
+        uint256,
```
chrismaree, line 46: perhaps add a comment saying this variable is unused and what it is

contracts/Polygon_SpokePool.sol
```diff
@@ -0,0 +1,81 @@
...
+        // Reset callValidated to false to disallow admin calls after this method exits.
+        callValidated = false;
+    }
+
+    /**************************************
+     *        INTERNAL FUNCTIONS          *
+     **************************************/
+
+    function _bridgeTokensToHubPool(RelayerRefundLeaf memory relayerRefundLeaf) internal override {
+        PolygonIERC20(relayerRefundLeaf.l2TokenAddress).safeIncreaseAllowance(relayerRefundLeaf.amountToReturn);
```
chrismaree, line 72: in the `Optimism_adapter` we use `IERC20(l1Token).approve(address(l1StandardBridge), amount);`. would it be better to change that to use `safeIncreaseAllowance` over `approve`?

chrismaree (approved):

contracts/Polygon_SpokePool.sol
```diff
@@ -0,0 +1,81 @@
...
+        polygonTokenBridger.send(relayerRefundLeaf.l2TokenAddress, relayerRefundLeaf.amountToReturn);
+
+        emit PolygonTokensBridged(relayerRefundLeaf.l2TokenAddress, address(this), relayerRefundLeaf.amountToReturn);
+    }
+
+    function _requireAdminSender() internal view override {
+        require(callValidated, "Must call processMessageFromRoot");
```
nicholaspai, line 79: This is a clever way to check that calls are only made via the fx tunnel. But could you not replace this call with a require that the msg.sender is the fxChild and remove the callValidated Boolean variable?

contracts/chain-adapters/Polygon_Adapter.sol
```diff
@@ -0,0 +1,74 @@
...
+            l1Weth.withdraw(amount);
+            rootChainManager.depositEtherFor{ value: amount }(to);
+        } else {
+            IERC20(l1Token).safeIncreaseAllowance(address(rootChainManager), amount);
+            rootChainManager.depositFor(to, l1Token, abi.encode(amount));
+        }
+        emit TokensRelayed(l1Token, l2Token, amount, to);
+    }
+
+    // Added to enable the Optimism_Adapter to receive ETH. used when unwrapping WETH.
```
nicholaspai, line 72: Nit: change Optimism to Polygon

nicholaspai (approved):
I don't really understand how the PolygonTokenBridger will work. Does each user need to deploy their own Bridger?

contracts/PolygonTokenBridger.sol
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    // Polygon side.
+    function send(PolygonIERC20 token, uint256 amount) public nonReentrant {
+        token.safeTransferFrom(msg.sender, address(this), amount);
+        token.withdraw(amount);
+    }
+
+    // Mainnet side.
+    function retrieve(IERC20 token) public nonReentrant {
```
nicholaspai, line 33: It's annoying that we will need a separate bot to execute this method but we could also call it in the executeRelayerRefund and executeSlowRelay methods which are overridable. This would be similar to how OptimismSpokePool works which also has the related problem of receiving ETH over the bridge but needing to have WETH to send to recipients.

nicholaspai (approved):
Just left one suggestion for resolving the issue (which may be a non issue) or calling TokenBridger.retrieve

contracts/Polygon_SpokePool.sol
mrice32: Fixed.

contracts/Polygon_SpokePool.sol
mrice32: Sure, I can do that.

mrice32: Done.

contracts/Polygon_SpokePool.sol
mrice32: Done.

contracts/Polygon_SpokePool.sol
mrice32: Yes! This was something that I considered. The particular case that I'm worried about (and maybe I shouldn't be) is that the polygon fxChild contract could make _other_ calls into contracts (now or in the future) on behalf of mainnet callers. If 1) those methods happen to be the same name/args as one of our admin methods or 2) (more likely) there's a way to partially or fully customize the selector used (hash of method name + arg types), someone could make a malicious call that comes from that contract but hasn't had its sender validated.

The types of attacks I'm worried about are ones in the style of the poly network hack, where someone is able to use user-inputs to force a contract to call another contract in a way that it wasn't intended to. https://twitter.com/kelvinfichter/status/1425217065690992641?s=20&t=OL5ecSja3ii-LktomykOOg.

I doubt this is possible on polygon today, but their contracts are upgradable and I just figure it's better to be extra careful.

Does that make sense? Thoughts?

contracts/chain-adapters/Polygon_Adapter.sol
mrice32: +1

contracts/PolygonTokenBridger.sol
mrice32: Will add.

contracts/PolygonTokenBridger.sol
mrice32: The retrieve method would need to be called on the mainnet side, right? executeRelayerRefund/executeSlowRelay are SpokePool methods, which would only happen on polygon, not mainnet right?

contracts/Polygon_SpokePool.sol
nicholaspai: I think that's a reasonable concern and one worth documenting! Two comments:
- I think you should add the `callValidated` flipping on/ff as a modifier for readability
- You should add a comment explaining why checking msg.sender == fxChild is not sufficient because of the reasons you explained above

contracts/PolygonTokenBridger.sol
nicholaspai: Good point. I thought this TokenBridger contract is meant to assist with tokens sent from HubPool --> Polygon_SpokePool, and back, so couldn't we use the above strategy to at least automate token retrievals on the Polygon side?

mrice32: I don't think we need any assistance on the polygon side. I thought the tokens get deposited/sent automatically, no?

contracts/Polygon_SpokePool.sol
mrice32: Done.

PR #64, opened 2022-03-02
improve: Change pool rebalance leaf count type to uint8

nicholaspai:
Max leaf count can only be uint8 to be consistent with size of `RootBundle.claimedBitMap` and is the same size passed into `proposeRootBundle`, so this is likely just fixing a mistake.

### DIFF
--- a/contracts/HubPool.sol
+++ b/contracts/HubPool.sol
@@ -50,11 +50,6 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
     // slowRelayRoot to a SpokePool. The latter two roots, once published to the SpokePool, contain
     // leaves that can be executed on the SpokePool to pay relayers or recipients.
     struct RootBundle {
-        // Number of pool rebalance leaves to execute in the poolRebalanceRoot. After this number
-        // of leaves are executed, a new root bundle can be proposed
-        uint8 unclaimedPoolRebalanceLeafCount;
-        // When root bundle challenge period passes and this root bundle becomes executable.
-        uint64 requestExpirationTimestamp;
         // Contains leaves instructing this contract to send funds to SpokePools.
         bytes32 poolRebalanceRoot;
         // Relayer refund merkle root to be published to a SpokePool.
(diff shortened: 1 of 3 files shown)


contracts/HubPool.sol
```diff
@@ -50,11 +50,11 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
     // slowRelayRoot to a SpokePool. The latter two roots, once published to the SpokePool, contain
     // leaves that can be executed on the SpokePool to pay relayers or recipients.
     struct RootBundle {
-        // When root bundle challenge period passes and this root bundle becomes executable.
-        uint64 requestExpirationTimestamp;
         // Number of pool rebalance leaves to execute in the poolRebalanceRoot. After this number
         // of leaves are executed, a new root bundle can be proposed
-        uint64 unclaimedPoolRebalanceLeafCount;
+        uint8 unclaimedPoolRebalanceLeafCount;
```
nicholaspai, line 55: reordered for style from small to big

chrismaree (approved):
o damn. good catch.

contracts/HubPool.sol
```diff
@@ -50,11 +50,11 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
...
+        // When root bundle challenge period passes and this root bundle becomes executable.
+        uint64 requestExpirationTimestamp;
```
mrice32, line 57: If we change this to a uint32 (which is safe until the year 2106), we should be able to group these variables with proposer and proposerBondRepaid, saving us a storage slot.

mrice32 (approved):
Looks good!

contracts/HubPool.sol
nicholaspai: sure

nicholaspai: Changed to uint32 and copied and pasted these two vars below `proposerBondRepaid`. Is that what you had in mind?

mrice32 (approved):
LGTM!

PR #65, opened 2022-03-03
feat: Make token bridge programmable per tokens

nicholaspai:


contracts/Optimism_SpokePool.sol
```diff
@@ -20,8 +20,12 @@ contract Optimism_SpokePool is CrossDomainEnabled, SpokePool {
     // ETH is an ERC20 on OVM.
     address public l2Eth = address(Lib_PredeployAddresses.OVM_ETH);
 
+    // Stores alternative token bridges to use for L2 tokens that don't go over the standard bridge.
```
chrismaree, line 23:
```suggestion
    // Stores alternative token bridges to use for L2 tokens that don't go over the standard bridge. This is needed
    // to suport non-standard ERC20 tokens on Optimism, such as DIA and SNX which both use custom bridges.
```

chrismaree (approved):
lgtm.

mrice32 (approved):

contracts/Optimism_SpokePool.sol
```diff
@@ -122,13 +137,17 @@ contract Optimism_SpokePool is CrossDomainEnabled, SpokePool {
...
+        IL2ERC20Bridge(
+            tokenBridges[relayerRefundLeaf.l2TokenAddress] == address(0)
+                ? Lib_PredeployAddresses.L2_STANDARD_BRIDGE
+                : tokenBridges[relayerRefundLeaf.l2TokenAddress]
+        ).withdrawTo(
+                relayerRefundLeaf.l2TokenAddress, // _l2Token. Address of the L2 token to bridge over.
+                hubPool, // _to. Withdraw, over the bridge, to the l1 pool contract.
+                relayerRefundLeaf.amountToReturn, // _amount.
+                l1Gas, // _l1Gas. Unused, but included for potential forward compatibility considerations
+                "" // _data. We don't need to send any data for the bridging action.
```
ericDeCourcy, line 149: since this can be used in some "nonstandard" ERC20 withdrawal case (as referenced for DIA and SNX's custom withdrawals above) - does it make sense to actually leave this data parameter open for input? If it's a standard optimism withdrawal then `data` can just be ignored, but otherwise it could be useful if any nonstandard withdrawal needs more parameters than the ones included already here

ericDeCourcy: After talking this out on slack, we agreed it's best to leave as is, and update the contract with an upgrade if needed in the future

PR #69, opened 2022-03-04
improve(spoke-pool): Slow relay execution should emit FillRelay event

nicholaspai:


chrismaree (approved):
good catch!

contracts/SpokePool.sol
```diff
@@ -625,6 +625,10 @@ abstract contract SpokePool is SpokePoolInterface, Testable, Lockable, MultiCall
         uint256 fillAmountPreFees = _fillRelay(relayHash, relayData, relayData.amount, relayerFeePct, true);
 
         _emitExecutedSlowRelayRoot(relayHash, fillAmountPreFees, relayData);
+
+        // Note: Set repayment chain ID to 0 to indicate that there is no repayment to be made. The off-chain data
+        // worker can use repaymentChainId=0 as a signal to ignore such relays for refunds.
+        _emitFillRelay(relayHash, fillAmountPreFees, 0, relayerFeePct, relayData);
```
mrice32, line 631: I think _not_ emitting this event was intentional. What benefit does this one provide over using the other event that's already being emitted?

test/SpokePool.SlowRelay.ts
```diff
@@ -151,12 +160,13 @@ describe("SpokePool Slow Relay Logic", async function () {
           )
         )
     )
-      .to.emit(spokePool, "ExecutedSlowRelayRoot")
```
mrice32, line 154: Why do we no longer check this event?

PR #70, opened 2022-03-04
changes for npm release

chrismaree:

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,3 +1,2 @@
 hardhat.config.ts
 scripts
-test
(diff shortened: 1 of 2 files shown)


.npmignore
```diff
@@ -1,3 +1,2 @@
 hardhat.config.ts
 scripts
-test
```
chrismaree, line 3: I want to the NPM package to include the tests so we can access some of the helpers and utils we've created along the way.

nicholaspai (approved):

PR #75, opened 2022-03-07
improve: Add isSlowRelay param to FilledRelay event

nicholaspai:
To reduce expected client event queries, overload the `FilledRelay` event to indicate both Slow and Fast relays. Add an `isSlowRelay` param to the event to clearly indicate whether a relay is slow or not. Currently the `repaymentChainId` is also set to 0 to indicate a slow relay but this PR allows for the possibility for the slow relayer to indicate a chain to be repaid on.


mrice32 (approved):
LGTM!

PR #76, opened 2022-03-07
fix: Check if spoke pool is initialized before executing root bundle

nicholaspai:


test/HubPool.Admin.ts
```diff
@@ -41,8 +41,15 @@ describe("HubPool Admin functions", function () {
   it("Only owner can disable L1 Tokens for liquidity provision", async function () {
```
nicholaspai, line 41: Unrelated to PR but these onlyOwner tests are missing

contracts/HubPool.sol
```diff
@@ -554,6 +554,13 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
         // Before interacting with a particular chain's adapter, ensure that the adapter is set.
         require(address(crossChainContracts[poolRebalanceLeaf.chainId].adapter) != address(0), "No adapter for chain");
 
+        // Make sure SpokePool address is initialized since `_sendTokensToChainAndUpdatePooledTokenTrackers` will not
```
chrismaree, line 557: nit: don't use "`".

nicholaspai: what should I use?

contracts/HubPool.sol
```diff
@@ -779,6 +787,7 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
     // Note this method does a lot and wraps together the sending of tokens and updating the pooled token trackers. This
     // is done as a gas saving so we don't need to iterate over the l1Tokens multiple times.
     function _sendTokensToChainAndUpdatePooledTokenTrackers(
+        address spokePool,
```
chrismaree, line 790: I was going to say that this is redundant butI actually think the way you have it now is cheaper as you only read the state once.

chrismaree (approved):

mrice32 (approved):

contracts/HubPool.sol
nicholaspai: exactly

PR #77, opened 2022-03-07
feat: make slow relay refund the user realyerFeePct

chrismaree:
The depositor should not have to pay the relayerFeePct in the case of a slow relay. This PR removes this from the amount the user pays.


contracts/SpokePool.sol
```diff
@@ -754,10 +754,12 @@ abstract contract SpokePool is SpokePoolInterface, Testable, Lockable, MultiCall
...
 
             // The user will fulfill the remainder of the relay, so we need to compute exactly how many tokens post-fees
-            // that they need to send to the recipient.
+            // that they need to send to the recipient. Note that if the relayer is filled using contract funds then
+            // this is a slow relay. In this case the depositor should get refunded the relayerFeePct. To do this we
+            // simply exclude it from the fees taken from the fillAmountPreFees, thereby refunding the user this amount.
             amountToSend = _computeAmountPostFees(
                 fillAmountPreFees,
-                relayData.realizedLpFeePct + updatableRelayerFeePct
+                relayData.realizedLpFeePct + (useContractFunds ? 0 : updatableRelayerFeePct)
```
nicholaspai, line 762: Should we emit the `relayerFeePct` as "0" in the `FilledRelay` event? Or should we keep it as is which implicitly says that  we are emitting the original `relayerFeePct` set by the depositor, not the actual one used in production.

FWIW I lean towards setting `relayerFeePct` to 0 and then renaming the param in the `FilledRelay` event from `relayerFeePct` to `appliedRelayerFeePct`

nicholaspai (approved):
One nit about changing the `FilledRelay` event to better reflect a slow relay. Also please change the title of this PR to conform with conventional commits

contracts/SpokePool.sol
nicholaspai: @mrice32 we do the same thing in the speed up case where we emit the updated relay fee pct

mrice32 (approved):
LGTM! Nice and simple

contracts/SpokePool.sol
mrice32: Awesome. SGTM

PR #78, opened 2022-03-08
feat: add emergency admin functions

mrice32:
This adds a pause and delete function to allow the governance system to pause proposals/executions and to delete an active proposal.

Working on tests.

### DIFF
--- a/contracts/HubPool.sol
+++ b/contracts/HubPool.sol
@@ -73,6 +73,9 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
     // can be submitted.
     RootBundle public rootBundleProposal;
 
+    // Whether the bundle proposal process is paused.
+    bool public paused;
+
     // Whitelist of origin token + ID to destination token routings to be used by off-chain agents. The notion of a
     // route does not need to include L1; it can be L2->L2 route. i.e USDC on Arbitrum -> USDC on Optimism as a "route".
     mapping(bytes32 => address) private whitelistedRoutes;
@@ -137,6 +140,15 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
     // Each root bundle proposal must stay in liveness for this period of time before it can be considered finalized.
     // It can be disputed only during this period of time. Defaults to 2 hours, like the rest of the UMA ecosystem.
     uint32 public liveness = 7200;
+
+    event Paused(bool indexed isPaused);
+
+    event EmergencyRootBundleDeleted(
+        bytes32 indexed poolRebalanceRoot,
+        bytes32 indexed relayerRefundRoot,
+        bytes32 slowRelayRoot,
+        address indexed proposer
+    );
 
     event ProtocolFeeCaptureSet(address indexed newProtocolFeeCaptureAddress, uint256 indexed newProtocolFeeCapturePct);
 
@@ -203,6 +215,11 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
         _;
     }
 
+    modifier unpaused() {
+        require(!paused, "Proposal process has been paused");
+        _;
+    }
+
     modifier zeroOptimisticOracleApproval() {
         _;
         bondToken.safeApprove(address(_getOptimisticOracle()), 0);
@@ -230,6 +247,33 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
     /*************************************************
      *                ADMIN FUNCTIONS                *
      *************************************************/
+
+    /**
+     * @notice Pauses the bundle proposal and execution process. This is intended to be used during upgrades or when
+     * something goes awry.
+     * @param pause true if the call is meant to pause the system, false if the call is meant to unpause it.
+     */
+    function setPaused(bool pause) public onlyOwner nonReentrant {
+        paused = pause;
+        emit Paused(pause);
+    }
+
+    /**
+     * @notice This allows for the deletion of the active proposal in case of emergency.
+     * @dev This is primarily intended to rectify situations where an unexecutable bundle gets through liveness in the
+     * case of a non-malicious bug in the proposal/dispute code. Without this function, the contract would be
+     * indefinitely blocked, migration would be required, and in-progress transfers would never be repaid.
+     */
+    function emergencyDeleteProposal() public onlyOwner nonReentrant {
+        if (!rootBundleProposal.proposerBondRepaid) bondToken.safeTransfer(rootBundleProposal.proposer, bondAmount);
+        emit EmergencyRootBundleDeleted(
+            rootBundleProposal.poolRebalanceRoot,
+            rootBundleProposal.relayerRefundRoot,
+            rootBundleProposal.slowRelayRoot,
+            rootBundleProposal.proposer
+        );
+        delete rootBundleProposal;
+    }
 
     /**
      * @notice Sends message to SpokePool from this contract. Callable only by owner.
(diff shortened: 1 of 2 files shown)


contracts/HubPool.sol
```diff
@@ -231,6 +248,33 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
      *                ADMIN FUNCTIONS                *
      *************************************************/
 
+    /**
+     * @notice Pauses the bundle proposal and execution process. This is intended to be used during upgrades or when
+     * something goes awry.
+     * @param pause true if the call is meant to pause the system, false if the call is meant to unpause it.
+     */
+    function setPaused(bool pause) public onlyOwner nonReentrant {
```
nicholaspai, line 256: nit: can we remove `nonReentrant`?

contracts/HubPool.sol
```diff
@@ -231,6 +248,33 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
...
+        emit Paused(pause);
+    }
+
+    /**
+     * @notice This allows for the deletion of the active proposal in case of emergency.
+     * @dev This is primarily intended to rectify situations where an unexecutable bundle gets through liveness in the
+     * case of a non-malicious bug in the proposal/dispute code. Without this function, the contract would be
+     * indefinitely blocked, migration would be required, and in-progress transfers would never be repaid.
+     */
+    function emergencyDeleteProposal() public onlyOwner nonReentrant {
```
nicholaspai, line 267: What if the `poolRebalanceLeafCount == 0` and a new proposal can come in? (i.e. the situation where `emeregencyDeleteProposal()` is called but `executeProposal` has already happened. Should we check for that or is this not harmful?

contracts/HubPool.sol
```diff
@@ -231,6 +248,33 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
...
+        if (!rootBundleProposal.proposerBondRepaid) bondToken.safeTransfer(rootBundleProposal.proposer, bondAmount);
+        emit EmergencyRootBundleDeleted(
+            rootBundleProposal.poolRebalanceRoot,
+            rootBundleProposal.relayerRefundRoot,
+            rootBundleProposal.slowRelayRoot,
+            rootBundleProposal.proposer
+        );
+        delete rootBundleProposal;
```
nicholaspai, line 275: should we be `delete`-ing the proposal first before making the external `safeTransfer` call? I'm thinking of the check-effects-interaction pattern

contracts/HubPool.sol
```diff
@@ -203,6 +215,11 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
         _;
     }
 
+    modifier unpaused() {
+        require(!paused, "Proposal process has been paused");
```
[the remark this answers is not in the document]
mrice32: It's exactly 32 :)

contracts/HubPool.sol
```diff
@@ -513,7 +557,7 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
         bytes32 poolRebalanceRoot,
         bytes32 relayerRefundRoot,
         bytes32 slowRelayRoot
-    ) public override nonReentrant noActiveRequests {
+    ) public override nonReentrant noActiveRequests unpaused {
```
nicholaspai, line 560: Should we be adding the `unpaused` modifier to other non-admin functions like `addLiquidity`, `removeLiquidity` and more subtle state-modifying methods like `sync` and `claimProtocolFeesCaptured`?

Also, what about `disputeRootBundle`?

nicholaspai (approved):
Implementation looks great, but I have some concerns about using the `unpaused` modifier in other user-facing methods

contracts/HubPool.sol
mrice32: I think the intention is to _only_ block proposals from moving forward, right? Do we want to block any normal LP or fee functionality?

nicholaspai: I agree with this. We also have other ways to pause liquidity methods via `enableL1TokenForLiquidityProvision`

nicholaspai (changes requested):
As discussed IRL, let's consider adding a method to delete a root bundle that was published to a spoke pool. This would involve adding an admin-only deleter method on the SpokePool and then optionally an onlyOwner helper on the HubPool to call that deleter

contracts/HubPool.sol
chrismaree: if we do then should do the same on all the `onlyOwner` functions.

contracts/HubPool.sol
chrismaree: I think no because this is called by the owner. it's up to the owner to ensure this won't break anything.

contracts/HubPool.sol
chrismaree: good point! transfer last.

contracts/HubPool.sol
chrismaree: I agree this is the only function that should have this.

chrismaree (commented):
looks good! just needs tests 👜

contracts/HubPool.sol
nicholaspai: Hm i'd still rather this folows check-effects-interaction since there's no downside to doing so.

contracts/HubPool.sol
nicholaspai: This saves a require statement but is maybe worth a comment explaining why we don't do any checks

mrice32: I don't think this really breaks anything. If poolRebalanceLeafCount == 0, the proposal could be deleted by just making a new proposal. So I don't really think adding that check here protects against anything.

contracts/HubPool.sol
mrice32: Deleting first just means we have to store some temp variables, making the implementation a bit more messy. Happy to do it if we think it's important, though.

mrice32:
> As discussed IRL, let's consider adding a method to delete a root bundle that was published to a spoke pool. This would involve adding an admin-only deleter method on the SpokePool and then optionally an onlyOwner helper on the HubPool to call that deleter

I added the admin function on the SpokePools. Is there any need for a function on the HubPool? Can't we just call it with the function on the hub pool that calls arbitrary spoke pool admin functions?

contracts/HubPool.sol
```diff
@@ -60,8 +60,6 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
         uint256 claimedBitMap;
         // Proposer of this root bundle.
         address proposer;
-        // Whether bond has been repaid to successful root bundle proposer.
-        bool proposerBondRepaid;
```
mrice32, line 64: Note: this wasn't being used outside of this PR. We were using unclaimedPoolRebalanceLeafCount in lieu of this variable, so I figured it made sense to remove.

contracts/HubPool.sol
nicholaspai: we don't have `nonReeentrant` on any public methods that don't make external calls, including some of the `onlyOwner` functions like `setLiveness`

contracts/HubPool.sol
nicholaspai: Ah ok no problem

nicholaspai: As long as this is reentrancy guarded

test/HubPool.Admin.ts
```diff
@@ -135,4 +135,39 @@ describe("HubPool Admin functions", function () {
   it("Only owner can set liveness", async function () {
     await expect(hubPool.connect(other).setLiveness(1000000)).to.be.reverted;
   });
+  it("Only owner can pause", async function () {
+    await expect(hubPool.connect(other).setPaused(true)).to.be.reverted;
+    await expect(hubPool.connect(owner).setPaused(true)).to.emit(hubPool, "Paused").withArgs(true);
+  });
+  it("Cannot propose while paused", async function () {
```
nicholaspai, line 142: For completeness, you should add a test that you can't execute a proposal while paused

nicholaspai, line 142: nit: move this "cannot propose while paused" test to the `HubPool.ProposeRootBundle.ts` test

nicholaspai (approved):
Great tests, lgtm!

test/HubPool.Admin.ts
mrice32: Will do!

mrice32: Done!

PR #79, opened 2022-03-09
fix: Remove EIP-191 signature recovery support and add destination chain ID to relay data

nicholaspai:
- By removing EIP191 support, we remove an attack vector whereby someone can forge a "speed up request" signature on the destination chain by deploying a custom contract. We also remove the possibility that a signature is not recoverable on the destination chain because the contract that signed the message and submitted it on the origin chain cannot be replicated on the destination side.
- Adding `destinationChainId` to the Relay Data is important so that slow relays can be associated deterministically to relays on a specific chain. Imagine a scenario where a slow relay to fill a relay on chain 100 is added to the `slowRelayRoot`. This slow relay is valid so the root bundle including the `slowRelayRoot` on the HubPool successfully passes liveness. Now, anyone can submit that slow relay leaf to a chain 200 to take funds out of a SpokePool


contracts/SpokePool.sol
```diff
@@ -107,6 +106,7 @@ abstract contract SpokePool is SpokePoolInterface, Testable, Lockable, MultiCall
         uint256 fillAmount,
         uint256 repaymentChainId,
         uint256 originChainId,
+        uint256 destinationChainId,
```
chrismaree, line 109: if you do this then please also add the `originChainId` to `FundsDeposited`

test/gas-analytics/SpokePool.Relay.ts
```diff
@@ -1,7 +1,7 @@
 import { toBNWei, SignerWithAddress, Contract, ethers, seedWallet, toBN } from "../utils";
 import { constructRelayParams, warmSpokePool, sendRelay } from "./utils";
 import * as constants from "../constants";
-import { spokePoolFixture, enableRoutes } from "../fixtures/SpokePool.Fixture";
```
chrismaree, line 4: 🍉

chrismaree (approved):
LGTM! makes sence as a change

contracts/SpokePool.sol
```diff
@@ -662,20 +669,17 @@ abstract contract SpokePool is SpokePoolInterface, Testable, Lockable, MultiCall
 
     // This function is isolated and made virtual to allow different L2's to implement chain specific recovery of
     // signers from signatures because some L2s might not support ecrecover, such as those with account abstraction
-    // like ZKSync.
+    // like ZKSync. To be safe, consider always reverting this function for L2s where ecrecover is different from how
```
mrice32, line 672: Does ZKSync not support standard ecrecover?

contracts/SpokePool.sol
```diff
@@ -467,18 +471,21 @@ abstract contract SpokePool is SpokePoolInterface, Testable, Lockable, MultiCall
         address destinationToken,
         uint256 amount,
         uint256 originChainId,
+        uint256 destinationChainId,
```
mrice32, line 474: Maybe we should omit this to save calldata and _assume_ that the current chainId is the one they intended? Thoughts?

mrice32 (commented):

contracts/SpokePool.sol
nicholaspai: I guess they do now, but its a special case so we need to be careful when integrating with zksync: https://v2-docs.zksync.io/dev/testnet/status.html#currently-supported-features

contracts/SpokePool.sol
nicholaspai: done

mrice32 (approved):

contracts/SpokePool.sol
nicholaspai: Yup done

chrismaree (approved):

PR #84, opened 2022-03-11
improve: Simplify ancillary data

nicholaspai:
Root bundle data can be grabbed deterministically by looking up the latest `ProposeRootBundle` event before the `DisputeRootBundle` event, so some of the ancillary data currently emitted by the HubPool is redundant.

The reason why we assume that voters will have to fetch a `ProposeRootBundle` event is that  the root bundle's `blockEvaluationNumbers` are not stored on-chain. If they were stored on-chain, we could emit all of the data necessary to evaluate a dispute in the ancillary data. But since `blockEvaluationNumbers` are not stored-onchain, voters will have to look up the `ProposeRootBundle` event regardless and therefore we should not expend gas emitting data already included in that event.

### DIFF
--- a/contracts/HubPool.sol
+++ b/contracts/HubPool.sol
@@ -753,10 +753,13 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
      * @notice Returns ancillary data containing the minimum data necessary that voters can use to identify
      * a root bundle proposal to validate its correctness.
      * @dev The root bundle that is being disputed was the most recently proposed one with a block number less than
-     * or equal to the dispute block time. All of this root bundles data can be found in the ProposeRootBundle event
+     * or equal to the dispute block time. All of this root bundle data can be found in the ProposeRootBundle event
      * params. Moreover, the optimistic oracle will stamp the requester's address (i.e. this contract address) meaning
      * that ancillary data for a dispute originating from another HubPool will always be distinct from a dispute
      * originating from this HubPool.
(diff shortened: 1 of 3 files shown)


chrismaree (approved):

mrice32:
@chrismaree @nicholaspai have we made a definitive decision as to whether we want to do this? Is all of this context easily derivable from events?

Maybe a compromise would just be putting the roots in, but removing the other fields. For instance, I think the claimed bitmap will always be 0. Similarly, I think we should at least rename the unclaimed leaves variable, since in this case this would just be the number of leaves, since this dispute would happen before execution starts.

Thoughts?

mrice32 (commented):
https://github.com/across-protocol/contracts-v2/pull/84#issuecomment-1067555127

nicholaspai:
@mrice32 I believe that fetching the `ProposeRootBundle` is required for voters because of the `bundleEvaluationNumbers` not being stored on-chain. See the PR description:

> The reason why we assume that voters will have to fetch a ProposeRootBundle event is that the root bundle's blockEvaluationNumbers are not stored on-chain. If they were stored on-chain, we could emit all of the data necessary to evaluate a dispute in the ancillary data. But since blockEvaluationNumbers are not stored-onchain, voters will have to look up the ProposeRootBundle event regardless and therefore we should not expend gas emitting data already included in that event.

Therefore all of the ancillary data that I've removed in this PR can be found in that event so there's really no point in emitting it here. Even if we emit the merkle roots in this ancillary data, voters will still need to look up the bundle block numbers right?

PR #85, opened 2022-03-11
fix(hub-pool): Enable data worker to ensure that roots are not redundantly relayed to SpokePools

nicholaspai:
Data worker can mark in `PoolRebalanceLeaf` whether it should send roots to the `SpokePool`. We assume that valid root bundles would only send roots to a SpokePool on a specific `chainId` once.

### DIFF
--- a/contracts/HubPool.sol
+++ b/contracts/HubPool.sol
@@ -60,8 +60,6 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
         uint256 claimedBitMap;
         // Proposer of this root bundle.
         address proposer;
-        // Keeps track of which SpokePools we have already relayed roots to.
-        mapping(uint256 => bool) relayedRootToSpokePool;
         // Number of pool rebalance leaves to execute in the poolRebalanceRoot. After this number
         // of leaves are executed, a new root bundle can be proposed
         uint8 unclaimedPoolRebalanceLeafCount;
--- a/contracts/HubPool.sol
+++ b/contracts/HubPool.sol
@@ -655,9 +655,7 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
         _sendTokensToChainAndUpdatePooledTokenTrackers(spokePool, chainId, l1Tokens, netSendAmounts, bundleLpFees);
 
         // Check bool used by data worker to prevent relaying redundant roots to SpokePool.
-        if (relayToSpokePool) {
-            _relayRootBundleToSpokePool(spokePool, chainId);
-        }
+        if (relayToSpokePool) _relayRootBundleToSpokePool(spokePool, chainId);
 
         // Transfer the bondAmount to back to the proposer, if this the last executed leaf. Only sending this once all
         // leafs have been executed acts to force the data worker to execute all bundles or they wont receive their bond.
--- a/contracts/HubPool.sol
+++ b/contracts/HubPool.sol
@@ -5,7 +5,6 @@ import "./MerkleLib.sol";
 import "./HubPoolInterface.sol";
 import "./Lockable.sol";
 
-import "./interfaces/AdapterInterface.sol";
 import "./interfaces/LpTokenFactoryInterface.sol";
 import "./interfaces/WETH9.sol";
 
@@ -99,7 +98,7 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
 
     // Heler contracts to facilitate cross chain actions between HubPool and SpokePool for a specific network.
     struct CrossChainContract {
-        AdapterInterface adapter;
+        address adapter;
         address spokePool;
     }
     // Mapping of chainId to the associated adapter and spokePool contracts.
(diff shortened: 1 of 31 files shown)


contracts/HubPool.sol
```diff
@@ -60,6 +60,8 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
         uint256 claimedBitMap;
         // Proposer of this root bundle.
         address proposer;
+        // Keeps track of which SpokePools we have already relayed roots to.
+        mapping(uint256 => bool) relayedRootToSpokePool;
```
chrismaree, line 64: I think this was left over from your previous implementation, no?

contracts/HubPoolInterface.sol
```diff
@@ -26,6 +26,10 @@ interface HubPoolInterface {
         // A positive number indicates that the HubPool owes the SpokePool funds. A negative number indicates that the
         // SpokePool owes the HubPool funds. See the comment above for the dynamics of this and netSendAmounts
         int256[] runningBalances;
+        // Used by data worker to mark which leaves should relay roots to SpokePools. We assume that the data worker
+        // would not mark this `True` for leaves that share `chainId` otherwise a root would get sent twice to a
```
chrismaree, line 30:
```suggestion
        // would not mark this True for leaves that share chainId otherwise a root would get sent twice to a
```

contracts/HubPool.sol
```diff
@@ -649,14 +655,27 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
         rootBundleProposal.unclaimedPoolRebalanceLeafCount--;
 
         _sendTokensToChainAndUpdatePooledTokenTrackers(spokePool, chainId, l1Tokens, netSendAmounts, bundleLpFees);
-        _relayRootBundleToSpokePool(spokePool, chainId);
+
+        // Check bool used by data worker to prevent relaying redundant roots to SpokePool.
+        if (relayToSpokePool) {
+            _relayRootBundleToSpokePool(spokePool, chainId);
+        }
```
chrismaree, line 662:
```suggestion
        if (relayToSpokePool) _relayRootBundleToSpokePool(spokePool, chainId);
```

contracts/HubPool.sol
```diff
@@ -194,6 +196,7 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
         address indexed proposer
     );
     event RootBundleExecuted(
+        bool relayToSpokePool,
```
chrismaree, line 199: does the ordering in this event impact the gas cost?

contracts/HubPool.sol
chrismaree, line 64: I think this was left over from your previous implementation, no?

```suggestion
```

chrismaree (approved):
lgtm once the nits are resolved.

contracts/HubPool.sol
nicholaspai: It does but I can't get the gas lower than putting it here. I tried placing it in other spots

contracts/HubPool.sol
nicholaspai: Removed

contracts/HubPool.sol
```diff
@@ -607,6 +609,7 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
         int256[] memory netSendAmounts,
         int256[] memory runningBalances,
         uint8 leafId,
+        bool relayToSpokePool,
```
mrice32, line 612: What was the reasoning for using a bool over an index? Would something like `perChainLeafIndex` be more clear to someone creating a bundle?

mrice32 (approved):

contracts/HubPool.sol
nicholaspai: I just don't see any utility for the index. If you could give me an example of how an index would be useful to the off-chain data worker, then I'd change to an index. But I can't think of any examples where the index is useful.

mrice32: I think it's just easier to understand since an index obviously increments for each leaf for a chain. Whereas you have to know that this bool goes true on the first element and then false on every following element. I think the bool makes a lot of sense _in the context of the HubPool_, but to a relayer, this bool might have less meaning since they aren't working from within the HubPool's send logic. Maybe to reverse the question, what's the reasoning for using a bool over an index?

Not strongly opinionated on this, however. I'm cool with whatever you think!

mrice32 (commented):
https://github.com/across-protocol/contracts-v2/pull/85#discussion_r827253697

contracts/HubPool.sol
nicholaspai: Hm ok i'll change to an index since it doesn't seem to matter too much and it gives us a bit more flexibility

nicholaspai: Ok changed @mrice32 but I don't love the name `groupIndex` nor the `perChainleafIndex`. Any other ideas?

mrice32: Yeah, the naming is difficult. I think groupIndex is totally fine, though, especially since you documented it so well.

PR #90, opened 2022-03-14
fix: HubPool should send ETH to caller of removeLiquidity if they specify so

nicholaspai:
Currently contract sends WETH to contracts even if they specify `sendEth = True` which is confusing and can lead to lost funds by contracts who cannot handle WETH and don't read the contract source code closely. This PR sends ETH or WETH to caller as specified

### DIFF
--- a/contracts/HubPool.sol
+++ b/contracts/HubPool.sol
@@ -467,7 +467,9 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
      * @param l1Token Token to redeem LP share for.
      * @param lpTokenAmount Amount of LP tokens to burn. Exchange rate between L1 token and LP token can be queried
      * via public exchangeRateCurrent method.
-     * @param sendEth Set to True if L1 token is WETH and user wants to receive ETH.
+     * @param sendEth Set to True if L1 token is WETH and user wants to receive ETH. Note that if caller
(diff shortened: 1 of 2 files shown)


contracts/HubPool.sol
```diff
@@ -482,9 +484,12 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
         // If they try access more funds that available (i.e l1TokensToReturn > liquidReserves) this will underflow.
         pooledTokens[l1Token].liquidReserves -= l1TokensToReturn;
 
-        if (sendEth) _unwrapWETHTo(payable(msg.sender), l1TokensToReturn);
-        else IERC20(l1Token).safeTransfer(msg.sender, l1TokensToReturn);
-
+        if (sendEth) {
+            weth.withdraw(l1TokensToReturn);
+            payable(msg.sender).transfer(l1TokensToReturn);
```
chrismaree, line 489:
```suggestion
            payable(msg.sender).transfer(l1TokensToReturn); // This will revert if the caller is a contract that does not implement a fallback function.
```

chrismaree (approved):
This is a good solution. we are saved by virtue of smart contracts that dont implement fallback functions reverting if you try send them eth. The way you have it now is most locally consistant and will enable both kinds of flow.

LGTM

PR #91, opened 2022-03-15
improve: [L02] validate function inputs

nicholaspai:
Issue:
- [L02] Lack of input validation
    - Throughout the codebase there are functions lacking sufficient input validation. For instance:
    - In the HubPool contract the various admin functions will accept 0 values for inputs. This could
result in the loss of funds and unexpected behaviors if null values are unintentionally provided.
    - In the HubPool contract the setProtocolFeeCapture function does not use the
noActiveRequests modifier. This could allow the protocol fee to be increased even for liquidity
providers that have already provided liquidity.
    - In the MerkleLib library the isClaimed1D function does not work as expected if an index is
greater than 255. In such a case, it will return true despite the fact that those values are not
actually claimed.
    - In the SpokePool contract the deposit function does not enforce the requirement suggested by
the deploymentTime comment which says that relays cannot have a quote time before
deploymentTime.
    - In the SpokePool contract the speedUpDeposit function does not restrict the
newRelayerFeePct to be less than 50% like the regular deposit does. In practice, the
_fillRelay function won't accept a fee that is too high, but this should still be enforced within
speedUpDeposit.
    - In the PolygonTokenBridger contract the "normal" use case of send involves the caller,
Polygon_SpokePool, evaluating if the token it is sending is wrapped matic in order to set the
isMatic flag appropriately. However, for any other caller, if they forget to set this flag while
sending wrapped matic, then their tokens would be unwrapped but not sent anywhere. For more
predictable behavior, consider checking for wrapped matic inline rather than relying on the
isMatic argument.
    - To avoid errors and unexpected system behavior, consider implementing require statements to validate
all user-controlled input, even that of admin accounts considering that some clients may default to
sending null parameters if none are specified.

Resolutions: 
- Validate `newProtocolFeeCaptureAddress` and `spokePool != address(0)` in `setProtocolFeeCapture` and `relaySpokePoolAdminFunction` respectively
- The active requests modifier has nothing to do with LPs, so we don't think we should add the `noActiveRequests` modifier to the `setProtocolFeeCapture` method as suggested. 
- Changes index type to uint8 for 1D claimed bitmap modifying functions since max index is 255.
- Remove `deploymentTime` state variable from SpokePool. This variable was meant as a convenience to an off-chain data worker evaluating fillRelay calls but it should have been placed on the HubPool if anywhere. Any fillRelay should deterministically map to a deposit on the origin SpokePool for a specific quote time. The quote time will be emitted in a FundsDeposited event with a unique depositId for that SpokePool contract. If that quote time is less than the HubPool’s deployment time, then computing the realizedLpFeePct for the fillRelay will always revert since it will require state that didn’t exist on an address prior to the HubPool’s deployment. This state variable is not useful on the SpokePool since the realizedLpFeePct is computed on the HubPool’s state, so it should be stored for convenience on L1 on the Hubpool but should have no effect on the SpokePool’s logic. However, we can do even better and state in the UMIP that no deposit quote time is valid that is greater than the WhitelistToken event involving it. [See this UMIP for Across_V1 for example](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-136.md):
    - “Any relayed transaction that is mined with earlier block timestamp than quoteTimestamp field from DepositData struct in the DepositRelayed event should be invalid.”
- Check modified relayer fee pct passed to speedUpDeposit is < 50%
- Rename `isMatic` to `isWrappedMatic` in `Polygon_TokenBridger` to reduce chance of user not setting this flag propoerly

### DIFF
--- a/contracts/HubPool.sol
+++ b/contracts/HubPool.sol
@@ -298,7 +298,6 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
         public
         override
         onlyOwner
-        noActiveRequests
     {
         require(newProtocolFeeCapturePct <= 1e18, "Bad protocolFeeCapturePct");
         require(newProtocolFeeCaptureAddress != address(0), "Bad protocolFeeCaptureAddress");
--- a/contracts/HubPool.sol
+++ b/contracts/HubPool.sol
@@ -5,7 +5,6 @@ import "./MerkleLib.sol";
 import "./HubPoolInterface.sol";
 import "./Lockable.sol";
 
-import "./interfaces/AdapterInterface.sol";
 import "./interfaces/LpTokenFactoryInterface.sol";
 import "./interfaces/WETH9.sol";
 
@@ -99,7 +98,7 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
 
     // Heler contracts to facilitate cross chain actions between HubPool and SpokePool for a specific network.
     struct CrossChainContract {
-        AdapterInterface adapter;
+        address adapter;
         address spokePool;
     }
     // Mapping of chainId to the associated adapter and spokePool contracts.
@@ -302,6 +301,7 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
         noActiveRequests
     {
         require(newProtocolFeeCapturePct <= 1e18, "Bad protocolFeeCapturePct");
+        require(newProtocolFeeCaptureAddress != address(0), "Bad protocolFeeCaptureAddress");
         protocolFeeCaptureAddress = newProtocolFeeCaptureAddress;
         protocolFeeCapturePct = newProtocolFeeCapturePct;
         emit ProtocolFeeCaptureSet(newProtocolFeeCaptureAddress, newProtocolFeeCapturePct);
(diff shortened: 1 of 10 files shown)


contracts/HubPool.sol
```diff
@@ -648,8 +646,33 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
...
+            spokePool,
+            chainId,
+            l1Tokens,
+            netSendAmounts,
+            bundleLpFees
+        );
+
+        // Relay root bundles to spoke pool on destination chain by
+        // performing delegatecall to use the adapter's code with this contract's context.
+        (bool success, ) = adapter.delegatecall(
```
nicholaspai, line 664: I think adding this inline is a bit clearer since we validate the `spokePool` address in this method and we should put logic and its validation together

contracts/HubPool.sol
```diff
@@ -851,14 +874,13 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
...
     // is done as a gas saving so we don't need to iterate over the l1Tokens multiple times.
     function _sendTokensToChainAndUpdatePooledTokenTrackers(
+        address adapter,
         address spokePool,
         uint256 chainId,
         address[] memory l1Tokens,
         int256[] memory netSendAmounts,
         uint256[] memory bundleLpFees
     ) internal {
-        AdapterInterface adapter = crossChainContracts[chainId].adapter;
```
nicholaspai, line 860: Save an SLOAD

mrice32 (approved):
LGTM!

PR #92, opened 2022-03-15
improve: [L06] Add reentrancy guards to all public methods

nicholaspai:


contracts/HubPool.sol
```diff
@@ -453,8 +454,8 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
         // Since _exchangeRateCurrent() reads this contract's balance and updates contract state using it, it must be
         // first before transferring any tokens to this contract to ensure synchronization.
         uint256 lpTokensToMint = (l1TokenAmount * 1e18) / _exchangeRateCurrent(l1Token);
-        ExpandedIERC20(pooledTokens[l1Token].lpToken).mint(msg.sender, lpTokensToMint);
```
nicholaspai, line 456: Moved token transfer "interaction" after any effects to follow CEI

contracts/HubPool.sol
```diff
@@ -476,12 +477,12 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
...
         require(address(weth) == l1Token || !sendEth, "Cant send eth");
         uint256 l1TokensToReturn = (lpTokenAmount * _exchangeRateCurrent(l1Token)) / 1e18;
-
-        ExpandedIERC20(pooledTokens[l1Token].lpToken).burnFrom(msg.sender, lpTokenAmount);
         // Note this method does not make any liquidity utilization checks before letting the LP redeem their LP tokens.
-        // If they try access more funds that available (i.e l1TokensToReturn > liquidReserves) this will underflow.
+        // If they try access more funds than available (i.e l1TokensToReturn > liquidReserves) this will underflow.
         pooledTokens[l1Token].liquidReserves -= l1TokensToReturn;
 
+        ExpandedIERC20(pooledTokens[l1Token].lpToken).burnFrom(msg.sender, lpTokenAmount);
```
nicholaspai, line 484: rearranging for CEI

contracts/HubPool.sol
```diff
@@ -725,6 +726,9 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
             customLiveness: liveness
         });
 
+        // Finally, delete the state pertaining to the active proposal so that another proposer can submit a new bundle.
+        delete rootBundleProposal;
```
nicholaspai, line 730: rearranging for CEI

contracts/Arbitrum_SpokePool.sol
```diff
@@ -108,5 +108,5 @@ contract Arbitrum_SpokePool is SpokePool {
     }
 
     // Apply AVM-specific transformation to cross domain admin address on L1.
-    function _requireAdminSender() internal override onlyFromCrossDomainAdmin {}
+    function _requireAdminSender() internal override onlyFromCrossDomainAdmin nonReentrant {}
```
chrismaree, line 111: so this is an internal method. why add it here? dont we only need it on the external/public functions?

contracts/HubPool.sol
```diff
@@ -334,7 +335,7 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
      * @notice Sets root bundle proposal liveness period. Callable only by owner.
      * @param newLiveness New liveness period.
      */
-    function setLiveness(uint32 newLiveness) public override onlyOwner {
+    function setLiveness(uint32 newLiveness) public override onlyOwner nonReentrant {
```
chrismaree, line 338: I personally feel that adding it to these admin functions is not needed at all. no real downside to having it to be as consistant as posible but it does not really do anything meaningful.

contracts/HubPool.sol
chrismaree: nice. doing this + having the guard is best posible. importantly, you are doing this _after_ the calls that modify tracking state so no risk in this change.

chrismaree (approved):

contracts/Arbitrum_SpokePool.sol
nicholaspai: Ah yes i'll add a comment. Basically we don't add the `nonReentrant` modifier to `onlyAdmin` functions in the base `SpokePool` contract because the `Polygon_SpokePool` will call these methods internally via the `processMessageFromRoot`. The other spoke pools like `Optimism_SpokePool` and `Arbitrum_SpokePool` have their admin functions triggered by an external contract so we should be reentrancy guarding those methods. However, in the `Polygon_SpokePool` case we need to reentrancy guard at the `processMessageFromRoot` method instead of at the admin functions.

contracts/HubPool.sol
nicholaspai: Also see this comment I had added in the PR: https://github.com/across-protocol/contracts-v2/blob/cc6853a368e19075b0803c77cc0190fdf85e6216/contracts/SpokePool.sol#L27

contracts/Arbitrum_SpokePool.sol
mrice32: But this isn't equivalent to adding a re-entrancy guard to the admin function, right? It guarantees that this function wasn't re-entered, but it doesn't help us if that admin function re-enters a different function since it doesn't wrap the entire method, no?

mrice32 (commented):
I think it's going to be difficult to make all the spoke pool admin methods non-reentrant when they're all called in such different ways.

The issue raised by OZ specifically applies to the HubPool, not the SpokePool right? So are these changes beyond the scope of the changes being requested?

mrice32 (approved):
LGTM!

PR #94, opened 2022-03-15
fix: [L05] Enabling token for Liquidity provisioning should not reset lastLpFeeUpdate

nicholaspai:
Issue:
- [L05] Liquidity provisioning can skew fee assessments
    - In the HubPool contract the enableL1TokenForLiquidityProvision function allows the contract
owner to enable an l1token to be added to the protocol for liquidity pooling.
    - This is allowed even if the l1token is already currently enabled.
    - As this function also sets the lastLpFeeUpdate variable to the then-current block.timestamp,
enabling an already enabled token will skip over the period of time since lastLpFeeUpdate was last
set. As a result, any LP fees that should have been assessed for that time period would simply never be
assessed.
    - Consider reverting if this function is called for an l1token that is already enabled.

Solution:
- Only set `lastLpFeeUpdate` upon first enabling


mrice32 (approved):
LGTM!

PR #95, opened 2022-03-15
improve: [L08] Add comment about shutting down contract before uint32 timestamps roll over in year 2106

nicholaspai:
Issue:
- [L08] Time is cast unsafely
    - In the HubPool function _updateAccumulatedLpFees, the return value of getCurrentTime() is cast to
a uint32 value. This means that the value will be truncated to fit within 32 bits, and at some point
around Feb 6, 2106, it will "roll over" and the value returned by casting to uint32 will drop down to 0.
This will set pooledToken.lastLpFeeUpdate to a much lower number than the previous
lastLpFeeUpdate. Any subsequent time _getAccumulatedFees is called, the
timeFromLastInteraction calculation will be exceedingly high, and all "undistributed" fees will be
accounted for as accumulated.
    - Again, note that this issue will only occur starting in the year 2106. Consider changing the size of the
cast from uint32 to a larger number, like uint64. This should be more than enough to not encounter
limits within a reasonably distant future. Alternatively, consider documenting the behavior and defining
a procedure for what to do if the system is still in operation when the uint32 limit is hit, or for shutting
down the system before the year 2106.

Solution:
- We use uint32 time-related variables to save gas costs by packing variables smartly, and we do intend to deprecate this contract by 2106. So we add a disclaimer at the top of the contract that it should be deprecated by 2106 and steps for doing so.

### DIFF
--- a/contracts/HubPool.sol
+++ b/contracts/HubPool.sol
@@ -64,7 +64,7 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
         // of leaves are executed, a new root bundle can be proposed
         uint8 unclaimedPoolRebalanceLeafCount;
         // When root bundle challenge period passes and this root bundle becomes executable.
-        uint32 requestExpirationTimestamp;
+        uint64 requestExpirationTimestamp;
     }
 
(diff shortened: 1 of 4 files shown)


mrice32 (approved):
LGTM!

PR #100, opened 2022-03-16
fix[N02] Move all structs to the same place

chrismaree:
**Problem:**
The PoolRebalanceLeaf struct is defined in HubPoolInterface.sol, while the RootBundle,
PooledToken, and CrossChainContract structs are all defined in the implementation, HubPool.sol.
Consider defining all structs for HubPool within the same contract.

**Resolution:**
pull the structs to all be in the interface.


nicholaspai (changes requested):
This should also move the PooledToken and CrossChainContract structs to the interface as per the issue description

chrismaree:
> This should also move the PooledToken and CrossChainContract structs to the interface as per the issue description

good point! fixed.

nicholaspai (approved):
LGTM

PR #101, opened 2022-03-16
fix[N03] Fixed inconsistant token metadata versioning

chrismaree:
**Problem:**
In the LpTokenFactory contract, the LP tokens it creates have inconsistent versioning in their
metadata.
While the token symbol is prepended with Av2 (ostensibly for "Across version 2"), the token name is
prepended only with "Across" and no version number.

**Resolution:**
LP token name now contains the version, like the symbol.


mrice32 (approved):
LGTM!

nicholaspai (approved):
LGTM

PR #102, opened 2022-03-16
fix[N04] Add additional documentation

chrismaree:
**Problem**:
Although most of the codebase is thoroughly documented, there are a few instances where
documentation is lacking. For instance:
In the HubPool contract the public unclaimedAccumulatedProtocolFees variable has no inline
documentation.
In the HubPool contract the inline documentation accompanying
PoolRebalanceLeaf.netSendAmounts, although lengthy, could benefit from additional
clarification around the case of negative values. It could clarify further that in such cases the
actual netSendAmounts value is ignored, but it should match the amountToReturn parameter in
the RelayerRefundLeaf.
Many of the functions in the MerkleLib library are missing NatSpec @return statements.

**Solution**: all places with missing documentation were updated.

### DIFF
--- a/contracts/HubPool.sol
+++ b/contracts/HubPool.sol
@@ -124,7 +124,7 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
     // the full amount of fees entitled to LPs in ~ 7.72 days, just over the standard L2 7 day liveness.
     uint256 public lpFeeRatePerSecond = 1500000000000;
 
-    // Mapping of l1TokenAddress to cumulative unclaimed protocol tokens that the sent to the protocolFeeCaptureAddress
+    // Mapping of l1TokenAddress to cumulative unclaimed protocol tokens that can be sent to the protocolFeeCaptureAddress
     // at any time. This enables the protocol to capture some percentage of LP fees which can be allocated elsewhere.
     mapping(address => uint256) public unclaimedAccumulatedProtocolFees;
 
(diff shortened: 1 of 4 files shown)


mrice32 (approved):

contracts/HubPool.sol
```diff
@@ -124,6 +124,8 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
     // the full amount of fees entitled to LPs in ~ 7.72 days, just over the standard L2 7 day liveness.
     uint256 public lpFeeRatePerSecond = 1500000000000;
 
+    // Mapping of l1TokenAddress to cumulative unclaimed protocol tokens that the sent to the protocolFeeCaptureAddress
```
nicholaspai, line 127:
```suggestion
    // Mapping of l1TokenAddress to cumulative unclaimed protocol tokens that can be sent to the protocolFeeCaptureAddress
```

contracts/HubPool.sol
```diff
@@ -124,6 +124,8 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
...
+    // at any time. This enables the protocol to capture some percentage of LP fees which can be allocated elsewhere.
```
nicholaspai, line 128:
```suggestion
    // at any time. This enables the protocol to reallocate some percentage of LP fees elsewhere.
```

nicholaspai (approved):
LGTM after committing my suggestions!

PR #103, opened 2022-03-16
fix[N05] Remove LPTokenFactory Magic numbers

chrismaree:
**Problem**

In the LpTokenFactory contract, when the createLpToken function is called, it creates a new ERC20
LP token and adds the msg.sender to the new token's minter and burner roles. These role
assignments use the magic values 1 and 2, which are the uint identifiers for the respective roles.

**Solution**

Magic numbers removed via `addMinter` and `addBurner` functions.


mrice32 (approved):

nicholaspai (approved):
LGTM

PR #104, opened 2022-03-16
fix[N06] Fix misleading comments

chrismaree:
**Problem:**

- HubPool lines 718-719 explain that the whitelistedRoute function returns whitelisted
destination tokens, but does not mention that if the token is not whitelisted then the function
returns address(0).
- The comments in the declaration of the PoolRebalanceLeaf struct appear to refer to a previous
version of the struct, making them hard to follow. For example, line 17 implies there are two
arrays above it (there is only one), and line 31 suggests there are multiple arrays below it (there
is only one).
- A comment about HubPool.executeRootBundle states that the function deletes the published
root bundle, however it does not.
- Within the LPTokenFactory contract, the comments on lines 24 and 25 should say "msg.sender"
or "the calling contract" rather than "this contract".
The comments above the lpFeeRatePerSecond variable suggest that LP fees are released
linearly. In fact, they are released sublinearly, because the _getAccumulatedFees function uses
a fraction of the undistributedLpFees (which decreases over time for any given loan), rather
than the total funds on loan.
- The comment in SpokePool above the definition of claimedBitmap state that there are 256x256
leaves per root. However, due to the indexing scheme in MerkleLib, there are a maximum of
2^248 different values of claimedWordIndex, with 256 different claimedBitIndexes. A more
clear comment might explain that there are 256x(2^248) leaves per root.

**Solution:**
Comments were updated.


chrismaree:
Note that this PR was forked from https://github.com/across-protocol/contracts-v2/pull/102/files which should be merged first to avoid merge conflicts.

mrice32:
@chrismaree going to hold off on reviewing this until the whitelisted routes update and the PR this is forked from are both merged, as I think they might require changes here.

contracts/HubPool.sol
```diff
@@ -125,7 +125,8 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
     bytes32 public identifier = "IS_ACROSS_V2_BUNDLE_VALID";
 
     // Interest rate payment that scales the amount of pending fees per second paid to LPs. 0.0000015e18 will pay out
-    // the full amount of fees entitled to LPs in ~ 7.72 days, just over the standard L2 7 day liveness.
+    // the full amount of fees entitled to LPs in ~ 7.72 days assuming no contract interactions. If someone interacts
+    // with the contract then the LP rewards are smeared sublinearly over the window.
```
nicholaspai, line 129: what does "smeared sublinearly" mean? this is a bit handwavy

nicholaspai: "because the _getAccumulatedFees function uses a fraction of the undistributedLpFees (which decreases over time for any given loan), rather than the total funds on loan."

contracts/HubPoolInterface.sol
```diff
@@ -23,9 +23,9 @@ interface HubPoolInterface {
         // deposits in this bundle. If non-zero then it must be set on the SpokePool's RelayerRefundLeaf amountToReturn
         // as -1 * this value to indicate if funds are being sent from or to the SpokePool.
         int256[] netSendAmounts;
-        // This is only here to be emitted in an event to track a running unpaid balance between the L2 pool and the L1 pool.
```
nicholaspai, line 26: nit: additional changes:
- Line 17: Remove "This array is grouped with the two above"
- Line 39: Replace "The following arrays" with "The bundleLpFees, netSendAmounts, and runningBalances"

contracts/SpokePool.sol
```diff
@@ -58,7 +58,7 @@ abstract contract SpokePool is SpokePoolInterface, Testable, Lockable, MultiCall
         // Merkle root of relayer refunds for successful relays.
         bytes32 relayerRefundRoot;
         // This is a 2D bitmap tracking which leaves in the relayer refund root have been claimed, with max size of
-        // 256x256 leaves per root.
+        // 256x(2^248) leaves per root.
```
nicholaspai, line 61:
```suggestion
        // 256x(2^248) leaves per root. Due to the indexing scheme in MerkleLib, there are a maximum of
        // 2^248 different values of claimedWordIndex with 256 different claimedBitIndexes.
```

nicholaspai (approved):
A few more changes then LGTM

contracts/HubPoolInterface.sol
chrismaree: o sorry. 17 was a result of a merge issue. this was taken out before but came back.
doing on 39.

PR #105, opened 2022-03-16
fix[N08] Propose fixes to some naming issues

chrismaree:
**Problem**:

We have identified some areas of the code which could benefit from better naming:
- In HubPoolInterface.liquidityUtilizationPostRelay, the parameter token should be
renamed to l1Token to better match other functions in the interface, as well as the function's
implementation in HubPool.
- In the RootBundle struct, requestExpirationTimestamp should be renamed to better indicate
that it ends the "challenge period". Consider renaming it to ChallengePeriodEndTimestamp or
similar.
- The RootBundleExecuted event in HubPool.sol only names one of its array parameters in
the plural form, but when the event is emitted, all array parameters are named in the plural
form. Consider changing the event definition so that all array parameters are pluralized.
- The name of function whitelistedRoute is vague and does not indicate what it's output will
be. Consider renaming it to something like destinationTokenFromRoute to better match the
return value.
- When weth is used in Polygon_SpokePool.sol, it refers to wrapped MATIC. Consider renaming
the weth variable in SpokePool.sol to wrapped_native_token to make it more generalizable.
- This will make Polygon_SpokePool less confusing and be more generalizeable for future
SpokePools.
- The executeSlowRelayRoot and executeRelayerRefundRoot functions are executing leaves
and should be renamed accordingly.
- The unclaimedPoolRebalanceLeafCount parameter of the ProposeRootBundle event should be
renamed to poolRebalanceLeafCount, since it's always the total number of leaves in the tree.
- The RootBundleCanceled event names the last parameter as disputedAncillaryData, but the
proposal is not necessarily disputed. It should just be ancillaryData.
- The _append function of the LpTokenFactory could be called _concatenate to better describe
its functionality.
- The onlyEnabledRoute modifier has a destinationId parameter that should be
destinationChainId to match the rest of the code base.

**Solution**: 

all recommendations have been made.


chrismaree:
Note that this PR requires a few others to go in first that it is dependent on to avoid conflicts. In particular:
- https://github.com/across-protocol/contracts-v2/pull/100
- https://github.com/across-protocol/contracts-v2/pull/101

contracts/SpokePool.sol
```diff
@@ -37,9 +37,9 @@ abstract contract SpokePool is SpokePoolInterface, Testable, Lockable, MultiCall
     // refunds and slow relays.
     address public hubPool;
 
-    // Address of WETH contract for this network. If an origin token matches this, then the caller can optionally
-    // instruct this contract to wrap ETH when depositing.
-    WETH9 public weth;
+    // Address of wrappedNativeToken contract for this network. If an origin token matches this, then the caller can
+    // optionally instruct this contract to wrap native tokens when depositing (ie ETH->WETH or MATIC->WMATIC).
+    WETH9 public wrappedNativeToken;
```
chrismaree, line 42: this change is potentially contentious. I'm curious what you think.

contracts/Polygon_SpokePool.sol
```diff
@@ -78,7 +77,7 @@ contract Polygon_SpokePool is IFxMessageProcessor, SpokePool {
     }
 
     /********************************************************
-     *    ARBITRUM-SPECIFIC CROSS-CHAIN ADMIN FUNCTIONS     *
+     *    POLYGON-SPECIFIC CROSS-CHAIN ADMIN FUNCTIONS     *
```
chrismaree, line 80: unrelated but saw this comment issue.

contracts/HubPool.sol
```diff
@@ -41,35 +41,6 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
     using SafeERC20 for IERC20;
     using Address for address;
 
-    // A data worker can optimistically store several merkle roots on this contract by staking a bond and calling
```
chrismaree, line 44: I'm not sure why it shows this diff here...previous PR moved this around.

contracts/HubPool.sol
```diff
@@ -209,17 +180,17 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
...
+        int256[] netSendAmounts,
+        int256[] runningBalances,
         address indexed caller
     );
     event SpokePoolAdminFunctionTriggered(uint256 indexed chainId, bytes message);
 
     event RootBundleDisputed(address indexed disputer, uint256 requestTime, bytes disputedAncillaryData);
 
-    event RootBundleCanceled(address indexed disputer, uint256 requestTime, bytes disputedAncillaryData);
+    event RootBundleCanceled(address indexed disputer, uint256 requestTime, bytes ancillaryData);
```
nicholaspai, line 193: You can remove this change since ancillaryData is no longer included in this event following #114

nicholaspai (approved):
This looks awesome but rebase master before you merge and check on any merge conflicts

PR #106, opened 2022-03-16
feat: add polygon deployment scripts and goerli/mumbai deployments

mrice32:


deploy/001_deploy_hubpool.ts
```diff
@@ -1,12 +1,16 @@
 import { L1_ADDRESS_MAP } from "./consts";
 
-const func = async function (hre: any) {
+// This import is needed to override the definition of the HardhatRuntimeEnvironment type.
+import "hardhat-deploy";
+import { HardhatRuntimeEnvironment } from "hardhat/types/runtime";
```
mrice32, line 5: Added this to the deployment scripts so we could use the correct hre type.

deploy/001_deploy_hubpool.ts
```diff
@@ -24,4 +28,4 @@ const func = async function (hre: any) {
   });
 };
 module.exports = func;
-func.tags = ["hubpool"];
+func.tags = ["HubPool", "mainnet"];
```
mrice32, line 31: Updated the tags so they're all the same style and so that it's easy to deploy all the contracts needed for a particular network.

nicholaspai (approved):

PR #107, opened 2022-03-16
fix[N09] Warning about non-standard tokens

chrismaree:
**Problem*:
Although tokens must be enabled to be used in the system, it is important to define what may make a
token troublesome so that which tokens can be whitelisted is easier to determine.
- ERC20 tokens which charge fees, or which can charge fees, will result in various accounting
issues as the amount transferred will not match the amount received by the contracts in the
system. Many spots in the code, such as in the addLiquidity function, assume the amount
transferred in equals the amount received.
- ERC777 tokens, which are ERC20-compatible, include hooks on transfers. These hooks are
configurable and may be configured to revert in some or all cases. In
SpokePool._executeRelayerRefundRoot, a failing transfer for one token could block all other
refunds for the specified leaf.
- Tokens which are upgradeable may change their implementations to become subject to the
above issues, even though they may not have been problematic before being upgraded.

**Solution**: Added documentation about risks of non-standard tokens.


mrice32 (approved):
LGTM!

contracts/HubPool.sol
```diff
@@ -819,6 +819,10 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
     /**
      * @notice Conveniently queries whether an origin chain + token => destination chain ID is whitelisted and returns
      * the whitelisted destination token.
+     * @dev Admin must be considerate to the compatibility of originToken and destinationToken within the protocol. Some
+     * token implementations will not function correctly within the Across v2 system. For example ERC20s that charge
+     * fees will break internal accounting, ERC777 can cause some functions to revert and upgradable tokens can be pose
```
nicholaspai, line 824:
```suggestion
     * fees will break internal accounting, ERC777 can cause some functions to revert and upgradable tokens can pose
```

nicholaspai (approved):
LGTM

PR #108, opened 2022-03-16
fix[N10] Make better use of immutable

chrismaree:
**Problem:**
Within the HubPool contract, the weth, finder, and lpTokenFactory variables are only ever assigned
a value in the constructor.

**Solution:**
made these variables that should never be changed immutable.


mrice32 (approved):

nicholaspai (approved):
LGTM but while you're at it, the following vars can also be made immutable:

`Optimism_SpokePool`
- l2Eth

`SpokePool`
- weth

chrismaree:
> LGTM but while you're at it, the following vars can also be made immutable:
> 
> `Optimism_SpokePool`
> 
> * l2Eth
> 
> `SpokePool`
> 
> * weth

good point! done.

PR #109, opened 2022-03-16
fix[N11] Fix Residual privleged roles

chrismaree:
**Problem:**

When the LpTokenFactory contract creates an ExpandedERC20 token contract, the factory becomes
the owner of that token contract. The factory then proceeds to assign the minter and burner roles to
the msg.sender. The factory remains the owner.

**Solution:**
Set associated roles and transfer ownership to the HubPool.


mrice32 (approved):

nicholaspai (approved):
LGTM

PR #115, opened 2022-03-17
fix[L04] Enforce chainId requirements in PolygonTokenBridger

mrice32:
**Problem:**

The PolygonTokenBridger contract's primary functions are only intended to be called either on l1 or
l2, but not both. In fact, calling the functions on the wrong chain could result in unexpected behavior
and unnecessary confusion.

In the best case, the functions will simply revert if called from the wrong chain because they will
attempt to interact with other contracts that do not exist on that chain. For example, calling the
receive function (by sending the contract some native asset) could trigger reverts on Polygon, but not
on Ethereum, because there is a WETH contract at the l1Weth address on the latter but not the former.

However, in the worst case, it is possible that such calls will not revert, but result in lost funds instead.
For example, if a WETH-like contract was later deployed to the l1Weth address on Polygon, then the
call would not revert. Instead, tokens would be sent to that contract and could remain stuck there.

Although the inline documentation details which function should be called on which chain, consider
having the functions in this contract actively enforce these requirements via limiting execution to the
correct block.chainid.

**Solution**

Add l1 and l2 chain ids to the polygon bridger, so create2 still guarantees matching addresses on both chains, but certain methods can be disabled depending on the env.

### DIFF
--- a/contracts/PolygonTokenBridger.sol
+++ b/contracts/PolygonTokenBridger.sol
@@ -40,14 +40,36 @@ contract PolygonTokenBridger is Lockable {
     // WETH contract on Ethereum.
     WETH9 public immutable l1Weth;
 
+    // Chain id for the L1 that this contract is deployed on or communicates with.
+    // For example: if this contract were meant to facilitate transfers from polygon to mainnet, this value would be
+    // the mainnet chainId 1.
+    uint256 public immutable l1ChainId;
+
+    // Chain id for the L2 that this contract is deployed on or communicates with.
+    // For example: if this contract were meant to facilitate transfers from polygon to mainnet, this value would be
+    // the polygon chainId 137.
+    uint256 public immutable l2ChainId;
+
+    modifier onlyChainId(uint256 chainId) {
+        _requireChainId(chainId);
+        _;
+    }
+
     /**
      * @notice Constructs Token Bridger contract.
      * @param _destination Where to send tokens to for this network.
      * @param _l1Weth Ethereum WETH address.
      */
-    constructor(address _destination, WETH9 _l1Weth) {
+    constructor(
+        address _destination,
+        WETH9 _l1Weth,
+        uint256 _l1ChainId,
+        uint256 _l2ChainId
+    ) {
         destination = _destination;
         l1Weth = _l1Weth;
+        l1ChainId = _l1ChainId;
+        l2ChainId = _l2ChainId;
     }
 
     /**
@@ -60,7 +82,7 @@ contract PolygonTokenBridger is Lockable {
         PolygonIERC20 token,
         uint256 amount,
         bool isWrappedMatic
-    ) public nonReentrant {
+    ) public nonReentrant onlyChainId(l2ChainId) {
         token.safeTransferFrom(msg.sender, address(this), amount);
 
         // In the wMatic case, this unwraps. For other ERC20s, this is the burn/send action.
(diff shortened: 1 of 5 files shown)


contracts/PolygonTokenBridger.sol
```diff
@@ -40,14 +40,36 @@ contract PolygonTokenBridger is Lockable {
...
+    // For example: if this contract were meant to facilitate transfers from polygon to mainnet, this value would be
+    // the mainnet chainId 1.
+    uint256 public immutable l1ChainId;
+
+    // Chain id for the L2 that this contract is deployed on or communicates with.
+    // For example: if this contract were meant to facilitate transfers from polygon to mainnet, this value would be
+    // the polygon chainId 137.
+    uint256 public immutable l2ChainId;
+
+    modifier onlyChainId(uint256 chainId) {
```
nicholaspai, line 53: nit instead of using modifier, does calling internal method _requireChainId directly in `send` and `retrieve` use less gas? I think its readable either way

test/chain-specific-spokepools/Polygon_SpokePool.ts
```diff
@@ -163,9 +165,10 @@ describe("Polygon Spoke Pool", function () {
   });
 
   it("PolygonTokenBridger retrieves and unwraps tokens correctly", async function () {
+    const chainId = await owner.getChainId();
     const polygonTokenBridger = await (
       await getContractFactory("PolygonTokenBridger", owner)
-    ).deploy(hubPool.address, weth.address);
+    ).deploy(hubPool.address, weth.address, chainId, chainId);
```
nicholaspai, line 171: I think you could test this by deploying new `PolygonTokenBridgers` with incorrect l1/l2 chainIDs and just testing that `send` and `retrieve` revert if the chain IDs are set unexpectedly

mrice32: Agree, will add a test for this!

contracts/PolygonTokenBridger.sol
mrice32: My understanding is that modifiers are usually inlined, so I expect that it would be the same either way.

test/chain-specific-spokepools/Polygon_SpokePool.ts
mrice32: Done.

chrismaree (approved):

contracts/PolygonTokenBridger.sol
```diff
@@ -74,12 +96,22 @@ contract PolygonTokenBridger is Lockable {
...
     }
 
     receive() external payable {
-        // Note: this should only happen on the mainnet side where ETH is sent to the contract directly by the bridge.
-        if (functionCallStackOriginatesFromOutsideThisContract()) l1Weth.deposit{ value: address(this).balance }();
+        if (functionCallStackOriginatesFromOutsideThisContract()) {
+            // This should only happen on the mainnet side where ETH is sent to the contract directly by the bridge.
+            _requireChainId(l1ChainId);
+            l1Weth.deposit{ value: address(this).balance }();
+        } else {
```
nicholaspai, line 108: more succinct: `_requireChainId(functionCallStackOriginatesFromOutsideThisContract() ? l1ChainId : l2ChainId)`

nicholaspai (approved):
LGTM with one nit suggestion (feel free to ignore it)

contracts/PolygonTokenBridger.sol
mrice32: Only reason I don't want to do this is that we still need to do the l1Weth.deposit call in a separate branch, which makes this feel a bit messier since we're doing the same branch twice.

PR #118, opened 2022-03-21
feat: Added rate model store from across v1

chrismaree:
We need the rate model store from across V1. This PR simply copy and pastes this contract with no changes other than one modification to an import.

### DIFF
--- a/contracts/RateModelStore.sol
+++ b/contracts/RateModelStore.sol
@@ -1,0 +1,29 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity ^0.8.0;
+
+import "@uma/core/contracts/common/implementation/MultiCaller.sol";
+import "@openzeppelin/contracts/access/Ownable.sol";
+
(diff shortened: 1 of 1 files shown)


nicholaspai (approved):
LGTM! Should we also use this contract to set other UMIP level constraints like:
- max # of L1 tokens in PoolRebalanceLeaves
- max # of refunds in RelayerRefundLeaves
- Other things?

Matt and I also toyed with the idea of storing the whitelist of tokens here but I can see arguments against that churn right now

chrismaree:
> LGTM! Should we also use this contract to set other UMIP level constraints like:
> 
> * max # of L1 tokens in PoolRebalanceLeaves
> * max # of refunds in RelayerRefundLeaves
> * Other things?
> 
> Matt and I also toyed with the idea of storing the whitelist of tokens here but I can see arguments against that churn right now

separate PR I think!

mrice32 (approved):

PR #119, opened 2022-03-21
improve: address solhint warnings

nicholaspai:
CI process does not currently include solhint which we can run via "yarn solhint ./contracts/**/*.sol"
﻿Signed-off-by: nicholaspai <npai.nyc@gmail.com>


mrice32 (approved):
LGTM!

PR #120, opened 2022-03-21
fix: Arbitrum Adapter needs to pay for L2 gas when relaying tokens

nicholaspai:
I noticed that `executeRootProposal` was reverting when attempting to send tokens via `relayTokens`. This PR adds the following fixes to get this to work:
- `bytes data` must be passed as the last param to `outboundTransfer` and the data must include `maxSubmissionCost`. Specific implementation details are in PR
- The total ETH to cover `maxSubmissionCost + gasPriceBid * gasLimit` must be sent as `msg.value` in `outboundTransfer` call
- The specific token's gateway, not router, must be approved to pull the tokens-to-be-bridged into escrow so the HubPool needs to set an allowance


mrice32 (approved):
LGTM!

PR #121, opened 2022-03-22
fix: store root bundles on SpokePool in mapping instead of array

nicholaspai:
Currently, every call to `relayRootBundle` pushes a new root bundle at the end of an array of `RootBundle` structs. This means that the SpokePool is vulnerable to replay attacks by the bridge operator who could publish duplicate root bundles to the SpokePool to easily drain it of its funds. This isn't as much of a risk in reality if the bridges we use are those we trust and that have big reputations that would deter them from replaying transactions. 

However, we might as well add replay protection if the code is simple enough--which it is. This PR changes the `relayRootBundle` implementation so that it accepts a `uint256 rootBundleID` that instructs the contract where to store the new root bundles. This allows the HubPool to pass in a unique ID to the cross-chain `relayRootBundle` call to prevent publishing duplicate root bundles. 

The HubPool uses the `requestExpirationTimestamp` as an easy UUID without having to store extra data.


chrismaree (commented):
this is pretty great @nicholaspai . simple reasoning but adds a layer of control and protection.

Do you not need to change the types passed in for `rootBundleId` when calling `executeRelayerRefundLeaf` and equally for `executeSlowRelayLeaf`?

nicholaspai:
> this is pretty great @nicholaspai . simple reasoning but adds a layer of control and protection .
> 
> Do you not need to change the types passed in for `rootBundleId` when calling `executeRelayerRefundLeaf` and equally for `executeSlowRelayLeaf`?

Good point, done! This was tricky as Solidity can implicitly convert uint32 => uint256 so there weren't compiler errors

contracts/HubPool.sol
```diff
@@ -671,7 +671,9 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
                     "relayMessage(address,bytes)",
                     spokePool, // target. This should be the spokePool on the L2.
                     abi.encodeWithSignature(
-                        "relayRootBundle(bytes32,bytes32)",
+                        "relayRootBundle(uint256,bytes32,bytes32)",
+                        // Use request expiration time as unique identifier for root bundle
+                        rootBundleProposal.challengePeriodEndTimestamp,
```
mrice32, line 676: Do we know how much additional gas this additional variable adds to the cross-chain bridge costs?

I think it would be preferable if we could find a setup where we could guarantee that each relayerRefundRoot was guaranteed to be unique and then just use that root as the key that points to a boolean + claim bitmap.

This has a few benefits:
- Less data being passed, which should save gas.
- No need to know this timestamp when trying to find the data on L2. You just have to pass the root (which you would need to create the proof), data, and proof.
- Conceptually simpler.

After thinking on this more, I think there's also an argument that replay protection should never be necessary unless we allow for sending the message multiple times on L1 (which we currently don't). The reasoning is pretty simple here: if you _don't_ have replay protection on a canonical bridge, double mints of bridgeable tokens would also presumably be possible, breaking a lot of core assumptions of the bridge.

Thoughts?

nicholaspai: How could you guarantee the relayer refund root, or any of the roots for that matter, unique without introducing some arbitrary nonce that the dataworker on L2 would have to be aware of? My point being that this nonce is as inconvenient for the dataworker to remember as the `challengePeriodEndTimestamp`.

mrice32: @nicholaspai I agree it would be arbitrary. It could simply be a counter that increments on each successful proposal. Right now, we already use information from the previous proposal, so I don't think this would be a huge lift.

I think both approaches, admittedly, are a bit clunky.

Going back to the last point. Do you agree that replay protection should be, by definition, guaranteed by any L2 that uses their own bridge to relay tokens (which we rely on anyway)?

nicholaspai: > Do you agree that replay protection should be, by definition, guaranteed by any L2 that uses their own bridge to relay tokens (which we rely on anyway)?

I'm a bit worried about this assumption which I think holds so far in practice with Optimism/Arbitrum/Polygon but I don't think its guaranteed. I'd rather be safe and just protect it and tradeoff dataworker UX. WDYT @chrismaree @mrice32

mrice32: @chrismaree thoughts?

PR #122, opened 2022-03-23
feat: release

chrismaree:
Updates released contracts.


nicholaspai (approved):

contracts/RateModelStore.sol
```diff
@@ -0,0 +1,29 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity ^0.8.0;
+
+import "@uma/core/contracts/common/implementation/MultiCaller.sol";
```
nicholaspai, line 4: why is this diff here? I thought we already added this contract

chrismaree: due to need to rebase. have fixed.

PR #128, opened 2022-03-25
fix: polygon testnet fixes

mrice32:
This PR includes three changes related to testnet testing against mumbai:
1. The PolygonTokenBridger's receive function needs to be gas-minimized because the polygon contracts provide minimal gas when sending ETH to it.
2. The approval in the Polygon adapter must be for the erc20 predicate, not the root chain manager that it directly calls.
3. The SpokePool must wrap matic that it receives from the plasma bridge.
4. The PolygonTokenBridger needs to be able to call startExitWithBurntTokens with custom data to complete the L1 exit when withdrawing matic via the plasma bridge on L1.
5. The Polygon Adapter contract must be able to send matic via the plasma bridge rather than the normal PoS bridge.

Note: the bi-directional transfer of ERC20s, WETH, and Matic has been tested on mumbai.


nicholaspai (approved):
Can you add a PR description explaining why we want to remove wrapping from the fallback functikn? LGTM otherwise I think allowing anyone to wrap this contracts ETH makes sense

chrismaree (approved):

contracts/PolygonTokenBridger.sol
```diff
@@ -59,15 +72,20 @@ contract PolygonTokenBridger is Lockable {
     /**
      * @notice Constructs Token Bridger contract.
      * @param _destination Where to send tokens to for this network.
-     * @param _l1Weth Ethereum WETH address.
+     * @param _polygonRegistry L1 registry that stores updated addresses of polygon contracts.
```
nicholaspai, line 75: Can you add a comment explaining how to set this on L2 since its described as an "L1 registry"

contracts/Polygon_SpokePool.sol
```diff
@@ -127,6 +127,16 @@ contract Polygon_SpokePool is IFxMessageProcessor, SpokePool {
         require(success, "delegatecall failed");
     }
 
+    /**
+     * @notice Allows the caller to trigger the wrapping of any unwrapped matic tokens.
+     * @dev Matic sends via L1 -> L2 bridging actions don't call into the contract receiving the tokens, so wrapping
+     * must be done via a separate transaction.
+     */
+    function wrap() public nonReentrant {
```
nicholaspai, line 135: Should `wrap` be triggered on any other external functions like `processMessageFromRoot` for convenience?

nicholaspai (approved):
LGTM, left some comments just requesting more detailed comments

contracts/PolygonTokenBridger.sol
mrice32: Sure. Can do. It should still be the L1 address to ensure that this gets deployed at an identical address on both chains.

contracts/Polygon_SpokePool.sol
mrice32: It could. Note: the functions that actually would need the tokens aren't triggered by processMessageFromRoot, though. executing relayer refunds and slow relays are the ones that would require this, but to add it to those, they might need to be overridden.

Will think a little about how to do this with little overhead.

contracts/PolygonTokenBridger.sol
mrice32: Done.

contracts/Polygon_SpokePool.sol
mrice32: Added this functionality in the last commit: https://github.com/across-protocol/contracts-v2/pull/128/commits/811ac20674d28189fd01297c05ce5b9e89f7a183. PTAL.

mrice32: I don't really think this extra check is worthwhile because:
1. We already check `address(this).balance > 0`, which is _extremely_ cheap. It costs 5 gas according to https://ethereum.org/en/developers/docs/evm/opcodes/ to read the balance. Reading the `weth` value from the contract, by comparison, would cost 100 - 2100 gas. I think these checks are the only difference between the two, since we will likely run _some_ matic execution step each time matic tokens are sent over to the spoke, meaning the number of wraps shouldn't really differ between the two.
2. Since this is on polygon, gas costs shouldn't really matter too much.

mrice32: As discussed offline, we'll leave this as is.

contracts/test/PolygonMocks.sol
```diff
@@ -15,3 +17,29 @@ contract FxStateSenderMock {
...
+    function depositERC20ForUser(
+        address token,
+        address user,
+        uint256 amount // solhint-disable-next-line no-empty-blocks
+    ) external {} // solhint-disable-line no-empty-blocks
+}
+
+contract PolygonRegistryMock {
+    // solhint-disable-next-line no-empty-blocks
+    function erc20Predicate() external returns (address predicate) {}
```
ericDeCourcy, line 31: I know this is a mock, but it looks like using the "registry" from [polygon's list](https://static.matic.network/network/mainnet/v1/index.json) may give the wrong value. The Registry is [0x33a02E6cC863D393d6Bf231B697b82F6e499cA71](https://etherscan.io/address/0x33a02E6cC863D393d6Bf231B697b82F6e499cA71#readContract), and you can see that function `erc20Predicate` returns [0x158d5fa3Ef8e4dDA8a5367deCF76b94E7efFCe95](https://etherscan.io/address/0x158d5fa3ef8e4dda8a5367decf76b94e7effce95). Just based on the tx's i see calling that contract, it's for exiting from Matic, but not bridging/depositing onto matic. 

Again since this is a mock, you can have it return whatever value you want. So not a problem. But it might lead to someone thinking that the actual mainnet contract will return the "deposit" predicate, which it doesn't. I believe the deposit predicate is [0x40ec5b33f54e0e8a33a975908c5ba1c14e5bbbdf](https://etherscan.io/address/0x40ec5b33f54e0e8a33a975908c5ba1c14e5bbbdf).

I'm thinking that Polygon_Adapter.sol is deployed on ETH mainnet, for transferring tokens over to Matic. If that's not true... this doesn't apply

PR #131, opened 2022-04-06
improve: Change indexed event params to assist offchain dataworker

nicholaspai:
Dataworker will compute roots for a given bundle block range and look up `SpokePool.RelayedRootBundle` events with matching roots, so we should index on those roots.

Additionally, the HubPool's `ProposeRootBundle` event doesn't need to index any of the root hashes. If anything, it should index the `bundleBlockRange` which is the key for the dataworker to reconstruct roots, but even this might be unneccessary.

### DIFF
--- a/contracts/HubPool.sol
+++ b/contracts/HubPool.sol
@@ -149,9 +149,9 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
     event ProposeRootBundle(
         uint32 challengePeriodEndTimestamp,
         uint64 poolRebalanceLeafCount,
-        uint256[] bundleEvaluationBlockNumbers,
+        uint256[] indexed bundleEvaluationBlockNumbers,
         bytes32 indexed poolRebalanceRoot,
-        bytes32 indexed relayerRefundRoot,
+        bytes32 relayerRefundRoot,
         bytes32 slowRelayRoot,
         address indexed proposer
     );
--- a/contracts/HubPool.sol
+++ b/contracts/HubPool.sol
@@ -149,9 +149,9 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
     event ProposeRootBundle(
         uint32 challengePeriodEndTimestamp,
         uint64 poolRebalanceLeafCount,
-        uint256[] indexed bundleEvaluationBlockNumbers,
+        uint256[] bundleEvaluationBlockNumbers,
         bytes32 indexed poolRebalanceRoot,
-        bytes32 relayerRefundRoot,
+        bytes32 indexed relayerRefundRoot,
         bytes32 slowRelayRoot,
         address indexed proposer
     );
--- a/contracts/SpokePool.sol
+++ b/contracts/SpokePool.sol
@@ -98,7 +98,11 @@ abstract contract SpokePool is SpokePoolInterface, Testable, Lockable, MultiCall
         address recipient,
         bool isSlowRelay
     );
-    event RelayedRootBundle(uint32 indexed rootBundleId, bytes32 relayerRefundRoot, bytes32 slowRelayRoot);
+    event RelayedRootBundle(
+        uint32 indexed rootBundleId,
+        bytes32 indexed relayerRefundRoot,
+        bytes32 indexed slowRelayRoot
+    );
     event ExecutedRelayerRefundRoot(
         uint256 amountToReturn,
         uint256 indexed chainId,


contracts/HubPool.sol
```diff
@@ -149,9 +149,9 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
     event ProposeRootBundle(
         uint32 challengePeriodEndTimestamp,
         uint64 poolRebalanceLeafCount,
-        uint256[] bundleEvaluationBlockNumbers,
+        uint256[] indexed bundleEvaluationBlockNumbers,
```
mrice32, line 152: I don't understand why we would need to index this param? We would only need to index by this if we _knew_ the bundle evaluation block numbers and wanted to find the event based on that?

contracts/SpokePool.sol
```diff
@@ -98,7 +98,11 @@ abstract contract SpokePool is SpokePoolInterface, Testable, Lockable, MultiCall
         address recipient,
         bool isSlowRelay
     );
-    event RelayedRootBundle(uint32 indexed rootBundleId, bytes32 relayerRefundRoot, bytes32 slowRelayRoot);
+    event RelayedRootBundle(
+        uint32 indexed rootBundleId,
+        bytes32 indexed relayerRefundRoot,
+        bytes32 indexed slowRelayRoot
```
mrice32, line 104: Not sure if this is _strictly_ required, but I think it's reasonable to add these as indexed params.

The reason I don't think it's required is that I would expect that we would, for instance, query the last `n` `rootBundleId`s, reconstruct those merkle roots based on finding the corresponding mainnet events based on the associated roots?

Or were you envisioning it starting with a mainnet query that goes back `n` blocks to determine the roots and then finding the associated rootBundleIds on the L2s for those roots?

Not strongly opinionated here, and I think this is reasonable to add regardless, just wanted to better understand the design.

mrice32 (commented):

contracts/HubPool.sol
phyllus-san: > We would only need to index by this if we knew the bundle evaluation block numbers and wanted to find the event based on that?

Yeah well I was thinking this would be a real situation, we can just ignore it for now.

contracts/SpokePool.sol
phyllus-san: > reconstruct those merkle roots based on finding the corresponding mainnet events based on the associated roots?

What do you mean by this?

I was envisioning the dataworker first looking up some `ProposeRootBundle` event on L1. Then it gets the `bundleBlockEvalNumbers` from that event, reconstructs roots for those blocks, and then finds a `RelayedRootBundle` event matching those roots

chrismaree: I think the spokepool & hubpool have all this information in their clients already and I’m pretty sure you can do this join without needing these to be indexed with all the data available in the clients you can look at `RelayedRootBundle` events on the spoke pool and relate it to `ProposeRootBundle` events by joining on the `slowRelayRoot` or `relayerRefundRoot` and from this you can find the `rootBundleId`.

i.e:

1. find all `RelayedRootBundle`. 
2. For each ^^, join to an an associated event on the `HubPool` via the `ProposeRootBundle `event.
3. from this, find the associated `bundleEvaluationBlockNumbers`
4. use ^^ to find all valid deposits in period to reconstruct tree.
5. construct each leaf and associated proof. 
6. simulate execution on target chain.
7. If it can be executed, execute.

alternative to 6 and 7 is to look at `ExecutedRelayerRefundRoot` and work out which are still to be executed.

nicholaspai: > reconstruct those merkle roots based on finding the corresponding mainnet events based on the associated roots?

What do you mean by this?

I was envisioning the dataworker first looking up some `ProposeRootBundle` event on L1. Then it gets the `bundleBlockEvalNumbers` from that event, reconstructs roots for those blocks, and then finds a `RelayedRootBundle` event matching those roots

nicholaspai: @chrismaree why do you need to look up ALL `RelayedRootBundle` though to execute leaves for 1 bundle block range?

contracts/HubPool.sol
nicholaspai: > We would only need to index by this if we knew the bundle evaluation block numbers and wanted to find the event based on that?

Yeah well I was thinking this would be a real situation, we can just ignore it for now.

mrice32 (approved):

contracts/SpokePool.sol
chrismaree: > @chrismaree why do you need to look up ALL RelayedRootBundle though to execute leaves for 1 bundle block range?

you dont need to but we have them anyways in the client :)

chrismaree (approved):

contracts/SpokePool.sol
nicholaspai: Hmm then I see your point that we don't need this indexed... what to do?

nicholaspai: Maybe someone else would want to index it since matching roots is the canonical way to identify `rootBundleId`?

PR #133, opened 2022-04-08
improve: Add appliedRelayerFeePct to FillRelay event

nicholaspai:
Emitting both the original and the updated `relayerFeePct` allows off-chain clients to distinguish between correctly and incorrectly sped up `fillRelay` calls with updated relayer fee %'s. Currently, the client has no way to determine if a `FillRelay` event emitted by a speed up relay (i.e. calling `fillRelayWithUpdatedFee`) was a valid fill or not. 

The current exploit would be to call `fillRelay` for a `deposit` with ALL of the correct params except for `relayerFeePct`, which the attacker would set to something lower than expected like `0`. Clients would not be able to easily detect if this modified `relayerFeePct=0` came from a correct `fillRelayWithUpdatedFee` call.

### DIFF
--- a/contracts/SpokePool.sol
+++ b/contracts/SpokePool.sol
@@ -82,7 +82,6 @@ abstract contract SpokePool is SpokePoolInterface, Testable, Lockable, MultiCall
         bytes depositorSignature
     );
     event FilledRelay(
-        bytes32 indexed relayHash,
         uint256 amount,
         uint256 totalFilledAmount,
         uint256 fillAmount,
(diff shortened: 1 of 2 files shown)


contracts/SpokePool.sol
```diff
@@ -795,19 +795,19 @@ abstract contract SpokePool is SpokePoolInterface, Testable, Lockable, MultiCall
         bytes32 relayHash,
         uint256 fillAmount,
         uint256 repaymentChainId,
-        uint64 relayerFeePct,
+        uint64 appliedRelayerFeePct,
         RelayData memory relayData,
         bool isSlowRelay
     ) internal {
         emit FilledRelay(
-            relayHash,
```
nicholaspai, line 803: I removed this to get around a `stack too deep error` but also realized we don't use `relayHash` in the current client code so its a reasonable delete

nicholaspai:
Note, this PR will require changing the `Fill` object interface in `relayer-v2/interfaces/SpokePool` and adding a test that valid speed up fills are included in `fillsToRefund` while invalid `fill` calls with all matching params except for the `relayerFeePct` are ignored.

mrice32 (approved):

PR #142, opened 2022-04-15
fix: [L02] prevent locking in the case of misparameterization

mrice32:
This addresses two potential cases where tokens get locked:
1. A caller incorrectly sets isWrappedMatic to false, but provides the wMatic token address.
2. ERC20 tokens are erroneously transferred to the contract.

In both cases, this ensures that the tokens make it to the HubPool contract.


nicholaspai:
Can you explain the failure case this addresses?

mrice32:
> Can you explain the failure case this addresses?

@nicholaspai great point. Added a description that describes this.

contracts/PolygonTokenBridger.sol
```diff
@@ -88,10 +88,13 @@ contract PolygonTokenBridger is Lockable {
...
         // In the wMatic case, this unwraps. For other ERC20s, this is the burn/send action.
-        token.withdraw(amount);
+        token.withdraw(token.balanceOf(address(this)));
 
         // This takes the token that was withdrawn and calls withdraw on the "native" ERC20.
-        if (isWrappedMatic) maticToken.withdraw{ value: amount }(amount);
+        if (isWrappedMatic) {
+            uint256 balance = address(this).balance;
+            maticToken.withdraw{ value: balance }(balance);
+        }
```
chrismaree, line 97: nit: does this need to be a separate state variable? could it not be:

```suggestion
        if (isWrappedMatic) maticToken.withdraw{ value: address(this).balance }(balance);
```

nicholaspai: seems reasonable and saves an SSTORE right?

mrice32: I don't think the above works, since balance doesn't exist. Maybe you meant:
```
if (isWrappedMatic) maticToken.withdraw{ value: address(this).balance }(address(this).balance);
```
?

We could do that. I'm not sure it would be cheaper, but that might not matter since this is on polygon and the call is already pretty cheap.

mrice32: Done.

nicholaspai (approved):
Nice and clean, LGTM

contracts/PolygonTokenBridger.sol
chrismaree: ye, sorry. it's hard to structure feedback in these comments with the proposed blocks. what you did is what I had intended.

PR #143, opened 2022-04-20
chrismaree/add block number to export

chrismaree:
This PR adds additional exports to our deployments file. they now contain the block number at which each contract on each network is exported. this is used by the relayer to decrease how many blocks lookback we need when initializing clients to decrease node burden.


deployments/deployments.json
```diff
@@ -1,41 +1,43 @@
...
-    "Arbitrum_Adapter": "0x18F4D98C7CeA6Ab934F2976c2a98009A529d8F49",
-    "Ethereum_Adapter": "0x682DEa71e7246910A4dec2396a53B24291EB0AD8",
-    "Ethereum_SpokePool": "0xB078bBb35f8E24c2431b9d2a88C0bC0c26CC1F92",
-    "HubPool": "0xa1b6DA4AaE90fA16F3A3338c8d1Dc70B4926FCa7",
-    "LpTokenFactory": "0x0668ab3839346ebf95d969b3e18B2a96b1CC2b02",
-    "Optimism_Adapter": "0x277C7ecB45851dcB49c494fB4D0d9d8228037f0C",
-    "PolygonTokenBridger": "0xF92B101f07df74B1f4f3160f9e8a25D48BA5d583",
-    "Polygon_Adapter": "0xd8176EBf6170513Aa232D065042fa480557232A4",
-    "RateModelStore": "0x18a200A2427e243154EC8217890Bc08062afc5A2"
+    "Arbitrum_Adapter": { "address": "0x18F4D98C7CeA6Ab934F2976c2a98009A529d8F49", "blockNumber": 10367195 },
```
chrismaree, line 3: notice this! we now have exported address and block number for all contracts on all chains :)

src/DeploymentUtils.ts
```diff
@@ -0,0 +1,19 @@
...
+export function getDeployedAddress(contractName: string, networkId: number): string {
+  try {
+    return (deployments as any)[networkId.toString()][contractName].address;
+  } catch (_) {
+    throw new Error(`Contract ${contractName} not found on ${networkId} in deployments.json`);
+  }
+}
+
+// Returns the deployment block number of any contract on any network.
+export function getDeployedBlockNumber(contractName: string, networkId: number): string {
```
chrismaree, line 13: the bots can use this method to see where to query from.

chrismaree:
> Does this introduce breaking changes to relayer-v2?

nope! latest code there requires these changes.

PR #146, opened 2022-04-22
refactor: Remove AncillaryData import from HubPool

nicholaspai:
AncillaryData import no longer needed since PR #114 removed logic to construct ancillary data using this lib


mrice32 (approved):
LGTM!

PR #155, opened 2022-05-10
feat: add reverse look up for contract info from address

chrismaree:
- adds a helper function to find info about a deployed contract from it's address.
- refines `getDeployedAddress` to let you ignore the network name when fetching a `SpokePool` address.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@across-protocol/contracts-v2",
-  "version": "0.0.54",
+  "version": "0.0.55",
   "author": "UMA Team",
   "license": "AGPL-3.0",
   "repository": {


src/DeploymentUtils.ts
```diff
@@ -1,19 +1,47 @@
...
 // Returns the deployed address of any contract on any network.
 export function getDeployedAddress(contractName: string, networkId: number): string {
   try {
-    return (deployments as any)[networkId.toString()][contractName].address;
+    return deployments[networkId.toString()][contractName].address; // First try match directly on the name provided.
   } catch (_) {
+    try {
+      // If this does not match then try search on a Regex on SpokePool to let the caller exclude the chain name.
+      for (const _contractName of Object.keys(deployments[networkId.toString()]))
+        if (/.*_SpokePool/.test(_contractName)) return deployments[networkId.toString()][_contractName].address;
```
daywiss, line 15: this could error because of a typo, and then the user ends up getting spoke pool address when maybe they meant somethign completely different, which could be bad. 

I'd suggest doing something where you are finding in an array of keys if they include "contractName".  Since this logic is duplicated, id suggest making it a function.

somethign like:
```
function findContractName(partial){
  const found = [...Object.keys(deployments[networkId])].find(name=>name.includes(partial))
  if(found) return deployments[networkId.toString()][found]
  throw new Error('Unable to find contract matching ' + partial)
}
```

src/DeploymentUtils.ts
```diff
@@ -1,19 +1,47 @@
...
     throw new Error(`Contract ${contractName} not found on ${networkId} in deployments.json`);
   }
 }
 
 // Returns the deployment block number of any contract on any network.
-export function getDeployedBlockNumber(contractName: string, networkId: number): string {
+export function getDeployedBlockNumber(contractName: string, networkId: number): number {
   try {
-    return (deployments as any)[networkId.toString()][contractName].blockNumber;
+    return deployments[networkId.toString()][contractName].blockNumber;
```
daywiss, line 24: this should have the same behavior as getDeployedAddress.  imo we might just want a way to look up the correct contract name, then you dont have to modify thsese 2 functions at all

src/DeploymentUtils.ts
```diff
@@ -1,19 +1,47 @@
...
   } catch (_) {
     throw new Error(`Contract ${contractName} not found on ${networkId} in deployments.json`);
   }
 }
+
+// Returns the chainId and contract name for a given contract address.
+export function getContractInfoFromAddress(contractAddress: string): { chainId: Number; contractName: string } {
```
daywiss, line 31: it looks like this solves my issues since im usually given the contract address and chain ( at least for spoke pools).   It would be a problem if hubpools also had different names, but they dont currently.

src/DeploymentUtils.ts
```diff
@@ -1,19 +1,47 @@
...
 
 // Returns the deployed address of any contract on any network.
 export function getDeployedAddress(contractName: string, networkId: number): string {
   try {
-    return (deployments as any)[networkId.toString()][contractName].address;
+    return deployments[networkId.toString()][contractName].address; // First try match directly on the name provided.
   } catch (_) {
+    try {
+      // If this does not match then try search on a Regex on SpokePool to let the caller exclude the chain name.
+      for (const _contractName of Object.keys(deployments[networkId.toString()]))
```
mrice32, line 14: I think this is confusing. Can we just make aliases in the generation? Aka save an identical deployment under a different name using deployments.save?

Something like:

```ts
  const spokePoolDeployment = await deploy("Optimism_SpokePool", {
    from: deployer,
    log: true,
    skipIfAlreadyDeployed: true,
    args: [
      hubPool.address, // Set hub pool as cross domain admin since it delegatecalls the Optimism_Adapter logic.
      hubPool.address,
      "0x0000000000000000000000000000000000000000", // timer
    ],
  });
  await hre.deployments.save("SpokePool", spokePoolDeployment);
```

Thoughts?

mrice32 (commented):
One comment!

src/DeploymentUtils.ts
chrismaree: the only way that would happen is if you added `SpokePool` to the search. this regex will ONLY pass if `*_SpokePool` is set. you are right though that you could do something silly like `HubPool_SpokePool` which would return the spoke pool on that chain.

chrismaree: however, I have modified this to export the name of the contract as `SpokePool` on each respective chain to avoid needing this workaround alltogether.

src/DeploymentUtils.ts
chrismaree: fixed. see above comments

src/DeploymentUtils.ts
chrismaree: ye, this method assumes that the deployment addresses on each chain are unique enabling you to go backwards from a known address to the contract name and chainId.

deployments/deployments.json
```diff
@@ -4,7 +4,7 @@
     "Arbitrum_Adapter": { "address": "0x937958992799bF4A9a656E6596FD10d7Da5c2216", "blockNumber": 14704380 },
     "Boba_Adapter": { "address": "0x33B0Ec794c15D6Cc705818E70d4CaCe7bCfB5Af3", "blockNumber": 14716798 },
     "Ethereum_Adapter": { "address": "0x527E872a5c3f0C7c24Fe33F2593cFB890a285084", "blockNumber": 14704381 },
-    "Ethereum_SpokePool": { "address": "0x931A43528779034ac9eb77df799d133557406176", "blockNumber": 14704425 },
```
chrismaree, line 7: @mrice32 see the changes made here.

scripts/processHardhatExport.ts
```diff
@@ -15,6 +15,7 @@ export async function run(): Promise<void> {
           if (!processedOutput[chainId]) processedOutput[chainId] = {};
           const address = castExport[chainId][0]?.contracts[contractName].address;
           const blockNumber = findDeploymentBlockNumber(castExport[chainId][0].name, contractName);
+          if (/.*_SpokePool/.test(contractName)) contractName = "SpokePool"; // Strip the network name from the spoke pool in the contractName.
```
chrismaree, line 18: change to modify the export without changing the artifacts.

src/DeploymentUtils.ts
mrice32: I don't think this will break anything. We're just adding a _second record_ of the same contract.

So you'll have "Polygon_SpokePool" and "SpokePool" on polygon with all the same details.

Hardhat will still have the artifact it expects, but we will also export a second artifact under a more generic name.

src/DeploymentUtils.ts
daywiss: I see, i missed that check

nicholaspai (approved):
LGTM sorry for not seeing this as I should have rebased off this into master

src/DeploymentUtils.ts
mrice32: Does the PolygonTokenBridger break this? It has the same address on multiple chains

chrismaree: Ye that contract would not work with this method but that's ok. This method is only used in the bots to enhance logs and the bots never directly call those polygon contracts so it won't be an issue. 

I'll add an error that is thrown if multiple addresses are found to prevent any confusion

PR #159, opened 2022-05-17
fix: address issue with weth on boba

chrismaree:


contracts/Boba_SpokePool.sol
```diff
@@ -13,6 +14,10 @@ contract Boba_SpokePool is Optimism_SpokePool {
      * @param _hubPool Hub pool address to set. Can be changed by admin.
      * @param timerAddress Timer address to set.
      */
+
+    address public immutable override l2Eth = address(0x4200000000000000000000000000000000000006);
```
nicholaspai, line 18: Maybe grab these from the Boba predeployed contracts that we can import

nicholaspai (approved):

mrice32 (approved):

PR #160, opened 2022-05-19
fix: add haircut proposal

chrismaree:
- Adds untested haircut logic that can scale the exchangeRateCurrent by some haircut amount, in the event of protocol loss of funds.
- Adds the ability to pause the addition/removal of liquidity. should be used in conjunction with haircutting to stop LPs from getting the wrong rate until the hair cut is done.

### DIFF
--- a/contracts/HubPool.sol
+++ b/contracts/HubPool.sol
@@ -176,7 +176,7 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
     }
 
     modifier unpaused() {
-        require(!paused, "Proposal process has been paused");
+        require(!paused, "Contract is paused");
         _;
     }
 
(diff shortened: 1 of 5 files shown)


contracts/HubPool.sol
```diff
@@ -420,6 +420,21 @@ contract HubPool is HubPoolInterface, Testable, Lockable, MultiCaller, Ownable {
...
+     * exchangeRateCurrent such that all LPs receive a pro rata loss of the the reserves. Should be used in conjunction
+     * with pause logic to prevent LPs from adding/withdrawing liquidity during the haircut process.
+     * Callable only by owner.
+     * @param l1Token Token to execute the haircut on.
+     * @param haircutAmount The amount of reserves to haircut the LPs by.
+     */
+    function haircutReserves(address l1Token, int256 haircutAmount) public onlyOwner nonReentrant {
+        // Note that we do not call sync first in this method. The Owner should call this manually before haircutting.
+        // This is done in the event sync is reverting due to too low balanced in the contract relative to bond amount.
+        pooledTokens[l1Token].utilizedReserves += haircutAmount;
```
nicholaspai, line 435: What if owner sets this amount too high? Does it break the contract permanently?

mrice32 (approved):

PR #162, opened 2022-05-23
feat: routes script can publish partial routes

mrice32:


chrismaree (approved):
LGTM! thanks for adding this

PR #165, opened 2022-05-27
feat: upgrade SNX L1 bridge

nicholaspai:

PR #166, opened 2022-05-28
fix: Standardize SNX custom Optimism bridge

kevinuma:
Separate SNX custom Optimism bridge logic into a separate contract that Optimism_Adapter can call.


nicholaspai (commented):
Hey I think #165 solves this L1 change needed, butwe need to think of a strategy for the L2 side (i.e. replacing OVM_SpokePool or coming up with a token bridge wrapper)

PR #167, opened 2022-05-29
improve + fix: Add a bridge wrapper for the non-standard SNX Optimism -> Ethereum bridge

kevinuma:
SNX bridge (https://github.com/Synthetixio/synthetix/blob/v2.71.0/contracts/SynthetixBridgeToBase.sol) is a non-standard Optimism bridge that implements withdrawTo differently (with 2 arguments instead of 4). In order to support it without hardcoding SNX token in the Ovm_SpokePool, this PR creates wrapper contract to standardize the SNX bridge. For any similar cases in the future, we can create BridgeWrapper contracts.


contracts/Ovm_SpokePool.sol
```diff
@@ -8,6 +8,8 @@ import "@eth-optimism/contracts/libraries/bridge/CrossDomainEnabled.sol";
 import "@eth-optimism/contracts/libraries/constants/Lib_PredeployAddresses.sol";
 import "@eth-optimism/contracts/L2/messaging/IL2ERC20Bridge.sol";
 
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
```
nicholaspai, line 11: Don't need to import this as its already imported in `SpokePool.sol`

contracts/Ovm_SpokePool.sol
```diff
@@ -141,17 +143,23 @@ contract Ovm_SpokePool is CrossDomainEnabled, SpokePool {
...
-            );
+
+        address l2TokenAddress = relayerRefundLeaf.l2TokenAddress;
+        address tokenBridge = tokenBridges[l2TokenAddress] == address(0)
+            ? Lib_PredeployAddresses.L2_STANDARD_BRIDGE
+            : tokenBridges[l2TokenAddress];
+        // In some cases such as SNX, the custom token bridge might require approval to transfer the ERC-20 before
+        // it can be burnt. We can grant approval and immediately revoke after the call to ensure no lingering
+        // approval remains.
+        IERC20(l2TokenAddress).approve(tokenBridge, relayerRefundLeaf.amountToReturn);
```
nicholaspai, line 154: Downside to this approach is extra gas per bridge call, but not sure there is a different way if we go through with this approach

contracts/bridge-wrappers/Synthetix_Optimism_BridgeWrapper.sol
```diff
@@ -0,0 +1,44 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity ^0.8.0;
```
nicholaspai, line 2: nit: Thoughts on naming this `SnxOptimismBridgeWrapper`

contracts/bridge-wrappers/Synthetix_Optimism_BridgeWrapper.sol
```diff
@@ -0,0 +1,44 @@
...
+    function withdrawTo(
+        address _l2Token,
+        address _to,
+        uint256 _amount,
+        uint32 _l1Gas,
+        bytes calldata _data
+    ) external {
+        IERC20 snxToken = IERC20(_l2Token);
+        // Caller should have already pre-approved this contract to pull SNX tokens.
+        snxToken.safeTransferFrom(msg.sender, address(this), _amount);
```
nicholaspai, line 41:
```suggestion
        IERC20(snxToken).safeTransferFrom(msg.sender, address(this), _amount);
```

Can just remove the first declaration of `snxToken` as I think this is just as readable

nicholaspai (commented):
I think this is exactly how I'd write the contracts so I'll wait to approve until after the unit tests are added! I think we shouldn't use this to upgrade an existing SpokePool like `Optimism`'s spoke pool but we can definitely use this spoke pool for the next chain's deployment

nicholaspai:
@chrismaree @mrice32 assuming we merge something like this we should plan to tell OZ about it and just make sure no risk is added

contracts/Ovm_SpokePool.sol
kevinuma: Agree but the gas increase should be minimal though, and also much cheaper considering this is Optimism as well.

contracts/Ovm_SpokePool.sol
kevinuma: We actually don't have to import it anywhere SafeErc20 is imported as that also pulls it in. I deleted it in SpokePOol and Ovm_SpokePool as well

contracts/bridge-wrappers/Synthetix_Optimism_BridgeWrapper.sol
kevinuma: Sounds good. I don't quite like the snakecase names anyway.

contracts/bridge-wrappers/Synthetix_Optimism_BridgeWrapper.sol
kevinuma: Done.

test/chain-specific-spokepools/Optimism_SpokePool.ts
```diff
@@ -141,4 +143,21 @@ describe("Optimism Spoke Pool", function () {
     const l2Eth = "0xDeadDeAddeAddEAddeadDEaDDEAdDeaDDeAD0000";
     expect(l2StandardBridge.withdrawTo).to.have.been.calledWith(l2Eth, hubPool.address, amountToReturn, 5000000, "0x");
   });
+
+  it.only("Bridge tokens to hub pool should not leave residual token approval", async function () {
```
nicholaspai, line 147: leftover .only

nicholaspai:
See this comment [here](https://github.com/across-protocol/contracts-v2/pull/165#issuecomment-1142597008) but this PR and that one can be closed until the new SNX bridge is deployed

mrice32 (commented):
Looks like we're gonna wait on SNX to upgrade. Feel free to re-request reviews if we do intend to merge this!

PR #172, opened 2022-06-29
feat: Arbitrum_RescueAdapter

nicholaspai:
This contract can be deployed to rescue ETH from the aliased HubPool address on Arbitrum﻿

### DIFF
--- a/contracts/chain-adapters/Arbitrum_RescueAdapter.sol
+++ b/contracts/chain-adapters/Arbitrum_RescueAdapter.sol
@@ -84,20 +84,24 @@ contract Arbitrum_RescueAdapter is AdapterInterface {
      * @param message Data to send to aliased hub pool.
      */
     function relayMessage(address, bytes memory message) external payable override {
+        uint256 valueToReturn = abi.decode(message, (uint256));
+
         uint256 requiredL1CallValue = _contractHasSufficientEthBalance();
 
+        // In the rescue ETH setup, we send the transaction to the refund address, we provide a call value equal to the
+        // amount we want to rescue, and we specify an empty calldata, since it's a simple ETH transfer.
         l1Inbox.createRetryableTicket{ value: requiredL1CallValue }(
-            aliasedL2HubPoolAddress, // destAddr destination L2 contract address
-            0, // l2CallValue call value for retryable L2 message
+            l2RefundL2Address, // destAddr destination L2 contract address
+            valueToReturn, // l2CallValue call value for retryable L2 message
             l2MaxSubmissionCost, // maxSubmissionCost Max gas deducted from user's L2 balance to cover base fee
             l2RefundL2Address, // excessFeeRefundAddress maxgas * gasprice - execution cost gets credited here on L2
             l2RefundL2Address, // callValueRefundAddress l2Callvalue gets credited here on L2 if retryable txn times out or gets cancelled
             l2GasLimit, // maxGas Max gas deducted from user's L2 balance to cover L2 execution
             l2GasPrice, // gasPriceBid price bid for L2 execution
-            message // data ABI encoded data of L2 message
+            "" // data ABI encoded data of L2 message
         );
 
-        emit MessageRelayed(aliasedL2HubPoolAddress, message);
+        emit MessageRelayed(aliasedL2HubPoolAddress, "");
     }
 
     /**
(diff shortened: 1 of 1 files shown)


contracts/chain-adapters/Arbitrum_RescueAdapter.sol
```diff
@@ -0,0 +1,127 @@
...
+     * @notice This contract must hold at least getL1CallValue() amount of ETH to send a message via the Inbox
+     * successfully, or the message will get stuck.
+     * @param message Data to send to aliased hub pool.
+     */
+    function relayMessage(address, bytes memory message) external payable override {
+        uint256 requiredL1CallValue = _contractHasSufficientEthBalance();
+
+        l1Inbox.createRetryableTicket{ value: requiredL1CallValue }(
+            aliasedL2HubPoolAddress, // destAddr destination L2 contract address
+            0, // l2CallValue call value for retryable L2 message
```
mrice32, line 91: This needs to be the amount you want to transfer out.

For instance, if you want to move 1ETH, this needs to be set to 1ETH.

contracts/chain-adapters/Arbitrum_RescueAdapter.sol
```diff
@@ -0,0 +1,127 @@
...
+     * @notice Send cross-chain message to aliased hub pool address on Arbitrum.
+     * @notice This contract must hold at least getL1CallValue() amount of ETH to send a message via the Inbox
+     * successfully, or the message will get stuck.
+     * @param message Data to send to aliased hub pool.
+     */
+    function relayMessage(address, bytes memory message) external payable override {
+        uint256 requiredL1CallValue = _contractHasSufficientEthBalance();
+
+        l1Inbox.createRetryableTicket{ value: requiredL1CallValue }(
+            aliasedL2HubPoolAddress, // destAddr destination L2 contract address
```
mrice32, line 90: This does not need to be specified. When a transaction is relayed to L2, it is _sent from_ the alias address. If you look here: https://arbiscan.io/address/0xd297fA914353c44B2e33EBE05F21846f1048CFeB, you can see all of our previous transactions coming _from_ that address.

Instead, this should be the address that you want to _receive_ the refunded ETH. In this case, it should probably be an immutable recipient address.

contracts/chain-adapters/Arbitrum_RescueAdapter.sol
```diff
@@ -0,0 +1,127 @@
...
+            0, // l2CallValue call value for retryable L2 message
+            l2MaxSubmissionCost, // maxSubmissionCost Max gas deducted from user's L2 balance to cover base fee
+            l2RefundL2Address, // excessFeeRefundAddress maxgas * gasprice - execution cost gets credited here on L2
+            l2RefundL2Address, // callValueRefundAddress l2Callvalue gets credited here on L2 if retryable txn times out or gets cancelled
+            l2GasLimit, // maxGas Max gas deducted from user's L2 balance to cover L2 execution
+            l2GasPrice, // gasPriceBid price bid for L2 execution
+            message // data ABI encoded data of L2 message
```
mrice32, line 97: This should be "".

contracts/chain-adapters/Arbitrum_RescueAdapter.sol
```diff
@@ -0,0 +1,127 @@
...
+        l2RefundL2Address = msg.sender;
+    }
+
+    /**
+     * @notice Send cross-chain message to aliased hub pool address on Arbitrum.
+     * @notice This contract must hold at least getL1CallValue() amount of ETH to send a message via the Inbox
+     * successfully, or the message will get stuck.
+     * @param message Data to send to aliased hub pool.
+     */
+    function relayMessage(address, bytes memory message) external payable override {
```
mrice32, line 86: We might want to enforce some restrictions on the length or type of data in message to ensure that if this is ever called by the HubPool to send roots to arbitrum that it reverts.

For instance, if we encode the ETH amount in the data portion, we probably want to ensure that the length is exactly long enough to store a uint256 and no more/no less.

mrice32 (commented):
This is great!

I think we should make this contract the most minimal contract possible and it could probably be used (and reused as many times as we need!) in a batch governance transaction:
1. Swap the existing arbitrum adapter for this contract.
2. Send a single relayMessage transaction to refund the ETH.
3. Swap the adapters back, so the normal production adapter is back in place.

mrice32 (approved):
@nicholaspai I fixed this up. Should be ready to go!

nicholaspai:
This looks great thanks for responding to all the issues you brought up and I think the decoding implementation of the message to enforce length is great !

PR #177, opened 2022-07-29
feat: add ethereum rescue adapter

mrice32:
This adapter is meant to rescue funds from a malfunctioning, broken, or misconfigured HubPool.

### DIFF
--- a/contracts/chain-adapters/Ethereum_RescueAdapter.sol
+++ b/contracts/chain-adapters/Ethereum_RescueAdapter.sol
@@ -1,0 +1,49 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity ^0.8.0;
+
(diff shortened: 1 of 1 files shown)


nicholaspai (approved):
This looks like it can't be front run. This would require calling relayMessage multiple times from HubPool, so if you want to consolidate into one call you could hardcode all ERC20 token addresses to rescue. Could even pass them in via construction... but this contract is prob not over generalizaing and just building a one time, super simple hardcodedcontract

mrice32:
> This looks like it can't be front run. This would require calling relayMessage multiple times from HubPool, so if you want to consolidate into one call you could hardcode all ERC20 token addresses to rescue. Could even pass them in via construction... but this contract is prob not over generalizaing and just building a one time, super simple hardcodedcontract

Yeah, I wanted to keep it as simple as possible to keep it low risk.

PR #181, opened 2022-08-10
improve(arbitrum-adapter): Use outboundTransferCustomRefund when sending tokens to L2

nicholaspai:
This [merged PR](https://github.com/OffchainLabs/arbitrum/pull/2107) is now live for most Arbitrum ERC20 gateways (except DAI), so we can use the new function outboundTransferCustomRefund to receive refunds at an EOA we control instead of the [aliased hub pool address](https://arbiscan.io/address/0xd297fa914353c44b2e33ebe05f21846f1048cfeb), which is the L1 sender of the tokens technically.

### DIFF
--- a/contracts/chain-adapters/Arbitrum_Adapter.sol
+++ b/contracts/chain-adapters/Arbitrum_Adapter.sol
@@ -29,6 +29,16 @@ interface ArbitrumL1ERC20GatewayLike {
         bytes calldata _data
     ) external payable returns (bytes memory);
 
+    function outboundTransferCustomRefund(
+        address _token,
+        address _refundTo,
+        address _to,
+        uint256 _amount,
+        uint256 _maxGas,
+        uint256 _gasPriceBid,
+        bytes calldata _data
+    ) external payable returns (bytes memory);
+
     function getGateway(address _token) external view returns (address);
 }
 
@@ -59,7 +69,7 @@ contract Arbitrum_Adapter is AdapterInterface {
     uint32 public immutable l2GasLimit = 2_000_000;
 
     // This address on L2 receives extra ETH that is left over after relaying a message via the inbox.
-    address public immutable l2RefundL2Address = 0x428AB2BA90Eba0a4Be7aF34C9Ac451ab061AC010;
+    address public immutable l2RefundL2Address;
 
     ArbitrumL1InboxLike public immutable l1Inbox;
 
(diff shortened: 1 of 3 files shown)


kevinuma (approved):

contracts/chain-adapters/Arbitrum_Adapter.sol
```diff
@@ -127,17 +137,32 @@ contract Arbitrum_Adapter is AdapterInterface {
...
-            l2GasPrice,
-            data
-        );
+        // Note: Legacy routers don't have the outboundTransferCustomRefund method, so default to using
+        // outboundTransfer(). Legacy routers are used for:
+        // - DAI
+        if (l1Token == 0x6B175474E89094C44Da98b954EedeAC495271d0F) {
+            // Note: outboundTransfer() will ultimately create a retryable ticket and set this contract's address as the
+            // refund address. This means that the excess ETH to pay for the L2 transaction will be sent to the aliased
+            // contract address on L2 and lost.
```
mrice32, line 146: nit: we don't lose the ETH, it's just hard to retrieve (we have a custom adapter for retrieving it now).

contracts/chain-adapters/Arbitrum_Adapter.sol
```diff
@@ -127,17 +137,32 @@ contract Arbitrum_Adapter is AdapterInterface {
...
-            to,
-            amount,
-            l2GasLimit,
-            l2GasPrice,
-            data
-        );
+        // Note: Legacy routers don't have the outboundTransferCustomRefund method, so default to using
+        // outboundTransfer(). Legacy routers are used for:
+        // - DAI
+        if (l1Token == 0x6B175474E89094C44Da98b954EedeAC495271d0F) {
```
mrice32, line 143: Is DAI the only token this applies to? Or just the only token _we currently support_ that this applies to?

mrice32 (approved):
LGTM!

contracts/chain-adapters/Arbitrum_Adapter.sol
nicholaspai: The latter, DAI is the only token we currently support that this applies to, I'll clarify the comment

contracts/chain-adapters/Arbitrum_Adapter.sol
nicholaspai: Updated

contracts/chain-adapters/Arbitrum_Adapter.sol
mrice32: Sounds good!

mrice32 (approved):

PR #183, opened 2022-08-18
chore: remove zksync-web3 from dependencies

nicholaspai:
We don't use zksync-web3 since we use the hardhat-deploy plugin's native support for zksync


mrice32 (approved):
LGTM!

PR #185, opened 2022-09-05
feat: add merkle distributor contract

amateima:
This PRs adds the contract source code and tests for the Merkle Distributor contract. It is based on the already audited contract from the UMA protocol [repository](https://github.com/UMAprotocol/protocol/blob/master/packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol)


package.json
```diff
@@ -7,6 +7,9 @@
     "type": "git",
     "url": "git+https://github.com/across-protocol/across-smart-contracts-v2.git"
   },
+  "engines": {
+    "node": ">=8.3.0"
```
nicholaspai, line 11: was this an issue for you?

nicholaspai (approved):

contracts/merkle-distributor/MerkleDistributor.sol
```diff
@@ -0,0 +1,257 @@
+// SPDX-License-Identifier: AGPL-3.0-only
```
mrice32, line 1: When you say "based on" do you mean directly copied? Or are there changes? If there are changes, we will likely need to have this re-audited.

test/merkle-distributor/MerkleDistributor.ts
```diff
@@ -0,0 +1,692 @@
+import { assert } from "chai";
```
mrice32, line 1: Same question here: is this copied code or should we review it? If copied, were there any specific changes we should look at?

mrice32 (commented):
Looks great! Just two high level questions -- trying to understand how much of this is copied code vs freshly written.

package.json
amateima: @nicholaspai 
The package builds just fine, but ESLint shows linting error `Rest/spread properties are not supported until Node.js 8.3.0. The configured version range is '>=8.0.0'`

contracts/merkle-distributor/MerkleDistributor.sol
amateima: The contract itself is identical

test/merkle-distributor/MerkleDistributor.ts
amateima: The tests structure is the same, only minor things were changed: replace web3 usage with ethers and update the assertion syntax (use `expect` instead of `assert` in order to have a similar syntax with the rest of the tests)

PR #186, opened 2022-10-10
chore: update `@uma/*` package dependencies

dohaki:
This PR updates some UMA package dependencies. We also need to update the deps in the @across-protocol/sdk-v2 repo after creating a new release of this package. These updates should fix the currently broken `@across-protocol/sdk-v2` package that errors with a `TypeError`:
```
/Users/dh.kim/Projects/across/merkle-distributor/node_modules/@uma/sdk/dist/src/clients/optimisticOracle/client.ts:16
export const contractInterface = new utils.Interface(getV1OptimisticOracleInterfaceAbi());
                                                     ^
TypeError: rateModelStore.getV1OptimisticOracleInterfaceAbi is not a function
    at Object.<anonymous> (/Users/dh.kim/Projects/across/merkle-distributor/node_modules/@uma/sdk/dist/src/clients/optimisticOracle/client.ts:16:54)
    at Module._compile (node:internal/modules/cjs/loader:1105:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1159:10)
    at Object.require.extensions.<computed> [as .js] (/Users/dh.kim/Projects/across/merkle-distributor/node_modules/ts-node/src/index.ts:1608:43)
    at Module.load (node:internal/modules/cjs/loader:981:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at Module.require (node:internal/modules/cjs/loader:1005:19)
    at require (node:internal/modules/cjs/helpers:102:18)
    at Object.<anonymous> (/Users/dh.kim/Projects/across/merkle-distributor/node_modules/@uma/sdk/dist/node/index.js:7:20)
    at Module._compile (node:internal/modules/cjs/loader:1105:14)
```
when trying to use it.

I did some digging and it seems to be rooted in conflicting versions of `@uma/contracts-node` and a breaking change somewhere between the versions **0.2.1** and **0.3.18** of that package. When installing a fresh latest version of the across sdk-v2 package it would install inside the root `node_modules`
```
- @uma/contracts-node -> 0.2.1 // latest 0.3.18
- @uma/sdk -> 0.23.4 // latest 0.29.3
    - node_modules
      - @uma/contracts-node -> 0.3.18
```
So the old source code of `@uma/sdk v0.23.4` would try to use the new code of `@uma/contracts-node v0.3.18` which are incompatible. Not sure why exactly the version resolution during installation behaves like that but updating the deps to the latest versions should fix the TypeError.


mrice32 (approved):
LGTM! Nice find and solid investigation.

amateima (approved):

PR #189, opened 2022-10-12
feat: goerli deployment MerkleDistributor

dohaki:
Goerli deployment of `MerkleDistributor`. See https://goerli.etherscan.io/address/0xc64226599444ca9b6d930603b507a04d8fa61add with owner 0x9A8f92a830A5cB89a3816e3D267CB7791c16b04D aka dev wallet


deployments/deployments.json
```diff
@@ -23,12 +23,13 @@
     "RateModelStore": { "address": "0x18a200A2427e243154EC8217890Bc08062afc5A2", "blockNumber": 10468968 }
   },
   "5": {
-    "AcrossConfigStore": { "address": "0x3215e3C91f87081757d0c41EF0CB77738123Be83", "blockNumber": 6860550 },
```
nicholaspai, line 26: I don't think you needed to redeploy any existing contracts right? Only deploy the new MerkleDistributor?

deployments/deployments.json
```diff
@@ -44,11 +45,13 @@
     "PolygonTokenBridger": { "address": "0x0330E9b4D0325cCfF515E81DFbc7754F2a02ac57", "blockNumber": 28604258 },
     "SpokePool": { "address": "0x69B5c72837769eF1e7C164Abc6515DcFf217F920", "blockNumber": 28604263 }
   },
+  "280": { "SpokePool": { "address": "0x467A26E2eAcf5f81b63016Dbb4F0828809e79fF5", "blockNumber": 1114933 } },
```
nicholaspai, line 48: What's this chainID?

nicholaspai (commented):
Left some questions

deployments/deployments.json
dohaki: Yea I didn't redeploy these, only the `MerkleDistributor`. But it seems that the artifacts within `deployments/goerli` were updated but the `deployments/deployments.json` didn't reflect these updates. When I initially ran `yarn`, it automatically updated this file.

deployments/deployments.json
dohaki: Seems to be zksync-goerli / alpha testnet. See https://github.com/across-protocol/contracts-v2/blob/master/deployments/zksync-goerli/.chainId. This was added automatically due to the same reason as above I think.

amateima (commented):
Looks good overall. I don't have questions other than what Nick pointed out about the modified existing contracts

mrice32 (commented):
This all looks great! Drive-by comment: @nicholaspai I thought we had talked about putting the contract code in the same repo as all of the merkle distribution scripts. Where did we end up landing on that?

Otherwise, this LGTM!

deployments/deployments.json
nicholaspai: What deploy command did you use? Ideally this PR should only add a single line for merkle distributor on goerli

nicholaspai:
> This all looks great! Drive-by comment: @nicholaspai I thought we had talked about putting the merkle distribution contract code in the same repo as all of the merkle distribution scripts. Where did we end up landing on that? Note: this may be overkill at this point since most of the work is done.
> 
> 
> 
> Otherwise, this LGTM!

We ultimately decided to just leave this contract here. The scripts will live separately. This wasn't a very opinionated decision and we can revisit this, but we wanted to avoid having to rewrite the hardhat config and other useful tasks

nicholaspai (approved):

PR #193, opened 2022-10-27
fix: Fix bug in MerkleDistributor that doesn't decrement remainingAmount for merkle windows correctly

nicholaspai:
I discovered this by while working on this [PR](https://github.com/across-protocol/across-token/pull/32/files) for the exact test case I've included. The problem arises when all of the claims _for different window indices_ are for the same token and recipient. If this `if` statement [here](https://github.com/across-protocol/contracts-v2/blob/f42e843e351d70692bd8575a44d0f563d1920d8b/contracts/merkle-distributor/MerkleDistributor.sol#L159) is not triggered, then `batchedAmount` accumulates the amount for each claim. It only gets decremented from a window's `remainingAmount` once the if statement is entered, which is either:

- On the last claim
- When the reward token changes between claims
- When the account changes between claims

If the reward token and account do not change, then its possible that `batchedAmount > window.remainingAmount`.

We should be decrementing remainingAmount for each claim that executes, this seems intuitive. 

The reason why I think we've never run into this is because we've never tried claiming in production across multiple windows. I don't think we've even used multiple windows in the past

### DIFF
--- a/contracts/merkle-distributor/MerkleDistributor.sol
+++ b/contracts/merkle-distributor/MerkleDistributor.sol
@@ -142,7 +142,6 @@ contract MerkleDistributor is Ownable {
         for (uint256 i = 0; i < claimCount; i++) {
             Claim memory _claim = claims[i];
             _verifyAndMarkClaimed(_claim);
-            merkleWindows[_claim.windowIndex].remainingAmount -= _claim.amount;
             batchedAmount += _claim.amount;
 
             // If the next claim is NOT the same account or the same token (or this claim is the last one),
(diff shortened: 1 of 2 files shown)


contracts/merkle-distributor/MerkleDistributor.sol
```diff
@@ -193,6 +193,15 @@ contract MerkleDistributor is Ownable {
         return claimedWord & mask == mask;
     }
 
+    /**
+     * @notice Returns rewardToken set by admin for windowIndex.
+     * @param windowIndex merkle root to check.
+     * @return address Reward token address
+     */
+    function getRewardTokenForWindow(uint256 windowIndex) public view returns (address) {
```
nicholaspai, line 201: Adding for usefulness in https://github.com/across-protocol/across-token/pull/32/files

mrice32 (approved):
LGTM! This is a very nice find! Great simple test, too.

PR #200, opened 2022-11-02
feat: Add claimFor method to AcrossMerkleDistributor

nicholaspai:
Allows a whitelisted account to claim on behalf of a user. Can be used to atomically claim and perform some action for a user, such as claim and stake.


contracts/merkle-distributor/AcrossMerkleDistributor.sol
```diff
@@ -61,4 +73,27 @@ contract AcrossMerkleDistributor is MerkleDistributor {
         require(whitelistedClaimers[msg.sender] || _claim.account == msg.sender, "invalid claimer");
         super.claim(_claim);
     }
+
```
mrice32, line 76: Maybe this should be done in a separate PR, but it's a little confusing that whitelisted claimers can claim directly to their own address _and_ call claimMulti to send tokens to the user.

If we end up going in this direction, do you think we should drop the overridden claimMulti and claim methods, just to reduce surface area and complexity?

mrice32 (approved):
LGTM!

contracts/merkle-distributor/AcrossMerkleDistributor.sol
nicholaspai: I think we still need to require that the claim.account is the caller so users can't reset each other's referral multipliers but i can remove the whitelist requirement. Thats more so to make testing easier but there's an easier way to do this

nicholaspai: Implemented the change where the whitelist only applies for `claimFor`: https://github.com/across-protocol/contracts-v2/pull/200/commits/e21526aab9f40fc1f1b6f03259905b627273dd29

mrice32 (approved):
LGTM!

contracts/merkle-distributor/AcrossMerkleDistributor.sol
```diff
@@ -39,26 +50,47 @@ contract AcrossMerkleDistributor is MerkleDistributor {
...
      * @param claims array of claims to claim.
      */
     function claimMulti(Claim[] memory claims) public override {
-        if (!whitelistedClaimers[msg.sender]) {
-            uint256 claimCount = claims.length;
-            for (uint256 i = 0; i < claimCount; i++) {
-                require(claims[i].account == msg.sender, "invalid claimer");
-            }
+        uint256 claimCount = claims.length;
+        for (uint256 i = 0; i < claimCount; i++) {
```
nicholaspai, line 58: Undoes changes in https://github.com/across-protocol/contracts-v2/pull/195/files that allow a whitelisted claimer to claim on behalf of claimer

contracts/merkle-distributor/AcrossMerkleDistributor.sol
```diff
@@ -39,26 +50,47 @@ contract AcrossMerkleDistributor is MerkleDistributor {
...
+    /**
+     * @notice Executes merkle leaf claim on behaf of user. This can only be called by a trusted
+     *         claimer address. This function is designed to be called atomically with other transactions
+     *         that ultimately return the claimed amount to the rightful recipient. For example,
+     *         AcceleratingDistributor could call this function and then stake atomically on behalf of the user.
+     * @dev    Caller must be in whitelistedClaimers struct set to "true".
+     * @param _claim leaf to claim.
+     */
+
+    function claimFor(Claim memory _claim) public {
```
nicholaspai, line 83: New claimFor method that whitelisted claimer can call

PR #204, opened 2022-11-21
refactor: Define and export MerkleTree instead of importing from @uma/protocol protocol

nicholaspai:
# Motivation
We can remove the `merkle-distributor` package from `@umaprotocol/protocol`

# Implementation
I thought about defining `MerkleTree` in sdk-v2 instead of here but sdk-v2 imports this package for other reasons so this is the more "upstream" repo


nicholaspai:
Could be replaced by https://github.com/UMAprotocol/protocol/pull/4255

PR #208, opened 2022-12-19
feat: Add upgradeability features to SpokePool

nicholaspai:
Adds "SpokeAdapter" contracts that can be reset by the admin of the SpokePool. This adapter contains the logic to send tokens back to HubPool, which can change overtime and for each token. This will reduce the chance of spoke pool upgrades in future.

The diff is large but it _should_ be 90% copy and pasting `SpokePool` logic into the `Spoke_Adapter`


contracts/Ethereum_SpokePool.sol
```diff
@@ -25,7 +29,7 @@ contract Ethereum_SpokePool is SpokePool, Ownable {
      **************************************/
 
     function _bridgeTokensToHubPool(RelayerRefundLeaf memory relayerRefundLeaf) internal override {
-        IERC20(relayerRefundLeaf.l2TokenAddress).transfer(hubPool, relayerRefundLeaf.amountToReturn);
+        IERC20(relayerRefundLeaf.l2TokenAddress).safeTransfer(hubPool, relayerRefundLeaf.amountToReturn);
```
mrice32, line 32: Can we put this in a separate PR since it's pretty distinct from the pause functionality?

mrice32 (commented):
One minor comment!

contracts/SpokePool.sol
```diff
@@ -367,7 +388,7 @@ abstract contract SpokePool is SpokePoolInterface, Testable, Lockable, MultiCall
         uint64 realizedLpFeePct,
         uint64 relayerFeePct,
         uint32 depositId
-    ) public nonReentrant {
+    ) public nonReentrant unpaused {
```
pxrl, line 391: Have we considered separate pause functionality for fills?

I think I can see a use case for pausing deposits but allowing fills to proceed. This would allow relayers to "drain" any remaining deposits that were made before the contract was paused, but which had not been filled in time. This would be especially useful in the case that we want to pause in order to perform administrative activities (as opposed to an emergency pause, in which case we may want to pause everything).

The reason I ask is that we really risk a messy situation if we end up with users funds stuck in a spoke pool with no way of obtaining a refund. I don't see any obvious way of making such users whole by paying them out-of-band and administratively nullifying the deposits, because then we'd need to manually orchestrate relayer refunds somehow.

Scenarios in which we might see "stuck" deposits include:
 - Users who have integrated the contracts directly.
 - Users of aggregators/other third-parties that have holes in their logic.

I don't anticipate this happening directly via the Across FE, since I presume we'd disable deposits in the FE a safe time before pausing the spoke pool.

nicholaspai: hmm this is a good point, I see no risk in having different pause functionality

contracts/interfaces/SpokeAdapterInterface.sol
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity ^0.8.0;
+
+/**
+ * @notice Sends cross chain tokens to from SpokePool to HubPool.
+ */
+
+interface SpokeAdapterInterface {
+    function bridgeTokensToHubPool(uint256 amountToReturn, address l2TokenAddress) external;
```
mrice32, line 9: This is really cool! I think this is a huge reduction in the chain-specific lines in each SpokePool implementation.

Do you think it's possible to remove _all_ chain-specific functionality from the SpokePools?

I was thinking that it would be really nice to be able to deploy the exact same SpokePool implementation to each chain along with an adapter that contains all the specialized functionality for: receiving messages from the hub, sending tokens to the HubPool, and anything else that might be specialized based on the chain.

To do this and handle _all_ necessary functionality, the adapter will have to _receive_ a lot of chain-specific function calls. I think this would ultimately either look like one of the two following architectures:
1. All the common SpokePool methods, like deposits and fills will have implementations in the main SpokePool. All specialized or customized functionality would live in the adapter. Any call that hits the SpokePool would be forwarded (via delegatecall) to the adapter by the fallback function if the SpokePool didn't implement it. Note: this is very very tricky from a permissioning perspective, since any methods on the adapter that are _only_ meant to be called by the SpokePool would need to be protected from unauthorized forwards.

2. A full-on proxy where the entire SpokePool implementation sits behind a Proxy, and that implementation can be swapped out by governance. This is a more common pattern and would probably be safer, but gives governance _full_ control over the implementation rather than just controlling some methods.

Thoughts on this general idea? Note: this gets complicated quickly, so I could also see us sticking with only making the token bridging upgradable.

@nicholaspai @pxrl

mrice32 (commented):
This implementation looks awesome! +1

I added one comment about a more expansive upgradable architecture just to get your thoughts.

contracts/SpokePool.sol
pxrl: Marking as resolved, since this topic continues here: https://github.com/across-protocol/contracts-v2/pull/212

contracts/interfaces/SpokeAdapterInterface.sol
nicholaspai: After our discussion offline I think we should favor going with a full [upgradeable proxy pattern](https://docs.openzeppelin.com/upgrades). The [UUPS](https://docs.openzeppelin.com/upgrades-plugins/1.x/#proxy-patterns) seems like a good one for our use case. We'd get the benefit of a well supported and tested proxy with less code change than this PR which refactors the contracts themselves to only have a tiny surface area be upgradeable.

It seems like the main change we'd have to make is to remove the constructor in favor of an `initializer` function, and generally inherit the `Intializable` contract

mrice32: 100% agreed! I also think this makes the spoke pool upgradability work _much much_ lower lift and less to audit. +1

mrice32 (commented):
Is the plan to update this PR with the proxy construction?

PR #209, opened 2022-12-19
feat: Support SNX bridging to Optimism via non-standard bridge

nicholaspai:
SNX doesn't conform to Optimism bridge standard, handle SNX explicitly:
- L1ToL2Bridge: https://github.com/Synthetixio/synthetix/blob/develop/contracts/SynthetixBridgeToOptimism.sol#L89
- L2ToL1Bridge: https://github.com/Synthetixio/synthetix/blob/develop/contracts/SynthetixBridgeToBase.sol#L50

### DIFF
--- a/contracts/Ovm_SpokePool.sol
+++ b/contracts/Ovm_SpokePool.sol
@@ -8,6 +8,11 @@ import "@eth-optimism/contracts/libraries/bridge/CrossDomainEnabled.sol";
 import "@eth-optimism/contracts/libraries/constants/Lib_PredeployAddresses.sol";
 import "@eth-optimism/contracts/L2/messaging/IL2ERC20Bridge.sol";
 
+// https://github.com/Synthetixio/synthetix/blob/5ca27785fad8237fb0710eac01421cafbbd69647/contracts/SynthetixBridgeToBase.sol#L50
+interface SynthetixBridgeToBase {
+    function withdrawTo(address to, uint256 amount) external;
+}
+
 /**
  * @notice OVM specific SpokePool. Uses OVM cross-domain-enabled logic to implement admin only access to functions. * Optimism and Boba each implement this spoke pool and set their chain specific contract addresses for l2Eth and l2Weth.
  */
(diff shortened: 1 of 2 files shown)


contracts/chain-adapters/Optimism_Adapter.sol
```diff
@@ -37,7 +41,7 @@ contract Optimism_Adapter is CrossDomainEnabled, AdapterInterface {
     address public immutable dai = 0x6B175474E89094C44Da98b954EedeAC495271d0F;
     address public immutable daiOptimismBridge = 0x10E6593CDda8c58a1d0f14C5164B376352a55f2F;
     address public immutable snx = 0xC011a73ee8576Fb46F5E1c5751cA3B9Fe0af2a6F;
-    address public immutable snxOptimismBridge = 0xCd9D4988C0AE61887B075bA77f08cbFAd2b65068;
+    address public immutable snxOptimismBridge = 0x39Ea01a0298C315d149a490E34B59Dbf2EC7e48F;
```
mrice32, line 44: Was our previous code just wrong?

contracts/Ovm_SpokePool.sol
```diff
@@ -141,17 +146,24 @@ contract Ovm_SpokePool is CrossDomainEnabled, SpokePool {
...
             relayerRefundLeaf.l2TokenAddress = l2Eth; // Set the l2TokenAddress to ETH.
         }
-        IL2ERC20Bridge(
-            tokenBridges[relayerRefundLeaf.l2TokenAddress] == address(0)
-                ? Lib_PredeployAddresses.L2_STANDARD_BRIDGE
-                : tokenBridges[relayerRefundLeaf.l2TokenAddress]
-        ).withdrawTo(
-                relayerRefundLeaf.l2TokenAddress, // _l2Token. Address of the L2 token to bridge over.
+        // Handle custom SNX bridge which doesn't conform to the standard bridge interface.
+        if (relayerRefundLeaf.l2TokenAddress == 0x8700dAec35aF8Ff88c16BdF0418774CB3D7599B4)
```
mrice32, line 150: +1 LGTM.

As a part of this spoke pool upgrade (different PR), we could consider trying to make our spoke pool bridging logic more adaptable. Maybe we could use a delegatecalled adapter setup like mainnet, so we can replace the bridging logic easily without having to move the entire contract over.

Thoughts?

mrice32 (approved):
Looks great!

contracts/chain-adapters/Optimism_Adapter.sol
nicholaspai: Yeah I got this new address from [here](https://github.com/ethereum-optimism/ethereum-optimism.github.io/blob/master/data/SNX/data.json) but it does look like 0xCd is the same contract that hasn't been used in 280 days... so I'm a bit confused how I got it

nicholaspai: Here's the original PR https://github.com/across-protocol/contracts-v2/pull/68

contracts/Ovm_SpokePool.sol
nicholaspai: yes definitely! Check out this [draft PR](https://github.com/across-protocol/contracts-v2/pull/208/files#diff-e3ad242c3b0904fa828d6a03174064c9133ca004a4ba53704b544b4ebda57826) where we can insert this custom SNX logic. I'm curious what you think of this draft architecture

nicholaspai: (if it doesn't link to the correct file, go down to the "OVM_SpokeAdapter")

mrice32 (approved):
LGTM!

PR #210, opened 2022-12-19
improve: Import MerkleTree from @uma/common instead of redefining

nicholaspai:


utils/MerkleTree.ts
```diff
@@ -1,145 +1,2 @@
...
-    if (!second) {
-      return first;
-    }
-
-    return keccak256(MerkleTree.sortAndConcat(first, second));
-  }
-
-  getRoot(): Buffer {
-    return this.layers[this.layers.length - 1][0];
-  }
```
mrice32, line 70: I think the implementations differ between this one and the one in @uma/common. I think we made some additions to this implementation to handle empty roots. The empty case is handled by returning the zero root (and this one has an "isEmpty()" method, whereas the one in @uma/common throws I think.

Maybe we should update the one in @uma/common before importing in everywhere in across so that it has these extra features. Thoughts?

mrice32 (commented):
Just one comment!

utils/MerkleTree.ts
nicholaspai: good point https://github.com/UMAprotocol/protocol/pull/4318

nicholaspai: Updated the @uma/common merkleTree implementation which is now exported in @uma/common:2.29

mrice32 (approved):
LGTM!

PR #212, opened 2022-12-20
feat: Enable pausing deposits and fills unilaterally

nicholaspai:
Useful for upgrades


pxrl:
This generally looks good to me.

I can't think of any reason why we'd want to disable _only_ fills on a specific chain, so I'm tempted to propose that we have "pauseDeposits" and "pause" (all). The reason for this is that in the event that we had some kind of emergency, it'd be a little bit painful to do 2x transactions per Spoke Pool.

An alternative could be to just implement a separate method to pause both deposits and fills in a single txn. That's probably safer. wdyt?

mrice32 (approved):
LGTM!

How do you think we should think about merging PRs like this, where we aren't sure if they're going to get into the next audit? Should we avoid merging until we are sure this is the code that will be deployed in the upgrade?

nicholaspai:
> LGTM!
> 
> How do you think we should think about merging PRs like this, where we aren't sure if they're going to get into the next audit? Should we avoid merging until we are sure this is the code that will be deployed in the upgrade?

Hm, why wouldn't this get into next audit?

mrice32:
> > LGTM!
> > How do you think we should think about merging PRs like this, where we aren't sure if they're going to get into the next audit? Should we avoid merging until we are sure this is the code that will be deployed in the upgrade?
> 
> Hm, why wouldn't this get into next audit?

@nicholaspai just prioritization! I think it's still unclear exactly what the prioritization will look like.

mrice32 (commented):
https://github.com/across-protocol/contracts-v2/pull/212#issuecomment-1362222990

nicholaspai:
> > > LGTM!
> > > How do you think we should think about merging PRs like this, where we aren't sure if they're going to get into the next audit? Should we avoid merging until we are sure this is the code that will be deployed in the upgrade?
> > 
> > 
> > Hm, why wouldn't this get into next audit?
> 
> @nicholaspai just prioritization! I think it's still unclear exactly what the prioritization will look like. Do you think we should definitely prioritize this feature over the others we've discussed for the audit?

I think the SpokePool upgrades are the highest priority and this would be a subset of that feature, so I do think its worth adding?

contracts/SpokePool.sol
```diff
@@ -151,10 +157,35 @@ abstract contract SpokePool is SpokePoolInterface, Testable, Lockable, MultiCall
...
+     * @param pause true if the call is meant to pause the system, false if the call is meant to unpause it.
+     */
+    function pauseDeposits(bool pause) public override onlyAdmin nonReentrant {
+        pausedDeposits = pause;
+        emit PausedDeposits(pausedDeposits);
+    }
+
+    function pauseFills(bool pause) public override onlyAdmin nonReentrant {
+        pausedFills = pause;
+        emit PausedFills(pausedDeposits);
```
pxrl, line 186: The event here looks incorrect - should emit the value of pausedFills instead of pausedDeposits.

contracts/SpokePool.sol
```diff
@@ -317,7 +348,7 @@ abstract contract SpokePool is SpokePoolInterface, Testable, Lockable, MultiCall
         uint64 newRelayerFeePct,
         uint32 depositId,
         bytes memory depositorSignature
-    ) public override nonReentrant {
+    ) public override nonReentrant unpausedDeposits {
```
pxrl, line 351: Is it desirable to make speedups subject to deposits not being paused?

As I see it, the point of pausing deposits is to stop additional funds from entering the Spoke Pool. In the case of a speedup, the funds are already there, and are presumably stuck because the deposit was never filled on the destination chain.

Additionally, as far as I can see, the only side-effect of a speedup is that the Spoke Pool emits another event. It's up to the bots as to whether they accept a speedup event, and it will implicitly be conditional on fills being enabled on the destination chain anyway.

Based on the above, I feel like we should not apply this modifier to speedUpDeposit(). wdyt?

test/SpokePool.Admin.ts
```diff
@@ -24,6 +24,19 @@ describe("SpokePool Admin Functions", async function () {
...
+    expect(await spokePool.pausedDeposits()).to.equal(false);
+    expect(await spokePool.pausedFills()).to.equal(false);
+    await expect(spokePool.connect(owner).pauseDeposits(true)).to.emit(spokePool, "PausedDeposits").withArgs(true);
+    await expect(spokePool.connect(owner).pauseFills(true)).to.emit(spokePool, "PausedFills").withArgs(true);
+    expect(await spokePool.pausedDeposits()).to.equal(true);
+    expect(await spokePool.pausedFills()).to.equal(true);
+    await expect(spokePool.connect(owner).pauseDeposits(false)).to.emit(spokePool, "PausedDeposits").withArgs(false);
+    await expect(spokePool.connect(owner).pauseFills(false)).to.emit(spokePool, "PausedFills").withArgs(false);
+    expect(await spokePool.pausedDeposits()).to.equal(false);
+    expect(await spokePool.pausedFills()).to.equal(false);
```
pxrl, line 37: I suspect that interleaving the test of pausedDeposits and pausedFills hid the error in SpokePool.sol @ line 186. It might be worth splitting the test up so that we are testing each modifier in isolation.

contracts/SpokePool.sol
nicholaspai: great catch

contracts/SpokePool.sol
nicholaspai: hmm I think its a bit easier to reason about from the bot's perspective of all deposit related events stop once a contract is paused, even though its philosophically ok to not pause speed ups. I was thinking pausing deposits and fills is more of a bot convenience so the owner of the contracts can effectively downgrade a contract by stopping all of its events

test/SpokePool.Admin.ts
nicholaspai: you are definitely correct, good reminder, I split this up

contracts/SpokePool.sol
```diff
@@ -121,6 +125,8 @@ abstract contract SpokePool is SpokePoolInterface, Testable, Lockable, MultiCall
         address caller
     );
     event EmergencyDeleteRootBundle(uint256 indexed rootBundleId);
+    event PausedDeposits(bool indexed isPaused);
+    event PausedFills(bool indexed isPaused);
```
pxrl, line 129: Minor: Do these need to be indexed? I imagine that there will be so few of them that there's not much advantage to filtering on the "paused" state.

contracts/SpokePool.sol
pxrl: Fair enough - I can definitely see the argument for keeping it simple.

There is a theoretical possibility for the edge case I described, and an annoying scenario would be that we have a number of underpriced deposits that are suddenly marooned in the Spoke Pool. In this case we'd need to either unpause to allow speedups to be submitted, or just take the hit and fill these relays at a discount. In theory the relayerFeePct could be very very low.

It's a theoretical possibility, and probably unlikely to be a reality. As long as we're OK with the possibility of it arising then I have no additional input on this one :)

contracts/SpokePool.sol
mrice32: +1

mrice32 (approved):
LGTM once @pxrl's comments are addressed!

mrice32:
> > > > LGTM!
> > > > How do you think we should think about merging PRs like this, where we aren't sure if they're going to get into the next audit? Should we avoid merging until we are sure this is the code that will be deployed in the upgrade?
> > > 
> > > 
> > > Hm, why wouldn't this get into next audit?
> > 
> > 
> > @nicholaspai just prioritization! I think it's still unclear exactly what the prioritization will look like. Do you think we should definitely prioritize this feature over the others we've discussed for the audit?
> 
> I think the SpokePool upgrades are the highest priority and this would be a subset of that feature, so I do think its worth adding?

Tbh, I don't think it's a big deal to go ahead and merge even if we later decide to drop this feature, as it would just be a simple revert. We have an upcoming meeting to figure this out, but regardless, LGTM, as that shouldn't hold this PR up IMO.

contracts/SpokePool.sol
nicholaspai: hmmm ok that's a valid concern, i don't love the idea that someone will try to use the speedup signatures to replay attack but I guess its not really an attack since they could have chosen to call `speedUp` on either the old or the new spoke pool

pxrl (approved):
LGTM!

PR #213, opened 2022-12-21
feat: Add Arbitrum_RescueAdapter helper scripts

nicholaspai:
To aid with future usage of this adapter


contracts/chain-adapters/Arbitrum_Adapter.sol
```diff
@@ -17,6 +17,17 @@ interface ArbitrumL1InboxLike {
         uint256 gasPriceBid,
         bytes calldata data
     ) external payable returns (uint256);
+
+    function createRetryableTicketNoRefundAliasRewrite(
```
mrice32, line 21: nit: should we use unsafeCreateRetryableTicket instead (I think the signature is the same)?

According to the [current Inbox implementation on mainnet](https://etherscan.io/address/0x5aed5f8a1e3607476f1f81c3d8fe126deb0afe94#code), createRetryableTicketNoRefundAliasRewrite is deprecated in favor of unsafeCreateRetryableTicket.

mrice32 (commented):
Looks good! Only one minor nit (note: since this is already deployed, we can go ahead and use the existing implementation for this operation, but I think we should update the contract in the repo with the non-deprecated method).

contracts/chain-adapters/Arbitrum_Adapter.sol
nicholaspai: This is a great call. `createRetryableTicketNoRefundAliasRewrite` calls `unsafeCreateRetryableTicket` so I agree we don't need to change deployed version

nicholaspai:
This adapter worked!
- L2 txn: https://arbiscan.io/tx/0xe01150352f64d34511abcafe60e834de26f5194907882670b32135ee301866cd
- L1 txn: https://etherscan.io/tx/0x33c315881a73b63dc62a175bc86830e48b461797c4c87f8590376572e98e43cd

pxrl (commented):
No comments from me :+1:.

mrice32 (approved):
LGTM!

PR #214, opened 2022-12-21
fix: Refactor Arbitrum_Adapter

nicholaspai:
We realized that we haven't updated the arbitrum adapter to use `outboundTransferCustomRefund` so this cleans up the adapter before updating it


contracts/chain-adapters/Arbitrum_Adapter.sol
```diff
@@ -64,10 +54,6 @@ contract Arbitrum_Adapter is AdapterInterface {
     // L2 Gas price bid for immediate L2 execution attempt (queryable via standard eth*gasPrice RPC)
     uint256 public immutable l2GasPrice = 5e9; // 5 gWei
 
-    // Gas limit for immediate L2 execution attempt (can be estimated via NodeInterface.estimateRetryableTicket).
-    // NodeInterface precompile interface exists at L2 address 0x00000000000000000000000000000000000000C8
-    uint32 public immutable l2GasLimit = 2_000_000;
```
nicholaspai, line 69: Gas limit for sending tokens should be a lot lower than for sending messages

mrice32: nit: any reason not to make these both constants rather than hardcoding them?

contracts/chain-adapters/Arbitrum_Adapter.sol
```diff
@@ -137,33 +123,17 @@ contract Arbitrum_Adapter is AdapterInterface {
...
 
-        // Note: Legacy routers don't have the outboundTransferCustomRefund method, so default to using
-        // outboundTransfer(). Legacy routers are used for the following tokens that are currently enabled:
-        // - DAI
-        if (l1Token == 0x6B175474E89094C44Da98b954EedeAC495271d0F) {
-            // Note: outboundTransfer() will ultimately create a retryable ticket and set this contract's address as the
-            // refund address. This means that the excess ETH to pay for the L2 transaction will be sent to the aliased
-            // contract address on L2, which we'd have to retrieve via a custom adapter
-            // (i.e. the Arbitrum_RescueAdapter).
-            l1ERC20GatewayRouter.outboundTransfer{ value: requiredL1CallValue }(
```
mrice32, line 148: Just double checking -- we have a confirmed DAI transfer where we see the refund working correctly (i.e. going to the hub pool alias address rather than the spoke pool)?

contracts/chain-adapters/Arbitrum_Adapter.sol
```diff
@@ -137,33 +123,17 @@ contract Arbitrum_Adapter is AdapterInterface {
...
-                l1Token,
-                l2RefundL2Address,
-                to,
-                amount,
-                l2GasLimit,
-                l2GasPrice,
-                data
-            );
-        }
+        // Excess funds will be sent to this contract's aliased arbitrum address.
```
mrice32, line 126: Is this logic identical to the logic that was here in the initial version of this adapter? I want to make sure we're really tight on what's being changed in these adapters since this code would be running in the HubPool context.

mrice32 (commented):
Looks great! Each refund is pretty small, so I think we should wait to hear back from arbitrum so we understand what's going on before making any changes in production.

contracts/chain-adapters/Arbitrum_Adapter.sol
nicholaspai: yes i confirmed it by checking the hub pool alias ETH balance the blockbefore and after the DAI transfer

contracts/chain-adapters/Arbitrum_Adapter.sol
nicholaspai: yep exactly the same

nicholaspai: ive checked against on-chain verified contract code [here](https://etherscan.io/address/0x937958992799bF4A9a656E6596FD10d7Da5c2216#code)

contracts/chain-adapters/Arbitrum_Adapter.sol
nicholaspai: you mean instead of `uint256 public immutable` do `uint256 public constant`? Is that a gas savings?

mrice32 (approved):
LGTM! As with the prev comment: https://github.com/across-protocol/contracts-v2/pull/214#pullrequestreview-1226755703, I think we should wait to hear back before moving forward with any modifications.

contracts/chain-adapters/Arbitrum_Adapter.sol
mrice32: Sorry, yeah, that too (constant is better than immutable!), but I just meant giving them names, which you did!

mrice32 (approved):
LGTM!

PR #215, opened 2022-12-22
feat: new release

nicholaspai:
Main new features: new license, bump dependencies, fix usdt in spoke pool and import merkletree


pxrl (approved):
The motivation for the bump is understood to be that some downstream Across repos want access to new goodies that were added after 2.01.

PR #216, opened 2022-12-22
feat: Index depositId in FilledRelay events

nicholaspai:
Will make certain dataworker functions more efficient. See `queryHistoricalMatchingFills` in https://github.com/across-protocol/relayer-v2/pull/382/files


pxrl (approved):
+1 major QoL upgrade!

PR #217, opened 2023-01-04
fix(Ethereum_SpokePool): Correct admin/owner comment

pxrl:


contracts/Ethereum_SpokePool.sol
```diff
@@ -32,7 +32,7 @@ contract Ethereum_SpokePool is SpokePool, Ownable {
         IERC20(relayerRefundLeaf.l2TokenAddress).safeTransfer(hubPool, relayerRefundLeaf.amountToReturn);
     }
 
-    // Admin is simply owner which should be same account that owns the HubPool deployed on this network. A core
-    // assumption of this contract system is that the HubPool is deployed on Ethereum.
+    // The SpokePool deployed to the same network as the HubPool must be owned by the HubPool.
+    // A core assumption of this contract system is that the HubPool is deployed on Ethereum.
```
nicholaspai, line 36: I think this is a good place to mention that `crossDomainAdmin` is also unused for this spoke pool and arbitrarily set to the deployer. There is no harm leaving it as the deployer

nicholaspai (approved):
Great suggestion! I have one additional comment I'd like to see

contracts/Ethereum_SpokePool.sol
pxrl: > I think this is a good place to mention that `crossDomainAdmin` is also unused for this spoke pool and arbitrarily set to the deployer. There is no harm leaving it as the deployer

Any thoughts on [this](https://github.com/across-protocol/contracts-v2/pull/217/commits/3be288a773d1106376d9343669423b1b6c23a971) instead?

nicholaspai: yea thats more consistent but still add a comment why we're setting `crossDomainAdmin` to that

nicholaspai (approved):

contracts/Ethereum_SpokePool.sol
```diff
@@ -14,6 +14,7 @@ contract Ethereum_SpokePool is SpokePool, Ownable {
 
     /**
      * @notice Construct the Ethereum SpokePool.
+     * @dev crossDomainAdmin is unused on Ethereum.
```
nicholaspai, line 17:
```suggestion
     * @dev crossDomainAdmin is unused on this contract.
```

nicholaspai (approved):

PR #218, opened 2023-01-04
!feat: Make SpokePools upgradeable 

nicholaspai:
This PR makes SpokePools compatible with OpenZeppelin's upgradeable proxy pattern. Some important changes to support upgradeability:

- `SpokePool's constructor` adds the `_disableInitializers` call to [prevent it from being left uninitialized](https://docs.openzeppelin.com/upgrades-plugins/1.x/writing-upgradeable#initializing_the_implementation_contract).
- All non-constant state variables that have initial values are [instead set in the `initialize` function](https://docs.openzeppelin.com/upgrades-plugins/1.x/writing-upgradeable#initializing_the_implementation_contract).
- Contracts that the `SpokePool` extends such as `Lockable` and `Testable` that have constructors replaces those with upgradeable versions that replace constructors with initializers.
- All unit tests involving the spoke pool are changed to interface with an upgradeable version of the spoke pool


contracts/upgradeable/LockableUpgradeable.sol
```diff
@@ -0,0 +1,80 @@
+// SPDX-License-Identifier: AGPL-3.0-only
```
mrice32, line 1: As discussed IRL, maybe we should just use the OZ lockable contract? Do you see any problem with that?

contracts/upgradeable/MultiCallerUpgradeable.sol
```diff
@@ -0,0 +1,35 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity ^0.8.0;
+
+/**
+ * @title MockSpokePool
+ * @notice Logic is 100% copied from "@uma/core/contracts/common/implementation/MultiCaller.sol" but one
```
mrice32, line 6: We should remember to exclude this contract from the audit scope considering the contract's logic has already been audited.

mrice32 (approved):
LGTM!

deployments/deployments.json
```diff
@@ -2,7 +2,8 @@
   "1": {
     "AcrossConfigStore": { "address": "0x3B03509645713718B78951126E0A6de6f10043f5", "blockNumber": 14717196 },
     "AcrossMerkleDistributor": { "address": "0xE50b2cEAC4f60E840Ae513924033E753e2366487", "blockNumber": 15976846 },
-    "Arbitrum_Adapter": { "address": "0x937958992799bF4A9a656E6596FD10d7Da5c2216", "blockNumber": 14704380 },
+    "Arbitrum_Adapter": { "address": "0x29528780E29abb8Af95a5e5a125b94766987543F", "blockNumber": 16243583 },
+    "Arbitrum_RescueAdapter": { "address": "0xC6fA0a4EBd802c01157d6E7fB1bbd2ae196ae375", "blockNumber": 16233939 },
```
pxrl, line 6: Seems unrelated - maybe split it out?

test/chain-specific-spokepools/Arbitrum_SpokePool.ts
```diff
@@ -23,14 +23,28 @@ describe("Arbitrum Spoke Pool", function () {
...
 
+  it("Only cross domain owner upgrade logic contract", async function () {
+    // TODO: Could also use upgrades.prepareUpgrade but I'm unclear of differences
+    const implementation = await hre.upgrades.deployImplementation(
+      await getContractFactory("Arbitrum_SpokePool", owner),
+      { kind: "uups" }
+    );
+
+    // upgradeTo fails unless called by cross domain admin
+    await expect(arbitrumSpokePool.upgradeTo(implementation)).to.be.reverted;
```
pxrl, line 44: Is it worthwhile specifying the expected revert message for these? Then it's clear exactly where it's being caught.

test/gas-analytics/SpokePool.RelayerRefundLeafExecution.ts
```diff
@@ -133,7 +133,7 @@ describe("Gas Analytics: SpokePool Relayer Refund Leaf Execution", function () {
       );
 
       // Store new tree.
-      await spokePool.connect(dataWorker).relayRootBundle(
+      await spokePool.connect(owner).relayRootBundle(
```
pxrl, line 136: Are these related to SpokePool upgradability?

contracts/SpokePool.sol
```diff
@@ -129,23 +135,37 @@ abstract contract SpokePool is SpokePoolInterface, Testable, Lockable, MultiCall
     event PausedDeposits(bool isPaused);
     event PausedFills(bool isPaused);
 
+    /**
+     * Do not leave an implementation contract uninitialized. An uninitialized implementation contract can be
+     * taken over by an attacker, which may impact the proxy. To prevent the implementation contract from being
+     * used, you should invoke the _disableInitializers function in the constructor to automatically lock it when
+     * it is deployed: */
```
pxrl, line 142:
```suggestion
     * it is deployed:
     */
```

contracts/SpokePool.sol
```diff
@@ -38,11 +44,11 @@ abstract contract SpokePool is SpokePoolInterface, Testable, Lockable, MultiCall
 
     // Address of wrappedNativeToken contract for this network. If an origin token matches this, then the caller can
     // optionally instruct this contract to wrap native tokens when depositing (ie ETH->WETH or MATIC->WMATIC).
-    WETH9 public immutable wrappedNativeToken;
+    WETH9 public wrappedNativeToken;
 
     // Any deposit quote times greater than or less than this value to the current contract time is blocked. Forces
     // caller to use an approximately "current" realized fee. Defaults to 10 minutes.
-    uint32 public depositQuoteTimeBuffer = 600;
+    uint32 public depositQuoteTimeBuffer;
```
pxrl, line 51: It [won't be possible to add any more state variables](https://docs.openzeppelin.com/upgrades-plugins/1.x/writing-upgradeable#modifying-your-contracts) to the base SpokePool after deployment. Have we ruled out adding some placeholders, just in case we find a use for something else in future?

deployments/deployments.json
nicholaspai: This was accidentally removed so i'm just adding it back here

test/gas-analytics/SpokePool.RelayerRefundLeafExecution.ts
nicholaspai: they're not, just different test fixes

contracts/upgradeable/LockableUpgradeable.sol
nicholaspai: Nope, I looked into Lockable. The main feature we add is the `_startReentrantGuardDisabled` internal function but that's not used by Across contracts so I think we're safe to use OZ lockable

contracts/SpokePool.sol
nicholaspai: Great point https://github.com/across-protocol/contracts-v2/pull/218/commits/6a9f5093193229800f4e46aaa52dd1d54e3fe667

What do you and @mrice32 think about adding 50 storage slots for future upgrades?

mrice32: Great call! Any reason not to make it way way bigger than we'd ever need? Like 1000? Just to be super safe

contracts/upgradeable/LockableUpgradeable.sol
mrice32: Awesome!

mrice32 (approved):
LGTM!

contracts/SpokePool.sol
nicholaspai: Here's a thought experiment. Imagine we wanted to support bridging with arbitrary data so that someone could bridge-and-swap. Say we had to upgrade the `deposit()` function (or add a new depositWithData function). Would this be possible with upgradeable proxies? How much of the extra storage slots would we need?

mrice32: It depends on the implementation. If the data is just hashed into the deposit data and added to an event, no storage changes would be necessary.

If we needed to store extra info about each deposit, deposits would need extra storage, but they aren't stored continuously, so that can be done without pre-reserving slots.

In this implementation, what extra contiguous storage would you imagine needing?

nicholaspai: I'm not sure, this was just a functionality that I discussed with Ryan about adding and wanted to make sure we're prepared for it. Basically bridge-and-swap would be a good UX addition

nicholaspai: What if we needed to add a new Base contract function, does that take up slots?

mrice32: Nope, functions don't take up slots or anything. They are just more bytecode in the implementation contract.

mrice32 (commented):
https://github.com/across-protocol/contracts-v2/pull/218#discussion_r1093504016

deployments/deployments.json
pxrl: It's preferable from a commit hygiene standpoint to split it out and submit it separately - at least so it's not ++confusing when someone looks back at the commit history and sees it modified as part of something totally unrelated. Just my 2c.

pxrl (approved):
Nice!

pxrl:
@nicholaspai One final thought: the OVM_SpokePool is also a base contract. Not sure whether we want to reserve some slots there as well?

nicholaspai:
> @nicholaspai One final thought: the OVM_SpokePool is also a base contract. Not sure whether we want to reserve some slots there as well?

Great call, i've added 50 slots for OVM_SpokePool and updated SpokePool's __gap to 1000

cc @mrice32

mrice32 (commented):
> > @nicholaspai One final thought: the OVM_SpokePool is also a base contract. Not sure whether we want to reserve some slots there as well?
> 
> Great call, i've added 100 slots for OVM_SpokePool and updated SpokePool's __gap to 1000
> 
> cc @mrice32

@nicholaspai why not just use a 1000 gap everywhere to give us tons of room and standardize? We have 2^256 slots to work with here, so there's no running out :).

mrice32 (approved):
LGTM!

nicholaspai:
@dohaki @mrice32 @pxrl I had to update this to take into account the new EIP712 contract that the SpokePool extends. This required changing that contract to be upgradeable with an initializer function instead of a constructor see this commit https://github.com/across-protocol/contracts-v2/pull/218/commits/22f5fbed937fa1be2d7e5bcd8bd8dbb834172cfa#diff-8c305f0568345cd237252c531582ce424adc9c7c9151cfefe304fd389b874c59 !

contracts/upgradeable/EIP712CrossChainUpgradeable.sol
```diff
@@ -72,6 +75,13 @@ abstract contract EIP712CrossChain {
...
-        return ECDSA.toTypedDataHash(_domainSeparatorV4(originChainId), structHash);
+        return ECDSAUpgradeable.toTypedDataHash(_domainSeparatorV4(originChainId), structHash);
     }
+
+    /**
+     * @dev This empty reserved space is put in place to allow future versions to add new
+     * variables without shifting down storage in the inheritance chain.
+     * See https://docs.openzeppelin.com/contracts/4.x/upgradeable#storage_gaps
+     */
+    uint256[1000] private __gap;
```
mrice32, line 86: Why is the gap at the bottom of the file below the functions? Shouldn't it be at the end of the list of variables, like in the other contracts?

contracts/upgradeable/MultiCallerUpgradeable.sol
```diff
@@ -0,0 +1,42 @@
...
+            results[i] = result;
+        }
+    }
+
+    /**
+     * @dev This empty reserved space is put in place to allow future versions to add new
+     * variables without shifting down storage in the inheritance chain.
+     * See https://docs.openzeppelin.com/contracts/4.x/upgradeable#storage_gaps
+     */
+    uint256[1000] private __gap;
```
mrice32, line 41: Same as above.

contracts/upgradeable/TestableUpgradeable.sol
```diff
@@ -0,0 +1,61 @@
...
+            return block.timestamp; // solhint-disable-line not-rely-on-time
+        }
+    }
+
+    /**
+     * @dev This empty reserved space is put in place to allow future versions to add new
+     * variables without shifting down storage in the inheritance chain.
+     * See https://docs.openzeppelin.com/contracts/4.x/upgradeable#storage_gaps
+     */
+    uint256[1000] private __gap;
```
mrice32, line 60: Same as above.

mrice32 (approved):
LGTM!

contracts/upgradeable/EIP712CrossChainUpgradeable.sol
nicholaspai: I'm not sure this is how OZ adds their gaps so I was following along https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/master/contracts/proxy/utils/UUPSUpgradeable.sol#L111

dohaki (approved):
Awesome LGTM 👍 I mainly looked at the changes related to EIP712

contracts/upgradeable/EIP712CrossChainUpgradeable.sol
mrice32: Makes sense. Either way is totally fine! However, I think this PR has it done both ways. Ovm_SpokePool and SpokePool seem to have it at the top. As long as it's consistent, LGTM!

mrice32 (commented):
https://github.com/across-protocol/contracts-v2/pull/218#discussion_r1097529817

contracts/upgradeable/EIP712CrossChainUpgradeable.sol
```diff
@@ -1,27 +1,35 @@
...
-import "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
+import "@openzeppelin/contracts-upgradeable/utils/cryptography/ECDSAUpgradeable.sol";
+import "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
 
 /**
  * @dev https://eips.ethereum.org/EIPS/eip-712[EIP 712] is a standard for hashing and signing of typed structured data.
  *
  * This contract is based on OpenZeppelin's implementation:
- * https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v4.8.0/contracts/utils/cryptography/EIP712.sol
+ * https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/master/contracts/utils/cryptography/EIP712Upgradeable.sol
```
pxrl, line 11:
```suggestion
 * https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/release-v4.8/contracts/utils/cryptography/EIP712Upgradeable.sol
```

Referring to master could be misleading in future, so it probably makes sense to refer to a specific version. I expect that we're just migrating from openzeppelin-contracts v4.8.0 to openzeppelin-contracts-upgradeable v4.8.0 - is that correct?

contracts/upgradeable/EIP712CrossChainUpgradeable.sol
```diff
@@ -1,27 +1,35 @@
...
  */
-abstract contract EIP712CrossChain {
+abstract contract EIP712CrossChainUpgradeable is Initializable {
     /* solhint-disable var-name-mixedcase */
-    bytes32 private immutable _HASHED_NAME;
-    bytes32 private immutable _HASHED_VERSION;
-    bytes32 private immutable _TYPE_HASH;
+    bytes32 private _HASHED_NAME;
+    bytes32 private _HASHED_VERSION;
+    bytes32 private constant _TYPE_HASH = keccak256("EIP712Domain(string name,string version,uint256 chainId)");
```
pxrl, line 22: Should this be assigned in initialize()?

https://docs.openzeppelin.com/upgrades-plugins/1.x/writing-upgradeable#avoid-initial-values-in-field-declarations

pxrl (approved):
Would be nice to be more specific in the comment referring to the OZ EIP712Upgradable contract, but otherwise :+1:

PR #219, opened 2023-01-05
feat(SpokePool): Initialize depositId

nicholaspai:
Constructor should be able to initialize deposit ID to support upgrades

The alternative implementation with less code diff would be to add an admin function `incrementDepositCount(uint32 count)` where the admin could increase the deposit count. I don't see this being useful outside of upgrades so I decided not to do it. 

Thoughts?

### DIFF
--- a/contracts/Arbitrum_SpokePool.sol
+++ b/contracts/Arbitrum_SpokePool.sol
@@ -29,6 +29,8 @@ contract Arbitrum_SpokePool is SpokePool {
 
     /**
      * @notice Construct the AVM SpokePool.
+     * @param _initialDepositId Starting deposit ID. Set to 0 unless this is a re-deployment in order to mitigate
+     * relay hash collisions.
      * @param _l2GatewayRouter Address of L2 token gateway. Can be reset by admin.
      * @param _crossDomainAdmin Cross domain admin to set. Can be changed by admin.
      * @param _hubPool Hub pool address to set. Can be changed by admin.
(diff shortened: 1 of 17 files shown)


mrice32 (approved):

mrice32 (approved):
LGTM!

Interesting comment on the admin call.
Advantages:
- More flexibility.

Disadvantages:
- More complexity.
- Possible to "skip" deposit ids in a single spoke pool. In the future, we could use skips as a sign that we missed a deposit/fill in the bots, but if these can be incremented after launch, that strategy wouldn't really work as nicely.

I can't think of a reason we'd want to do this. Can you @pxrl?

pxrl:
An alternative solution could be to have the contract pause deposits automatically in the constructor, which would then allow the depositId to be manually adjusted via a separate txn, before unpausing again. This might make it marginally easier to re-deploy successive contracts because it'd be one less thing to modify each time.

nicholaspai:
> An alternative solution could be to have the contract pause deposits automatically in the constructor, which would then allow the depositId to be manually adjusted via a separate txn, before unpausing again. This might make it marginally easier to re-deploy successive contracts because it'd be one less thing to modify each time.
> 
> FWIW as background, I like the concept of the contract being deployed in a "guarded" state to allow time for any manual post-deployment testing and config, before flicking a switch to make it accessible to all others.

This is the case today as `enabledDeposits` for all indices will be `false`

PR #220, opened 2023-01-05
improve: Update speed up signature 

nicholaspai:
Upon redeployment of SpokePool may be helpful to avoid signature collisions


mrice32 (approved):

mrice32 (approved):
I think we'll be protected from collisions by the deposit id increment anyway, but I see no harm in being extra careful and modifying the string regardless.

PR #221, opened 2023-01-18
feat: support erc1271 for verifying relay fee update sigs

dohaki:
Closes ACX-584

As noted by @nicholaspai: _This does open the possibility that someone could forge an EIP1271 signature and set the relayer fee to 50%, and then relay the honest depositor's deposit, stealing ~50% of the total deposit. We should comment on this and debate whether this risk is worth it_

### DIFF
--- a/contracts/SpokePool.sol
+++ b/contracts/SpokePool.sol
@@ -10,6 +10,7 @@ import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
 import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import "@openzeppelin/contracts/utils/Address.sol";
 
+import "@openzeppelin/contracts/utils/cryptography/SignatureChecker.sol";
 import "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
 import "@uma/core/contracts/common/implementation/Testable.sol";
 import "@uma/core/contracts/common/implementation/MultiCaller.sol";
@@ -722,10 +723,8 @@ abstract contract SpokePool is SpokePoolInterface, Testable, Lockable, MultiCall
         bytes32 ethSignedMessageHash,
         bytes memory depositorSignature
     ) internal view virtual {
-        // Note: We purposefully do not support EIP-1271 signatures (meaning that multisigs and smart contract wallets
-        // like Argent are not supported) because of the possibility that a multisig that signed a message on the origin
-        // chain does not have a parallel on this destination chain.
-        require(depositor == ECDSA.recover(ethSignedMessageHash, depositorSignature), "invalid signature");
+        bool isValid = SignatureChecker.isValidSignatureNow(depositor, ethSignedMessageHash, depositorSignature);
+        require(isValid, "invalid signature");
     }
 
     function _computeAmountPreFees(uint256 amount, uint64 feesPct) private pure returns (uint256) {
--- a/contracts/SpokePool.sol
+++ b/contracts/SpokePool.sol
@@ -723,6 +723,13 @@ abstract contract SpokePool is SpokePoolInterface, Testable, Lockable, MultiCall
         bytes32 ethSignedMessageHash,
         bytes memory depositorSignature
     ) internal view virtual {
+        // Note:
+        // - We don't need to worry about reentrancy from a contract deployed at the depositor address since the method
+        //   `SignatureChecker.isValidSignatureNow` is a view method. Re-entrancy can happen, but it cannot affect state.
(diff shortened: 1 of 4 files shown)


test/SpokePool.Relay.ts
```diff
@@ -175,199 +175,220 @@ describe("SpokePool Relayer Logic", async function () {
...
-    const { signature } = await modifyRelayHelper(
-      consts.modifiedRelayerFeePct,
-      consts.firstDepositId.toString(),
-      spokePoolChainId.toString(),
-      depositor
-    );
+    await testUpdatedFeeSignaling(depositor.address);
+  });
+  it("EIP1271 - Can signal to relayer to use updated fee", async function () {
+    await testUpdatedFeeSignaling(erc1271.address);
```
nicholaspai, line 181: +1 like this test layout

test/SpokePool.Relay.ts
```diff
@@ -175,199 +175,220 @@ describe("SpokePool Relayer Logic", async function () {
...
+    await testFillRelayWithUpdatedFee(depositor.address);
+  });
+  it("EIP1271 - Can fill relay with updated fee by including proof of depositor's agreement", async function () {
+    await testFillRelayWithUpdatedFee(erc1271.address);
+  });
+  it("Updating relayer fee signature verification failure cases", async function () {
+    await testUpdatedFeeSignatureFailCases(depositor.address);
+  });
+  it("EIP1271 - Updating relayer fee signature verification failure cases", async function () {
+    await testUpdatedFeeSignatureFailCases(erc1271.address);
```
nicholaspai, line 193: Can you add a test case for the `erc1271` contract in particular that tries to re-enter the SpokePool via a malicious `isValidSignature` function? This way we can confirm that the SpokePool is protected against reentrancy via this avenue

nicholaspai: Hm actually maybe we should add a comment that reentrancy might be impossible:

"https://github.com/across-protocol/contracts-v2/commit/6e69fdcef92eb9867409e3fd58387aaef73b81e1#diff-c828c513da79ba8aaef90514612391ab755989974a14d8e480ae02ebd9cfb4beL683"

contracts/SpokePool.sol
```diff
@@ -722,10 +723,8 @@ abstract contract SpokePool is SpokePoolInterface, Testable, Lockable, MultiCall
         bytes32 ethSignedMessageHash,
         bytes memory depositorSignature
     ) internal view virtual {
-        // Note: We purposefully do not support EIP-1271 signatures (meaning that multisigs and smart contract wallets
-        // like Argent are not supported) because of the possibility that a multisig that signed a message on the origin
-        // chain does not have a parallel on this destination chain.
-        require(depositor == ECDSA.recover(ethSignedMessageHash, depositorSignature), "invalid signature");
+        bool isValid = SignatureChecker.isValidSignatureNow(depositor, ethSignedMessageHash, depositorSignature);
```
nicholaspai, line 726: Maybe we should also add a comment about how `isValidSignatureNow` can change its behavior from block N to block M: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/cryptography/SignatureChecker.sol#L21

As long as the signature is valid at the time of calling `fillRelayWithUpdatedFee` then the signature is treated as valid.

mrice32: IMO, we should note that:
- EIP-1271 signatures are supported. This means that a signature valid now, may not be valid later.
- For an EIP-1271 signature to work, the depositor contract address must map to a deployed contract on the _destination_ chain that can validate the signature.
- Regular signatures from an EOA are also supported.

test/SpokePool.Relay.ts
mrice32: Since isValidSignature is a view method, re-entrancy can happen, but it cannot affect state.

mrice32 (approved):
LGTM!

amateima (approved):

contracts/SpokePool.sol
dohaki: Thanks for the input! I added them [here](https://github.com/across-protocol/contracts-v2/pull/221/commits/15cdaf60828edd6e2c56f3ef8c2fcdae31cfedd3)

test/SpokePool.Relay.ts
dohaki: Thanks for the input! I added them [here](https://github.com/across-protocol/contracts-v2/pull/221/commits/15cdaf60828edd6e2c56f3ef8c2fcdae31cfedd3)

PR #222, opened 2023-01-23
feat: eip-712 support

dohaki:
Closes ACX-590

This allows the usage of EIP-712 structured and typed data signatures for updating relayer fees. The implementation of `EIP712CrossChain` is based on OZ's implementation of the EIP-712 specification but is less strict to allow the verification of signatures cross-chain.

### DIFF
--- a/contracts/EIP712CrossChain.sol
+++ b/contracts/EIP712CrossChain.sol
@@ -19,8 +19,6 @@ abstract contract EIP712CrossChain {
     bytes32 private immutable _HASHED_NAME;
     bytes32 private immutable _HASHED_VERSION;
     bytes32 private immutable _TYPE_HASH;
-    // NOTE: We use a hardcoded address to allow cross-chain verification
-    address public immutable _VERIFYING_CONTRACT = 0xCcCCccccCCCCcCCCCCCcCcCccCcCCCcCcccccccC;
 
     /* solhint-enable var-name-mixedcase */
 
@@ -39,9 +37,7 @@ abstract contract EIP712CrossChain {
     constructor(string memory name, string memory version) {
         bytes32 hashedName = keccak256(bytes(name));
         bytes32 hashedVersion = keccak256(bytes(version));
-        bytes32 typeHash = keccak256(
-            "EIP712Domain(string name,string version,uint256 chainId,address verifyingContract)"
-        );
+        bytes32 typeHash = keccak256("EIP712Domain(string name,string version,uint256 chainId)");
         _HASHED_NAME = hashedName;
         _HASHED_VERSION = hashedVersion;
         _TYPE_HASH = typeHash;
@@ -59,8 +55,8 @@ abstract contract EIP712CrossChain {
         bytes32 nameHash,
         bytes32 versionHash,
         uint256 originChainId
-    ) private view returns (bytes32) {
-        return keccak256(abi.encode(typeHash, nameHash, versionHash, originChainId, _VERIFYING_CONTRACT));
+    ) private pure returns (bytes32) {
+        return keccak256(abi.encode(typeHash, nameHash, versionHash, originChainId));
     }
 
     /**
--- a/contracts/EIP712CrossChain.sol
+++ b/contracts/EIP712CrossChain.sol
@@ -45,6 +45,8 @@ abstract contract EIP712CrossChain {
 
     /**
      * @dev Returns the domain separator depending on the `originChainId`.
+     * @param originChainId Chain id of network where message originates from.
+     * @return bytes32 EIP-712-compliant domain separator.
      */
     function _domainSeparatorV4(uint256 originChainId) internal view returns (bytes32) {
         return keccak256(abi.encode(_TYPE_HASH, _HASHED_NAME, _HASHED_VERSION, originChainId));
@@ -65,6 +67,9 @@ abstract contract EIP712CrossChain {
      * bytes32 digest = _hashTypedDataV4(structHash, originChainId);
      * address signer = ECDSA.recover(digest, signature);
      * ```
+     * @param structHash Hashed struct as defined in https://eips.ethereum.org/EIPS/eip-712#definition-of-hashstruct.
+     * @param originChainId Chain id of network where message originates from.
+     * @return bytes32 Hash digest that is recoverable via `EDCSA.recover`.
      */
     function _hashTypedDataV4(bytes32 structHash, uint256 originChainId) internal view virtual returns (bytes32) {
         return ECDSA.toTypedDataHash(_domainSeparatorV4(originChainId), structHash);
--- a/contracts/EIP712CrossChain.sol
+++ b/contracts/EIP712CrossChain.sol
@@ -47,16 +47,7 @@ abstract contract EIP712CrossChain {
      * @dev Returns the domain separator depending on the `originChainId`.
      */
     function _domainSeparatorV4(uint256 originChainId) internal view returns (bytes32) {
-        return _buildDomainSeparator(_TYPE_HASH, _HASHED_NAME, _HASHED_VERSION, originChainId);
-    }
-
-    function _buildDomainSeparator(
-        bytes32 typeHash,
-        bytes32 nameHash,
-        bytes32 versionHash,
-        uint256 originChainId
-    ) private pure returns (bytes32) {
-        return keccak256(abi.encode(typeHash, nameHash, versionHash, originChainId));
+        return keccak256(abi.encode(_TYPE_HASH, _HASHED_NAME, _HASHED_VERSION, originChainId));
     }
 
     /**
--- a/contracts/EIP712CrossChain.sol
+++ b/contracts/EIP712CrossChain.sol
@@ -1,0 +1,85 @@
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity ^0.8.0;
+
+import "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
+
+/**
+ * @dev https://eips.ethereum.org/EIPS/eip-712[EIP 712] is a standard for hashing and signing of typed structured data.
+ *
+ * This contract is based on OpenZeppelin's implementation:
+ * https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v4.8.0/contracts/utils/cryptography/EIP712.sol
+ *
+ * NOTE: Modified version that allows to build a domain separator that relies on a different chain id than the chain this
+ * contract is deployed to. An example use case we want to support is:
+ * - User A signs a message on chain with id = 1
+ * - User B executes a method by verifying user A's EIP-712 compliant signature on a chain with id != 1
+ */
+abstract contract EIP712CrossChain {
+    /* solhint-disable var-name-mixedcase */
+    bytes32 private immutable _HASHED_NAME;
+    bytes32 private immutable _HASHED_VERSION;
+    bytes32 private immutable _TYPE_HASH;
+    // NOTE: We use a hardcoded address to allow cross-chain verification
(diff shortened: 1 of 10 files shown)


contracts/SpokePool.sol
```diff
@@ -137,14 +137,18 @@ abstract contract SpokePool is SpokePoolInterface, Testable, Lockable, MultiCall
...
         uint32 _initialDepositId,
         address _crossDomainAdmin,
         address _hubPool,
         address _wrappedNativeTokenAddress,
-        address timerAddress
-    ) Testable(timerAddress) {
+        address timerAddress,
+        string memory eip712DomainName,
+        string memory eip712DomainVersion
+    ) Testable(timerAddress) EIP712CrossChain(eip712DomainName, eip712DomainVersion) {
```
nicholaspai, line 151: nit: I think you should pass in the hardcoded `eip712DomainName` and `eip712DomainVersion` here rather than have to change the constructor of Arbitrum/Ovm/PolygonSpokePool contracts. They all would use the same name and version

contracts/EIP712CrossChain.sol
```diff
@@ -0,0 +1,85 @@
...
+ * - User A signs a message on chain with id = 1
+ * - User B executes a method by verifying user A's EIP-712 compliant signature on a chain with id != 1
+ */
+abstract contract EIP712CrossChain {
+    /* solhint-disable var-name-mixedcase */
+    bytes32 private immutable _HASHED_NAME;
+    bytes32 private immutable _HASHED_VERSION;
+    bytes32 private immutable _TYPE_HASH;
+    // NOTE: We use a hardcoded address to allow cross-chain verification
+    address public immutable _VERIFYING_CONTRACT = 0xCcCCccccCCCCcCCCCCCcCcCccCcCCCcCcccccccC;
```
nicholaspai, line 23: Would we need to change this in production? Isn't this contract the destination SpokePool's address?

nicholaspai (commented):
Left some high level comments !

contracts/EIP712CrossChain.sol
dohaki: This is again the same problem as with the chain id. Because this value is part of the domain separator, we either need to hardcode it across all chains / spoke pool contracts OR make it also a parameter. I decided to hardcode it because:
1. The spec defines this only with the keyword _may_, so my interpretation is that this is optional. See https://eips.ethereum.org/EIPS/eip-712#definition-of-domainseparator
2. If this would be a parameter, then the function signatures of the methods `speedUpDeposit` and `fillRelayWithUpdatedFee` would need to change. Because we would need to somehow pass the address of the respective spoke pool (same value for corresponding chains)?

contracts/SpokePool.sol
dohaki: changed here https://github.com/across-protocol/contracts-v2/pull/222/commits/e8a9c8d006e0d444dce07c329181740e774d8c13

contracts/EIP712CrossChain.sol
nicholaspai: hmmm this is an interesting point @dohaki . So one way around this is to just hardcode the verifying contract so that its essentially ignored. Does that break wallet clients ability to parse this message? I just don't want them to reject it because the `verifyingContract` isn't actually the contract that is verifying it

nicholaspai: As discussed offline, it seems that some of the fields like `verifyingContract` are optional in the EIP so we don't need to use that one. I still think we should keep `chainId = originChainId`. If we remove `verifyingContract` i don't think we expose ourselves to any replay attacks where someone submits the signature to a different spoke pool on the `originChain`. The SpokePools now allow fills to be paused including the `fillRelayWithUpdatedFee` so that will revert on deprecated SpokePools. Additionally, even if the `fillRelayWithUpdatedFee` succeeds, the relayer will note receive a refund if the SpokePool isn't connected to the HubPool anymore.

test/fixtures/SpokePool.Fixture.ts
```diff
@@ -297,17 +305,29 @@ export async function modifyRelayHelper(
...
+      chainId: Number(originChainId),
+      verifyingContract: "0xCcCCccccCCCCcCCCCCCcCcCccCcCCCcCcccccccC",
+    },
+    message: {
+      newRelayerFeePct: modifiedRelayerFeePct,
+      depositId,
+      originChainId,
+    },
+  };
+  const signature = await depositor._signTypedData(typedData.domain, typedData.types, typedData.message);
```
nicholaspai, line 329: 1. This is how we'd sign the speed up in the FE after this change goes through? Is there a non underscore function we can use instead?
2. Do we need any additional unit tests? Like verifying that the typed data has the expected fields and can be parsed by a wallet client?

dohaki: > 1. This is how we'd sign the speed up in the FE after this change goes through? Is there a non underscore function we can use instead?

Yes, we would use this in the FE to speed up. This is just a wrapper method provided by [ethers](https://docs.ethers.org/v5/api/signer/#Signer-signTypedData) and will eventually be replaced by `signTypedData` without an underscore. Alternatively, we could just directly use the JSON RPC like:
```js
const signature = await depositor.provider.send("eth_signTypedData_v4", [fromAddress, encodedMsgParams]
```
where we need to do some additional encoding. This approach is a bit verbose though.

> 2. Do we need any additional unit tests? Like verifying that the typed data has the expected fields and can be parsed by a wallet client?

Not sure if I understand correctly. Verifying whether the typed data has the expected fields is already covered in our tests, no? Basically, the test case is where if the signed message is not equal to the expected message then the verification fails. I am also not sure how to write a unit test on whether or not a wallet client can parse the typed data. Doesn't this depend on the implementation of the wallet? And can't we assume that if a wallet supports the method `eth_signTypedData_v4` then it can parse any message supported by the spec?

nicholaspai: > eventually be replaced by signTypedData without an underscore

Any idea when this is going to happen? Not that it blocks us i'm just curious what's holding this up

nicholaspai: > Verifying whether the typed data has the expected fields is already covered in our tests, no? Basically, the test case is where if the signed message is not equal to the expected message then the verification fails

Do we need to perform additional checks that the signed message is in the correct EIP712 format?

contracts/EIP712CrossChain.sol
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    function _buildDomainSeparator(
+        bytes32 typeHash,
+        bytes32 nameHash,
+        bytes32 versionHash,
+        uint256 originChainId
+    ) private pure returns (bytes32) {
+        return keccak256(abi.encode(typeHash, nameHash, versionHash, originChainId));
+    }
```
mrice32, line 60: nit: can we combine these methods to simplify since we're not intending to use _buildDomainSeparator outside of _domainSeparatorV4?

mrice32 (approved):
LGTM!

test/fixtures/SpokePool.Fixture.ts
dohaki: > > eventually be replaced by signTypedData without an underscore
> 
> Any idea when this is going to happen? Not that it blocks us i'm just curious what's holding this up

No idea tbh... Couldn't find anything on this in the repo.



> > Verifying whether the typed data has the expected fields is already covered in our tests, no? Basically, the test case is where if the signed message is not equal to the expected message then the verification fails
> 
> Do we need to perform additional checks that the signed message is in the correct EIP712 format?

Is this possible? The signature is in the end just the same bytes string as a message signed with `signer.signMessage`.

amateima (approved):
Nice work 👍

contracts/SpokePool.sol
```diff
@@ -698,20 +700,26 @@ abstract contract SpokePool is SpokePoolInterface, Testable, Lockable, MultiCall
...
+            keccak256(
+                abi.encode(
+                    keccak256(
+                        "UpdateRelayerFeeMessage(uint64 newRelayerFeePct,uint32 depositId,uint256 originChainId)"
+                    ),
+                    newRelayerFeePct,
+                    depositId,
+                    originChainId
+                )
+            ),
```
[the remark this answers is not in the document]
dohaki: Yea, I agree that this is a bit nested. But tbh I don't know what to put in local variables because this hash structure is defined like that in the EIP. I could do 
```solidity
bytes32 hashStruct = keccak256(...);
bytes32 expectedTypedDataV4Hash = _hashTypedDataV4(hashStruct);
```
but not sure if this makes it really more readable?

contracts/SpokePool.sol
```diff
@@ -144,7 +144,7 @@ abstract contract SpokePool is SpokePoolInterface, Testable, Lockable, MultiCall
         address _hubPool,
         address _wrappedNativeTokenAddress,
         address timerAddress
-    ) Testable(timerAddress) {
+    ) Testable(timerAddress) EIP712CrossChain("ACROSS-V2", "1.0.0") {
```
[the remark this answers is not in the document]
dohaki: Good question. I would assume so yes

contracts/EIP712CrossChain.sol
```diff
@@ -0,0 +1,72 @@
...
+     *     keccak256(bytes(mailContents))
+     * ));
+     * bytes32 digest = _hashTypedDataV4(structHash, originChainId);
+     * address signer = ECDSA.recover(digest, signature);
+     * ```
+     */
+    function _hashTypedDataV4(bytes32 structHash, uint256 originChainId) internal view virtual returns (bytes32) {
+        return ECDSA.toTypedDataHash(_domainSeparatorV4(originChainId), structHash);
+    }
+}
```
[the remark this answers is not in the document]
dohaki: Done here https://github.com/across-protocol/contracts-v2/pull/222/commits/566b4ddc976ac8e05fc2723a74d771102117cb56!

contracts/EIP712CrossChain.sol
dohaki: Done here https://github.com/across-protocol/contracts-v2/pull/222/commits/21aab9c2453ce1a06c4833ef7cf3e77a1d2b9d1f

contracts/SpokePool.sol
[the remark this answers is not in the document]
nicholaspai: Is there a gas difference if you change the syntax? Can run the test like so `REPORT_GAS=true hardhat test`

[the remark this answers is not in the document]
dohaki: @nicholaspai Thanks for the suggestion. Yes, with the additional local variable the methods `speedUpDeposit` and `fillRelayWithUpdatedFee` cost 10 gas more.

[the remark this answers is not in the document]
nicholaspai: this is an interesting case where smart contract engineering is different from off-chain engineering. Temp variables increase readability but incur a real cost! In this case I'd opt for the reduction in readability to save gas

nicholaspai:
> @nicholaspai I e2e tested a deployed version of the contract with the frontend and everything seems fine 👍 When you are happy, then I can merge

https://github.com/across-protocol/contracts-v2/pull/222#discussion_r1093406199

Just one comment then LGTM

nicholaspai (commented):
https://github.com/across-protocol/contracts-v2/pull/222#discussion_r1093406199

dohaki:
> [#222 (comment)](https://github.com/across-protocol/contracts-v2/pull/222#discussion_r1093406199)

@nicholaspai Thanks! I haven't actually put the nested stuff in a local variable. So this PR has the version with lower gas costs

nicholaspai (approved):

PR #224, opened 2023-01-24
feat: Add Foundry support

nicholaspai:
Adds a simple unit test + documentation. Hoping we can start to add some unit tests here eventually but this is the 0->1 PR that will make it easy for anyone to add solidity tests.

One awesome feature of forge is the native [Fuzz](https://book.getfoundry.sh/forge/fuzz-testing) and [Fork](https://book.getfoundry.sh/forge/fork-testing)

Very useful for fast compiling contracts and checking file size.

Tests are NOT added to CI in this PR.


contracts/interfaces/WETH9.sol
```diff
@@ -8,5 +8,5 @@ interface WETH9 {
 
     function balanceOf(address guy) external view returns (uint256 wad);
 
-    function transfer(address guy, uint256 wad) external;
+    function transfer(address guy, uint256 wad) external returns (bool);
```
nicholaspai, line 11: Didn't even realize we defined the interface incorrectly until I tried to use this in a Soldidity test

remappings.txt
```diff
@@ -0,0 +1,2 @@
+forge-std/=lib/forge-std/src/
+@prb/test/=lib/prb-test/src/
```
nicholaspai, line 2: [this is a more advanced version of the `forg-std/Test` contract](https://github.com/PaulRBerg/prb-test)

PR #225, opened 2023-01-25
improve: Default depositQuoteTimeBuffer=3600

nicholaspai:
This default lines up with what is currently set in the live SpokePools, as depositors realized eventually that making this buffer less strict improved depositor UX and fees didn't change much within the 
hour. This reduces the chance of a deposit reverting if the API providing a suggested fee is stale for an hour.


mrice32 (approved):
LGTM!

PR #226, opened 2023-02-01
feat: Permissioned Across-v2 bond token

pxrl:
This PR creates a new bond token, to be used as the collateral required by the HubPool on the following functions:
 - proposeRootBundle()
 - disputeRootBundle()

The new bond token (symbol "ABT") hosts an internal set of approved proposer addresses in order to impose constraints on attempted inbound ABT transfers to the HubPool. Permissioned addresses are unconstrained in their transfer of ABT, whereas all other addresses must _not_ be the proposer of a root bundle in order to transfer ABT to the HubPool. This permits any ABT holder to submit a dispute of an existing root bundle proposal, whilst preventing non-approved addresses from proposing.

ABT inherits from the WETH9 contract with minimal modifications necessary to override the transferFrom() function.

Fixes ACX-699 ACX-700


deploy/019_deploy_bond_token.ts
```diff
@@ -0,0 +1,22 @@
+import "hardhat-deploy";
```
pxrl, line 1: This deployment script has not been tested, so may well contain errors.

contracts/BondToken.sol
```diff
@@ -0,0 +1,80 @@
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity ^0.8.0;
+
+import "@openzeppelin/contracts/access/Ownable.sol";
+import "@openzeppelin/contracts/utils/Address.sol";
+
+import "./HubPoolInterface.sol";
+import "./WETH9.sol";
+
+interface _HubPool is HubPoolInterface {
```
nicholaspai, line 10: Prefer sticking to existing naming convention and not prefixing this with a _. Maybe name HubPoolExtendedInterface or something, or even IHubPool

contracts/BondToken.sol
```diff
@@ -0,0 +1,80 @@
...
+     * @notice Emitted on proposer permissions update.
+     */
+    event ProposerModified(address proposer, bool enabled);
+
+    /**
+     * @notice BondToken constructor.
+     * @param _hubPool Address of the target HubPool contract.
+     */
+    constructor(_HubPool _hubPool) {
+        name = "Across v2 Bond Token";
```
nicholaspai, line 44:
```suggestion
        name = "Across Bond Token";
```
Not sure V2 is necessary. Do you think we should name it Proposer Bond instead of just Bond?

contracts/BondToken.sol
```diff
@@ -0,0 +1,80 @@
...
+        symbol = "ABT";
```
nicholaspai, line 45: nit: can we come up with a catchier symbol? :P

contracts/BondToken.sol
```diff
@@ -0,0 +1,80 @@
...
+     * @param amt Amount to transfer.
+     * @return True on success.
+     */
+    function transferFrom(
+        address src,
+        address dst,
+        uint256 amt
+    ) public override returns (bool) {
+        if (dst == address(hubPool)) {
+            require(proposers[src] || hubPool.rootBundleProposal().proposer != src, "Transfer not permitted");
```
nicholaspai, line 76: Does this mean proposer's can't dispute themselves? I don't think we should block such behavior

contracts/WETH9.sol
```diff
@@ -0,0 +1,77 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity ^0.8.0;
+
+// Copied from the verified code from Etherscan:
+// https://etherscan.io/address/0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2#code
+// And then updated for Solidity version 0.6. Specific changes:
```
nicholaspai, line 6:
```suggestion
// And then updated for Solidity version 0.8. Specific changes:
```

deploy/019_deploy_bond_token.ts
nicholaspai: you can simulate this on local hardhat network by running `yarn hardhat deploy --tags BondToken` without specifying the `--network` flag.

nicholaspai (commented):
Did a once over and left some comments

contracts/BondToken.sol
```diff
@@ -0,0 +1,80 @@
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity ^0.8.0;
+
+import "@openzeppelin/contracts/access/Ownable.sol";
+import "@openzeppelin/contracts/utils/Address.sol";
+
+import "./HubPoolInterface.sol";
+import "./WETH9.sol";
```
pxrl, line 8: WETH9 was copied here in order to mark the transferFrom() function virtual.

contracts/BondToken.sol
pxrl: This check only requires that one of the conditions is true, so proposers can always dispute without impediment. There's a test that verifies that - see test/BondToken.e2e.ts line 72.

contracts/WETH9.sol
pxrl: That comment was actually pre-existing from the variant that I copied from the UMA repository, so I think it shold stay as-is, because the only change I made for newer versions of solidity was marking transferFrom() as virtual. This is unfortunately the downside of having copied and modified the content in the same commit :(

contracts/BondToken.sol
pxrl: "AMT" => "Across Moon Token" ?

contracts/BondToken.sol
pxrl: Preface: Naming things is hard :)

My thinking with the "v2" naming is that this token has a finite lifespan - it will only be in use as long as the current (Across v2) HubPool is deployed. That's why I preferred to include v2 in the name.

contracts/BondToken.sol
pxrl: Good point - I'll find something better :+1:

contracts/WETH9.sol
pxrl: Maybe I should change the comment to identify the UMA repo commit where I got it from :thinking:.

contracts/BondToken.sol
pxrl: Implemented: https://github.com/across-protocol/contracts-v2/pull/226/commits/91c1780c121f2782eed0e552092c6f017cf23f42

contracts/BondToken.sol
mrice32: This is _extremely_ counterintuitive, but it's really clever. This is the sort of logic that I think Certora would be really good for verifying. We would probably want to verify that any address with tokens can _always_ send a dispute.

I'm not super worried about this edge case, but this logic does seem to mean that if I proposed and then I was removed from the whitelist, I would then be no longer able to dispute myself.

Just out of curiousity, why did you not use the the check we use in the HubPool, that [unclaimedPoolRebalanceLeafCount](https://github.com/across-protocol/contracts-v2/blob/598b17f019f80540e7d5d0ae7bad15235edb1e62/contracts/HubPoolInterface.sol#L68) != 0 to detect disputes vs proposals. See the [noActiveRequests](https://github.com/across-protocol/contracts-v2/blob/598b17f019f80540e7d5d0ae7bad15235edb1e62/contracts/HubPool.sol#L173-L176) modifier.

EDIT: That last idea doesn't work. I didn't quite get the nuance you picked up on, that the variables are written _before_ the transfer. I think the way you've done it and checking the timestamp is probably the only way I could see this working. Very clever.

contracts/WETH9.sol
```diff
@@ -0,0 +1,95 @@
...
+        allowance[msg.sender][guy] = wad;
+        emit Approval(msg.sender, guy, wad);
+        return true;
+    }
+
+    function transfer(address dst, uint256 wad) public returns (bool) {
+        return transferFrom(msg.sender, dst, wad);
+    }
+
+    function transferFrom(
```
mrice32, line 76: nit: can we add a comment that this is the _only_ difference between this contract and the real WETH9?

contracts/WETH9.sol
```diff
@@ -0,0 +1,95 @@
...
+/**
+ * Imported as at commit 33d01d471437e1ab6861e4545ea4bb3895fd4d74 from:
+ * UMAprotocol/protocol/packages/core/contracts/financial-templates/common/WETH9.sol
+ * Updated with the following changes post-import:
+ * - Marked transferFrom() as virtual
+ */
+
+pragma solidity ^0.8.0;
+
+contract WETH9 {
```
mrice32, line 29: nit: we probably don't want two contracts in the repo called WETH9 -- this causes issues in hardhat I think. Any reason we can't just delete the interface or rename it to WETH9Interface?

mrice32 (dismissed):
This is a really clever implementation. Nice work, @pxrl!

contracts/BondToken.sol
pxrl: > I'm not super worried about this edge case, but this logic does seem to mean that if I proposed and then I was removed from the whitelist, I would then be no longer able to dispute myself.

That was also my intuition, but then I tried to implement a test for it and found the opposite - it works unexpected. This is similarly because the pending root bundle is deleted before the bond token transferFrom() function is called. It's a happy accident, I guess.

pxrl: > > I'm not super worried about this edge case, but this logic does seem to mean that if I proposed and then I was removed from the whitelist, I would then be no longer able to dispute myself.
> 
> That was also my intuition, but then I tried to implement a test for it and found the opposite - it works unexpectedly. This is similarly because the pending root bundle is deleted before the bond token transferFrom() function is called. It's a happy accident, I guess.
> 
> For reference, see the test at line 78 in test/BondToken.e2e.ts.

> Just out of curiousity, why did you not use the the check we use in the HubPool, that [unclaimedPoolRebalanceLeafCount](https://github.com/across-protocol/contracts-v2/blob/598b17f019f80540e7d5d0ae7bad15235edb1e62/contracts/HubPoolInterface.sol#L68) != 0 to detect disputes vs proposals. See the [noActiveRequests](https://github.com/across-protocol/contracts-v2/blob/598b17f019f80540e7d5d0ae7bad15235edb1e62/contracts/HubPool.sol#L173-L176) modifier.

That was my first attempt as well, but it doesn't work. At the time of evaluation, the proposed root bundle _is_ the version submitted by the proposer, so it seems reliably non-zero in case of both proposals and disputes.

contracts/WETH9.sol
pxrl: Have updated the comment to clarify that we sourced it from UMAprotocol/protocol (w/ commit hash), and went further by restoring the proper license header. We've previously improperly removed attribution to Dapphub, and it's a requirement of the GPL that we retain the pre-existing license header.

contracts/WETH9.sol
pxrl: Good point - I'd totally missed the existing interface. I've split this out - see https://github.com/across-protocol/contracts-v2/pull/227.

contracts/BondToken.sol
pxrl: Have clarified that in the WETH9 implementation.

contracts/WETH9.sol
pxrl: Added here: https://github.com/across-protocol/contracts-v2/pull/226/commits/edbee53ba48609c8c17a2031b254261f0343c302.

contracts/BondToken.sol
pxrl: > This is _extremely_ counterintuitive

I've added a @dev comment to the BondToken.safeTransferFrom() implementation to call this out more explicitly. See https://github.com/across-protocol/contracts-v2/pull/226/commits/5215b60ab414e4d1d0f9feac6ab4a0b3aba4fea6.

test/BondToken.e2e.ts
```diff
@@ -0,0 +1,141 @@
...
+      } else {
+        // Proposal successful; bondAmount is transferred from proposer to HubPool.
+        await expect(proposeRootBundle(hubPool, dataworker)).to.emit(hubPool, "ProposeRootBundle");
+        expect((await hubPool.rootBundleProposal()).proposer).to.equal(dataworker.address);
+        expect(await bondToken.balanceOf(hubPool.address)).to.equal(hubPoolBal.add(consts.bondAmount));
+        expect((await bondToken.balanceOf(dataworker.address)).eq(dataworkerBal.sub(consts.bondAmount))).to.be.true;
+      }
+    }
+  });
+
```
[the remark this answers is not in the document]
pxrl: Implemented: https://github.com/across-protocol/contracts-v2/pull/226/commits/d860e618113c11af1d45863d94dd86080a74224f.

deploy/019_deploy_bond_token.ts
pxrl: Resolved: https://github.com/across-protocol/contracts-v2/pull/226/commits/2ae7164d0996e9c752a9ae5c3bee6f493a16bf86.

contracts/BondToken.sol
nicholaspai: I generally don't like Versions in token names, look at UMA Voting Token V1 for example

contracts/BondToken.sol
nicholaspai: Will think on this more haha

contracts/BondToken.sol
nicholaspai: I had the same question as Matt but see your point now!

contracts/WETH9.sol
nicholaspai: Nice +1

contracts/BondToken.sol
pxrl: "v2" removed: https://github.com/across-protocol/contracts-v2/pull/226/commits/ee0aedb78f29999d9518bec709beeaa13749e9a4

I do think it's a different scenario to UMA v1, but I can live with either :)

contracts/BondToken.sol
mrice32: Wow, it keeps getting weirder @pxrl. Really would love some validation on this!

mrice32 (approved):
LGTM!

PR #227, opened 2023-02-01
chore: Import WETH9

pxrl:
Done in advance of the new Across Bond Token (name tbc)
implementation, which will inherit from WETH9.

 - Rename the existing WETH9 -> WETH9Interface
 - Import WETH9 from UMAproject/protocol.
 - WETH9 licensing corrections.

The reason for retaining the WETH9 interface description is that it's
less restrictive than the WETH9 contract that is being imported.
Specifically, Solidity is unhappy that the WETH9 contract identifies
payable functions, and that has ripple effects for the existing users
of the WETH9 interface. We might consider addressing those points,
but the focus here is just to shuffle the furniture to make space for
the new bond token.

Further changes to come once this is in.


mrice32 (commented):
Thanks for adding this! Some systems care about the _name of the contract_, not the name of the file. It's good cleanliness to name the file WETH9Interface.sol, but to make sure there's no conflicts in the hardhat artifacts or in other flattening procedures, we should change the name of the interface in the solidity WETH9 -> WETH9Interface.

pxrl:
> Thanks for adding this! Some systems care about the _name of the contract_, not the name of the file. It's good cleanliness to name the file WETH9Interface.sol, but to make sure there's no conflicts in the hardhat artifacts or in other flattening procedures, we should change the name of the interface in the solidity WETH9 -> WETH9Interface.

Fair point - have renamed to WETH9 interface.

What's your gut feel on this? Is it cleaner to just nuke the interface definition and use the newly imported version? I'm not sure either way.

nicholaspai (approved):

PR #229, opened 2023-02-04
improve: Make changes after running Slither on SpokePool contracts

nicholaspai:
[Slither](https://github.com/crytic/slither) is a Solidity static analysis framework written in Python 3. It runs a
suite of vulnerability detectors, prints visual information about contract details, and provides an API to easily write
custom analyses. Slither enables developers to find vulnerabilities, enhance their code comprehension, and quickly
prototype custom analyses.

Spire-Contracts has been analyzed using `Slither@0.9.2` and no major bugs was found. To rerun the analytics, run:

```sh
slither contracts/SpokePool.sol
\ --solc-remaps @=node_modules/@
\ --solc-args "--optimize --optimize-runs 1000000"
\ --filter-paths "node_modules"
\ --exclude naming-convention
```

You can replace `SpokePool.sol` with the specific contract you want to analyze. For this PR I ran this on the following contracts:
- Ovm_SpokePool.sol
- Polygon_SpokePool.sol
- Optimism_SpokePool.sol
- Boba_SpokePool.sol
- Ethereum_SpokePool.sol
- SpokePool.sol
- PolygonTokenBridger

This covered any contracts imported by those contracts. Notably I did not make any changes recommended to HubPool.sol, even though there were some, because we don't have plans to redeploy that soon.


contracts/Ovm_SpokePool.sol
```diff
@@ -59,6 +59,7 @@ contract Ovm_SpokePool is SpokePool {
         l1Gas = 5_000_000;
         __SpokePool_init(_initialDepositId, _crossDomainAdmin, _hubPool, _wrappedNativeToken, _timerAddress);
         messenger = Lib_PredeployAddresses.L2_CROSS_DOMAIN_MESSENGER;
+        require(_l2Eth != address(0), "Invalid L2 ETH address");
```
nicholaspai, line 62: Question: Slither recommends we add these zero address checks. It does add a fixed gas cost to construction but adds some safety. Thoughts?

contracts/PolygonTokenBridger.sol
```diff
@@ -38,7 +38,7 @@ interface MaticToken {
  *  This ultimately allows create2 to generate deterministic addresses that don't depend on the transaction count of the
  * sender.
  */
-contract PolygonTokenBridger is Lockable {
+contract PolygonTokenBridger is ReentrancyGuard {
```
nicholaspai, line 41: Slither had some complaints about Lockable but i figured this was a good time to use OZ's ReentrancyGuard instead

contracts/SpokePool.sol
```diff
@@ -290,6 +290,11 @@ abstract contract SpokePool is
      * to ensure that a small input range doesn't limit which indices this method is able to reach.
      */
     function emergencyDeleteRootBundle(uint256 rootBundleId) public override onlyAdmin nonReentrant {
+        // Deleting a struct containing a mapping does not delete the mapping in Solidity, therefore the bitmap's
```
nicholaspai, line 293: This warning was the most interesting in this PR IMO

contracts/SpokePool.sol
```diff
@@ -332,11 +337,17 @@ abstract contract SpokePool is
...
         // in underflow. This isn't a problem as the deposit will revert, but the error might be unexpected for clients.
+
+        //slither-disable-next-line timestamp
         require(
             getCurrentTime() >= quoteTimestamp - depositQuoteTimeBuffer &&
                 getCurrentTime() <= quoteTimestamp + depositQuoteTimeBuffer,
             "invalid quote time"
         );
+
+        // Increment count of deposits so that deposit ID for this spoke pool is unique.
```
nicholaspai, line 348: Move the storage variable modification to `numberOfDeposits` above the external call `safeTransferFrom` to align with CEI pattern

contracts/interfaces/WETH9Interface.sol
```diff
@@ -8,5 +8,5 @@ interface WETH9Interface {
 
     function balanceOf(address guy) external view returns (uint256 wad);
 
-    function transfer(address guy, uint256 wad) external;
+    function transfer(address guy, uint256 wad) external returns (bool);
```
nicholaspai, line 11: This interface was incorrect and inconsistent with how we implement WETH9

contracts/PolygonTokenBridger.sol
```diff
@@ -90,6 +90,8 @@ contract PolygonTokenBridger is Lockable {
         uint256 _l1ChainId,
         uint256 _l2ChainId
     ) {
+        require(_destination != address(0), "invalid dest address");
+        require(_l2WrappedMatic != address(0), "invalid wmatic address");
```
pxrl, line 94: I'm not sure about the value of these checks. 0 is only one possible invalid address, but there are ~infinitely many other invalid addresses as well. The only way we can really be confident that the address is set correctly is via post-deployment verification.

I'm not firmly against it or anything, it just seems a bit redundant and might give a false sense of security.

contracts/Ovm_SpokePool.sol
pxrl: I'm not sure about the value of these checks. 0 is only one possible invalid address, but there are ~infinitely many other invalid addresses as well. The only way we can really be confident that the address is set correctly is via post-deployment verification.

I'm not firmly against it or anything, it just seems a bit redundant and might give a false sense of security.

pxrl (commented):
Good initiative!

contracts/SpokePool.sol
mrice32: I agree with @pxrl. Just to add one point: we will never reuse this array slot (because deleting an element of an array does not actually reduce the array.length I believe, meaning the slot is still reserved), so the mapping will never be read or written to again, by definition, no? So we shouldn't need to worry about this at all right?

pxrl: > Just to add one point: we will never reuse this array slot (because deleting an element of an array does not actually reduce the array.length I believe, meaning the slot is still reserved), so the mapping will never be read or written to again, by definition, no? So we shouldn't need to worry about this at all right?

You're right - I'd understood that the array element was actually removed, but I can see that it's only zeroed. So my 4th point was garbage!

README.md
```diff
@@ -41,6 +41,25 @@ NODE_URL_1=https://mainnet.infura.com/xxx yarn hardhat deploy --tags HubPool --n
...
+prototype custom analyses.
+
+Spire-Contracts has been analyzed using `Slither@0.9.2` and no major bugs was found. To rerun the analytics, run:
+
+```sh
+slither contracts/SpokePool.sol
+\ --solc-remaps @=node_modules/@
+\ --solc-args "--optimize --optimize-runs 1000000"
+\ --filter-paths "node_modules"
+\ --exclude naming-convention
```
pxrl, line 58: Should we make a shortcut for this via package.json?

contracts/Ovm_SpokePool.sol
nicholaspai: i think I agree with your comment and will remove these zero address checks

contracts/PolygonTokenBridger.sol
mrice32: I don't think this contract will be in scope for the audit. Are we okay making this change?

mrice32 (approved):
LGTM!

contracts/PolygonTokenBridger.sol
nicholaspai: good point, removed

PR #232, opened 2023-02-08
feat: add succinct contracts

mrice32:
This includes the Succinct SpokePool and Adapter contracts. Note: succinct is only a message bridge, so the methods that relay tokens have been left as no-ops. Those will either need to be overridden, or be left as no-ops and never used on those chains (in favor of arbitrage-style mechanisms). Right now, I have these as no-ops rather than reverting so they don't prevent other related actions from going through. Might be a good idea to have them emit a warning event that we can watch for.

Note: I used the succinct contracts in this repo: https://github.com/succinctlabs/telepathy-contracts.


contracts/Succinct_SpokePool.sol
```diff
@@ -0,0 +1,72 @@
...
+
+contract Succinct_SpokePool is SpokePool, ITelepathyHandler {
+    // Address of the succinct contract.
+    address public succinctTargetAmb;
+
+    // Warning: this variable should _never_ be touched outside of this contract. It is intentionally set to be
+    // private. Leaving it set to true can permanently disable admin calls.
+    bool private adminCallValidated;
+
+    uint16 public hubChainId;
```
nicholaspai, line 16: nit: can we make this a constant = 1 and temporarily override when deploying to goerli? Not a big deal but save some gas

contracts/Succinct_SpokePool.sol
```diff
@@ -0,0 +1,72 @@
...
+        // Validate msg.sender as succinct, the x-chain sender as being the hubPool (the admin) and the source chain as
+        // 1 (mainnet).
+        require(
+            msg.sender == succinctTargetAmb && _senderAddress == hubPool && _sourceChainId == hubChainId,
+            "Invalid message"
+        );
+
+        // This operates similarly to a re-entrancy guard. It is set after validation to tell methods called by this
+        // method that the call has been validated as an admin and is safe.
+        require(!adminCallValidated, "Re-entered handleTelepathy");
```
nicholaspai, line 51: Seems like a good candidate to use a `modifier` even though it adds contract storage and potentailly adds gas, it makes it more readable

nicholaspai (approved):
This is pretty straightforward to me LGTM

contracts/Succinct_SpokePool.sol
mrice32: I don't think we can do that (unless we make it immutable) because it would be stored in the implementation contract and not the proxy, right?

contracts/Succinct_SpokePool.sol
mrice32: Done.

PR #233, opened 2023-02-14
feat: universal bridge adapter

mrice32:
This modifies the SpokePool to support incentives to move tokens to other chains. This is done by establishing a defined ordering of deposits and fills on each chain and rewarding depositors and relayers for moving the tokens towards the right running balance.

Specific changes:
- To avoid frontrunning and allow depositors and relayers to lock in their reward, a running count of the aggregate amount of funds is tracked for both deposits and fills. Depositors and relayers can then pass in a maxCount to cap the amount of capital coming in ahead of them that they did not anticipate.
- Both LP fees and relayer fees can now be negative.
- The slow fill process/struct includes a new value, called payoutAdjustment. This enables the fees/rewards that would typically be passed on to a relayer to be passed on to the end-user in the case of a slow fill.
- When relayers take refunds on other chains, their impact is not considered on this chain or the other chain. This means that the dataworker will need to "deprioritize" their ordering and give them the least favored positioning within the bundle.

### DIFF
--- a/contracts/SpokePool.sol
+++ b/contracts/SpokePool.sol
@@ -876,7 +876,10 @@ abstract contract SpokePool is
         // We limit the relay fees to prevent the user spending all their funds on fees. Note that 0.5e18 (i.e. 50%)
         // fees are just magic numbers. The important point is to prevent the total fee from being 100%, otherwise
         // computing the amount pre fees runs into divide-by-0 issues.
-        require(updatableRelayerFeePct < 0.5e18 && relayData.realizedLpFeePct < 0.5e18, "invalid fees");
+        require(
+            SignedMath.abs(updatableRelayerFeePct) < 0.5e18 && SignedMath.abs(relayData.realizedLpFeePct) < 0.5e18,
+            "invalid fees"
+        );
 
         // Check that the relay has not already been completely filled. Note that the relays mapping will point to
         // the amount filled so far for a particular relayHash, so this will start at 0 and increment with each fill.
--- a/contracts/SpokePool.sol
+++ b/contracts/SpokePool.sol
@@ -15,6 +15,7 @@ import "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
 import "@openzeppelin/contracts-upgradeable/utils/AddressUpgradeable.sol";
 import "@openzeppelin/contracts-upgradeable/proxy/utils/UUPSUpgradeable.sol";
 import "@openzeppelin/contracts-upgradeable/security/ReentrancyGuardUpgradeable.sol";
+import "@openzeppelin/contracts/utils/math/SignedMath.sol";
 
 /**
  * @title SpokePool
@@ -96,7 +97,7 @@ abstract contract SpokePool is
         uint256 amount,
         uint256 originChainId,
         uint256 destinationChainId,
-        uint64 relayerFeePct,
+        int64 relayerFeePct,
         uint32 indexed depositId,
         uint32 quoteTimestamp,
         address indexed originToken,
@@ -104,7 +105,7 @@ abstract contract SpokePool is
         address indexed depositor
     );
     event RequestedSpeedUpDeposit(
-        uint64 newRelayerFeePct,
+        int64 newRelayerFeePct,
         uint32 indexed depositId,
         address indexed depositor,
         bytes depositorSignature
@@ -116,9 +117,9 @@ abstract contract SpokePool is
         uint256 repaymentChainId,
         uint256 originChainId,
         uint256 destinationChainId,
-        uint64 relayerFeePct,
-        uint64 appliedRelayerFeePct,
-        uint64 realizedLpFeePct,
+        int64 relayerFeePct,
+        int64 appliedRelayerFeePct,
+        int64 realizedLpFeePct,
         uint32 indexed depositId,
         address destinationToken,
         address indexed relayer,
@@ -337,9 +338,8 @@ abstract contract SpokePool is
         require(enabledDepositRoutes[originToken][destinationChainId], "Disabled route");
 
         // We limit the relay fees to prevent the user spending all their funds on fees.
-        require(relayerFeePct < 0.5e18, "invalid relayer fee");
-
-        require(amount < 1e36, "amount too large");
+        require(SignedMath.abs(relayerFeePct) < 0.5e18, "Invalid relayer fee");
+        require(amount < 1e36, "Amount too large");
         require(maxCount >= depositCounter[originToken], "Above max count");
 
         // This function assumes that L2 timing cannot be compared accurately and consistently to L1 timing. Therefore,
@@ -403,7 +403,7 @@ abstract contract SpokePool is
      */
     function speedUpDeposit(
         address depositor,
-        uint64 newRelayerFeePct,
+        int64 newRelayerFeePct,
         uint32 depositId,
         bytes memory depositorSignature
     ) public override nonReentrant {
@@ -453,8 +453,8 @@ abstract contract SpokePool is
         uint256 maxTokensToSend,
         uint256 repaymentChainId,
         uint256 originChainId,
-        uint64 realizedLpFeePct,
-        uint64 relayerFeePct,
+        int64 realizedLpFeePct,
+        int64 relayerFeePct,
         uint32 depositId,
         uint256 maxCount
     ) public nonReentrant unpausedFills {
@@ -520,9 +520,9 @@ abstract contract SpokePool is
         uint256 maxTokensToSend,
         uint256 repaymentChainId,
         uint256 originChainId,
-        uint64 realizedLpFeePct,
-        uint64 relayerFeePct,
-        uint64 newRelayerFeePct,
+        int64 realizedLpFeePct,
+        int64 relayerFeePct,
+        int64 newRelayerFeePct,
         uint32 depositId,
         bytes memory depositorSignature,
         uint256 maxCount
@@ -586,8 +586,8 @@ abstract contract SpokePool is
         address destinationToken,
         uint256 amount,
         uint256 originChainId,
-        uint64 realizedLpFeePct,
-        uint64 relayerFeePct,
+        int64 realizedLpFeePct,
+        int64 relayerFeePct,
         uint32 depositId,
         uint32 rootBundleId,
         int256 payoutAdjustment,
@@ -719,8 +719,8 @@ abstract contract SpokePool is
         uint256 amount,
         uint256 originChainId,
         uint256 destinationChainId,
-        uint64 realizedLpFeePct,
-        uint64 relayerFeePct,
+        int64 realizedLpFeePct,
+        int64 relayerFeePct,
         uint32 depositId,
         uint32 rootBundleId,
         int256 payoutAdjustment,
@@ -785,7 +785,7 @@ abstract contract SpokePool is
     function _verifyUpdateRelayerFeeMessage(
         address depositor,
         uint256 originChainId,
-        uint64 newRelayerFeePct,
+        int64 newRelayerFeePct,
         uint32 depositId,
         bytes memory depositorSignature
     ) internal view {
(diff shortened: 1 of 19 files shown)


nicholaspai (commented):
meta level comment: I think it'd be useful to draft the UMIP changes that would lead to this implementation change. I recall when writing the original UMIP that we wish we had made changes to the contract to make the UMIP easier to write

contracts/SpokePool.sol
```diff
@@ -319,14 +330,18 @@ abstract contract SpokePool is
...
+        uint32 quoteTimestamp,
+        uint256 maxCount
     ) public payable override nonReentrant unpausedDeposits {
         // Check that deposit route is enabled.
         require(enabledDepositRoutes[originToken][destinationChainId], "Disabled route");
 
         // We limit the relay fees to prevent the user spending all their funds on fees.
-        require(relayerFeePct < 0.5e18, "invalid relayer fee");
+        require(SignedMath.abs(relayerFeePct) < 0.5e18, "Invalid relayer fee");
+        require(amount < 1e36, "Amount too large");
```
nicholaspai, line 342: why this change?

contracts/SpokePool.sol
```diff
@@ -319,14 +330,18 @@ abstract contract SpokePool is
...
+        require(depositCounter[originToken] <= maxCount, "Above max count");
```
nicholaspai, line 343: Could you not re-use the `depositId` to prevent front-running instead of using the new `depositCounter` per token? i.e. make user pass in `maxDepositId`

contracts/SpokePool.sol
```diff
@@ -706,7 +750,16 @@ abstract contract SpokePool is
         // Note: use relayAmount as the max amount to send, so the relay is always completely filled by the contract's
         // funds in all cases. As this is a slow relay we set the relayerFeePct to 0. This effectively refunds the
         // relayer component of the relayerFee thereby only charging the depositor the LpFee.
-        uint256 fillAmountPreFees = _fillRelay(relayHash, relayData, relayData.amount, 0, true);
+        uint256 fillAmountPreFees = _fillRelay(
```
nicholaspai, line 753: Does _fillRelay have any new divide by zero checks we need to add now that the fees can be negative?

contracts/SpokePool.sol
```diff
@@ -883,12 +959,40 @@ abstract contract SpokePool is
...
+        uint256 startingFillAmount,
+        uint256 endingFillAmount,
+        uint256 totalFillAmount,
+        int64 realizedLPFeePct,
+        address token,
+        bool localRepayment,
+        bool useContractFunds
+    ) internal {
+        // If this is a slow fill or it's an initial 0-fill, do nothing, as these should not impact the count.
+        if (useContractFunds || endingFillAmount == 0) return;
```
nicholaspai, line 972: I don't think we allow 0 fills, only 1 wei fills

contracts/SpokePool.sol
```diff
@@ -883,12 +959,40 @@ abstract contract SpokePool is
...
+
+        // If this is the first fill and it's partial, assume the rest of the fill will be slow filled (refunded on this chain).
+        if (startingFillAmount == 0 && totalFillAmount - endingFillAmount > 0) {
+            fillCounter[token] += _computeAmountPostFees(totalFillAmount - endingFillAmount, realizedLPFeePct);
+        }
+
+        // If this is not the first fill, remove the partial fill that was previously assumed.
+        if (startingFillAmount != 0) {
```
nicholaspai, line 980: What if there were multiple partial fills that were not the first fill?

hardhat.config.ts
```diff
@@ -23,19 +23,19 @@ dotenv.config();
 // the following config is true.
 const compileZk = process.env.COMPILE_ZK === "true";
 
-const solcVersion = "0.8.17";
+const solcVersion = "0.8.18";
```
nicholaspai, line 26: +1

hardhat.config.ts
```diff
@@ -23,19 +23,19 @@ dotenv.config();
...
 const mnemonic = getMnemonic();
 
 // Compilation settings are overridden for large contracts to allow them to compile without going over the bytecode
 // limit.
 const LARGE_CONTRACT_COMPILER_SETTINGS = {
   version: solcVersion,
-  settings: { optimizer: { enabled: true, runs: 200 } },
+  settings: { optimizer: { enabled: true, runs: 200 }, viaIR: true },
```
nicholaspai, line 33: What does `viaIR` do?

test/MerkleLib.utils.ts
```diff
@@ -120,10 +120,10 @@ export async function constructSingleChainTree(token: string, scalingSize = 1, r
   return { tokensSendToL2, realizedLpFees, leaves, tree };
 }
 
-export async function buildSlowRelayTree(relays: RelayData[]) {
```
nicholaspai, line 123: Did you just rename `RelayData` to `SlowFill` for better readability?

nicholaspai (commented):
Did a first pass review, haven't reviewed tests yet only solidity files

hardhat.config.ts
mrice32: It essentially allows us to get around stack too deep errors. It modifies the compilation process to allow solidity to use an intermediate representation. Not sure of the details other than this is what the compiler recommends when you encounter stack too deep errors.

contracts/SpokePool.sol
mrice32: I meant to add the same check on fillRelay. Will add that now.

The idea is that if someone can send an invalid fill of arbitrary size, they could theoretically force the fillCounter to overflow, blocking all fills thereafter (because adding to it would revert). So we want to cap the size of the amount on the fill side to make it infeasible for an attacker to force an overflow on the fill side. We mirror the check on the deposit side to avoid a scenario where a deposit could theoretically be unfillable.

With this cap, someone would need to create 5.78e40 invalid fills to force an overflow, which is effectively infeasible.

contracts/SpokePool.sol
mrice32: It's possible, yes. However, I don't think it works as well. Here's the example I've been thinking about:

In normal operation, many users may be wanting to deposit in the same pool at the same time. Instead of setting the deposit count to exactly the current value, they add some cushion to avoid transactions frequently reverting (kind of like uniswap slippage parameters). Let's say my fee is roughly the same as long as there isn't more than 100k of volume that slips in before me. I can then set the current depositCounter + 100k.

Similarly, if I were using depositIds, I could set it to exactly the current value, but if _any_ transaction slips in front of me, my transaction would revert. This would be like setting your uniswap slippage to 0. This would be fine if transactions are well spread out, but if transactions are happening often and someone takes a minute or so to sign their transaction, they will encounter reverts a lot. If I wanted to do something similar to the above for depositIds, I would basically be limiting the _number_ of transactions that come in front of me. Those transactions could be for $1 or $1MM. I have no real way to guarantee that these deposits won't change my quote by more than X.

Does that make sense?

contracts/SpokePool.sol
mrice32: But it is possible to send 0 fills, right?

contracts/SpokePool.sol
mrice32: We're not removing the entire slow fill amount, only the amount that was replaced, so I _think_ this should handle this correctly (but please lmk if you think it's wrong!).

contracts/SpokePool.sol
mrice32: I can't think of any -- can you? The total fees cannot be 100% or -100%. They are capped at +-50%, so I think the remaining amount is guaranteed to always be >0.

test/MerkleLib.utils.ts
mrice32: No, it used to be that RelayData and SlowFill were the same data struct. That's no longer true. We have the payoutAdjustment parameter that is _not_ in the normal relayData struct (which is what we hash to identify relays).

contracts/SpokePool.sol
mrice32: Done.

nicholaspai: ok this seems safe

contracts/SpokePool.sol
nicholaspai: hmm I see what you're saying, so `depositCounter` is a much more granular way of protecting against front-running. I just want to balance the gas cost against this added protection as our gas efficiency is a big advantage we have

contracts/SpokePool.sol
nicholaspai: it is yes

hardhat.config.ts
nicholaspai: oh wow didnt know about that, we should always have used that then?

test/MerkleLib.utils.ts
nicholaspai: Ah i see

hardhat.config.ts
mrice32: It's a new thing somewhere in the 0.8.X releases.

contracts/SpokePool.sol
mrice32: That's fair. I think we should look at gas costs pre and post this change to get a sense of how much it hurts us.

contracts/SpokePool.sol
```diff
@@ -854,8 +918,24 @@ abstract contract SpokePool is
...
+            if (payoutAdjustment != 0) {
+                require(int256(amountToSend) + payoutAdjustment >= 0, "payoutAdjustment too small");
+                amountToSend = uint256(int256(amountToSend) + payoutAdjustment);
+            }
         }
 
+        // Update fill counter.
+        _updateCountFromFill(
+            relayFills[relayHash],
+            relayFills[relayHash] + fillAmountPreFees,
```
nicholaspai, line 931: just curious, why `fillAmountPreFees` here?

nicholaspai (approved):
This LGTM but can you explain the following new params:
- local repayment: can repayment chain still be set to any chain? When is this true vs false? What's the purpose?
- payoutAdjustment: How will this be used by dataworker?

mrice32:
> This LGTM but can you explain the following new params:
> 
> * local repayment: can repayment chain still be set to any chain? When is this true vs false? What's the purpose?
> * payoutAdjustment: How will this be used by dataworker?

I added explanations in the description. Will also look into adding more comments about this in a follow up!

contracts/SpokePool.sol
mrice32: Because the method is responsible for removing the LP fee from that to compute the "liabilities" that this chain will have to pay out as a refund. Does that make sense?

contracts/SpokePool.sol
```diff
@@ -884,12 +964,40 @@ abstract contract SpokePool is
...
+        // If this is a slow fill or it's an initial 0-fill, do nothing, as these should not impact the count.
+        if (useContractFunds || endingFillAmount == 0) return;
+
+        // If this is the first fill and it's partial, assume the rest of the fill will be slow filled (refunded on this chain).
+        if (startingFillAmount == 0 && totalFillAmount - endingFillAmount > 0) {
+            fillCounter[token] += _computeAmountPostFees(totalFillAmount - endingFillAmount, realizedLPFeePct);
+        }
+
+        // If this is not the first fill, remove the partial fill that was previously assumed.
+        if (startingFillAmount != 0) {
```
pxrl, line 985: Maybe this is premature, but is there a gas optimisation by making these two if statements mutually exclusive? For example:

```
        if (startingFillAmount == 0) {
            if (totalFillAmount - endingFillAmount > 0) {
                fillCounter[token] += _computeAmountPostFees(totalFillAmount - endingFillAmount, realizedLPFeePct);
            }
        } else {
        // If this is not the first fill, remove the partial fill that was previously assumed (startingFillAmount > 0).
```

hardhat.config.ts
pxrl: There are a few instances where additional functions have been added to work around stack too deep issues. Would it be worth revisiting those now that this option is available?

PR #234, opened 2023-02-16
improve: move __gap to bottom of contracts

nicholaspai:
Makes it more obvious that __gap should always be at end of contract storage.


mrice32 (approved):
LGTM!

PR #237, opened 2023-02-17
improve(SpokePool): Remove Testable/Timer base contract

nicholaspai:
This base contract adds unneccessary gas costs to `getCurrentTime` calls and `setCurrentTime` is never used in production. This PR replaces those base contracts with a `getCurrentTime` implementation directly in the `SpokePool` contract.

Tests then use `MockSpokePool` which extends the `getCurrentTime` functionality and also adds a `setCurrentTime` call


deployments/deployments.json
```diff
@@ -1,7 +1,6 @@
 {
   "1": {
     "AcrossConfigStore": { "address": "0x3B03509645713718B78951126E0A6de6f10043f5", "blockNumber": 14717196 },
-    "AcrossMerkleDistributor": { "address": "0xE50b2cEAC4f60E840Ae513924033E753e2366487", "blockNumber": 15976846 },
```
pxrl, line 4: Could this be an unintentional change?

pxrl (approved):
Nice to clean this up. Not sure about the change to deployments.json, but otherwise LGTM :+1:

deployments/deployments.json
nicholaspai: Oops yeah

mrice32 (approved):
LGTM!

PR #239, opened 2023-02-20
feat: add message to across

mrice32:


contracts/SpokePool.sol
```diff
@@ -16,6 +16,15 @@ import "@openzeppelin/contracts-upgradeable/utils/AddressUpgradeable.sol";
 import "@openzeppelin/contracts-upgradeable/proxy/utils/UUPSUpgradeable.sol";
 import "@openzeppelin/contracts-upgradeable/security/ReentrancyGuardUpgradeable.sol";
 import "@openzeppelin/contracts/utils/math/SignedMath.sol";
+import "@openzeppelin/contracts/utils/Address.sol";
+
+interface AcrossMessageHandler {
```
nicholaspai, line 21: nit add comment here about whats expected of integrators before mergimg

contracts/SpokePool.sol
```diff
@@ -801,23 +867,25 @@ abstract contract SpokePool is
             // EIP-712 compliant hash struct: https://eips.ethereum.org/EIPS/eip-712#definition-of-hashstruct
             keccak256(
                 abi.encode(
-                    keccak256("UpdateRelayerFeeMessage(int64 newRelayerFeePct,uint32 depositId,uint256 originChainId)"),
-                    newRelayerFeePct,
+                    UPDATE_DEPOSIT_DETAILS_HASH,
```
nicholaspai, line 870: +1

contracts/SpokePool.sol
```diff
@@ -833,6 +901,22 @@ abstract contract SpokePool is
         require(isValid, "invalid signature");
     }
 
+    function _verifySlowFill(
```
nicholaspai, line 904: Moving it here just for readability?

contracts/SpokePool.sol
```diff
@@ -948,20 +1025,28 @@ abstract contract SpokePool is
...
                 IERC20Upgradeable(relayData.destinationToken).safeTransferFrom(
                     msg.sender,
                     relayData.recipient,
                     amountToSend
                 );
             else IERC20Upgradeable(relayData.destinationToken).safeTransfer(relayData.recipient, amountToSend);
         }
+
+        if (Address.isContract(relayData.recipient) && relayData.message.length > 0) {
+            AcrossMessageHandler(relayData.recipient).handleAcrossMessage(
```
nicholaspai, line 1044: So if this fails, then relay is not filled at all and depositor can update the recipient/data, right?

mrice32: Exactly!

contracts/SpokePool.sol
mrice32: Yep. Just to make the flow super simple and similar in each of the fill methods -- check some single-line requirements and then process the fill.

PR #243, opened 2023-02-22
feat: Document the SpokePool upgrade process

nicholaspai:
Explain high level how SpokePool upgrades will work

### DIFF
--- a/SpokePoolUpgrades.md
+++ b/SpokePoolUpgrades.md
@@ -26,3 +26,11 @@ If SpokePool is not an upgradeable proxy, then upgrading is more involved.
 
 3. Each of the above calldata's needs to be loaded into a HubPool call to `relaySpokePoolAdminFunction(uint256 chainId,bytes calldata)`.
 4. Add all of the transactions (i.e. each of the `relaySpokePoolAdminFunctions`) into an atomic transaction that is executed to finalize the "upgrade".
+
+## Addendum:
+
+### Adding new tokens to SpokePools
+
+This process has some overlap with upgrading spoke pools in that each new spoke pool address needs to be aware of other L2 chains and tokens that it can have deposit paths to/from.
+
+Adding new tokens is enabled mostly via the script [enableL1TokenAcrossEcosystem.ts](./tasks/enableL1TokenAcrossEcosystem.ts).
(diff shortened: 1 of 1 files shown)


SpokePoolUpgrades.md
```diff
@@ -0,0 +1,36 @@
...
+
+1. Deploy new SpokePool implementation contract on L2.
+2. Call `upgradeTo` on existing SpokePool proxy with new implementation address. The existing state should carry over
+   and the system should just work.
+
+## When upgrading without a proxy:
+
+If SpokePool is not an upgradeable proxy, then upgrading is more involved.
+
+1. Deploy new SpokePool contract.
```
mrice32, line 19: Should there be anything in here about dealing with assets that are in the spoke pool? Should they be transferred to the new spoke pool? Or back to the hub? You could send and execute a series of "relayer refunds" to transfer assets to the new spoke pool.

SpokePoolUpgrades.md
```diff
@@ -0,0 +1,36 @@
+# An Overview of how SpokePools would be upgraded
+
+## Before executing upgrade
+
+To be safe, make sure there is no pending bundle in the HubPool. To be extra safe, disable every single deposit route
+and only re-enable them after executing the upgrade. This is probably over-cautious when upgrading via proxy but
+definitely should be done when upgrading without a proxy.
```
mrice32, line 7: Are there any concerns about un-relayed deposits or outstanding partial fills?

mrice32 (approved):
This is awesome! I definitely think we should use this file as a place to build out the entire upgrade process. Great idea to commit it to the repo!

nicholaspai:
Now that we just went through process we should document:

Upgrading/Adding totally new spoke pools
- Deploy spoke pools
- Transfer Ethereum_SpokePool ownership to HubPool
- Make sure all bundles are executed and there is none pending.
- Transfer all remaining spoke pool balance to new spoke pool
- `setCrossChainContracts`
- `setEnableDeposits=true` for every single route
- `whitelistTokens` for Arbitrum
- `setTokenBridge` for Optimism

Upgrading proxy spoke pools
- Run upgradeTo script to produce bytecode to pass through owner of HubPool
- Need to `relaySpokePoolAdmin` with data that contains `upgradeTo` instructions for each chain's proxy

PR #244, opened 2023-02-22
improve: Remove excess brackets

nicholaspai:
We use the `viaIR` compiler flag so we don't need these brackets

### DIFF
--- a/contracts/SpokePool.sol
+++ b/contracts/SpokePool.sol
@@ -1105,37 +1105,31 @@ abstract contract SpokePool is
 
     // The following internal methods emit events with many params to overcome solidity stack too deep issues.
     function _emitFillRelay(RelayExecution memory relayExecution, uint256 fillAmountPreFees) internal {
-        RelayExecutionInfo memory relayExecutionInfo;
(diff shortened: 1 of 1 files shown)


pxrl (approved):

contracts/SpokePool.sol
```diff
@@ -1105,37 +1105,31 @@ abstract contract SpokePool is
...
-        {
-            relayExecutionInfo = RelayExecutionInfo({
-                relayerFeePct: relayExecution.updatedRelayerFeePct,
-                recipient: relayExecution.updatedRecipient,
-                message: relayExecution.updatedMessage,
-                isSlowRelay: relayExecution.slowFill,
-                payoutAdjustmentPct: relayExecution.payoutAdjustmentPct
-            });
-        }
+        RelayExecutionInfo memory relayExecutionInfo = RelayExecutionInfo({
```
mrice32, line 1108: Great call -- I figured these would be necessary even with viaIR, but I was wrong!

mrice32 (approved):
LGTM!

PR #245, opened 2023-02-23
fix(arbitrum-send-tokens-adapter): Use outboundTransferCustomRefund

nicholaspai:
This will refund relayer instead of aliased hub pool address on Arbitrum


mrice32 (approved):
LGTM!

pxrl (approved):

PR #250, opened 2023-03-15
improve(SpokePool): Optimize which params are `indexed`

nicholaspai:
This should be safe to merge post-audit since it only affects event params `indexed` property.

`FundsDeposited` and `FilledRelay` should index on `originChainId`/`destinationChainid` and `depositId` since this is all thats needed to uniquely identify a deposit.

This would make the relayer's job easier for certain functions like looking up a specific deposit for a fill, if it knows which `SpokePool` contract to look up events on, it can index on `depositId` and `destinationChainId` to efficiently find the deposit.

Conversely, if looking up a fill for a deposit, then the relayer would want to index on `originChainId` and `depositId` assuming it knows which `SpokePool` was the destination one and is the one it needs to query events on.

See [this](https://github.com/across-protocol/relayer-v2/blob/master/src/clients/SpokePoolClient.ts#L316) comment for example


mrice32 (approved):
Good call, makes sense!

PR #251, opened 2023-03-15
improve: [L-01] Incorrect or misleading documentation

nicholaspai:
## From Audit

Several instances of docstrings or comments in the codebase were found to be erroneous. In
particular:

- In the SpokePool contract's `_fillRelay` function, the comment on line 1003
appears unrelated to the if statement on the following line. Consider revising the
comment.
- The docstring for `pauseDeposits` in the SpokePool contract says it "pauses deposit
and fill functions", but pausing fill functions is performed by calling `pauseFills` , which
does not have a docstring.

Consider revising the `pauseDeposits` docstring and adding a docstring to `pauseFills`.

### DIFF
--- a/contracts/SpokePool.sol
+++ b/contracts/SpokePool.sol
@@ -262,8 +262,10 @@ abstract contract SpokePool is
     function _authorizeUpgrade(address newImplementation) internal override onlyAdmin {}
 
     /**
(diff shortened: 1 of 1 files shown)


mrice32 (approved):
LGTM!

PR #254, opened 2023-03-15
fix: [L04] updatedRelayerFeePct can be lower than expected

nicholaspai:
## From Audit

In the SpokePool contract, the speedUpDeposit function performs the following check on
the updatedRelayerFeePct variable:
require(updatedRelayerFeePct < 0.5e18, "invalid relayer fee");
Unlike the equivalent check in the deposit function, in this case the SignedMath.abs
function is not used, which allows the updatedRelayerFeePct value to reach or go below
the lower limit of -0.5e18 . An attempt by the relayer to fill this request will be rejected by the
_fillRelay function's fee check.
Consider using the SignedMath.abs function to ensure the updatedRelayerFeePct
argument provided to speedUpDeposit is within the expected limits.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -4,3 +4,5 @@ cache
 coverage*
 gasReporterOutput.json
 typechain
+artifacts-zk
+cache-zk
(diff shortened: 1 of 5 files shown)


mrice32 (approved):
LGTM!

PR #255, opened 2023-03-15
improve: [N01] Move all interfaces under `interfaces` directories

nicholaspai:
## From Audit

The Across Protocol V2 codebase contains the interfaces directory, but the
HubPoolInterface.sol and SpokePoolInterface.sol files do not reside in that
location. Instead, they are located in the parent directory.
Consider moving all non-external interface files to the interfaces directory, so that
interfaces can be more easily located.

## Developer's response

I appreciate this push as I've always found it inconsistent to have `interface` files not in `interface/` folders. I've added `interface` folders to each main contract directory


mrice32 (approved):
LGTM!

PR #256, opened 2023-03-15
improve: [N02] Multiple conditions in a single require statement

nicholaspai:
## From Audit

There is a require statement with multiple conditions in the handleTelepathy function of
the Succinct_SpokePool contract.
Consider isolating each condition in its own separate require statement with a corresponding
error message, to more easily identify the specific condition that failed.

## Developer's response

This change allowed me to make unit tests for `Succinct_SpokePool` more strict and check error messages


mrice32 (approved):
LGTM!

PR #257, opened 2023-03-15
improve: [N03] Redundant condition in if statement

nicholaspai:
## From Audit

In the SpokePool contract, the _updateCountFromFill function is responsible for
updating the fillCounter data when a fill takes place. No update is necessary in the
following three cases: when the action is either a slow fill or an initial zero-fill, or when a partial
fill for that request has already happened.
However, in the case of a zero-fill the execution flow will not reach this function, which makes
the condition endingFillAmount == 0 redundant.
Consider removing the redundant condition for increased clarity, readability, and gas savings.
For extra safety, consider adding a comment where the function returns if there is a zero-fill,
denoting that any updates to this part should take the function _updateCountFromFill into
account.

### DIFF
--- a/contracts/SpokePool.sol
+++ b/contracts/SpokePool.sol
@@ -1008,6 +1008,12 @@ abstract contract SpokePool is
         // This allows the caller to add in frontrunning protection for quote validity.
         require(fillCounter[relayData.destinationToken] <= relayExecution.maxCount, "Above max count");
 
+        // If the max tokens to send is 0, then the fillAmountPreFees will be 0 so we exit early.
+        // This also means that if the ending filled amount is 0, then we won't enter the
+        // function _updateCountFromFill. Therefore, _updateCountFromFill doesn't explicilty check
+        // that the ending fill amount is 0 (i.e. that the fillAmountPreFees is 0).
+        // If this if statement were to be removed, then _updateCountFromFill would need to be updated to
+        // check that the ending fill amount is 0.
         if (relayExecution.maxTokensToSend == 0) return 0;
 
         // Derive the amount of the relay filled if the caller wants to send exactly maxTokensToSend tokens to
@@ -1048,7 +1054,6 @@ abstract contract SpokePool is
         // Update fill counter.
         _updateCountFromFill(
             relayFills[relayExecution.relayHash],
-            relayFills[relayExecution.relayHash] + fillAmountPreFees,
             relayData.amount,
             relayData.realizedLpFeePct,
             relayData.destinationToken,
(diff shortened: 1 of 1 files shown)


contracts/SpokePool.sol
```diff
@@ -1000,13 +1008,19 @@ abstract contract SpokePool is
...
+        // If this if statement were to be removed, then _updateCountFromFill would need to be updated to
+        // check that the ending fill amount is 0.
         if (relayExecution.maxTokensToSend == 0) return 0;
 
         // Derive the amount of the relay filled if the caller wants to send exactly maxTokensToSend tokens to
         // the recipient. For example, if the user wants to send 10 tokens to the recipient, the full relay amount
         // is 100, and the fee %'s total 5%, then this computation would return ~10.5, meaning that to fill 10.5/100
         // of the full relay size, the caller would need to send 10 tokens to the user.
+        // This is equivalent to the amount to be sent by the relayer before fees have been taken out.
         fillAmountPreFees = _computeAmountPreFees(
```
nicholaspai, line 1024: @mrice32 Is it possible that `_computeAmountPreFees` could set `fillAmountPreFees` to 0if the `maxTokensToSend` is really small or the `(relayData.realizedLpFeePct + relayExecution.updatedRelayerFeePct)` is really large?

If so, then this PR would no longer catch the case where  `relayFills[relayExecution.relayHash] = 0 = relayFills[relayExecution.relayHash] + fillAmountPreFees`.

I think it should be fine but wanted to get eyes on this

mrice32: Great call-out. I'll need to think on this.

mrice32 (commented):
@nicholaspai I'm not sure we can remove this check.

contracts/SpokePool.sol
```diff
@@ -1025,6 +1025,12 @@ abstract contract SpokePool is
         // This allows the caller to add in frontrunning protection for quote validity.
         require(fillCounter[relayData.destinationToken] <= relayExecution.maxCount, "Above max count");
 
+        // If the max tokens to send is 0, then the fillAmountPreFees will be 0 so we exit early.
+        // This also means that if the ending filled amount is 0, then we won't enter the
+        // function _updateCountFromFill. Therefore, _updateCountFromFill doesn't explicilty check
+        // that the ending fill amount is 0 (i.e. that the fillAmountPreFees is 0).
```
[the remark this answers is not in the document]
nicholaspai: i think this is true, that's what I was hesitant about.

On other effects, I don't think so. We don't really check `fillAmountPreFees` anywhere else (which is the maxTokensToSend multiplied by some % which could be rounded down to 0)

nicholaspai:
https://github.com/across-protocol/contracts-v2/pull/257#discussion_r1137465413

https://github.com/across-protocol/contracts-v2/pull/257#discussion_r1142719023

PR #259, opened 2023-03-15
improve: [N05] Unecessary modulo operation

nicholaspai:
## From Audit

The setClaimed1D function in MerkleLib.sol performs a modulo 256 operation on
index to yield a result in the range [0, 255]. However, index is a uint8 variable, so the
operation index % 256 is equivalent to index .
Consider removing the redundant modulo operation.


mrice32 (approved):
LGTM! This one is interesting, uncaught from the initial audit.

PR #264, opened 2023-03-20
fix(SpokePool): fix handling of refunds on other chains and partial fills

nicholaspai:
In the current code, refunds taken on other chains create two problems:
1. They cannot be placed in time on the fill chain. When combined with slow fills, there is no consistent way to "place" them. Previously, it was assumed that they would be placed at the initial fill, but the a refund on another chain could cause this fill amount (liability) to decrease, meaning it retroactively changes. This means depositors can be frontrun by relayers without protection!
2. They cannot be placed in time on the chain where the refund is taken because there is no explicit transaction.

This fixes this problem by combining the following fixes:
1. Disallowing a partial fill to be refunded on a different chain.
2. Disallows 0-fills completely to avoid ambiguity around what a partial fill is.
3. Disabling fillCount incrementing for fills where the refund is taken on another chain.
4. Creating a requestRefund method and event so a relayer can slot in their transaction on the other chain. This increments the fill counter there, providing users on that chain adequate frontrunning protection.


contracts/SpokePool.sol
```diff
@@ -1062,6 +1165,7 @@ abstract contract SpokePool is
             }
         }
 
+        // TODO: Add in some limit to `fillAmount` to mitigate censor attacks on the fill counter.
```
mrice32, line 1168: Note: in other places, we typically do this by requiring that the `amount` value passed into any of the fill (or deposit) methods is below some value rather than putting a check in this function.

contracts/SpokePool.sol
```diff
@@ -668,6 +682,95 @@ abstract contract SpokePool is
...
+            relayFills[relayHash],
+            relayFills[relayHash] + fillAmount,
+            amount,
+            realizedLpFeePct,
+            destinationToken,
+            false // Slow fills should never match with a Refund. This should be enforced by off-chain bundle builders.
+        );
+
+        // TODO: Add in some limit to `fillAmount` to mitigate censor attacks on the fill counter.
+        emit Refund(
```
mrice32, line 757: What about RequestRefund?

mrice32 (approved):
I think this looks good!

contracts/SpokePool.sol
nicholaspai: Yeah i'm good with that

mrice32:
@nicholaspai I'm not ignoring this PR -- I just want to figure out how we're dealing with partial fills + refunds on other chains before coming back to it!

nicholaspai:
> @nicholaspai I'm not ignoring this PR -- I just want to figure out how we're dealing with partial fills + refunds on other chains before coming back to it!

oh same I was going to follow up after our discussions offline

contracts/SpokePool.sol
```diff
@@ -1035,6 +1094,10 @@ abstract contract SpokePool is
             relayExecution.maxTokensToSend,
             (relayData.realizedLpFeePct + relayExecution.updatedRelayerFeePct)
         );
+        // If fill amount minus fees, which is possible with small fill amounts and negative fees, then
+        // revert.
+        require(fillAmountPreFees > 0, "fill amount pre fees is 0");
```
mrice32, line 1099: With this, do you think we could/should drop the maxTokensToSend check above?

mrice32 (approved):
Looks great!

contracts/SpokePool.sol
```diff
@@ -668,6 +682,83 @@ abstract contract SpokePool is
...
+        _updateCountFromFill(
+            totalFilledAmount - fillAmount, // Starting fill amount for fill on destination chain
+            totalFilledAmount, // Ending fill amount
+            amount,
+            realizedLpFeePct,
+            destinationToken,
+            false // Slow fills should never match with a Refund. This should be enforced by off-chain bundle builders.
+        );
+
+        emit RequestRefund(
```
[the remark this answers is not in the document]
nicholaspai: @mrice32 what do you think of this?

contracts/SpokePool.sol
nicholaspai: I'm also curious if you think we should block deposits with amount == 0?

contracts/SpokePool.sol
```diff
@@ -667,6 +674,58 @@ abstract contract SpokePool is
...
+    function requestRefund(
+        address relayer,
+        address destinationToken,
+        uint256 amount,
+        uint256 originChainId,
+        int64 realizedLpFeePct,
+        uint32 depositId,
+        uint256 maxCount
+    ) external nonReentrant {
+        // TODO: We don't check for `amount > 0` but maybe we should? `fillRelay` reverts
```
james-a-morris, line 704: Nit: would it make sense to perform the check here as a sanity check?

contracts/SpokePool.sol
mrice32: It's a good question..... I guess is there any marginal value in blocking 0 if you can still send 1 wei deposits that might round to 0?

contracts/SpokePool.sol
nicholaspai: Yeah I added a check that `amount > 0`

contracts/SpokePool.sol
```diff
@@ -667,6 +675,60 @@ abstract contract SpokePool is
...
+     * This should be account receiving any refund stemming from this event.
+     * @param destinationToken This chain's destination token equivalent for original deposit destination token.
+     * @param amount Original deposit amount.
+     * @param originChainId Original origin chain ID.
+     * @param realizedLpFeePct Original realized LP fee %.
+     * @param depositId Original deposit ID.
+     * @param maxCount Max count to protect the refund recipient from frontrunning.
+     * @param fillBlock Block number of FilledRelay event on destinationChain we want to match with this Refund.
+     */
+    function requestRefund(
```
mrice32, line 699: Is there any concern about duplicates? Will duplicates break anything?

contracts/SpokePool.sol
```diff
@@ -667,6 +675,60 @@ abstract contract SpokePool is
...
+        address relayer,
+        address destinationToken,
+        uint256 amount,
+        uint256 originChainId,
+        int64 realizedLpFeePct,
+        uint32 depositId,
+        uint256 maxCount,
+        uint256 fillBlock
```
mrice32, line 707: Do you think fillBlock actually helps our validation? I'm not convinced it does if you can put it indefinitely far into the future.

Ideally, we'd like a way to know if it's valid or invalid the moment it's submitted or block it from being submitted if it's referencing something in the future... do you think there's a value we could use for that? Maybe block timestamp? And require that the timestamp is strictly in the past? Or even require that the timestamp is > some configurable number of seconds before the current block?

Thoughts?

contracts/SpokePool.sol
nicholaspai: Duplicates shouldn't break anything as long as bots have a determinsitic way to choose one, so say, pick the earliest match?

contracts/SpokePool.sol
nicholaspai: Why would setting a fillBlock far into the future work? I was thinking `fillBlock` would have to be exactly correct to be valid, not as a lower limit

contracts/SpokePool.sol
mrice32: I think the duplicate count should be emitted in the event so the bot can be absolutely sure that it doesn't double refund even if missing an event.

mrice32: I have added duplicate counting to this PR.

contracts/SpokePool.sol
mrice32: Related to the duplicate counting, I think there will probably need to be rules around how the timing of the fill block relates to the timing of the refund request, but the duplicate counting ensures that you can submit twice safely with the understanding that the first must be deemed invalid for the second to be valid.

mrice32 (approved):

contracts/SpokePool.sol
```diff
@@ -667,6 +681,81 @@ abstract contract SpokePool is
...
+     * This should be account receiving any refund stemming from this event.
+     * @param destinationToken This chain's destination token equivalent for original deposit destination token.
+     * @param amount Original deposit amount.
+     * @param originChainId Original origin chain ID.
+     * @param realizedLpFeePct Original realized LP fee %.
+     * @param depositId Original deposit ID.
+     * @param maxCount Max count to protect the refund recipient from frontrunning.
+     */
+    function requestRefund(
+        address relayer,
```
pxrl, line 706: What was the thinking around having the relayer address as an argument to this function, rather than just using msg.sender?

As is, it seems that this allows anyone to arbitrarily request a refund on behalf of another relayer. I.e. it could be probably be used to the advantage of a prospective depositor on the same chain by bundling a refund request for _another_ relayer before their own deposit in order to maximise their benefit/minimise their penalty.

Maybe this is actually desirable - i.e. it's another case of pushing the risk onto the relayer, and it provides others with additional ways of keeping the SpokePool at its target balance.

nicholaspai: Do you think we should require that only the `relayer` can call `requestRefund`? This is an interesting point and I think it has pros and cons so maybe we just remove the edge case entirely and `require(msg.sender == relayer)`

contracts/SpokePool.sol
```diff
@@ -1115,15 +1218,15 @@ abstract contract SpokePool is
 
     function _updateCountFromFill(
         uint256 startingFillAmount,
-        uint256 endingFillAmount,
+        bool localRepayment,
```
nicholaspai, line 1221: so a first partial fill that is not a local repayment (i.e. with repayment on another chain) should NOT increment the fill counter, right? This is a good catch, seems really important actually and would have been bad if we missed it. Did you add a test for it?

contracts/SpokePool.sol
```diff
@@ -1115,15 +1218,15 @@ abstract contract SpokePool is
...
         uint256 totalFillAmount,
         int64 realizedLPFeePct,
         address token,
         bool useContractFunds
     ) internal {
-        // If this is a slow fill, it's an initial 0-fill, or a partial fill has already happened, do nothing, as these
-        // should not impact the count.
-        if (useContractFunds || endingFillAmount == 0 || startingFillAmount > 0) return;
+        // If this is a slow fill, or a partial fill has already happened, do nothing, as these
```
nicholaspai, line 1227:
```suggestion
        // If this is a slow fill, a first partial fill with repayment on another chain, or a partial fill has already happened, do nothing, as these
```

test/SpokePool.Relay.ts
```diff
@@ -89,6 +105,40 @@ describe("SpokePool Relayer Logic", async function () {
...
+    // Changed consts.amountToRelay to relayData.amount to make it a full fill
+    await expect(
+      spokePool.connect(relayer).fillRelay(...getFillRelayParams(relayData, relayData.amount, consts.repaymentChainId))
+    )
+      .to.emit(spokePool, "FilledRelay")
+      .withArgs(
+        relayData.amount,
+        relayData.amount,
+        relayData.amount,
+        consts.repaymentChainId,
```
nicholaspai, line 128: Should you check in this test that `fillCounter` is not incremented because its not a `localRepayment`?

contracts/SpokePool.sol
pxrl: > Do you think we should require that only the `relayer` can call `requestRefund`?

The comments give me the impression that we assume that a relayer will only ever self-request, so I did get curious about the other scenarios. I was actually wondering about the case where a relayer might fill and hold onto its repayment request until the SpokePool on the repayment chain was above its target balance. In this case there could theoretically be many bundles between making a fill and requesting a refund...though if it's permissionless to request a refund on behalf of another relayer then that relayer would probably be more interested in requesting its refund as soon as the fill is mined (and the block number has a low probability of a re-org).

> This is an interesting point and I think it has pros and cons so maybe we just remove the edge case entirely and `require(msg.sender == relayer)`

That would definitely simplify the reasoning. In that case, would it just be better to remove the relayer argument to the requestRefund() function?

nicholaspai: Hm these are interesting examples. Requiring msg.sender == relayer would open up possibility that relayers waits a long time to request a refund. Although it would also be annoying to build a bot that goes through all Fills and requests refunds for them. Maybe we should build this sort of bot?

nicholaspai: Maybe we can enforce in the UMIP that the refund request is between -2 and +2 hours of the Fill based on a "reasonable" computation of the Fill time. This would be a bit hard to enforce in court but the spirit of the law would protect the system but opoortunistic relayers

test/SpokePool.Relay.ts
nicholaspai: I've done that^

contracts/SpokePool.sol
nicholaspai: @pxrl @james-a-morris @mrice32 as discussed off line there will be a rule added to the UMIP that the Refund event takes place no more than 20 minutes after the Fill event. This gives a margin of error for the destination and repayment chain to have timestamps that are off by 20 minutes (highly unlikely) and protects the dataworker a bit when deciding whether to validate a refund.

contracts/SpokePool.sol
nicholaspai: Had a good discussion with @mrice32 about this, its actually not a problem if the relayer waits until its profitable for them to execute the refund. That's exactly what the repayment chain SpokePool wants--to have relayers withdraw funds when its over balance and hold off when its under balance.

The change is implemented here not as a `require` statement, since we don't know the `relayer` address from the destination chain, so its up to the off chain validator to check this

contracts/SpokePool.sol
nicholaspai: as discussed offline, i've merged the above comment

nicholaspai:
> LGTM - I think we just have to resolve the linter issue

Which linter issue?

contracts/SpokePool.sol
mrice32: @nicholaspai 100% agree!

mrice32 (approved):

pxrl (approved):
Nice - LGTM!

contracts/SpokePool.sol
```diff
@@ -138,7 +143,16 @@ abstract contract SpokePool is
         bytes message,
         RelayExecutionInfo updatableRelayData
     );
-
+    event RefundRequest(
```
pxrl, line 146:
```suggestion
    event RefundRequested(
```

contracts/SpokePool.sol
```diff
@@ -667,6 +681,80 @@ abstract contract SpokePool is
         _emitFillRelay(relayExecution, fillAmountPreFees);
     }
 
+    /**
+     * @notice Caller signals to the system that they want a refund on this chain, which they set as the
+     * `repaymentChainId` on the original fillRelay() call on the `destinationChainId`. An observer should be
+     * be able to 1-to-1 match the emitted RefundRequest event with the FilledRelay event on the `destinationChainId`.
```
pxrl, line 687:
```suggestion
     * be able to 1-to-1 match the emitted RefundRequested event with the FilledRelay event on the `destinationChainId`.
```

test/SpokePool.Relay.ts
```diff
@@ -109,6 +162,105 @@ describe("SpokePool Relayer Logic", async function () {
...
+        destErc20.address,
+        relayData.amount,
+        consts.originChainId,
+        relayData.realizedLpFeePct,
+        relayData.depositId,
+        11, // Use any block number for this test.
+        maxFillCount
+      )
+    )
+      .to.emit(spokePool, "RefundRequest")
```
pxrl, line 188:
```suggestion
      .to.emit(spokePool, "RefundRequested")
```

test/SpokePool.Relay.ts
```diff
@@ -109,6 +162,105 @@ describe("SpokePool Relayer Logic", async function () {
...
```
pxrl, line 252:
```suggestion
      .to.emit(spokePool, "RefundRequested")
```

contracts/SpokePool.sol
```diff
@@ -667,6 +681,80 @@ abstract contract SpokePool is
...
+     * be able to 1-to-1 match the emitted RefundRequested event with the FilledRelay event on the `destinationChainId`.
+     * @dev This function could be used to artificially inflate the `fillCounter`, allowing the caller to "frontrun"
+     * and cancel pending fills in the mempool. This would in the worst case censor fills at the cost of the caller's
+     * gas costs. We don't view this as a major issue as the fill can be resubmitted and obtain the same incentive,
+     * since incentives are based on validated refunds and would ignore these censoring attempts. This is no
+     * different from calling `fillRelay` and setting msg.sender = recipient.
+     * @dev Caller needs to pass in `fillBlock` that the FilledRelay event was emitted on the `destinationChainId`.
+     * This is to make it hard to request a refund before a fill has been mined and to make lookups of the original
+     * fill as simple as possible.
+     * @param destinationToken This chain's destination token equivalent for original deposit destination token.
```
pxrl, line 696: Thought bubble: Naming this "destinationToken" might be confusing because we also have "destinationChainId" as used in deposits and fills, and "destination" means something different in these two contexts:
 - destinationChainId refers to the chain where the fill is/was made.
 - destinationToken refers to the chain where the refund will be paid out, and should be the address of the refund token on the refund chain, not the fill token on the destination chain. 

Since this address should refer to the address of the token to be refunded, wdyt about naming this "refundToken" ?

nicholaspai: this is a good point. I like `refundToken` or `repaymentToken`

nicholaspai: https://github.com/across-protocol/contracts-v2/pull/264/commits/7fda67d9188e57a6ce7bcb5bb6e1df49a64e3ac4

contracts/SpokePool.sol
```diff
@@ -675,6 +690,83 @@ abstract contract SpokePool is
...
+    ) external nonReentrant {
+        // Prevent unrealistic amounts from increasing fill counter too high.
+        require(amount <= MAX_TRANSFER_SIZE, "Amount too large");
+
+        // This allows the caller to add in frontrunning protection for quote validity.
+        require(fillCounter[refundToken] <= maxCount, "Above max count");
+
+        // Track duplicate refund requests.
+        bytes32 refundHash = keccak256(
+            abi.encode(msg.sender, refundToken, amount, originChainId, realizedLpFeePct, depositId, fillBlock)
```
mrice32, line 731: Do you need to add the destinationChainId to this abi.encode to ensure that refundHash doesn't match for different destinationChainIds?

mrice32 (approved):
Only one comment!

pxrl (approved):
No further comments from me.

james-a-morris:
> > LGTM - I think we just have to resolve the linter issue
> 
> Which linter issue?

This [one](https://github.com/across-protocol/contracts-v2/actions/runs/4536384152/jobs/7992996007), but it's been since resolved 🙂

contracts/SpokePool.sol
nicholaspai: Great call, done

PR #267, opened 2023-03-23
fix: fix potential underpayment in slow fill with negative LP fees

mrice32:
There was one major issue and a related minor issue discovered around how payoutAdjustmentPct was handled:
1.  In the case where there is a slow fill on a deposit with a negative LP fee, the maxSendAmount passed into the _fillRelay function could cause fillAmountPreFees to be less than the amount, which means it could be less than amountRemainingInRelay. In this case, two errors would occur: a) the slow fill would not complete the fill and b) the payoutAdjustmentPct would be ignored because it is inside the if statement comparing fillAmountPreFees to amountRemainingInRelay.
2. The payoutAdjustmentPct could be set arbitrarily high, potentially causing math overflows. Note: I don't think this would cause any mathematical _errors_ but it could certainly trigger cryptic overflows.

To fix this:
1. The maxSendAmount in the slow fill method was set to a number that is much 1e4 times larger than the max transfer size. Because the minimum LP fee is -50%, this should be more than sufficient to ensure that it is larger than any possible transfer.
2. The payoutAdjustmentPct logic was moved out from the if statement that it was previously inside.
3. A check was added to ensure that the final payout was still smaller than the maxSendAmount when payoutAdjustmentPct != 0.
4. The payoutAdjustmentPct was limited to being 1e2 or smaller. This means it can never cause overflows.

### DIFF
--- a/contracts/SpokePool.sol
+++ b/contracts/SpokePool.sol
@@ -95,6 +95,8 @@ abstract contract SpokePool is
     // 1e40 * 2e18 = 2e58 << 2^255 ~= 5e76
     uint256 public constant SLOW_FILL_MAX_TOKENS_TO_SEND = 1e40;
 
+    // Set max payout adjustment to something
+
     bytes32 public constant UPDATE_DEPOSIT_DETAILS_HASH =
         keccak256(
             "UpdateDepositDetails(uint32 depositId,uint256 originChainId,int64 updatedRelayerFeePct,address updatedRecipient,bytes updatedMessage)"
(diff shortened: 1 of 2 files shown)


contracts/SpokePool.sol
```diff
@@ -1049,16 +1055,18 @@ abstract contract SpokePool is
...
+            // If payoutAdjustmentPct is positive, then the recipient will receive more than the amount they
+            // were originally expecting. If it is negative, then the recipient will receive less.
+            // -1e18 is -100%. Because we cannot pay out negative values, that is the minimum.
+            require(relayExecution.payoutAdjustmentPct >= -1e18, "payoutAdjustmentPct too small");
 
-                // Note: since _computeAmountPostFees is typicaly intended for fees, the signage must be reversed.
-                amountToSend = _computeAmountPostFees(amountToSend, -relayExecution.payoutAdjustmentPct);
-            }
+            // Note: since _computeAmountPostFees is typicaly intended for fees, the signage must be reversed.
+            amountToSend = _computeAmountPostFees(amountToSend, -relayExecution.payoutAdjustmentPct);
```
mrice32, line 1069: @nicholaspai technically payoutAdjustmentPct could be as large as 2^255, which would make the amountToSend extremely large. Do you think we should add any upper bounds in the contract to payoutAdjustment?

Alternatively, do you think we should ensure that the final amountToSend is <= the maxSendAmount passed into this function. Feels a bit odd that the contract allows the payout to be larger than maxSendAmount in the slow relay case (even though slow relays are trusted).

Thoughts?

nicholaspai: i think adding an upper limit to prevent math issues is definitely a good idea. I can't see any reason why payoutAdjustmentPct would even be greater than say 1000% for example. Even that is stretching it

We have limits on `realizedLpFeePct` and `relayerFeePct` in _fillRelay so why not also have one for this pct to give us peace of mind?

nicholaspai (approved):

contracts/SpokePool.sol
```diff
@@ -89,6 +89,12 @@ abstract contract SpokePool is
 
     uint256 public constant MAX_TRANSFER_SIZE = 1e36;
 
+    // Note: this needs to be larger than the max transfer size to ensure that all slow fills are fillable, even if
+    // their fees are negative.
+    // It's important that it isn't too large, however, as it should be multipliable by ~2e18 without overflowing.
+    // 1e40 * 2e18 = 2e58 << 2^255 ~= 5e76
+    uint256 public constant SLOW_FILL_MAX_TOKENS_TO_SEND = 1e40;
```
nicholaspai, line 96: This is a good solution. I think we should limit payoutAdjustmentPct based on the diff between `SLOW_FILL_MAX_TOKENS_TO_SEND` and `MAX_TRANSFER_SIZE`

contracts/SpokePool.sol
pxrl: The new check on payoutAdjustmnetPct LGTM :+1:

contracts/SpokePool.sol
```diff
@@ -1049,16 +1057,26 @@ abstract contract SpokePool is
...
+            // This is a sanity check to ensure the payouts do not grow too large via some sort of issue in bundle
+            // construction.
+            require(relayExecution.payoutAdjustmentPct <= 100e18, "payoutAdjustmentPct too large");
+
+            // Note: since _computeAmountPostFees is typicaly intended for fees, the signage must be reversed.
+            amountToSend = _computeAmountPostFees(amountToSend, -relayExecution.payoutAdjustmentPct);
+
+            // Note: this error should never happen, since the maxTokensToSend is expected to be set much higher than
+            // the amount, but it is here as a sanity check.
+            require(amountToSend <= relayExecution.maxTokensToSend, "Somehow hit maxTokensToSend!");
```
nicholaspai, line 1079: +1

PR #268, opened 2023-03-30
improve(SpokePool): Apply rounding consistently to partial and full fills

nicholaspai:
Rounding errors are possible when computing `fillAmountPreFees` and we want to enforce that the error is considered in `amountToSend` both to partial and full fills. Currently `amountToSend` is only reset to its postFee amount for situations where `maxTokensToSend > amountRemainingInRelay` which is often the case for slow fills. We should apply the error consistently regardless of the type of fill.


mrice32 (approved):
LGTM!

PR #269, opened 2023-03-30
improve(docstrings): Follow up fixes

nicholaspai:
- In Polygon_Adapter.sol, the user parameter for the depositEtherFor function is documented using "@dev" instead of "@param user".
- In SpokePool.sol, within the RelayExecution struct's relay parameter description, "parametersin" should be "parameters in".
- In SpokePool.sol, the RelayExecution struct's docstring for parameter payoutAdjustmentPct lists it as "payoutAdjustment".

### DIFF
--- a/contracts/chain-adapters/Polygon_Adapter.sol
+++ b/contracts/chain-adapters/Polygon_Adapter.sol
@@ -13,7 +13,7 @@ import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
 interface IRootChainManager {
     /**
      * @notice Send msg.value of ETH to Polygon
-     * @dev Recipient of ETH on Polygon.
+     * @param user Recipient of ETH on Polygon.
      */
     function depositEtherFor(address user) external payable;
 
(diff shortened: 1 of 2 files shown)


contracts/chain-adapters/Polygon_Adapter.sol
```diff
@@ -13,7 +13,7 @@ import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
 interface IRootChainManager {
     /**
      * @notice Send msg.value of ETH to Polygon
-     * @dev Recipient of ETH on Polygon.
+     * @param user Recipient of ETH on Polygon.
```
pxrl, line 16: Where does "user" come from? I saw it in the request from OZ, but I can't find it described in natspec.

nicholaspai: its in the function input params

pxrl: Oof...sorry - I didn't bring my brain to this review :facepalm:.

pxrl (approved):
All good - sorry for the redundant question!

PR #276, opened 2023-04-17
fix: Execute fill with updated message and recipient

nicholaspai:
When sped up, updated message and recipient are not used

### DIFF
--- a/contracts/SpokePool.sol
+++ b/contracts/SpokePool.sol
@@ -1212,7 +1212,7 @@ abstract contract SpokePool is
         // net movement of funds.
         // Note: this is important because it means that relayers can intentionally self-relay in a capital efficient
         // way (no need to have funds on the destination).
-        if (msg.sender == relayData.recipient) return fillAmountPreFees;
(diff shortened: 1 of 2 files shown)


mrice32 (approved):
Very nice find!

contracts/SpokePool.sol
```diff
@@ -1223,24 +1223,28 @@ abstract contract SpokePool is
...
                     msg.sender,
-                    relayData.recipient,
+                    relayExecution.updatedRecipient,
+                    amountToSend
+                );
+            else
+                IERC20Upgradeable(relayData.destinationToken).safeTransfer(
+                    relayExecution.updatedRecipient,
                     amountToSend
                 );
```
james-a-morris, line 1240: Nit: should we add the `tsconfig` setting to require brackets after conditionals?

james-a-morris (commented):
Left a comment

contracts/SpokePool.sol
nicholaspai: do you know which config we should set for that? I marginally favor requiring brackets but not a big deal

PR #277, opened 2023-04-20
refactor: Consolidate new SpokePool/proxy deployment

pxrl:
This change aims to factor out quite a bit of duplicated logic.

For context, the "deployer" and the use of getNamedAccounts() was left out since it seems to be unused. hardhat.config.ts unconditionally calls getMnemonic(), which sources the private key from the environment, so the use of getNamedAccounts() may confuse and distract from the real address being used.


.github/workflows/publish.yml
```diff
@@ -23,16 +23,16 @@ jobs:
         id: release
         uses: manovotny/github-releases-for-automated-package-publishing-action@v2.0.1
 
-      # Perform installs, run tests, run a build step, etc. here, as needed.
-      - run: yarn
-
```
nicholaspai, line 28: Is this a merge conflict change?

scripts/setupArbitrumSpokePool.ts
```diff
@@ -1,6 +1,6 @@
 // @notice Logs ABI-encoded function data that can be relayed from HubPool to ArbitrumSpokePool to set it up.
 
-import { getContractFactory, ethers } from "../utils/utils";
+import { getContractFactory, ethers, hre } from "../utils/utils";
```
pxrl, line 3:
```suggestion
import { getContractFactory, ethers } from "../utils/utils";
```

utils/index.ts
```diff
@@ -1,3 +1,3 @@
 export * from "./utils";
 export * from "./MerkleTree";
-export * from "./constants";
+export * from "./constants";
```
pxrl, line 3:
```suggestion
export * from "./constants";

```

.github/workflows/publish.yml
pxrl: Not sure, but it's not present in the PR any more. I've handled all the merge conflicts, and now the actual intended changes are all that remain.

nicholaspai (approved):

PR #278, opened 2023-04-21
fix: Prevent partial fills with messages

nicholaspai:
Protects against edge cases where a `message` is sent multiple times to the `recipient`, who holds an unexpected amount of tokens.

### DIFF
--- a/contracts/SpokePool.sol
+++ b/contracts/SpokePool.sol
@@ -1193,6 +1193,13 @@ abstract contract SpokePool is
             "invalid repayment chain"
         );
 
+        // If this is a partial fill, require that the fill has no message. This prevents situations in which
+        // a message is sent multiple times for multiple partial fills.
+        require(
+            relayExecution.relay.amount == fillAmountPreFees || relayExecution.updatedMessage.length == 0,
+            "invalid partial fill message"
+        );
+
         // Update fill counter.
         _updateCountFromFill(
             relayFills[relayExecution.relayHash],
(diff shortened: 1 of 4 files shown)


contracts/SpokePool.sol
```diff
@@ -1193,6 +1193,13 @@ abstract contract SpokePool is
             "invalid repayment chain"
         );
 
+        // If this is a partial fill, require that the fill has no message. This prevents situations in which
+        // a message is sent multiple times for multiple partial fills.
+        require(
+            relayExecution.relay.amount == fillAmountPreFees || relayExecution.updatedMessage.length == 0,
+            "invalid partial fill message"
```
pxrl, line 1200:
```suggestion
            "invalid partial fill (message)"
```

pxrl (approved):
I think this is OK and protects against the scenario identified. It would be nice to be able to support partial fills in order to allow multiple relayers to contribute to a complete fill, but I can't see a simple and fair way to make that work at this point.

contracts/SpokePool.sol
nicholaspai: you think this improves the error message a bit? I'm not convinced but curious what you were thinking?

pxrl: > you think this improves the error message a bit? I'm not convinced but curious what you were thinking?

Yeah..even though I know exactly what the issue is, I have trouble parsing "invalid partial fill message" in my head. How about this?

```suggestion
            "invalid message on partial fill"
```

nicholaspai:
Closed in favor of more flexible PR: https://github.com/across-protocol/contracts-v2/pull/279

PR #282, opened 2023-04-26
feat: Release 2.2.0

nicholaspai:
note: 2.1.0 was accidentally released and has been deprecated

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@across-protocol/contracts-v2",
-  "version": "2.1.0",
+  "version": "2.2.0",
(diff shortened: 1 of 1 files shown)


index.ts
```diff
@@ -1,3 +1,3 @@
 export * from "./typechain";
 export * from "./src/DeploymentUtils";
-export * from "./utils/MerkleTree";
+export * from "./utils";
```
nicholaspai, line 3: This will allow `constants` to be exported so we can import from `sdk-v2` and re-export

james-a-morris (approved):

PR #284, opened 2023-04-28
feat: Remove direct "hardhat" imports, update types in deploy/ scripts

nicholaspai:
The HRE should only be constructed once, and for the deploy scripts we should try to load that existing object


james-a-morris (approved):
I think the additional typing is a good move. +1

PR #285, opened 2023-04-28
feat: add contract verifier

mrice32:
The idea behind this contract is primarily to stop users from depositing ETH into a SpokePool address that does not exist on a particular chain -- this can happen if the wallet and dapp have different ideas about which chain the transaction is destined for.


contracts/SpokePoolVerifier.sol
```diff
@@ -0,0 +1,37 @@
...
+import "@openzeppelin/contracts/utils/Address.sol";
+import "./interfaces/SpokePoolInterface.sol";
+
+/**
+ * @notice SpokePoolVerifier is a contract that verifies that the SpokePool exists on this chain before sending ETH to it.
+ */
+contract SpokePoolVerifier {
+    using Address for address;
+
+    function deposit(
```
james-a-morris, line 13: Should we add natspec docs to this?

james-a-morris (approved):
Only a small nit.

deploy/023_deploy_spoke_pool_verifier.ts
```diff
@@ -0,0 +1,19 @@
...
+  await deploy("SpokePoolVerifier", {
+    from: deployer,
+    log: true,
+    skipIfAlreadyDeployed: true,
+    deterministicDeployment: "0x1234", // Salt for the create2 call.
+  });
+};
+
+module.exports = func;
+func.tags = ["SpokePoolVerifier", "polygon", "mainnet", "optimism", "arbitrum", "zksync"];
```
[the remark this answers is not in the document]
mrice32: I think there is a native token!

contracts/SpokePoolVerifier.sol
mrice32: yes!

mrice32:
@james-a-morris @pxrl I will fix your comments in a follow-up!

PR #289, opened 2023-05-26
refactor(UBA): sideload HRE environment variable

james-a-morris:
This change essentially side-loads the HRE variables to prevent the need to directly interface with the Hardhat Runtime Environment.

### DIFF
--- a/deploy/003_deploy_optimism_spokepool.ts
+++ b/deploy/003_deploy_optimism_spokepool.ts
@@ -1,4 +1,4 @@
-import { deployNewProxy } from "../utils";
+import { deployNewProxy } from "../utils/utils.hre";
 import { DeployFunction } from "hardhat-deploy/types";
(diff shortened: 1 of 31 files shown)


mrice32 (approved):
LGTM! Great fix.

Two other concerns that I don't think we need to worry about now, but might come up in the future:
1. We're exporting test utils (things that use chai, etc). This is a bit weird, and probably means we have to include test libraries like chai or hardhat in our regular dependencies and not dev dependencies. That won't break anything, but it increases the size of the exported package and is pretty nonstandard for a package.
2. In some of these utilities, we're importing nodejs-specific packages that will break if imported into a FE. These include path, and fs. I'm not sure if this is problematic today, but might end up causing breakages in the future.

I think this is good to go as is and we can worry about those others in the future if/when we need to!

nicholaspai (approved):

PR #291, opened 2023-06-02
fix: Export utils.hre

pxrl:
relayer-v2 is dependent on these exports.

FWIW, I'm not 100% sure that this isn't just re-introducing the problem we previously had over in the scraper api. Would be good to discuss :+1:

### DIFF
--- a/utils/index.ts
+++ b/utils/index.ts
@@ -1,3 +1,4 @@
 export * from "./utils";
+export * from "./utils.hre";
 export * from "./MerkleTree";
 export * from "./constants";


james-a-morris (commented):
This was the change I was trying to avoid in the previous push so that we don't have the `HRE` exposed to the scraper. We may want to loop in @mrice32 on this one.

pxrl:
We may alternatively be able to make relayer-v2 directly dependent on hardhat, but it seems like the wrong approach to me because it has an inherent coupling to the contracts-v2 repo.

Otherwise, I'm not super knowledgeable about how this _should_ work in practice, so happy to take guidance.

utils/index.ts
```diff
@@ -1,3 +1,4 @@
 export * from "./utils";
+export * from "./utils.hre";
```
mrice32, line 2: Exporting this directly into the index.ts means that _every importer_ will have to have a hardhat config and will automatically start hardhat in process. This is a bad pattern for a package. Hardhat shouldn't be used this way.

What in the relayer repo depends on this? This file exports deployNewProxy and the hre object as far as I can see, and I can't find deployNewProxy anywhere in the relayer repo. Are we sure the relayer-v2 repo isn't just importing `hre` from here instead of just importing it on its own: `import hre from "hardhat";`?

Side note: IMO we should also just avoid re-exporting hre in the future and just pull it in directly via importing hardhat wherever needed. Re-exports are an antipattern unless the dependency needs to be modified or initialized. I think that's how we got here, and I think we just did that for "cleanliness" so we could condense imports (which doesn't seem worthwhile in hindsight.

pxrl: Understood - makes sense. I was missing some of the context on how hardhat is intended to be used and had assumed to re-exported from contracts-v2 for a reason.

It works fine to import hardhat directly in the places that need it within relayer-v2.

pxrl:
Abandoned in light of @mrice32's input. I've made an alternative proposal in relayer-v2: https://github.com/across-protocol/relayer-v2/pull/710.

PR #292, opened 2023-06-06
feat(UBA): Support UBA pool rebalance leaves in utils

pxrl:
Post-UBA, runningBalances will be 2x the length of the other arrays. To avoid having to teach contacts-v2 about ConfigStore versioning, just loosen the constraints to permit pre- and post-UBA formats.

Fixes ACX-1136

### DIFF
--- a/test/MerkleLib.utils.ts
+++ b/test/MerkleLib.utils.ts
@@ -55,11 +55,14 @@ export function buildRelayerRefundLeaves(
 
 export async function buildPoolRebalanceLeafTree(poolRebalanceLeaves: PoolRebalanceLeaf[]) {
   for (let i = 0; i < poolRebalanceLeaves.length; i++) {
-    // The 4 provided parallel arrays must be of equal length.
+    // 3 of the provided parallel arrays must be of equal length.
     expect(poolRebalanceLeaves[i].l1Tokens.length)
       .to.equal(poolRebalanceLeaves[i].bundleLpFees.length)
-      .to.equal(poolRebalanceLeaves[i].netSendAmounts.length)
-      .to.equal(poolRebalanceLeaves[i].runningBalances.length);
+      .to.equal(poolRebalanceLeaves[i].netSendAmounts.length);
+
+    // runningBalances must be 1x or 2x as long as the other arrays (pre/post UBA).
+    expect([1, 2].includes(poolRebalanceLeaves[i].runningBalances.length / poolRebalanceLeaves[i].l1Tokens.length)).to
+      .be.true;
   }
 
   const paramType = await getParamType("MerkleLibTest", "verifyPoolRebalance", "rebalance");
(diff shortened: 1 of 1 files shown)


pxrl:
nb. I unfortunately can't test this locally at the moment because contracts-v2 is horribly broken on my development server. Typechain is throwing a lot of errors :face_with_head_bandage:

pxrl:
Returned to draft because it's failing test.

test/MerkleLib.utils.ts
```diff
@@ -55,11 +55,14 @@ export function buildRelayerRefundLeaves(
...
-    // The 4 provided parallel arrays must be of equal length.
+    // 3 of the provided parallel arrays must be of equal length.
     expect(poolRebalanceLeaves[i].l1Tokens.length)
       .to.equal(poolRebalanceLeaves[i].bundleLpFees.length)
-      .to.equal(poolRebalanceLeaves[i].netSendAmounts.length)
-      .to.equal(poolRebalanceLeaves[i].runningBalances.length);
+      .to.equal(poolRebalanceLeaves[i].netSendAmounts.length);
+
+    // runningBalances must be 1x or 2x as long as the other arrays (pre/post UBA).
+    expect([1, 2].includes(poolRebalanceLeaves[i].runningBalances.length / poolRebalanceLeaves[i].l1Tokens.length)).to
```
nicholaspai, line 64: What if L1 tokens is 0 length? Should we be instead multiplying the length of L1 tokens by 1 or 2 and checking equivalence so we don't get a divide by 0?

nicholaspai (approved):
https://github.com/across-protocol/contracts-v2/pull/292#discussion_r1219492898

One comment

test/MerkleLib.utils.ts
```diff
@@ -54,12 +54,16 @@ export function buildRelayerRefundLeaves(
...
-    // The 4 provided parallel arrays must be of equal length.
-    expect(poolRebalanceLeaves[i].l1Tokens.length)
-      .to.equal(poolRebalanceLeaves[i].bundleLpFees.length)
-      .to.equal(poolRebalanceLeaves[i].netSendAmounts.length)
-      .to.equal(poolRebalanceLeaves[i].runningBalances.length);
+  for (const leaf of poolRebalanceLeaves) {
+    const { l1Tokens, bundleLpFees, netSendAmounts, runningBalances } = leaf;
+
+    // l1Tokens, bundleLpFees and netSendAmounts must always be of equal length.
+    expect(l1Tokens.length).to.equal(bundleLpFees.length).to.equal(netSendAmounts.length);
```
nicholaspai, line 61: didn't know you could do this. Have you tested that this double `.equal()` actually works and doesn't return early after the first `.equal()` returns true?

nicholaspai (approved):

test/MerkleLib.utils.ts
pxrl: > didn't know you could do this. Have you tested that this double `.equal()` actually works and doesn't return early after the first `.equal()` returns true?

This is basically the same logic as was there before, so I assumed it worked. I'll verify it now.

pxrl: @nicholaspai Looks like it fails as expected, with the following test:

```
 expect(0).to.equal(0).to.equal(1);
```

2023-06-06-160923_1791x504_scrot

PR #294, opened 2023-06-07
fix(OVM_SpokePool): Support Bedrock ETH withdrawals

nicholaspai:
OVM Bedrock has changed its withdrawal pattern https://community.optimism.io/docs/developers/bedrock/how-is-bedrock-different/#two-phase-withdrawals

as well as its contracts https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts-bedrock/contracts/L2/L2StandardBridge.sol#L179

### DIFF
--- a/contracts/Ovm_SpokePool.sol
+++ b/contracts/Ovm_SpokePool.sol
@@ -147,7 +147,6 @@ contract Ovm_SpokePool is SpokePool {
         SpokePoolInterface.RelayerRefundLeaf memory relayerRefundLeaf,
         bytes32[] memory proof
     ) public override(SpokePool) nonReentrant {
-        // Leave any ETH we need to send back to L1 unwrapped.
         if (relayerRefundLeaf.l2TokenAddress == address(wrappedNativeToken)) _depositEthToWeth();
 
         _executeRelayerRefundLeaf(rootBundleId, relayerRefundLeaf, proof);


contracts/Ovm_SpokePool.sol
```diff
@@ -137,7 +147,9 @@ contract Ovm_SpokePool is SpokePool {
         SpokePoolInterface.RelayerRefundLeaf memory relayerRefundLeaf,
         bytes32[] memory proof
     ) public override(SpokePool) nonReentrant {
-        if (relayerRefundLeaf.l2TokenAddress == address(wrappedNativeToken)) _depositEthToWeth();
+        // Leave any ETH we need to send back to L1 unwrapped.
+        if (relayerRefundLeaf.l2TokenAddress == address(wrappedNativeToken))
```
mrice32, line 151: What if there isn't enough unwrapped ETH to cover it? For instance, if we have 10 deposits of 1 ETH, they would be auto-converted to WETH. If the amountToReturn were > 0, this would fail, right? We would need some way to _unwrap_ the WETH, right?

Should we set this up so that it wraps _or_ unwraps to achieve a specific amount of ETH that we want (which would _usually_ be 0)?

mrice32 (commented):
Looks good! Just one comment

contracts/Ovm_SpokePool.sol
nicholaspai: Yeah I'm gonna revert this change. The issue I was trying to address is that in the Optimism_SpokePool unit test, we don't actually unwrap any WETH because WETH is a faked contract. Therefore the payable call to `withdrawTo` reverts in the tests. I think one way around this is to make WETH a real contract

mrice32 (approved):
LGTM!

contracts/Ovm_SpokePool.sol
mrice32: Gotcha, makes sense!

PR #300, opened 2023-06-14
Create PermissionSplitter

nicholaspai:
This contract is a proxy that owns another contract and can split permissions by role based on what method is being called. All methods are always callable by the owner of the proxy.


contracts/PermissionSplitterProxy.sol
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    function _init(address _target) internal virtual {
+        _grantRole(DEFAULT_ADMIN_ROLE, msg.sender);
+        __setTarget(_target);
+    }
+
+    function _isAllowedToCall(address caller, bytes calldata callData) internal view virtual returns (bool) {
+        bytes4 selector;
+        if (callData.length < 4) {
```
james-a-morris, line 45: Nit: would we want to store this in a constant? The 4 seems like a magic number

james-a-morris (commented):
Left a few comments related to documentation and a small nit. LGTM

mrice32 (approved):

PR #304, opened 2023-06-20
feat: add POOL to constants

dohaki:
Based on addresses listed here https://dev.pooltogether.com/protocol/deployments/governance


james-a-morris (approved):
LGTM

PR #306, opened 2023-06-22
feat(erc1155): harden script to only mint to recipients with balance 0

dohaki:
Adds check to mint script to skip addresses that already have received a token

### DIFF
--- a/scripts/mintERC1155.ts
+++ b/scripts/mintERC1155.ts
@@ -27,18 +27,31 @@ async function main() {
 
   for (let i = skip; i < validRecipients.length; i = i + RECIPIENTS_CHUNK_SIZE) {
     const recipientsChunk = validRecipients.slice(i, i + RECIPIENTS_CHUNK_SIZE);
-    const airdropTx = await erc1155.airdrop(tokenId, recipientsChunk, 1);
-    console.log(
-      `Minting token with id ${tokenId} to ${recipientsChunk.length} recipients in index range ${i} - ${
-        i + RECIPIENTS_CHUNK_SIZE - 1
-      }...`
-    );
-    console.log("Tx hash:", airdropTx.hash);
-    await airdropTx.wait();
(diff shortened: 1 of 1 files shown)


scripts/mintERC1155.ts
```diff
@@ -27,18 +27,31 @@ async function main() {
...
+      console.log(`Skipping ${alreadyMintedRecipients.length} already minted recipients...`);
+    }
+
+    if (recipientsToMint.length === 0) {
+      console.log(`No recipients to mint for. Skipping chunk...`);
+      continue;
+    }
+
+    const airdropTx = await erc1155.airdrop(tokenId, recipientsToMint, 1);
+    console.log(`Minting token with id ${tokenId} to ${recipientsToMint.length} recipients...`);
```
james-a-morris, line 51: Nit: should we place this before we run the airdrop tx? If it fails to perform the airdrop we would have this as the last log

james-a-morris (approved):
Left a single nit. this is great code

scripts/mintERC1155.ts
dohaki: Good catch. Changed

scripts/mintERC1155.ts
```diff
@@ -27,18 +27,31 @@ async function main() {
...
+    if (recipientsToMint.length === 0) {
+      console.log(`No recipients to mint for. Skipping chunk...`);
+      continue;
+    }
+
+    console.log(`Minting token with id ${tokenId} to ${recipientsToMint.length} recipients...`);
+    const airdropTx = await erc1155.airdrop(tokenId, recipientsToMint, 1);
+    console.log("Tx hash:", airdropTx.hash);
+    await airdropTx.wait();
+    console.log(`Successfully minted token to ${recipientsToMint.length} recipients in chunk`);
```
pxrl, line 54: Consider dropping the txn hash here?

PR #308, opened 2023-06-29
improve(SpokePool): Remove .transfer()

nicholaspai:
Uses safe, audited library function from openzeppelin instead of sending ETH via `.transfer()`, which is [not recommended](https://twitter.com/zksync/status/1644139364270878720) for zksync which has a very different gas fee model.

The library code can be found [here](https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/master/contracts/utils/AddressUpgradeable.sol#L41)

## Side Effects
- The updated `openzeppelin/contracts-upgradeable` import has added a `delegatecall` within a function of the library contract `AddressUpgradeable`.
- Since we use the above contract in the `SpokePool` proxy, we either need to unilaterally allow unsafe `delegatecall` or copy the `AddressUpgradeable` into the repo and add a specific check to that contract to allow that `delegatecall`.
- If someone knows a better way to handle these cases (we don't even use the function in `AddressUpgradeable` that makes the `delegatecall`!) please follow up with a PR!


mrice32 (approved):
LGTM! Not sure I 100% agree on removing the warning (the delegatecall vulnerability in proxy contracts is sort of non intuitive), but up to you!

PR #313, opened 2023-07-12
fix: [M01] add comment documenting that this contract will not protect zkSync

mrice32:


nicholaspai (approved):

contracts/SpokePoolVerifier.sol
```diff
@@ -6,6 +6,9 @@ import "./interfaces/SpokePoolInterface.sol";
 
 /**
  * @notice SpokePoolVerifier is a contract that verifies that the SpokePool exists on this chain before sending ETH to it.
+ * @dev This contract must be deployed via Create2 to the same address on all chains. That way, an errant transaction sent
+ * to the wrong chain will be blocked by this contract rather than hitting a dead address. This means that this contract
+ * will not work to protect chains, like zkSync, where Create2 address derivations don't match other chains.
```
nicholaspai, line 11:
```suggestion
 * will not work to protect chains, like zkSync, where Create2 address derivations don't match other chains.
 * Source: https://era.zksync.io/docs/reference/architecture/differences-with-ethereum.html#create-create2
```

james-a-morris (commented):
From a CYA perspective this makes sense to me. I'm not sure I'm in a position to approve/disapprove this PR.

PR #316, opened 2023-07-17
chore: Add Sepolia testnet details

pxrl:
These contract addresses are needed by relayer-v2 for testing.


hardhat.config.ts
```diff
@@ -162,12 +168,12 @@ const config: HardhatUserConfig = {
       kovan: process.env.ETHERSCAN_API_KEY!,
       rinkeby: process.env.ETHERSCAN_API_KEY!,
       goerli: process.env.ETHERSCAN_API_KEY!,
+      // sepolia: process.env.ETHERSCAN_API_KEY!, hardhat-etherscan is unhappy with custom definitions.
       optimisticEthereum: process.env.OPTIMISM_ETHERSCAN_API_KEY!,
       arbitrumOne: process.env.ARBITRUM_ETHERSCAN_API_KEY!,
       polygon: process.env.POLYGON_ETHERSCAN_API_KEY!,
       polygonMumbai: process.env.POLYGON_ETHERSCAN_API_KEY!,
     },
-    customChains: [],
```
pxrl, line 170: I had to remove the definition for `sepolia` because we inherit a very old version of `hardhat-etherscan` from `@uma/common`, and it doesn't support `customChains`.

I can't see any obvious way of preventing UMA's hardhat dependencies from executing, and intuitively it seems wrong that we inherit this requirement to _our_ environment from an external package. I've commented `sepolia` out for now, but it'd be good to address the underlying issue.

```
yarn run v1.22.19                                                                                               
$ hardhat compile && tsc && rsync -a --include '*/' --include '*.d.ts' --exclude '*' ./typechain ./dist/        
Error in plugin @nomiclabs/hardhat-etherscan: Etherscan API token "sepolia" is for an unsupported network                                                                                                                        
                                                                                                                
Learn more at https://hardhat.org/plugins/nomiclabs-hardhat-etherscan.html#multiple-api-keys-and-alternative-block-explorers                                                                                                     
                                                                                                                                                                                                                                 
NomicLabsHardhatPluginError: Etherscan API token "sepolia" is for an unsupported network                                                                                                                                         
                                                                                                                                                                                                                                 
Learn more at https://hardhat.org/plugins/nomiclabs-hardhat-etherscan.html#multiple-api-keys-and-alternative-block-explorers
```

hardhat.config.ts
```diff
@@ -120,7 +120,7 @@ const config: HardhatUserConfig = {
       url: "https://goerli-rollup.arbitrum.io/rpc",
       saveDeployments: true,
       accounts: { mnemonic },
-      companionNetworks: { l1: "goerli" },
+      companionNetworks: { l1: "sepolia" },
```
nicholaspai, line 123: Are you sure you want to commit this change? I don't think `arbitrum-goerli` can actually communicate with `sepolia`, only `goerli` system contracts.
```suggestion
      companionNetworks: { l1: "goerli" },
```

hardhat.config.ts
nicholaspai, line 170: I need this for `ZkSync` did this auto remove?

pxrl: Ah; no. I removed it because we can't actually use it, per the issue above about hardhat-etherscan. I'll restore it.

nicholaspai: We actually should not being using `hardhat-etherscan` anymore to verify. The latest package we should be using is `hardhat-verify`: https://github.com/across-protocol/contracts-v2/blob/master/package.json#L53

hardhat.config.ts
pxrl: Yeah - this can remain as goerli. Communication obviously isn't possible, but I didn't want anyone to accidentally refer to the wrong HubPool instance when looking at the new deployments on Etherscan.

hardhat.config.ts
pxrl: Correct - we're _not_ actually intending to use `hardhat-etherscan`. We seem to be inheiriting it via `@uma/common`, and I haven't found any way to inhibit that.

hardhat.config.ts
nicholaspai: ah yeah but i would rather keep this committed as linked to `goerli`

nicholaspai (approved):

PR #322, opened 2023-07-25
fix: [N-01] Constants not using UPPER_CASE format

pxrl:


contracts/chain-adapters/ZkSync_Adapter.sol
```diff
@@ -72,17 +72,17 @@ contract ZkSync_Adapter is AdapterInterface {
...
     // Note, this value can change and will require an updated adapter.
-    uint256 public constant l1GasToL2GasPerPubDataLimit = 800;
+    uint256 public constant L1_GAS_TO_L2_GAS_PER_PUB_DATA_LIMIT = 800;
 
     // This address receives any remaining fee after an L1 to L2 transaction completes.
-    // If refund recipient = address(0) then L2 msg.sender is used, unelss msg.sender is a contract then its address
+    // If refund recipient = address(0) then L2 msg.sender is used, unless msg.sender is a contract then its address
     // gets aliased.
-    address public constant l2RefundAddress = 0x428AB2BA90Eba0a4Be7aF34C9Ac451ab061AC010;
+    address public constant L2_REFUND_ADDRESS = 0x428AB2BA90Eba0a4Be7aF34C9Ac451ab061AC010;
```
pxrl, line 85: nb. The change here is also related to N-03, so it may need to change again.

mrice32 (approved):
Everything looks great! I would probably just change the interfaces, too, as they suggest, even if it does feel a bit weird. The linter also picks up on these, so I think the suggestion is technically correct.

mrice32:
> Everything looks great! I would probably just change the interfaces, too, as they suggest, even if it does feel a bit weird. The linter also picks up on these, so I think the suggestion is technically correct.

@pxrl I checked the other contracts, and I agree the way you did this was consistent. Due to the change in #326, I dropped the refund address change. So I think this should be ready to go.

PR #326, opened 2023-07-27
fix: [N-03] Lack of Documentation (removed hardcoded refund address)

mrice32:
This PR removes the hardcoded refund address for zkSync, replaces it with an immutable value, and adds a default address in the deployment script that is the most common dataworker address.

### DIFF
--- a/contracts/chain-adapters/ZkSync_Adapter.sol
+++ b/contracts/chain-adapters/ZkSync_Adapter.sol
@@ -82,7 +82,7 @@ contract ZkSync_Adapter is AdapterInterface {
     // This address receives any remaining fee after an L1 to L2 transaction completes.
     // If refund recipient = address(0) then L2 msg.sender is used, unelss msg.sender is a contract then its address
     // gets aliased.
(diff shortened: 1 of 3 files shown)


pxrl (commented):
Nice improvement to make it configurable via the constructor.

I've proposed a PR to update the tests so they verify that the output l2RefundAddress matches the what was supplied in the constructor. I couldn't propose it directly on this PR because it's not  the scope of the diff. See https://github.com/across-protocol/contracts-v2/pull/329

contracts/chain-adapters/ZkSync_Adapter.sol
```diff
@@ -101,9 +101,11 @@ contract ZkSync_Adapter is AdapterInterface {
     /**
      * @notice Constructs new Adapter.
      * @param _l1Weth WETH address on L1.
+     * @param _l2RefundAddress address that recieves excess gas refunds on L2.
      */
-    constructor(WETH9Interface _l1Weth) {
+    constructor(WETH9Interface _l1Weth, address _l2RefundAddress) {
```
james-a-morris, line 106: Good call on making this dynamic

james-a-morris (approved):

PR #328, opened 2023-07-28
fix: [M-01] zkSync limits

mrice32:
This PR adds a proxy contract that could be used to bypass zk sync deposit limits by address. To allow for this proxy to be swapped out, the ETH sending address was separated from the address used to send contract calls, so one could be swapped out without affecting the other.

### DIFF
--- a/contracts/chain-adapters/ZkSync_Adapter.sol
+++ b/contracts/chain-adapters/ZkSync_Adapter.sol
@@ -53,6 +53,63 @@ interface ZkBridgeLike {
     ) external payable returns (bytes32 txHash);
 }
 
+// Note: this contract just forwards the calls from the HubPool to ZkSync to avoid limits.
+// A modified ZKSync_Adapter should be deployed with this address swapped in for all zkSync addresses.
+contract LimitBypassProxy is ZkSyncInterface, ZkBridgeLike {
+    using SafeERC20 for IERC20;
+    ZkSyncInterface public constant zkSync = ZkSyncInterface(0x32400084C286CF3E17e7B677ea9583e60a000324);
+    ZkBridgeLike public constant zkErc20Bridge = ZkBridgeLike(0x57891966931Eb4Bb6FB81430E6cE0A03AAbDe063);
+
+    function l2TransactionBaseCost(
+        uint256 _l1GasPrice,
+        uint256 _l2GasLimit,
+        uint256 _l2GasPerPubdataByteLimit
+    ) external view returns (uint256) {
+        return zkSync.l2TransactionBaseCost(_l1GasPrice, _l2GasLimit, _l2GasPerPubdataByteLimit);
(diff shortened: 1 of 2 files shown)


nicholaspai (commented):
This seems like the most straightforward implementation of this and I like that the default case doesn't use the proxy. In the case where we do get limited, how would the remedy work. 

Would we deploy the proxy and try again? and if we get limited again, then we redeploy yet again? I'm kind of skeptical that the HubPool would ever get limited and we'd subsequently be able to bypass the limit with another address. 

It seems like if zksync truly wanted to censor us they could keep updating their limits for our new proxy addresses so the only way around it in that case would be to split the sends up by multiple callers... but if it ever comes to that we probably have bigger problems

contracts/chain-adapters/ZkSync_Adapter.sol
```diff
@@ -53,6 +53,63 @@ interface ZkBridgeLike {
     ) external payable returns (bytes32 txHash);
 }
 
+// Note: this contract just forwards the calls from the HubPool to ZkSync to avoid limits.
+// A modified ZKSync_Adapter should be deployed with this address swapped in for all zkSync addresses.
+contract LimitBypassProxy is ZkSyncInterface, ZkBridgeLike {
+    using SafeERC20 for IERC20;
+    ZkSyncInterface public constant zkSync = ZkSyncInterface(0x32400084C286CF3E17e7B677ea9583e60a000324);
```
james-a-morris, line 60: `Solhint` is indicating that this should be in uppercase snake case `ZKSYNC`. Should we modify the solhint?

james-a-morris (approved):
LGTM - there are multiple instances of SolHint issues and I wonder if we should modify the linter config or modify the code to conform to it

contracts/chain-adapters/ZkSync_Adapter.sol
mrice32: @james-a-morris @pxrl Since we generally don't snake case contract interfaces, I removed the solhint warning for now to appease the linter.

mrice32:
> This seems like the most straightforward implementation of this and I like that the default case doesn't use the proxy. In the case where we do get limited, how would the remedy work.
> 
> Would we deploy the proxy and try again? and if we get limited again, then we redeploy yet again? I'm kind of skeptical that the HubPool would ever get limited and we'd subsequently be able to bypass the limit with another address.
> 
> It seems like if zksync truly wanted to censor us they could keep updating their limits for our new proxy addresses so the only way around it in that case would be to split the sends up by multiple callers... but if it ever comes to that we probably have bigger problems

Agreed, we would basically be deploying a new adapter with a new proxy address over and over. An auto-deploy would be a little better, but would also be more complex, so I wanted to avoid doing something overly complicated for something that is unlikely to be needed.

PR #331, opened 2023-07-28
style: SNAKE_CASE for constants in Arbitrum_Adapter

pxrl:
Prompted by OZ zkSync audit finding [N-01]. No functional change, but it does appease SOLHINT.


mrice32 (commented):
Personally, I think it's better to avoid unnecessary semantic changes in contracts that are already deployed in the interest of it more closely matching what's on-chain. If we were to make substantial changes to this contract and have it re-audited, that might be a better time to make this sort of change.

PR #332, opened 2023-07-28
fix: [N-01] Adjust zkSync interfaces

pxrl:
This change should move the zkSync contract and deployment scripts closer to the existing patterns established in this repo. As a side benefit, we no longer see SNAKE_CASE complaints about this from SOLHINT.


pxrl:
It just occurred to me that I'd overlooked this comment:

```
    // Hardcode the following ZkSync system contract addresses to save gas on construction. This adapter can be
    // redeployed in the event that the following addresses change.
```

Closing this accordingly...

PR #340, opened 2023-08-07
fix: zkSync SpokePool deployment script

pxrl:
This fix uses zkSync-specific packages, which are necessary for contract deployment on zkSync. The zkEthBridge configurable is also removed, since it is no longer needed for SpokePool deployment on zkSync. Note also that there is explicit allowance for delegatecall in this script. It currently incorrectly resolves the upstream AddressUpgradeable contract, rather than the local AddressLibUpgradeable. See the following commit for more details: 02cff2cc68e76f09ccbb735612f771523adf911b

The relevant zkSync documentation can be found here:
https://era.zksync.io/docs/tools/hardhat/hardhat-zksync-deploy.html
https://era.zksync.io/docs/tools/hardhat/hardhat-zksync-upgradable.html


deploy/consts.ts
```diff
@@ -93,9 +93,12 @@ export const L2_ADDRESS_MAP: { [key: number]: { [contractName: string]: string }
     fxChild: "0xCf73231F28B7331BBe3124B907840A94851f9f11",
   },
   280: {
-    zkErc20Bridge: "0x92131f10c54f9b251a5deaf3c05815f7659bbe02",
-    zkEthBridge: "0x2c5d8a991f399089f728f1ae40bd0b11acd0fb62",
-    l2Weth: "0xD3765838f9600Ccff3d01EFA83496599E0984BD2",
+    zkErc20Bridge: "0x00ff932A6d70E2B8f1Eb4919e1e09C1923E7e57b",
+    l2Weth: "0x20b28B1e4665FFf290650586ad76E977EAb90c5D",
```
pxrl, line 97: Flip ordering.

```suggestion
    l2Weth: "0x20b28B1e4665FFf290650586ad76E977EAb90c5D",
    zkErc20Bridge: "0x00ff932A6d70E2B8f1Eb4919e1e09C1923E7e57b",
```

.upgradable/zkSync-era.json
pxrl, line 1: @nicholaspai Not sure whether we actually need this, but I added it just to be on the safe side. I saw that you previously added `.upgradable` for the zksync-goerli deployments, but the fact that it's hidden in a `.` folder makes me unsure if it really needs to be part of the repo.

PR #356, opened 2023-09-13
improve(SpokePool): Prevent future deposit quoteTimestamps

pxrl:
improve(SpokePool): Prevent future deposit quoteTimestamps
    
Future quoteTimestamps create unusual challenges for the bots that
process Deposit events. Fundamentally, it's impossible to compute the
correct realizedLpFeePct for a deposit with a future quoteTimestamp, so
all implementations must correctly park these early deposits and pull
them out at the right time. Given that we are now seeing diversity in
relayer implementations, it seems improbable that all will implement
this correctly, and the downside is that they will lose funds for a
slight mistake here.
    
From the depositor side, for the most part it is hard to justify
deliberately setting a quoteTimestamp that is in the future.
    
The scenarios where this might impact users is:
 - A chain lags slightly behind the current time, meaning that deposits
   will fail due to very small time deltas.
 - A deposit from a multisig normally wants the widest possible time
   range to collect the necessary signatures, and is technically losing
   half of that window.
    
For the former, this could be managed by the API, which supplies the
suggested quoteTimestamp to the depositor. In cases where the chain has
stalled, the API could simply throw a descriptive error.
    
For the latter, this could be mitigated by doubling the current deposit
window.

References ACX-1533

### DIFF
--- a/contracts/SpokePool.sol
+++ b/contracts/SpokePool.sol
@@ -443,7 +443,7 @@ abstract contract SpokePool is
         // It is assumed that cross-chain timestamps are normally loosely in-sync, but clock drift can occur. If the
         // SpokePool time stalls or lags significantly, it is still possible to make deposits by setting quoteTimestamp
         // within the configured buffer. The owner should pause deposits if this is undesirable. This will underflow if
-        // quoteTimestamp is less than depositQuoteTimeBuffer; this is safe but will throw an unintuitive error.
+        // quoteTimestamp is more than depositQuoteTimeBuffer; this is safe but will throw an unintuitive error.
 
         // slither-disable-next-line timestamp
         require(getCurrentTime() - quoteTimestamp <= depositQuoteTimeBuffer, "invalid quoteTimestamp");
--- a/contracts/SpokePool.sol
+++ b/contracts/SpokePool.sol
@@ -439,17 +439,16 @@ abstract contract SpokePool is
         require(amount <= MAX_TRANSFER_SIZE, "Amount too large");
         require(depositCounter[originToken] <= maxCount, "Above max count");
 
-        // This function assumes that L2 timing cannot be compared accurately and consistently to L1 timing. Therefore,
-        // block.timestamp is different from the L1 EVM's. Therefore, the quoteTimestamp must be within a configurable
-        // buffer of this contract's block time to allow for this variance.
-        // Note also that quoteTimestamp cannot be less than the buffer otherwise the following arithmetic can result
-        // in underflow. This isn't a problem as the deposit will revert, but the error might be unexpected for clients.
+        // Require that quoteTimestamp has a maximum age so that depositors pay an LP fee based on recent HubPool usage.
+        // It is assumed that cross-chain timestamps are normally loosely in-sync, but clock drift can occur. If the
+        // SpokePool time stalls or lags significantly, it is still possible to make deposits by setting quoteTimestamp
+        // within the configured buffer. The owner should pause deposits if this is undesirable. This will underflow if
+        // quoteTimestamp is less than depositQuoteTimeBuffer; this is safe but will throw an unintuitive error.
 
-        //slither-disable-next-line timestamp
+        // slither-disable-next-line timestamp
         require(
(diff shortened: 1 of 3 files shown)


contracts/SpokePool.sol
```diff
@@ -439,17 +439,16 @@ abstract contract SpokePool is
...
+        // within the configured buffer. The owner should pause deposits if this is undesirable. This will underflow if
+        // quoteTimestamp is less than depositQuoteTimeBuffer; this is safe but will throw an unintuitive error.
 
-        //slither-disable-next-line timestamp
+        // slither-disable-next-line timestamp
         require(
-            getCurrentTime() >= quoteTimestamp - depositQuoteTimeBuffer &&
-                getCurrentTime() <= quoteTimestamp + depositQuoteTimeBuffer,
-            "invalid quote time"
+            getCurrentTime() >= quoteTimestamp && getCurrentTime() <= quoteTimestamp + depositQuoteTimeBuffer,
```
james-a-morris, line 450: The `getCurrentTime()` should not change between these two calls. Would storing them in a local variable make sense so we don't have to call `getCurrentTime()` twice?

james-a-morris, line 450: Potential Refactor:
```suggestion
            quoteTimestamp - getCurrentTime() <= depositQuoteTimeBuffer
```

Reasoning:
The original expression is really looking to check if the quote timestamp exists in the window between `[c_t, c_(t+b)]`, or visually:
```
---------------- c ------------------
       [                    ]
      ct                ct+b

c = quote timestamp
ct = current time
b = deposit time buffer
```

If we apply algebra, we can simplify to:
```
c_t <= c <= c_t + b
0 <= c - c_t <= b
```

Because `getCurrentTime()` and `quoteTimestamp()` are both unsigned ints, I believe any case where `c - c_t` where the answer _should_ be negative will result in an underflow and thus make the resulting value incredibly large. 

As a result, we can leverage this underflow effectively to ignore the lower comparison and we are left with the revised expression.

cc @nicholaspai

pxrl: The buffer is backwards-looking rather than forwards-looking, so `quoteTimestamp` must always be _less_ than `getCurrentTime()` and we're actually verifying that `ct-b <= c <= ct`. But adjusting the change to `getCurrentTime() - quoteTimestamp <= depositQuoteTimeBuffer` works and avoids the 2nd call to `getCurrentTime()`. This saves almost 100 gas on a deposit :+1:

nb. The change to the comment above was also incorrect; there was no more possibility of underflow, but if we apply the above change there will be once more. I think that's probably OK, since we only expect advanced users to end up submitting future timestamps, so they should also be advanced enough to identify where it reverts.

pxrl: I'd understood that it was counter-intuitively cheaper to call `getCurrentTime()` twice than to store it in a local variable, but I have now run the gas analytics over the two options and it doesn't show any difference for me. Per your comment below, we should however be OK to remove the 2nd call to `getCurrentTime()` and just compute that the delta between the current time and `quoteTimestamp` is less than the configured buffer.

contracts/SpokePool.sol
```diff
@@ -439,18 +439,14 @@ abstract contract SpokePool is
...
-        require(
-            getCurrentTime() >= quoteTimestamp - depositQuoteTimeBuffer &&
-                getCurrentTime() <= quoteTimestamp + depositQuoteTimeBuffer,
-            "invalid quote time"
-        );
+        // Require that quoteTimestamp has a maximum age so that depositors pay an LP fee based on recent HubPool usage.
+        // It is assumed that cross-chain timestamps are normally loosely in-sync, but clock drift can occur. If the
+        // SpokePool time stalls or lags significantly, it is still possible to make deposits by setting quoteTimestamp
+        // within the configured buffer. The owner should pause deposits if this is undesirable. This will underflow if
+        // quoteTimestamp is less than depositQuoteTimeBuffer; this is safe but will throw an unintuitive error.
```
nicholaspai, line 446:
```suggestion
        // quoteTimestamp is more than depositQuoteTimeBuffer; this is safe but will throw an unintuitive error.
```

nicholaspai (approved):
One comment change but LGTM

PR #357, opened 2023-09-19
feat: deposit() wrapper with current quoteTimestamp

pxrl:
depositNow() is a new deposit() wrapper that removes the need for the caller to supply a quoteTimestamp. Instead, the value of getCurrentTime() is used.

This is introduced to address the issue of multisig depositors having only a narrow window to collect the necessary signatures for a deposit. Instead, LP fee certainty is traded for flexibility around when the deposit is able to be submitted. The depositor is still able to query the expected LP fee via the Across API, so the chance of paying more fees than anticipated is fairly low.

Ref ACX-1544

### DIFF
--- a/contracts/SpokePool.sol
+++ b/contracts/SpokePool.sol
@@ -481,6 +481,42 @@ abstract contract SpokePool is
     }
 
     /**
+     * @notice This is a simple wrapper for deposit() that sets the quoteTimestamp to the current SpokePool timestamp.
+     * @notice This function is intended for multisig depositors who can accept some LP fee uncertainty in order to lift
+     * the quoteTimestamp buffer constraint.
+     * @dev Re-orgs may produce invalid fills if the quoteTimestamp moves across a change in HubPool utilisation.
+     * @dev The existing function modifiers are already enforced by deposit(), so no additional modifiers are imposed.
+     * @param recipient Address to receive funds at on destination chain.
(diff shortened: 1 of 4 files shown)


contracts/SpokePool.sol
```diff
@@ -480,6 +480,42 @@ abstract contract SpokePool is
...
+    function depositNow(
+        address recipient,
+        address originToken,
+        uint256 amount,
+        uint256 destinationChainId,
+        int64 relayerFeePct,
+        bytes memory message,
+        uint256 maxCount
+    ) public payable {
+        deposit(
```
nicholaspai, line 507: I think you'll need to pass through the `msg.value` to `deposit`

nicholaspai (commented):
Looks good but one change required I think https://github.com/across-protocol/contracts-v2/pull/357/files#r1330205507

contracts/SpokePool.sol
pxrl: msg.value propagates because the call to `deposit()` is internal, but I've added tests to verify this: https://github.com/across-protocol/contracts-v2/pull/357/commits/9499d4eddf420d5471e19d3814b9b2058c527420.

nicholaspai (approved):

mrice32 (approved):
LGTM!

PR #358, opened 2023-09-27
chore: Add Polygon Mumbai deployments

pxrl:
This was requested by a project that wants to test bridging and possibly
also messaging.


mrice32 (approved):
LGTM!

pxrl:
> QQ: this looks to be a purely additive change for mumbai. Should we also open PRs to reference/account for these changes?

Not sure what you mean by this, but I think we'll bump the contacts-v2 version naturally as part of the concurrent SpokePool work that's going on, so these changes will be included in that, and then we can ripple them through to sdk-v2 and relayer-v2 accordingly :+1:

PR #360, opened 2023-09-28
fix: Update BAL address on Base

nicholaspai:
BAL team originally deployed the address we're updating away from and then updated to this new one. All BAL dApps should be using this new address now. Because of this conflict, Across has previously decided against enabling BAL on Base. Therefore this PR change is the first step to enabling BAL on Base


james-a-morris (approved):
Approved to remove blocker

pxrl (approved):
Address checks out on Uniswap's [tokenlist](https://gateway.ipfs.io/ipns/tokens.uniswap.org)

2023-09-29-111026_1269x229_scrot


2023-09-29-110919_1212x543_scrot

PR #363, opened 2023-10-13
refactor: Simplify Deposit test + helpers

pxrl:
Simplify deposit generation by converting getDepositParams() to accept an object, rather than a fixed list of parameters. This makes it cleaner for the caller to opt-out of supplying various inputs.

This was prompted by a need to update the deposit() and getDepositParams() utilities such that the caller can supply a custom message.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@across-protocol/contracts-v2",
-  "version": "2.4.6",
+  "version": "2.4.5",
   "author": "UMA Team",
   "license": "AGPL-3.0-only",
   "repository": {


test/fixtures/SpokePool.Fixture.ts
```diff
@@ -74,22 +82,23 @@ export async function deposit(
...
-        recipient.address,
-        token.address,
-        amountToDeposit,
-        destinationChainId,
-        depositRelayerFeePct,
-        quoteTimestamp ?? (await spokePool.getCurrentTime())
-      )
-    );
+  quoteTimestamp ??= await spokePool.getCurrentTime();
+  await spokePool.connect(depositor).deposit(
```
pxrl, line 91: This looks incorrect - it should probably use `BigNumber.toNumber()` instead.

test/fixtures/SpokePool.Fixture.ts
```diff
@@ -116,12 +125,12 @@ export async function fillRelay(
...
-  depositAmount: BigNumber = consts.amountToDeposit,
-  amountToRelay: BigNumber = consts.amountToRelay,
-  realizedLpFeePct: BigNumber = consts.realizedLpFeePct,
-  relayerFeePct: BigNumber = consts.depositRelayerFeePct
+  depositId = consts.firstDepositId,
+  originChainId = consts.originChainId,
+  depositAmount = consts.amountToDeposit,
+  amountToRelay = consts.amountToRelay,
+  realizedLpFeePct = consts.realizedLpFeePct,
+  relayerFeePct = consts.depositRelayerFeePct
```
pxrl, line 133: Opportunistic cleanup here.

test/fixtures/SpokePool.Fixture.ts
```diff
@@ -222,24 +231,25 @@ export function getRelayHash(
...
-    _maxCount ? _maxCount.toString() : consts.maxUint256.toString(),
+    args.recipient ?? randomAddress(),
+    args.originToken,
+    args.amount.toString(),
+    args.destinationChainId.toString(),
+    args.relayerFeePct.toString(),
+    args.quoteTimestamp.toString(),
+    args.message ?? "0x",
+    args?.maxCount?.toString() ?? consts.maxUint256.toString(),
   ];
```
pxrl, line 253: This was the entry point for this change - we need to be able to generate deposits with a custom message.

test/fixtures/SpokePool.Fixture.ts
pxrl: Fixed: https://github.com/across-protocol/contracts-v2/pull/363/commits/217dc9ea8508e0e90f440c5ddfd97a4d9c8b875f

nicholaspai (approved):
LGTM thanks for doing this. We had a TODO in the recent clean up to copy these utilities to the relayer repo. Should we do it now?

pxrl:
> LGTM thanks for doing this. We had a TODO in the recent clean up to copy these utilities to the relayer repo. Should we do it now?

It's not clear to me whether it's better to copy, or just invest effort in fixing the utils here. If we copied then it'd probably have to go to sdk-v2, since we also depend on the functionality there. I'm just not a big fan of having duplicates of these utilities.

PR #367, opened 2023-11-02
feat(SpokePool): Add depositFor public function to override `depositor` address in FundsDeposited event

nicholaspai:
Fixes ACX-1669

# Motivation

Enables deposits sent through third parties (aggregators, SpokePoolVerifier) to be sped up by end user

# Implementation

Logic in `deposit` is copied into internal function `_deposit` that takes one additional param `depositor`.


contracts/SpokePool.sol
```diff
@@ -906,6 +958,62 @@ abstract contract SpokePool is
...
+        emit FundsDeposited(
+            amount,
+            chainId(),
+            destinationChainId,
+            relayerFeePct,
+            newDepositId,
+            quoteTimestamp,
+            originToken,
+            recipient,
+            depositor,
```
nicholaspai, line 1012: This is the only line changed, setting `depositor` instead of `msg.sender` to `FundsDeposited.depositor`

nicholaspai:
I added a new minor version bump here so we can more easily tag the latest `deployed` version

contracts/SpokePool.sol
mrice32: Yeah, wanted to triple check that the transferFrom lines weren't changed. +1

mrice32 (approved):
LGTM!

PR #369, opened 2023-11-09
feat: Create USSLib

nicholaspai:
# Breaking changes from existing `SpokePool` types of same name:

Fixes ACX-1672

## RelayData
- `relayerFeePct`/ `realizedLpFeePct` replaced with `outputAmount`. The total fee is implied to be difference between `inputAmount` - `outputAmount` * `exchangeRateInputOutputTokens`
- Fill `expiryTimestamp`/`fillDeadline`
- Exclusive `relayer`

## RelayExecution
- No `maxCount`
- Updatable deposit fields are optimistically set to:
    - outputAmount
    - recipient
    - message

# Optimizing compilation of contracts

Unfortunately, adding more functions to the `SpokePool` will push it over the contract bytecode size unless we decrease the optimizer. Not sure how to get around this since we want to maintain old functionality so we can only add to the size. I've bumped it down by 100x to the point where the compiler doesn't warn anymore

### DIFF
--- a/contracts/SpokePool.sol
+++ b/contracts/SpokePool.sol
@@ -7,7 +7,7 @@ import "./interfaces/SpokePoolInterface.sol";
 import "./upgradeable/MultiCallerUpgradeable.sol";
 import "./upgradeable/EIP712CrossChainUpgradeable.sol";
 import "./upgradeable/AddressLibUpgradeable.sol";
-import "./USSLib.sol";
+import "./USSSpokePoolInterface.sol";
 
 import "@openzeppelin/contracts-upgradeable/token/ERC20/IERC20Upgradeable.sol";
 import "@openzeppelin/contracts-upgradeable/token/ERC20/utils/SafeERC20Upgradeable.sol";
@@ -37,6 +37,7 @@ interface AcrossMessageHandler {
  * submits a proof that the relayer correctly submitted a relay on this SpokePool.
  */
 abstract contract SpokePool is
+    USSSpokePoolInterface,
     SpokePoolInterface,
     UUPSUpgradeable,
     ReentrancyGuardUpgradeable,
@@ -628,8 +629,8 @@ abstract contract SpokePool is
         address depositRefundCallbackAddress,
         // TODO: Running into stack-too-deep errors when emitting FundsDeposited with all of the parameters
         // so I've packed them for now into input and output token structs
-        USSLib.InputToken memory inputToken,
-        USSLib.OutputToken memory outputToken,
+        InputToken memory inputToken,
+        OutputToken memory outputToken,
         uint256 destinationChainId,
         address exclusiveRelayer,
         uint32 quoteTimestamp,
@@ -641,7 +642,7 @@ abstract contract SpokePool is
         // Do stuff
         // - Pull funds from depositor
 
-        emit USSLib.USSFundsDeposited(
+        emit USSFundsDeposited(
             inputToken,
             outputToken,
             destinationChainId,
@@ -910,8 +911,8 @@ abstract contract SpokePool is
         address depositor,
         address recipient,
         address exclusiveRelayer,
-        USSLib.InputToken memory inputToken,
-        USSLib.OutputToken memory outputToken,
+        InputToken memory inputToken,
+        OutputToken memory outputToken,
         uint256 repaymentChainId,
         uint256 originChainId,
         uint32 depositId,
@@ -920,8 +921,8 @@ abstract contract SpokePool is
     ) public nonReentrant unpausedFills {
         // Validate input params
 
-        USSLib.RelayExecution memory relayExecution = USSLib.RelayExecution({
-            relay: USSLib.RelayData({
+        USSRelayExecution memory relayExecution = USSRelayExecution({
+            relay: USSRelayData({
                 depositor: depositor,
                 recipient: recipient,
                 relayer: exclusiveRelayer,
@@ -940,7 +941,7 @@ abstract contract SpokePool is
             updatedRecipient: recipient,
             updatedMessage: message,
             repaymentChainId: repaymentChainId,
-            executionType: USSLib.RelayExecutionType.FastFill,
+            executionType: RelayExecutionType.FastFill,
             payoutAdjustmentPct: 0
         });
         relayExecution.relayHash = keccak256(abi.encode(relayExecution.relay));
@@ -951,7 +952,7 @@ abstract contract SpokePool is
 
         // Trigger `message` callback if appropriate.
 
-        emit USSLib.USSFilledRelay(
+        emit USSFilledRelay(
             inputToken,
             outputToken,
             repaymentChainId,
@@ -965,7 +966,7 @@ abstract contract SpokePool is
             // updatedRecipient,
             recipient,
             // executionType,
-            USSLib.RelayExecutionType.FastFill,
+            RelayExecutionType.FastFill,
             // updatedOutputTokenAmount
             outputToken.amount,
             // payout adjustment pct
--- a/contracts/SpokePool.sol
+++ b/contracts/SpokePool.sol
@@ -941,7 +941,7 @@ abstract contract SpokePool is
             updatedRecipient: recipient,
             updatedMessage: message,
             repaymentChainId: repaymentChainId,
-            executionType: RelayExecutionType.FastFill,
+            slowFill: false,
             payoutAdjustmentPct: 0
         });
         relayExecution.relayHash = keccak256(abi.encode(relayExecution.relay));
@@ -965,8 +965,8 @@ abstract contract SpokePool is
             message,
             // updatedRecipient,
             recipient,
-            // executionType,
-            RelayExecutionType.FastFill,
+            // slowFill,
+            false,
             // updatedOutputTokenAmount
             outputToken.amount,
             // payout adjustment pct
--- a/contracts/SpokePool.sol
+++ b/contracts/SpokePool.sol
@@ -631,7 +631,6 @@ abstract contract SpokePool is
         address exclusiveRelayer,
         uint256 inputAmount,
         uint256 outputAmount,
-        uint256 originChainId,
         uint256 destinationChainId,
         uint32 quoteTimestamp,
         uint32 fillDeadline,
@@ -646,7 +645,6 @@ abstract contract SpokePool is
         // - Pull funds from depositor
 
         emit USSLib.FundsDeposited(
-            originChainId,
             destinationChainId,
             inputAmount,
             outputAmount,
@@ -971,7 +969,6 @@ abstract contract SpokePool is
             relayExecution.relay.outputAmount,
             relayExecution.repaymentChainId,
             relayExecution.relay.originChainId,
-            relayExecution.relay.destinationChainId,
             relayExecution.relay.depositId,
             relayExecution.relay.fillDeadline,
             relayExecution.relay.inputToken,
--- a/contracts/SpokePool.sol
+++ b/contracts/SpokePool.sol
@@ -626,33 +626,29 @@ abstract contract SpokePool is
         address depositor,
         address recipient,
         address depositRefundCallbackAddress,
-        address inputToken,
-        address outputToken,
-        address exclusiveRelayer,
-        uint256 inputAmount,
-        uint256 outputAmount,
+        // TODO: Running into stack-too-deep errors when emitting FundsDeposited with all of the parameters
+        // so I've packed them for now into input and output token structs
+        USSLib.InputToken memory inputToken,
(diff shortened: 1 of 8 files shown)


hardhat.config.ts
```diff
@@ -38,7 +38,7 @@ const LARGE_CONTRACT_COMPILER_SETTINGS = {
 
 const config: HardhatUserConfig = {
   solidity: {
-    compilers: [{ version: solcVersion, settings: { optimizer: { enabled: true, runs: 1000000 }, viaIR: true } }],
+    compilers: [{ version: solcVersion, settings: { optimizer: { enabled: true, runs: 10000 }, viaIR: true } }],
```
nicholaspai, line 41: See PR desc. about contract size :(

contracts/USSLib.sol
```diff
@@ -0,0 +1,113 @@
...
+        RefundDeposit // Deposit refunds are included similarly to Slow relay leaves but are distinguished for the
+        // convenience of off-chain actors who may want to treat slow fill execution differently from deposit
+        // refund execution. For example, deposit refunds are expected to have `message` fields with helpful
+        // information to define the deposit for the refund callback address, but typically, an off-chain actor
+        // would not want to execute slow fill leaves with messages because they might be rebating the slow fill
+        // recipient their relayer fees unintentionally.
+    }
+    // This struct represents the data to fully specify a **unique** relay. If any portion of this data differs,
+    // the relay is considered to be completely distinct. This data is hashed and inserted to the SpokePool to
+    // prevent collisions.
```
pxrl, line 20:
```suggestion
// This struct represents the data to fully specify a **unique** relay. This data is hashed and saved by the SpokePool                                                                         
// to prevent collisions. If any portion of this data differs, the relay is considered to be completely distinct.
```

contracts/USSLib.sol
```diff
@@ -0,0 +1,113 @@
...
+        uint256 outputAmount;
+        int256 payoutAdjustmentPct;
+        bytes message;
+    }
+
+    // TODO: I know within Solidity we can name this event the same as SpokePool.FundsDeposited but does that make
+    // it harder at the typescript level to distinguish the events?
+    event FundsDeposited(
+        uint256 indexed originChainId,
+        uint256 destinationChainId,
```
pxrl, line 82: Do we really want to index `originChainId` ? It's a constant on the SpokePool, whereas `destinationChainId` could be interesting to filter for.

contracts/USSLib.sol
```diff
@@ -0,0 +1,113 @@
...
+    );
+
+    event FilledRelay(
+        uint256 inputAmount,
+        uint256 outputAmount,
+        uint256 repaymentChainId,
+        uint256 indexed originChainId,
+        uint256 destinationChainId,
+        uint32 indexed depositId,
+        uint32 fillDeadline,
```
pxrl, line 104: Assume this is a 1:1 copy of `expiryTimestamp` ?

contracts/SpokePool.sol
```diff
@@ -862,6 +908,82 @@ abstract contract SpokePool is
...
+            updatedOutputAmount: outputAmount,
+            updatedRecipient: recipient,
+            updatedMessage: message,
+            repaymentChainId: repaymentChainId,
+            executionType: USSLib.RelayExecutionType.FastFill,
+            payoutAdjustmentPct: 0
+        });
+        relayExecution.relayHash = keccak256(abi.encode(relayExecution.relay));
+
+        // Validate
```
pxrl, line 954: Does this include evaluating `exclusiveRelayer` against `msg.sender` ? If so, would that instead be better placed at the very top of the function?

contracts/USSLib.sol
```diff
@@ -0,0 +1,113 @@
...
+pragma solidity ^0.8.0;
+
+// Contains structs and functions used by SpokePool contracts to facilitate universal settlement.
+library USSLib {
+    enum RelayExecutionType {
+        FastFill, // Fast relay execution is the default and is used for most relays by calling
+        // `fillRelay()` or `fillRelayWithUpdatedDeposit`.
+        SlowFill, // Slow relay execution can be done after the dataworker includes a slow relay leaf in the root bundle
+        // proposal, which propogates the leaf to this SpokePool and someone calls executeSlowRelayLeaf().
+        RefundDeposit // Deposit refunds are included similarly to Slow relay leaves but are distinguished for the
```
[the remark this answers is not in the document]
nicholaspai: that's more consistent with the other terms +1

contracts/USSLib.sol
nicholaspai: hmm good point

contracts/USSLib.sol
nicholaspai: it is yeah i wasn't sure what to name it or whether to keep them all the same name, any thoughts? In my mind `fillDeadline` == `expiryTimestamp` but the former might be more intuitive

contracts/USSLib.sol
nicholaspai: this was suggested by the linter so i can't change it

contracts/USSLib.sol
```diff
@@ -0,0 +1,113 @@
...
+        address depositRefundCallbackAddress,
+        bytes message
+    );
+
+    event FilledRelay(
+        uint256 inputAmount,
+        uint256 outputAmount,
+        uint256 repaymentChainId,
+        uint256 indexed originChainId,
+        uint256 destinationChainId,
```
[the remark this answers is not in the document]
nicholaspai: 79f8a135e5c037f8c6410a9ed44de41db0c7c301

contracts/SpokePool.sol
nicholaspai: yep 912db111aa03ccebd0a953f058cba0a8aff0985c

contracts/USSLib.sol
```diff
@@ -0,0 +1,125 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.0;
+
+// Contains structs and functions used by SpokePool contracts to facilitate universal settlement.
+library USSLib {
```
mrice32, line 5: So the library is just a way for us to namespace all of this so it's clear what the "old version" was and what the new version is?

Alternatively, we could just put all this in an interface and prefix all the names with USS, which might result in shorter, cleaner code. Will think a bit on the best way to organize this.

contracts/SpokePool.sol
```diff
@@ -1053,6 +1166,7 @@ abstract contract SpokePool is
             // limit the refund count in valid proposals to be ~800 so any RelayerRefundLeaves with > 800 refunds should
             // not make it to this stage.
 
+            // TODO: I think we can remove this if we bump solidity to >=0.8.22
```
mrice32, line 1169: +1

mrice32 (commented):
Looks awesome! Just a few questions.

contracts/SpokePool.sol
```diff
@@ -617,6 +618,49 @@ abstract contract SpokePool is
...
     }
 
+    /********************************************
+     *         USS DEPOSITOR FUNCTIONS          *
+     ********************************************/
+
+    function depositUSS(
+        address depositor,
+        address recipient,
+        address depositRefundCallbackAddress,
```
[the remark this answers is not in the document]
nicholaspai: No, `message` would only be sent to the `recipient` in the case of a successful deposit. If a deposit is refunded, the `message` is thrown away by the dataworker. Instead, the `depositRefundCallbackAddress`'s `handleAcrossMessage` is called which should be programmed to identify the deposit and refund accordingly. I imagine the dataworker would construct the SlowFIllLeaf for a deposit refund to the `depositRefundCallbackAddress` by setting `message = keccak256(abi.encode(relayData))` to help the callback contract identify the deposit that was refunded

contracts/SpokePool.sol
```diff
@@ -617,6 +618,49 @@ abstract contract SpokePool is
...
+            outputAmount,
+            newDepositId,
+            quoteTimestamp,
+            fillDeadline,
+            inputToken,
+            outputToken,
+            depositor,
+            recipient,
+            exclusiveRelayer,
+            depositRefundCallbackAddress,
```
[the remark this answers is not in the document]
nicholaspai: Agreed, would be very happy to remove this logic/parameter

contracts/USSLib.sol
nicholaspai: The library is only meant for namespace yes. Do you think an interface is cleaner than a library? I'm unopinionated

package.json
```diff
@@ -23,6 +23,7 @@
     "lint": "yarn prettier --list-different",
     "lint-fix": "yarn prettier --write",
     "prettier": "prettier .",
+    "clean": "rm -rf node_modules cache cache-zk artifacts artifacts-zk dist typechain",
```
pxrl, line 26:
```suggestion
    "clean": "for dir in node_modules cache cache-zk artifacts artifacts-zk dist typechain; do mv \"${dir}\" \"_${dir}\"; rm -rf \"_${dir}\" &; done",
```

pxrl: ^^ fwiw the above simply moves the dirs to somewhere that we don't care about (prefixes with _), and then starts a background rm -rf on the newly moved directory. This is faster in the case where there are a lot of files to be nuked.

test/SpokePool.Deposit.ts
```diff
@@ -7,6 +7,7 @@ import {
   depositRelayerFeePct as relayerFeePct,
   maxUint256,
 } from "./constants";
+import { ZERO_ADDRESS } from "@uma/common";
```
pxrl, line 10: Ethers exports this directly so we can avoid a dependency on @uma/common here.

```suggestion

const { AddressZero: ZERO_ADDRESS } = ethers.utils;
```

package.json
nicholaspai: I'm no expert on bash so i'm curious why your changes are better? I noticed you used something similar in other repos but I didn't understand why

contracts/SpokePool.sol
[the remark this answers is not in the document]
mrice32: Can't that be spoofed? Can't I send a fake fill with keccack256(abi.encode(somebodyElsesRelay)) as the message to the origin chain to trigger an invalid refund?

IMO, if we do want to do a refund callback, we should probably have it call a different method, like `recieveAcrossRefund` that can't be spoofed by anybody and must be triggered by a validated refund.

contracts/USSLib.sol
mrice32: I think it's cleaner as long as we're cool with prefixing everything, yeah.

mrice32 (commented):
Responded to your responses!

contracts/SpokePool.sol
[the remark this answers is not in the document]
nicholaspai: Hmm but the dataworker would be matching a deposit with a fill and if the fill can't be founded, it should refund the deposit and construct the relay data using the deposit information, not the fill's.

And on this point:

> IMO, if we do want to do a refund callback, we should probably have it call a different method, like recieveAcrossRefund that can't be spoofed by anybody and must be triggered by a validated refund.

How would that work to call that function if we're implementing the deposit refund as a slow fill leave execution?

contracts/USSLib.sol
nicholaspai: Hm yeah I think its cleaner too. For now I've defined functions and events/structs in `USSSpokePoolInterface.sol`.

Currently the existing `SpokePool` file implements the new functions. I thought about making a new `USSSpokePool` that either extends the SpokePool or is extended by the SpokePool. However, this has tradeoffs:

## abstract contract `USSSpokePool` is `SpokePool`

Requires replacing a lot of `SpokePool` references in contracts like `Arbitrum_SpokePool`, `Ovm_SpokePool` etc. that all are expecting the extend the `SpokePool` but would now want to extend `USSSpokePool`. This isn't that bad but touches a lot of files.

## abstract contract `SpokePool` is `USSSpokePool`

This would require declaring a `USSSpokePool` abstract contract that extends a lot of the same interfaces that `SpokePool` extends like `NonReentrancyGuardUpgradeable` and `EIP712CrossChainUpgradeable` while also adding some declarations like `using SafeERC20Upgradeable for IERC20Upgradeable` and `using AddressLibUpgradeable for address`. This is a bit verbose but probably cleaner than the first option because it contains the extra code to one file, `USSSpokePool`. These changes would also apply to declaring `USSSpokePool` as a library.

contracts/SpokePool.sol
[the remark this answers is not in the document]
nicholaspai: Also I think `handleAcrossMessage` would need to be a protected function that is only callable by the SpokePool. Not ideal...

mrice32 (approved):
This scaffolding looks great! LGTM!

PR #371, opened 2023-11-20
feat(SpokePool): Modify existing deposit() function to emit USSFundsDeposited event and add new depositUSS function

nicholaspai:
Fixes #ACX-1684 and #ACX-1685

## Modifications to existing deposit function: deposit():
- At upgrade: backwards compatible deposit() function will emit the new event. The deposit2() function (with new params) will emit the new event as well. There will be no deposit function post upgrade that emits the old event. The idea behind modifying the existing function is that post-upgrade, no more “old deposits” are emitted.
- Instead of the depositor using the relayerFeePct parameter to set the relayer’s portion of the fee, depositor uses the relayerFeePct to mark a “total fee” percentage. The recipient therefore knows that they will receive `amount * (1 - relayerFeePct)` tokens. The funds deposited event emits outputTokenAmount equal to this amount * (1 - relayerFeePct).
- outputToken: new parameter, defaults to some default token address like 0x0 to indicate that the same input and output token should be used
- Should hardcode committedFiller to 0x0
- Set expiryTimestamp to MAX_UINT. We never want to process deposit refunds for “old deposits”. We should add a rule in the dataworker that deposit refunds only count for type 2 deposits.

## Provides new public function deposit2() for depositors:
- expiryTimestamp: integer - This is going to be relative to the destination chain. Don’t allow this to be more than ~3 bundles into the future, so 12 hours maybe. This way the dataworker/relayer doesn’t have to maintain more a lookback longer than this.
- Emits FundsDeposited2 event

### DIFF
--- a/contracts/SpokePool.sol
+++ b/contracts/SpokePool.sol
@@ -515,16 +515,7 @@ abstract contract SpokePool is
         bytes memory message,
         uint256 maxCount
     ) public payable {
-        deposit(
-            recipient,
-            originToken,
-            amount,
-            destinationChainId,
-            relayerFeePct,
-            uint32(getCurrentTime()),
-            message,
-            maxCount
-        );
+        deposit(recipient, originToken, amount, destinationChainId, relayerFeePct, getCurrentTime(), message, maxCount);
     }
 
     /**
@@ -560,7 +551,7 @@ abstract contract SpokePool is
             amount,
             destinationChainId,
             relayerFeePct,
-            uint32(getCurrentTime()),
+            getCurrentTime(),
             message,
             maxCount
         );
@@ -626,8 +617,6 @@ abstract contract SpokePool is
     function depositUSS(
         address depositor,
         address recipient,
-        // TODO: Running into stack-too-deep errors when emitting FundsDeposited with all of the parameters
-        // so I've packed them for now into input and output token structs
         InputToken memory inputToken,
         OutputToken memory outputToken,
         uint256 destinationChainId,
@@ -636,10 +625,39 @@ abstract contract SpokePool is
         uint32 fillDeadline,
         bytes memory message
     ) public payable override nonReentrant unpausedDeposits {
-        // Validate
+        // Check that deposit route is enabled for the input token. There are no checks required for the output token
+        // which is pulled from the relayer at fill time and passed through this contract atomically to the recipient.
+        require(enabledDepositRoutes[inputToken.token][destinationChainId], "Disabled route");
 
-        // Do stuff
-        // - Pull funds from depositor
+        //@dev The following checks are copied from _deposit. Are they still needed?
+        require(inputToken.amount <= MAX_TRANSFER_SIZE, "Amount too large");
+        require(outputToken.amount <= MAX_TRANSFER_SIZE, "Amount too large");
+
+        // Require that quoteTimestamp has a maximum age so that depositors pay an LP fee based on recent HubPool usage.
+        // It is assumed that cross-chain timestamps are normally loosely in-sync, but clock drift can occur. If the
+        // SpokePool time stalls or lags significantly, it is still possible to make deposits by setting quoteTimestamp
+        // within the configured buffer. The owner should pause deposits if this is undesirable. This will underflow if
+        // quoteTimestamp is more than depositQuoteTimeBuffer; this is safe but will throw an unintuitive error.
+
+        // slither-disable-next-line timestamp
+        require(getCurrentTime() - quoteTimestamp <= depositQuoteTimeBuffer, "invalid quoteTimestamp");
+
+        // fillDeadline is relative to the destination chain.
+        // Don’t allow fillDeadline to be more than ~3 bundles into the future, so 12 hours maybe.
+        // This way the dataworker/relayer doesn’t have to maintain more a lookback longer than this.
+        // TODO: We could choose to skip this check in favor of adding a rule in the dataworker to enforce this
+        // constraint.
+        require(fillDeadline <= getCurrentTime() + 12 hours, "invalid fillDeadline");
+
+        // If the address of the origin token is a wrappedNativeToken contract and there is a msg.value with the
+        // transaction then the user is sending ETH. In this case, the ETH should be deposited to wrappedNativeToken.
+        if (inputToken.token == address(wrappedNativeToken) && msg.value > 0) {
+            require(msg.value == inputToken.amount, "msg.value must match amount");
+            wrappedNativeToken.deposit{ value: msg.value }();
+            // Else, it is a normal ERC20. In this case pull the token from the user's wallet as per normal.
+            // Note: this includes the case where the L2 user has WETH (already wrapped ETH) and wants to bridge them.
+            // In this case the msg.value will be set to 0, indicating a "normal" ERC20 bridging action.
+        } else IERC20Upgradeable(inputToken.token).safeTransferFrom(msg.sender, address(this), inputToken.amount);
 
         emit USSFundsDeposited(
             inputToken,
@@ -1062,8 +1080,8 @@ abstract contract SpokePool is
      * @notice Gets the current time.
      * @return uint for the current timestamp.
      */
-    function getCurrentTime() public view virtual returns (uint256) {
-        return block.timestamp; // solhint-disable-line not-rely-on-time
+    function getCurrentTime() public view virtual returns (uint32) {
+        return uint32(block.timestamp); // solhint-disable-line not-rely-on-time
     }
 
     /**************************************
@@ -1112,16 +1130,23 @@ abstract contract SpokePool is
             // In this case the msg.value will be set to 0, indicating a "normal" ERC20 bridging action.
         } else IERC20Upgradeable(originToken).safeTransferFrom(msg.sender, address(this), amount);
 
-        emit FundsDeposited(
-            amount,
-            chainId(),
+        emit USSFundsDeposited(
+            InputToken({ token: originToken, amount: amount }), // inputToken
+            OutputToken({ token: address(0), amount: _computeAmountPostFees(amount, relayerFeePct) }),
+            // outputToken:
+            // - setting token to 0x0 will signal to off-chain validator that the "equivalent"
+            // token as the inputToken for the destination chain should be replaced here.
+            // - amount will be the deposit amount less relayerFeePct, which should now be set
+            // equal to realizedLpFeePct + gasFeePct + capitalCostFeePct where (gasFeePct + capitalCostFeePct)
+            // is equal to the old usage of `relayerFeePct`.
             destinationChainId,
-            relayerFeePct,
             newDepositId,
             quoteTimestamp,
-            originToken,
-            recipient,
+            type(uint32).max, // fillDeadline. Older deposits don't expire.
             depositor,
+            recipient,
+            address(0), // exclusiveRelayer. Setting this to 0x0 will signal to off-chain validator that there
+            // is no exclusive relayer.
             message
         );
     }
--- a/contracts/test/MockSpokePool.sol
+++ b/contracts/test/MockSpokePool.sol
@@ -10,7 +10,7 @@ import "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
  */
 contract MockSpokePool is SpokePool, OwnableUpgradeable {
     uint256 private chainId_;
-    uint256 private currentTime;
+    uint32 private currentTime;
 
     function initialize(
         uint32 _initialDepositId,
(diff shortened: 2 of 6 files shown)


contracts/SpokePool.sol
```diff
@@ -636,10 +634,39 @@ abstract contract SpokePool is
...
+
+        // slither-disable-next-line timestamp
+        require(getCurrentTime() - quoteTimestamp <= depositQuoteTimeBuffer, "invalid quoteTimestamp");
+
+        // fillDeadline is relative to the destination chain.
+        // Don’t allow fillDeadline to be more than ~3 bundles into the future, so 12 hours maybe.
+        // This way the dataworker/relayer doesn’t have to maintain more a lookback longer than this.
+        // TODO: We could choose to skip this check in favor of adding a rule in the dataworker to enforce this
+        // constraint.
+        require(fillDeadline <= getCurrentTime() + 12 hours, "invalid fillDeadline");
```
mrice32, line 659: nit: prefer making this an immutable variable that can be set on deployment, so it can, theoretically, be changed on redeploy.

contracts/SpokePool.sol
```diff
@@ -636,10 +634,39 @@ abstract contract SpokePool is
...
+        require(outputToken.amount <= MAX_TRANSFER_SIZE, "Amount too large");
+
+        // Require that quoteTimestamp has a maximum age so that depositors pay an LP fee based on recent HubPool usage.
+        // It is assumed that cross-chain timestamps are normally loosely in-sync, but clock drift can occur. If the
+        // SpokePool time stalls or lags significantly, it is still possible to make deposits by setting quoteTimestamp
+        // within the configured buffer. The owner should pause deposits if this is undesirable. This will underflow if
+        // quoteTimestamp is more than depositQuoteTimeBuffer; this is safe but will throw an unintuitive error.
+
+        // slither-disable-next-line timestamp
+        require(getCurrentTime() - quoteTimestamp <= depositQuoteTimeBuffer, "invalid quoteTimestamp");
```
mrice32, line 652: Something we might want to change on the spoke pools, generally, is to make depositQuoteTimeBuffer an immutable variable. We will have to redeploy to change, but it will make every deposit a bit cheaper.

contracts/SpokePool.sol
```diff
@@ -636,10 +634,39 @@ abstract contract SpokePool is
...
+        // fillDeadline is relative to the destination chain.
+        // Don’t allow fillDeadline to be more than ~3 bundles into the future, so 12 hours maybe.
+        // This way the dataworker/relayer doesn’t have to maintain more a lookback longer than this.
+        // TODO: We could choose to skip this check in favor of adding a rule in the dataworker to enforce this
+        // constraint.
+        require(fillDeadline <= getCurrentTime() + 12 hours, "invalid fillDeadline");
+
+        // If the address of the origin token is a wrappedNativeToken contract and there is a msg.value with the
+        // transaction then the user is sending ETH. In this case, the ETH should be deposited to wrappedNativeToken.
+        if (inputToken.token == address(wrappedNativeToken) && msg.value > 0) {
```
mrice32, line 663: Not sure how we didn't see this earlier, but wrappedNativeToken should also be immutable, since this basically never changes.

mrice32 (approved):
Looks great! Only minor comments

contracts/SpokePool.sol
nicholaspai: +1 this also means we can remove some code!

contracts/SpokePool.sol
nicholaspai: So good point, like `depositQuoteTimeBuffer` above, but I think we have a problem converting this to an immutable variable. Immutable variables must be set in constructors or inline, which presents problems if we want a different value set here for each of the spokepools. Additionally there might be other issues with immutable variables in upgradeable contracts https://forum.openzeppelin.com/t/upgradable-contracts-instantiating-an-immutable-value/28763

nicholaspai: https://docs.openzeppelin.com/upgrades-plugins/1.x/faq#why-cant-i-use-immutable-variables

contracts/SpokePool.sol
```diff
@@ -60,8 +60,8 @@ abstract contract SpokePool is
     WETH9Interface public wrappedNativeToken;
 
     // Any deposit quote times greater than or less than this value to the current contract time is blocked. Forces
-    // caller to use an approximately "current" realized fee. Defaults to 1 hour.
-    uint32 public depositQuoteTimeBuffer;
+    // caller to use an approximately "current" realized fee.
+    uint32 public constant depositQuoteTimeBuffer = 3600;
```
nicholaspai, line 64: TODO: Figure out if this is upgrade-safe

nicholaspai: If not, we'll need to define a new constant DEPOSIT_QUOTE_TIME_BUFFER at the bottom of the contract's variables

contracts/SpokePool.sol
mrice32: I think OZ's docs are good for the common case, but they _don't apply to this use case_, assuming the tools still work with constructors and immutable variables in the implementation contract.

1. OZ says constructors should be replaced with initializers, which is true! However, I don't see why the implementation contract can't still have a constructor that sets immutable variables in the bytecode.
2. OZ says this means that the immutable variable will be shared by all proxies that point at the implementation. In this case, that's fine, desirable even! There is only have 1 proxy pointing at 1 implementation per chain and the implementation can be upgraded anytime to change the variable.

mrice32 (approved):
Looks good! Just responded to your comments about storage layout and immutables.

contracts/SpokePool.sol
nicholaspai: As per our offline discussions, https://github.com/across-protocol/contracts-v2/commit/be1c03926b89280542a94b566166628a5000ae98 implements the gas optimizations that convert storage vars to immutable/constant ones.

contracts/SpokePool.sol
nicholaspai: https://github.com/across-protocol/contracts-v2/pull/371#discussion_r1406843568

I don't think we ever actually change `depositQuoteTimeBuffer` in the tests and just leave it hardcoded to 3600 as set in the constructor. So IMO there's no functional difference making this constant or immutable since we can just as easily change it going forwards

nicholaspai: Nevermind, as discussed offline, `immutable` for some variables like `depositQuoteTimeBuffer` allows us to set a different value for diff. child contracts which could come in handy given the block time disparity between chains

contracts/Arbitrum_SpokePool.sol
```diff
@@ -27,23 +27,28 @@ contract Arbitrum_SpokePool is SpokePool {
     event SetL2GatewayRouter(address indexed newL2GatewayRouter);
     event WhitelistedTokens(address indexed l2Token, address indexed l1Token);
 
+    /// @custom:oz-upgrades-unsafe-allow constructor
```
mrice32, line 30: 👌

contracts/SpokePool.sol
```diff
@@ -55,13 +55,19 @@ abstract contract SpokePool is
...
+    // Note: storage variables prefixed with DEPRECATED used to be variables that could be set by the cross-domain
+    // admin. Admins ended up not changing these in production, so to reduce gas in deposit/fill functions,
+    // we are converting them to private variables to maintain the  contract storage layout and replacing them with
+    // immutable or constant variables, because retrieving a constant value is cheaper than retrieving
+    // a storage variable. Please see out the immutable/constant variable section.
+
     // Address of wrappedNativeToken contract for this network. If an origin token matches this, then the caller can
     // optionally instruct this contract to wrap native tokens when depositing (ie ETH->WETH or MATIC->WMATIC).
-    WETH9Interface public wrappedNativeToken;
+    WETH9Interface private DEPRECATED_wrappedNativeToken;
```
mrice32, line 66: nit: might be useful to replace the comments above this and DEPRECATED_depositQuoteTimeBuffer to explain that these variables are no longer used, but are preserved to keep the slot spacing the same.

mrice32 (approved):
LGTM!

contracts/SpokePool.sol
nicholaspai: ah yes, i already copied these comments down to the new immutable/constants var section so i can remove these

contracts/Boba_SpokePool.sol
```diff
@@ -19,12 +26,6 @@ contract Boba_SpokePool is Ovm_SpokePool {
...
         address _hubPool
     ) public initializer {
-        __OvmSpokePool_init(
-            _initialDepositId,
-            _crossDomainAdmin,
-            _hubPool,
-            0x4200000000000000000000000000000000000006,
-            0xDeadDeAddeAddEAddeadDEaDDEAdDeaDDeAD0000
-        );
+        __OvmSpokePool_init(_initialDepositId, _crossDomainAdmin, _hubPool, 0x4200000000000000000000000000000000000006);
```
pxrl, line 29: This is academic since we don't support Boba, but was there a reason not to use `Lib_PreeployAddresses.OVM_ETH` instead of the hardcoded `0x42...06` address?

contracts/SpokePool.sol
```diff
@@ -55,13 +55,13 @@ abstract contract SpokePool is
...
-    WETH9Interface public wrappedNativeToken;
-
-    // Any deposit quote times greater than or less than this value to the current contract time is blocked. Forces
-    // caller to use an approximately "current" realized fee. Defaults to 1 hour.
-    uint32 public depositQuoteTimeBuffer;
+    // Note: storage variables prefixed with DEPRECATED used to be variables that could be set by the cross-domain
+    // admin. Admins ended up not changing these in production, so to reduce gas in deposit/fill functions,
+    // we are converting them to private variables to maintain the contract storage layout and replacing them with
+    // immutable or constant variables, because retrieving a constant value is cheaper than retrieving
+    // a storage variable. Please see out the immutable/constant variable section.
```
mrice32, line 62: Nit: typo "please see out the"

contracts/SpokePool.sol
```diff
@@ -55,13 +55,13 @@ abstract contract SpokePool is
...
     address public hubPool;
 
-    // Address of wrappedNativeToken contract for this network. If an origin token matches this, then the caller can
-    // optionally instruct this contract to wrap native tokens when depositing (ie ETH->WETH or MATIC->WMATIC).
-    WETH9Interface public wrappedNativeToken;
-
-    // Any deposit quote times greater than or less than this value to the current contract time is blocked. Forces
-    // caller to use an approximately "current" realized fee. Defaults to 1 hour.
-    uint32 public depositQuoteTimeBuffer;
+    // Note: storage variables prefixed with DEPRECATED used to be variables that could be set by the cross-domain
```
mrice32, line 58: Nit: I would say "the following two storage variables..." since this does not really apply to all deprecated variables, just these two.

mrice32 (approved):
LGTM!

deploy/003_deploy_optimism_spokepool.ts
```diff
@@ -9,7 +9,9 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
   // with deprecated spoke pool.
   // Set hub pool as cross domain admin since it delegatecalls the Adapter logic.
   const constructorArgs = [1_000_000, hubPool.address, hubPool.address];
```
pxrl, line 11: It's a bit confusing with 2x `constructorArgs` in the call to `deployNewProxy()`. Thoughts on renaming this initializer args?

```suggestion
  const initArgs = [1_000_000, hubPool.address, hubPool.address];
```

contracts/SpokePool.sol
nicholaspai: But doesn't this accurately describe that this applies to "storage variables prefixed with DEPRECATED"?

mrice32 (approved):
LGTM!

contracts/SpokePool.sol
```diff
@@ -636,10 +661,37 @@ abstract contract SpokePool is
...
-        // Validate
+        // Check that deposit route is enabled for the input token. There are no checks required for the output token
+        // which is pulled from the relayer at fill time and passed through this contract atomically to the recipient.
+        require(enabledDepositRoutes[inputToken.token][destinationChainId], "Disabled route");
 
-        // Do stuff
-        // - Pull funds from depositor
+        //@dev The following checks are copied from _deposit. Are they still needed?
+        require(inputToken.amount <= MAX_TRANSFER_SIZE, "Amount too large");
+        require(outputToken.amount <= MAX_TRANSFER_SIZE, "Amount too large");
```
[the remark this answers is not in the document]
pxrl: Just to circle back to this one, I don't think there's any marginal risk here. In an OFA system, either the depositor sets the outputToken to the amount that the winning relayer bid on, or it doesn't. I don't see how capping the max token amount protects against this.

PR #372, opened 2023-11-26
improve(SpokePool): Convert wrappedNativeToken from variable to overridable public function

nicholaspai:
The motivation is to remove an unneccessary SLOAD to read `wrappedNativeToken` on each deposit since this address should never change after a SpokePool is deployed.

Unfortunately, we can't easily convert the existing variable to be an `immutable` variable since upgradeable contracts are not compatible easily with immutable variables [docs](https://docs.openzeppelin.com/upgrades-plugins/1.x/faq#why-cant-i-use-immutable-variables).

This variable is tricky to change to a constant also because it needs to change for each different child `SpokePool` contract implementation.

One way around it is is to rename but not remove the existing `wrappedNativeToken` variable to preserve the storage layout of the contract. Note, I **think** this is safe to do but I'm not sure. Then, declare an overridable `public` function `wrappedNativeToken` that we can reimplement in each child SpokePool to return a constant value. We also could make this function `internal` if that saves more gas versus calling a `public` function
@[nicholaspai](https://github.com/across-protocol/contracts-v2/commits?author=nicholaspai)
nicholaspai committed 1 hour ago

### DIFF
--- a/contracts/SpokePool.sol
+++ b/contracts/SpokePool.sol
@@ -57,7 +57,7 @@ abstract contract SpokePool is
 
     // Address of wrappedNativeToken contract for this network. If an origin token matches this, then the caller can
     // optionally instruct this contract to wrap native tokens when depositing (ie ETH->WETH or MATIC->WMATIC).
-    WETH9Interface public wrappedNativeToken;
+    WETH9Interface public DEPRECATED_wrappedNativeToken;
 
     // Any deposit quote times greater than or less than this value to the current contract time is blocked. Forces
     // caller to use an approximately "current" realized fee.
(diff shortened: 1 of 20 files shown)


contracts/SpokePool.sol
```diff
@@ -57,7 +57,7 @@ abstract contract SpokePool is
 
     // Address of wrappedNativeToken contract for this network. If an origin token matches this, then the caller can
     // optionally instruct this contract to wrap native tokens when depositing (ie ETH->WETH or MATIC->WMATIC).
-    WETH9Interface public wrappedNativeToken;
+    WETH9Interface public DEPRECATED_wrappedNativeToken;
```
nicholaspai, line 60: upgrading a proxy with this impl. means we can't remove this storage slot but I think we can rename it

nicholaspai:
Replaced this PR with https://github.com/across-protocol/contracts-v2/commit/be1c03926b89280542a94b566166628a5000ae98

PR #373, opened 2023-11-28
feat(SpokePool): Add USSRelayerRefund leaf struct with fillsRefunded merkle root hash

nicholaspai:
The fillsRefunded merkle root can be used an off-chain bot/script that needs to know which Deposits were successfully or unsuccessfully (i.e. not filled validly before fillDeadline). For example, a filler might have collateral escrowed that could be accordingly slashed/returned.
 
Refactors the `executeRelayerRefundLeaf` logic so that shared logic can be used in new function `executeRelayerRefundLeafUSS`.

Fixes ACX-1724


contracts/SpokePool.sol
```diff
@@ -1322,8 +1370,33 @@ abstract contract SpokePool is
         emit SetHubPool(newHubPool);
     }
 
+    function _preExecuteLeafHook(address l2TokenAddress) internal virtual {
```
nicholaspai, line 1373: I think this is a nice pattern we should already be using, where certain SpokePool children perform hooks before diving into the base executeRelayerRefund logic. By implementing an overridable hook and making sure its called before executeSlow/RelayerRefundLeaf functions, we can reduce the amount of functions that the children SpokePools have to implement

contracts/interfaces/USSSpokePoolInterface.sol
```diff
@@ -38,6 +38,28 @@ interface USSSpokePoolInterface {
         int256 payoutAdjustmentPct;
     }
 
+    struct USSRelayerRefundLeaf {
```
nicholaspai, line 41: Copied exactly from `RelayerRefundLeaf` plus two params at the end. Wish I could use the following syntax:

```
struct USSRelayerRefundLeaf is RelayerRefundLeaf 

or

struct USSRelayerRefundLeaf {
...RelayerRefundLeaf,
}
```

contracts/SpokePool.sol
```diff
@@ -1232,28 +1297,11 @@ abstract contract SpokePool is
...
-            relayerRefundLeaf.amountToReturn,
-            relayerRefundLeaf.chainId,
-            relayerRefundLeaf.refundAmounts,
-            rootBundleId,
-            relayerRefundLeaf.leafId,
-            relayerRefundLeaf.l2TokenAddress,
-            relayerRefundLeaf.refundAddresses,
-            msg.sender
-        );
+            emit TokensBridged(amountToReturn, chainId(), leafId, l2TokenAddress, msg.sender);
```
pxrl, line 1303: ooc, is it still relevant to emit `msg.sender` in this event? We're incidentally executing via Multicall3 so it doesn't really give any useful information at present. The same applies for the `...Executed` events.

Side note: We could make the bot a bit sharper so it avoids Multicall3 in cases where there is only one leaf to be executed, and in fact I think we only get a benefit from Multicall3 on the HubPool chain.

tokensbridged

relayerRefund

contracts/interfaces/USSSpokePoolInterface.sol
```diff
@@ -38,6 +38,28 @@ interface USSSpokePoolInterface {
...
+        // The associated L2TokenAddress that these claims apply to.
+        address l2TokenAddress;
+        // Must be same length as refundAmounts and designates each address that must be refunded.
+        address[] refundAddresses;
+        // Merkle tree containing each of the fills refunded by this leaf and accounted for in refundAmounts
+        // and refundAddresses. The MerkleLeaf structure should be defined in the ACROSS-V2 UMIP, as there is no
+        // logic in this contract that supports proving leaf inclusion in this root.
+        bytes32 fillsRefundedRoot;
+        // IPFS hash of the file containing all leaves in the fillsRefundedRoot.
+        string fillsRefundedIpfsHash;
```
pxrl, line 60: Do we want to lock this in as an IPFS hash, or could we just call it something more generic like `fillsRefundedHash` ? In principle that might allow us to  swap out the storage layer for something else later on via an UMIP change, without having contrdictory naming.

contracts/interfaces/USSSpokePoolInterface.sol
```diff
@@ -101,6 +123,19 @@ interface USSSpokePoolInterface {
...
+        uint256 amountToReturn,
+        uint256 indexed chainId,
+        uint256[] refundAmounts,
+        uint32 indexed rootBundleId,
+        uint32 indexed leafId,
+        address l2TokenAddress,
+        address[] refundAddresses,
+        bytes32 fillsRefundedRoot,
+        string fillsRefundedIpfsHash,
+        address caller
```
pxrl, line 136: Same question as for the non-USS events - is the caller relevant if we assume Multicall3 continues to be used?

contracts/interfaces/USSSpokePoolInterface.sol
```diff
@@ -101,6 +123,19 @@ interface USSSpokePoolInterface {
         bytes updatedMessage
     );
 
+    event USSExecutedRelayerRefundRoot(
```
pxrl, line 126: This is pure bikeshedding, but I am kinda partial to `<verb>USS<noun>()`. For example: `ExecutedUSSRelayerRefundRoot()`.

pxrl (commented):
This looks good - my comments are mostly on minor topics :+1:

contracts/SpokePool.sol
```diff
@@ -1094,8 +1095,92 @@ abstract contract SpokePool is
...
+
+    /**
+     * @notice Executes a relayer refund leaf stored as part of a root bundle. Will send the relayer the amount they
+     * sent to the recipient plus a relayer fee.
+     * @param rootBundleId Unique ID of root bundle containing relayer refund root that this leaf is contained in.
+     * @param relayerRefundLeaf Contains all data necessary to reconstruct leaf contained in root bundle and to
+     * refund relayer. This data structure is explained in detail in the SpokePoolInterface.
+     * @param proof Inclusion proof for this leaf in relayer refund root in root bundle.
+     */
+    function executeRelayerRefundLeafUSS(
```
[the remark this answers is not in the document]
nicholaspai: I did try refactoring as much as possible, how would you condense more? Note that I wanted to preserve the order of verifying the MerkleLeaf is included and unclaimed before distributing funds. This means that we need to make a call to `MerkleLib.verifyRelayerRefundLeaf{USS}()` in the middle of two internal function calls at the least

contracts/SpokePool.sol
nicholaspai: its probably overkill yeah, `msg.sender` is not useful for the most part because we execute everything ourselves but what about a world where there are multiple executors? This would make it easier to track perhaps?

nicholaspai: 5770b3f821959a0c97221f73764ec380e14a7c43

I don't think this neccessarily breaks bots but we should remove from sdk-v2/interfaces/SpokePool

https://github.com/across-protocol/sdk-v2/blob/master/src/interfaces/SpokePool.ts#L231

nicholaspai: > Side note: We could make the bot a bit sharper so it avoids Multicall3 in cases where there is only one leaf to be executed, and in fact I think we only get a benefit from Multicall3 on the HubPool chain.

Also, don't we handle that here?

https://github.com/across-protocol/relayer-v2/blob/master/src/clients/MultiCallerClient.ts#L407

contracts/interfaces/USSSpokePoolInterface.sol
nicholaspai: Removed `caller` in this event but will keep in the original SpokePool event to reduce breaking changes in relayer-v2

contracts/SpokePool.sol
pxrl: > Isn't that what we are doing currently by making execute{Relayer,Slow}RefundRoot() virtual?

It's not virtual anymore though - it's only marked as `override`.

virtual

> And for the use case, would this be used in the case where the merkle leaf math is somehow different per chain?

Yeah - I was wondering about something like that. I don't know how realistic it is that we'd see that.

contracts/SpokePool.sol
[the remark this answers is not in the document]
pxrl: I was just thinking about factoring out the code before and after the `require()` statement, but it might be a case of diminishing returns since the implementation is pretty simple.

contracts/Ovm_SpokePool.sol
```diff
@@ -104,66 +104,17 @@ contract Ovm_SpokePool is SpokePool {
...
-            chainId(),
-            realizedLpFeePct,
-            relayerFeePct,
-            depositId,
-            rootBundleId,
-            message,
-            payoutAdjustment,
-            proof
-        );
+    function _preExecuteLeafHook(address l2TokenAddress) internal override {
```
pxrl, line 114: Should we also make this virtual since it can be inherited by the downstream chain-specific contracts ?

contracts/SpokePool.sol
```diff
@@ -1233,28 +1294,11 @@ abstract contract SpokePool is
...
-            relayerRefundLeaf.amountToReturn,
-            relayerRefundLeaf.chainId,
-            relayerRefundLeaf.refundAmounts,
-            rootBundleId,
-            relayerRefundLeaf.leafId,
-            relayerRefundLeaf.l2TokenAddress,
-            relayerRefundLeaf.refundAddresses,
-            msg.sender
-        );
+            emit TokensBridged(amountToReturn, chainId(), leafId, l2TokenAddress);
```
pxrl, line 1300: We switch here from `relayerRefundLeaf.chainId` to `chainId()`. Is that just to reduce the number of arguments that need to be supplied to this function?

pxrl (approved):
A couple of questions but otherwise lgtm.

contracts/SpokePool.sol
nicholaspai: Its not virtual anymore because the `_preExecuteLeafHook` is now the `virtual` method. The `_execute{Relayer,Slow}RefundRoot()` are otherwise the same currently so I moved all shared logic that could be overridden in the children in to this new hook virtual method

nicholaspai: > Yeah - I was wondering about something like that. I don't know how realistic it is that we'd see that.

I think we can solve that if and when we ever see that but that would cause other issues if the root submitted to the HubPool results in different proofs for different chains

contracts/Ovm_SpokePool.sol
nicholaspai: I think we can get to that if we see a need to but its not needed now and would mean extra override declarations in the child contracts

contracts/SpokePool.sol
nicholaspai: hmm no I think its either an oversight or perhaps a gas optimization. Let me play with it and see whether the MLOAD or the block.chainId emits more gas

nicholaspai: Turns out passing in _chainId is quite a bit cheaper, perhaps the internal function call is expensive

contracts/interfaces/USSSpokePoolInterface.sol
nicholaspai: I'm good with that

PR #379, opened 2023-12-07
feat(SpokePool): Implement fillRelay/executeSlowFill

nicholaspai:
- One optimization here is to emit whether the FilledRelay completed a slow fill or not, making it easier for the dataworker to subtract excesses from a spoke pool without having to query for a historical RequestedSlowFill event
- Changes events so that they don't include structs, which is nicer for off-chain users to parse
- Due to the number of params passed to the fill functions, pack them into USSRelayData struct but keep params unpacked for events

Resolves  #ACX-1686
Resolves  #ACX-1688


contracts/SpokePool.sol
```diff
@@ -938,72 +949,78 @@ abstract contract SpokePool is
...
+            relayData.originChainId,
+            relayData.depositId,
+            relayData.fillDeadline,
+            relayData.exclusivityDeadline,
+            relayData.exclusiveRelayer,
+            msg.sender,
+            relayData.depositor,
+            relayData.recipient,
+            relayData.message,
+            replacedSlowFillExecution
```
nicholaspai, line 1001: I think this flag `replacedSlowFillExecution` could be useful for the dataworker to determine whether to return excess funds to HubPool:
- If `true` then return full `outputAmount` to HubPool, else SlowFill leaf executed and do nothing

contracts/SpokePool.sol
```diff
@@ -651,7 +651,7 @@ abstract contract SpokePool is
     ) public payable override nonReentrant unpausedDeposits {
         // Check that deposit route is enabled for the input token. There are no checks required for the output token
         // which is pulled from the relayer at fill time and passed through this contract atomically to the recipient.
-        require(enabledDepositRoutes[inputToken][destinationChainId], "Disabled route");
+        if (!enabledDepositRoutes[inputToken][destinationChainId]) revert DisabledRoute();
```
pxrl, line 654: Is this done to emit the same revert reason from multiple places in the contract?

contracts/SpokePool.sol
```diff
@@ -955,18 +955,17 @@ abstract contract SpokePool is
...
+        // Exclusivity deadline is inclusive and is the latest timestamp that the exclusive relayer has sole right
+        // to fill the relay.
         // solhint-disable-next-line not-rely-on-time
-        require(relayData.fillDeadline >= block.timestamp, "Fill deadline expired");
-        // solhint-disable-next-line not-rely-on-time
-        require(
-            relayData.exclusivityDeadline >= block.timestamp && relayData.exclusiveRelayer == msg.sender,
-            "Not exclusive relayer"
-        );
+        if (relayData.exclusivityDeadline >= block.timestamp && relayData.exclusiveRelayer != msg.sender) {
```
pxrl, line 961: Is it cheaper to reverse these, since the vast majority of fills will fall within the exclusivity deadline? It should save evaluating the 2nd statement in most cases.

```suggestion
        if (relayData.exclusiveRelayer != msg.sender && relayData.exclusivityDeadline >= block.timestamp) {
```

contracts/SpokePool.sol
```diff
@@ -1330,19 +1304,9 @@ abstract contract SpokePool is
         // Send each relayer refund address the associated refundAmount for the L2 token address.
         // Note: Even if the L2 token is not enabled on this spoke pool, we should still refund relayers.
         uint256 length = refundAmounts.length;
-        for (uint256 i = 0; i < length; ) {
+        for (uint256 i = 0; i < length; i++) {
```
pxrl, line 1307: Pre-increment _may_ be cheaper, but may also be optimised away by the compiler.

```suggestion
        for (uint256 i = 0; i < length; ++i) {
```

contracts/SpokePool.sol
nicholaspai: reverts used to be cheaper on gas cost and bytecode but I'm not actually sure that's the case anymore. Regardless I do think its nice to re-use revert reasons and the name of the revert error can be descriptive and as long as you want while the require message is limited to 32 chars

contracts/SpokePool.sol
nicholaspai: that's a good point

contracts/SpokePool.sol
nicholaspai: [relayData.exclusivityDeadline >= block.timestamp && relayData.exclusiveRelayer != msg.sender](https://docs.soliditylang.org/en/v0.8.11/control-structures.html#revert)

I think they're equivalent?

pxrl: The above link is for reversion (also note 0.8.11 rather than 0.8.22/23).

In any case, there are still many blog posts claiming that pre-increment is a bit cheaper than post-incremement.
https://medium.com/@solidity101/100daysofsolidity-gas-optimization-techniques-in-solidity-9a3eb4b327c5
https://medium.com/coinmonks/gas-saving-techniques-for-smart-contract-development-5daa7da22370
https://betterprogramming.pub/solidity-gas-optimizations-and-tricks-2bcee0f9f1f2

I haven't verified this by simulating, and it probably seems like something the optimiser should detect and handle. But it also doesn't hurt to just use pre-increment, since there are no side-effects.

contracts/SpokePool.sol
```diff
@@ -938,72 +949,69 @@ abstract contract SpokePool is
...
-                inputAmount: inputToken.amount,
-                outputAmount: outputToken.amount,
-                originChainId: originChainId,
-                destinationChainId: chainId(),
-                depositId: depositId,
-                fillDeadline: fillDeadline,
-                message: message
-            }),
+        // No need to override passed in destination chain id to be equal to chainId(),
+        // since this is not emitted in the FilledRelay event. So we do not have to fear a replay attack.
```
mrice32, line 966: I don't quite understand this? What has changed that allows us to be less worried about replay attacks or people spoofing the chain id?

contracts/SpokePool.sol
```diff
@@ -938,72 +949,69 @@ abstract contract SpokePool is
...
+     * @notice Request Across to send LP funds to this contract to fulfill a slow fill relay
+     * for a deposit in the next bundle.
+     * @dev Slow fills are not possible unless the input and output tokens are "equivalent", i.e.
+     * they route to the same L1 token via PoolRebalanceRoutes.
+     * @param relayData USSRelayData struct containing all the data needed to identify the deposit that should be
+     * slow filled. If any of the params are missing or different then Across will not include a slow
+     * fill for the intended deposit.
+     */
+    function requestUSSSlowFill(USSRelayData memory relayData) public override nonReentrant unpausedFills {
+        // TODO: Should we require that the exclusivity deadline has passed? In practice, I would never expect
```
mrice32, line 993: I think this raises a broader question: what happens if someone requests a slow fill _and_ the relay gets filled normally afterward? Is this effectively the same thing that happens today where a slow fill gets processed but it gets fast-filled before the slow fill can go through?

contracts/SpokePool.sol
```diff
@@ -938,72 +949,69 @@ abstract contract SpokePool is
...
+        // an exclusive relayer to request a slow fill, but there could be reasons?
+        // solhint-disable-next-line not-rely-on-time
+        if (relayData.fillDeadline < block.timestamp) revert ExpiredFillDeadline();
+
+        bytes32 relayHash = keccak256(abi.encode(relayData));
+        if (relayFills[relayHash] != uint256(FillStatus.Unfilled)) revert InvalidSlowFill();
```
mrice32, line 999:
```suggestion
        if (relayFills[relayHash] != uint256(FillStatus.Unfilled)) revert InvalidSlowFillRequest();
```

contracts/SpokePool.sol
```diff
@@ -1262,19 +1304,9 @@ abstract contract SpokePool is
...
-
-            // OK because we assume refund array length won't be > types(uint256).max.
-            // Based on the stress test results in /test/gas-analytics/SpokePool.RelayerRefundLeaf.ts, the UMIP should
-            // limit the refund count in valid proposals to be ~800 so any RelayerRefundLeaves with > 800 refunds should
-            // not make it to this stage.
-
-            // TODO: I think we can remove this if we bump solidity to >=0.8.22
-            unchecked {
-                ++i;
-            }
```
mrice32, line 1277: ❤️

contracts/SpokePool.sol
```diff
@@ -1621,6 +1667,118 @@ abstract contract SpokePool is
...
+        // status. However, we also use this slot to track whether this fill had a slow fill requested. Therefore
+        // we can include a bool in the FilledRelay event making it easy for the dataworker to compute if this
+        // fill was a fast fill that replaced a slow fill and therefore this SpokePool has excess funds that it
+        // needs to send back to the HubPool.
+        if (relayFills[relayHash] == uint256(FillStatus.Filled)) revert RelayFilled();
+        relayFills[relayHash] = uint256(FillStatus.Filled);
+
+        uint256 amountToSend = relayExecution.updatedOutputAmount;
+
+        // This can only happen in a slow fill, where the contract is funding the relay.
```
mrice32, line 1699: If we want to remove all UBA/dynamic fee logic until later, we should be able to safely remove all this payout adjustment logic, which should free up some bytecode!

contracts/SpokePool.sol
```diff
@@ -1621,6 +1667,118 @@ abstract contract SpokePool is
...
+
+        // If relayer and receiver are the same address, there is no need to do any transfer, as it would result in no
+        // net movement of funds.
+        // Note: this is important because it means that relayers can intentionally self-relay in a capital efficient
+        // way (no need to have funds on the destination).
+        // If this is a slow fill, we can't exit early since we still need to send funds out of this contract
+        // since there is no "relayer".
+        bool isSlowFill = relayExecution.slowFill;
+        address recipientToSend = relayExecution.updatedRecipient;
+        if (msg.sender == recipientToSend && !isSlowFill) return;
```
mrice32, line 1723: Does this mean we won't emit the FilledUSSRelay in the case that it's a self-relay? That seems like an issue, no?

contracts/SpokePool.sol
```diff
@@ -1621,6 +1667,118 @@ abstract contract SpokePool is
...
+            if (!isSlowFill) IERC20Upgradeable(outputToken).safeTransferFrom(msg.sender, address(this), amountToSend);
+            _unwrapwrappedNativeTokenTo(payable(recipientToSend), amountToSend);
+            // Else, this is a normal ERC20 token. Send to recipient.
+        } else {
+            // Note: Similar to note above, send token directly from the contract to the user in the slow relay case.
+            if (!isSlowFill) IERC20Upgradeable(outputToken).safeTransferFrom(msg.sender, recipientToSend, amountToSend);
+            else IERC20Upgradeable(outputToken).safeTransfer(recipientToSend, amountToSend);
+        }
+
+        // @TODO Can we remove the third param in handleAcrossMessage since we no longer support partial fills? Will
```
mrice32, line 1742: Ideally, yes. Maybe we should ask that integrators implement both, deprecate the old one and schedule a switch-over date at the launch of the USS. Thoughts?

contracts/SpokePool.sol
```diff
@@ -1621,6 +1667,118 @@ abstract contract SpokePool is
...
+            relayData.exclusivityDeadline,
+            relayData.exclusiveRelayer,
+            relayer,
+            relayData.depositor,
+            relayData.recipient,
+            relayData.message,
+            replacedSlowFillExecution
+        );
+
+        emit USSRelayExecution(
```
mrice32, line 1773: Is this event new? Why two events? (Sorry if we've already discussed this and I'm forgetting)

mrice32 (commented):
Looks awesome! A few minor comments

contracts/SpokePool.sol
nicholaspai: the point I'm trying to make here is that `destinationChainId` isn't used in `_fillRelay(USS)` and its also not emitted in `FilledUSSRelay`, but I think I'm overlooking the fact actually that its part of the relayHash, oops

contracts/SpokePool.sol
nicholaspai: I think that's the goal yeah. If a slow fill is requested but then followed by a fast fill in a subsequent bundle, then funds need to be returned to Hub. If the fast fill is in same bundle, nothing to do.

contracts/SpokePool.sol
nicholaspai: I think we still need to use payoutAdjustmentPct to rebate the recipient their relayer fee component, right? Because we can't modify the output amount without changing the relayHash

contracts/SpokePool.sol
nicholaspai: Ahh yes good catch. Great catch.

contracts/SpokePool.sol
```diff
@@ -638,50 +638,61 @@ abstract contract SpokePool is
...
         // SpokePool time stalls or lags significantly, it is still possible to make deposits by setting quoteTimestamp
-        // within the configured buffer. The owner should pause deposits if this is undesirable. This will underflow if
-        // quoteTimestamp is more than depositQuoteTimeBuffer; this is safe but will throw an unintuitive error.
+        // within the configured buffer. The owner should pause deposits/fills if this is undesirable.
+        // This will underflow if quoteTimestamp is more than depositQuoteTimeBuffer;
+        // this is safe but will throw an unintuitive error.
+
         // slither-disable-next-line timestamp
-        require(getCurrentTime() - quoteTimestamp <= depositQuoteTimeBuffer, "invalid quoteTimestamp");
+        if (getCurrentTime() - quoteTimestamp > depositQuoteTimeBuffer) revert InvalidQuoteTimestamp();
```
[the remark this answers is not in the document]
nicholaspai: That's right. I have been trying to get more code in per PR due to speed of reviews currently but I'll be more proactive in asking for reviews

contracts/SpokePool.sol
nicholaspai: Because the normal FilledRelay event has too many params including these such that the params emitted would have to be packed into a struct which is not nice for integrators. Moreover these params emitted here are not as commonly used. In fact the relayer-v2 repo doesn't use them so I thought it'd be better to emit them separately. 

The trade off of course is gas cost

contracts/SpokePool.sol
nicholaspai: +1. I'm not sure how much an num costs bytecode-wise but a custom error is encoded like a function signature:

"Using a custom error instance will usually be much cheaper than a string description, because you can use the name of the error to describe it, which is encoded in only four bytes. A longer description can be supplied via NatSpec which does not incur any costs."

https://docs.soliditylang.org/en/latest/control-structures.html#revert

contracts/SpokePool.sol
nicholaspai: do you understand it differently?

contracts/SpokePool.sol
nicholaspai: That's an interesting idea. @pxrl do you think that's something that could be useful? Also do we have an accurate idea of the total integrator set?

contracts/SpokePool.sol
mrice32: I think another tradeoff is that it is pretty tricky to associate this second event with the info in the first and vice-versa, making the event much harder to use.

contracts/SpokePool.sol
mrice32: Great point. Would it be simpler to just pass it in as an updatedOutputAmount to this function rather than a payout adjustment? Or would that require too many structural changes to the slow relay?

contracts/SpokePool.sol
mrice32: Nope, that was my understanding, just making sure.

In my mind the slow fill is sort of the fallback path, so I wouldn't want exclusivity to be able to block a slow fill. In the worst case, if the exclusivity params are fat fingered, it would be nice to be able to fall back on slow fills for a stuck deposit. So I would argue for fewer limitations.

Thoughts?

mrice32 (commented):
Replied

contracts/SpokePool.sol
nicholaspai: I'd agree that we don't really want restrictions on this function wrt exclusivity

contracts/SpokePool.sol
nicholaspai: I think updatedOutputAmount and payoutAdjustment are the same in terms of calldata and extra data stored in the relay execution struct, right? If so then I'd opt for the former because it's a more intuitive name and you can eliminate the percentage multiplication logic that probably adds gas

contracts/SpokePool.sol
nicholaspai: Hmm so maybe I'll revisit packing these updated params into a structure that the relayer can otherwise ignore in normal behavior

contracts/SpokePool.sol
nicholaspai: oh wait nm, updatedOutputAmount is already in the struct! this is def an improvement +1

contracts/SpokePool.sol
nicholaspai: Removed this comment

contracts/SpokePool.sol
nicholaspai: 03ba553e4a85d2f4940c2bb7ec216e44dd4ace56

contracts/SpokePool.sol
nicholaspai: Consolidated the event a bit here d5b4023329fb05baa5d065332b6a124f09c7ecce

contracts/SpokePool.sol
nicholaspai: 015c116193aa1c716a5e4f81035712395fce5235

mrice32 (approved):
LGTM!

contracts/SpokePool.sol
pxrl: My inclination is get this in soon with a breaking change, since we only have a few integrators at the moment. Ideally we can coordinate this with them so that they're ready to update when the new interface goes live. I'll check with the integrators directly.

PR #381, opened 2023-12-13
feat(scroll): adapter and spokepool

james-a-morris:
This PR creates the Chain Adapter for Scroll. 

#### Deployments:
| Network | Contract Name | Address|
| -------- | ----------------|---------|
| Sepolia  | Scroll Adapter   | [0x99EC530a761E68a377593888D9504002Bd191717](https://sepolia.etherscan.io/address/0x99EC530a761E68a377593888D9504002Bd191717) |
| Sepolia | Mock Hubpool (for l1->l2 token transfers) | [0xa9F2Ba6333B78375b967d39CFdB38feE28288F2c](https://sepolia.etherscan.io/address/0xa9F2Ba6333B78375b967d39CFdB38feE28288F2c)|
| Sepolia | Hubpool (for SetDepositRoute) | [0x14224e63716afAcE30C9a417E0542281869f7d9e](https://sepolia.etherscan.io/address/0x14224e63716afAcE30C9a417E0542281869f7d9e) |
| Sepolia Scroll | Spoke Pool | [0x76f3fe966F91602129cb278043239afBB7B7646A](https://sepolia.scrollscan.com/address/0x76f3fe966F91602129cb278043239afBB7B7646A)

| Type | L1 Transaction      | L2 Txn Confirmation |
| ----------- |  ----------- | ----------- |
| ERC-20      | [0xc2a...8d8](https://sepolia.etherscan.io/tx/0xc2a03e66d6404ee889c45b971e499a1d6260c2cf496324597e0abafcfdffc8d8) | [0xddd...125](https://sepolia.scrollscan.com/tx/0xddd99c4751a487285bc7d9bbf23d191cd0428c7338add8bb5d0d8ac71ccc3125) |
| WETH         | [0xb42...5c1](https://sepolia.etherscan.io/tx/0xb423e65ab1060d2b7b8148ce008f07265dce0ea4356cb12abf7f8b9d9fc335c1) | [0xb1c...532](https://sepolia.scrollscan.com/tx/0xb1c6520a872a139cb2fd64cd999084a4e9a5b307282c241725bdfd8d765d6532)
| Set Deposit Route | [0x4a9...546](https://sepolia.etherscan.io/tx/0x4a985de26cd3ab33058ebcff8848d1e0ebc58f9561b9464d25de0c39a134c546) | [0xbba...af0](https://sepolia.scrollscan.com/tx/0xbbad1a6198d4b1a97c942717db3f6fc7b90b3384fdb7e1b0516fa5037794caf0)

---
Closes ACX-1753, ACX-1757, ACX-1765, ACX-1780


contracts/chain-adapters/Scroll_Adapter.sol
```diff
@@ -0,0 +1,77 @@
...
+     * to the target contract on L2.
+     * @param target L2 address to send message to.
+     * @param message Message to send to `target`.
+     */
+    function relayMessage(address target, bytes calldata message) external payable {
+        // We can specifically send a message with 0 value to the Scroll Bridge
+        // and it will not forward any ETH to the target contract on L2. However,
+        // we need to set the payable value to msg.value to ensure that the Scroll
+        // Bridge has enough gas to forward the message to L2.
+        l1ScrollMessenger.sendMessage{ value: msg.value }(target, 0, message, l2GasLimit);
```
nicholaspai, line 40: What happens with the excess value if we set the second param to 0? Do we get refunded on L1?

james-a-morris: Correct - the bridge calls our fallback function with the excess value.

nicholaspai, line 40: Does the `msg.value` have to account in any way for the l2 gas cost? Also, does this contract have to hold ETH to pay for the message like the `arbitrum_adapter` does or does it do something more like the `ovm_adapter` where the L2 recipient pays for the gas

contracts/chain-adapters/Scroll_Adapter.sol
```diff
@@ -0,0 +1,77 @@
...
+        require(_l2Token == l2Token, "l2Token Mismatch");
+
+        // Bump the allowance
+        IERC20(l1Token).safeIncreaseAllowance(address(_l1GatewayRouter), amount);
+
+        // The scroll bridge handles arbitrary ERC20 tokens and is mindful of
+        // the official WETH address on-chain. We don't need to do anything specific
+        // to differentiate between WETH and a separate ERC20.
+        // Note: This happens due to the L1GatewayRouter.getERC20Gateway() call
+        _l1GatewayRouter.depositERC20{ value: msg.value }(l1Token, to, amount, l2GasLimit);
```
nicholaspai, line 74: OOC: Does the recipient receive WETH when sending WETH or ETH on the L2?

nicholaspai (commented):
This looks great, how did it perform with integration tests?

contracts/chain-adapters/Scroll_Adapter.sol
nicholaspai: where is that documented OOC?

nicholaspai: for reference, when executing pool rebalance leaves we need to send ETH to the HubPool: https://github.com/across-protocol/relayer-v2/blob/master/src/dataworker/Dataworker.ts#L1828

deploy/026_deploy_scroll_adapter.ts
pxrl, line 1: Beware of not colliding on 026 w/ @dohaki on his PR: https://github.com/across-protocol/contracts-v2/pull/382/files#diff-717b139563fe4a8cdcfced14928b8c23bb6d72bbc281eb20299bc09ec2568fc4

pxrl (commented):
Some quick initial comments.

contracts/Scroll_SpokePool.sol
```diff
@@ -0,0 +1,33 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.0;
```
pxrl, line 2: Pretty sure we've bumped up to ^0.8.22 or 0.8.23 elsewhere - might be worth following that here.

james-a-morris: Not sure we want to do this. The `^0.8.0` denotes we want the latest version within `^0.8.X`

deploy/026_deploy_scroll_adapter.ts
james-a-morris: Thanks for bringing this to my attention. I'll make sure to rebase with master before that merging.

contracts/chain-adapters/Scroll_Adapter.sol
mrice32: Similar comment: https://github.com/across-protocol/contracts-v2/pull/382#discussion_r1433463127.

We probably don't want to require the executor to send funds with the call to the HubPool.

contracts/chain-adapters/Scroll_Adapter.sol
```diff
@@ -0,0 +1,77 @@
...
+     * to be paid by the sender to forward the message to L2. However, it will not send msg.value
+     * to the target contract on L2.
+     * @param target L2 address to send message to.
+     * @param message Message to send to `target`.
+     */
+    function relayMessage(address target, bytes calldata message) external payable {
+        // We can specifically send a message with 0 value to the Scroll Bridge
+        // and it will not forward any ETH to the target contract on L2. However,
+        // we need to set the payable value to msg.value to ensure that the Scroll
+        // Bridge has enough gas to forward the message to L2.
```
mrice32, line 39: Are we able to compute the cost of l2Gas? For instance, if we know we're sending 250k L2 gas, are we able to compute the cost of that gas in the transaction so we don't get a refund back (or underpay)?

contracts/chain-adapters/Scroll_Adapter.sol
mrice32, line 74: Same comment about msg.value as above. We should compute the amount we need to send and use the funds already loaded in the contract instead of expecting the caller to provide it, as that's generally the pattern we've used previously and is simpler for the offchain bots.

contracts/Scroll_SpokePool.sol
```diff
@@ -0,0 +1,64 @@
...
+     **************************************/
+
+    function _bridgeTokensToHubPool(uint256 amountToReturn, address l2TokenAddress) internal virtual override {
+        IL2GatewayRouter _l2GatewayRouter = l2GatewayRouter;
+
+        // The scroll bridge handles arbitrary ERC20 tokens and is mindful of
+        // the official WETH address on-chain. We don't need to do anything specific
+        // to differentiate between WETH and a separate ERC20.
+        // Note: This happens due to the L2GatewayRouter.getERC20Gateway() call
+        _l2GatewayRouter.withdrawERC20{ value: msg.value }(l2TokenAddress, hubPool, amountToReturn, l1GasLimit);
```
mrice32, line 57: We have to send ETH on every withdrawal? This is sort of atypical? Are they finalizing these withdrawals for us?

contracts/Scroll_SpokePool.sol
```diff
@@ -0,0 +1,64 @@
...
+     */
+    function initialize(
+        IL2GatewayRouter _l2GatewayRouter,
+        uint32 _initialDepositId,
+        address _crossDomainAdmin,
+        address _hubPool
+    ) public initializer {
+        __SpokePool_init(_initialDepositId, _crossDomainAdmin, _hubPool);
+        l2GatewayRouter = _l2GatewayRouter;
+        l1GasLimit = 250_000;
```
mrice32, line 43: What happens if you supply too little L1 gas?

mrice32 (commented):
Looks great! A few questions

contracts/chain-adapters/Scroll_Adapter.sol
james-a-morris: It's documented here: https://docs.scroll.io/en/developers/l1-and-l2-bridging/eth-and-erc20-token-bridge/

```
depositETH and depositERC20 are payable functions, the amount of ETH sent to these functions will be used to pay for L2 fees. If the amount is not enough, the transaction will not be sent. All excess ETH will be sent back to the sender. 0.00001 ETH should be more than enough to process a token deposit.
```

contracts/chain-adapters/Scroll_Adapter.sol
james-a-morris: They will receive WETH. The `depositERC20` converts WETH -> ETH, bridges the ETH, then converts `ETH -> WETH` on scroll

contracts/Scroll_SpokePool.sol
james-a-morris: My understanding from their documentation is that they are finalizing the withdrawals for us. The docs aren't fully clear, but we are paying a relayer fee with the payable value:

```
    /// @notice Withdraw of some token to a recipient's account on L1.
    /// @dev Make this function payable to send relayer fee in Ether.
    /// @param token The address of token in L2.
    /// @param to The address of recipient's account on L1.
    /// @param amount The amount of token to transfer.
    /// @param gasLimit Unused, but included for potential forward compatibility considerations.
    function withdrawERC20(
        address token,
        address to,
        uint256 amount,
        uint256 gasLimit
    ) external payable;
```

contracts/Scroll_SpokePool.sol
james-a-morris: Per docs, this can actually be set to 0 because it's unused for now. I was including it for symmetry but it's not needed L2 -> L1

Docs:
```
    /// @notice Withdraw of some token to a recipient's account on L1.
    /// @dev Make this function payable to send relayer fee in Ether.
    /// @param token The address of token in L2.
    /// @param to The address of recipient's account on L1.
    /// @param amount The amount of token to transfer.
    /// @param gasLimit Unused, but included for potential forward compatibility considerations.
    function withdrawERC20(
        address token,
        address to,
        uint256 amount,
        uint256 gasLimit
    ) external payable;
```

As a note, it does look like the `relayerFee` is sent along to facilitate the finalization

contracts/chain-adapters/Scroll_Adapter.sol
james-a-morris: I can't seem to find any explicit docs - I will forward via TG

contracts/Scroll_SpokePool.sol
mrice32: I would set it to 0 then, and we can always finalize ourselves, right?

contracts/chain-adapters/Scroll_Adapter.sol
mrice32: If they are giving a synchronous refund, there should be some way to compute the cost -- hopefully just a public function.

contracts/Scroll_SpokePool.sol
james-a-morris: Yep! The updated version sets it to 0

contracts/chain-adapters/Scroll_Adapter.sol
nicholaspai: can you document this link in the code if not already?

nicholaspai: Since adapters and chain specific spoke pools are network specific I like leaving links to technical documentation to help with future upgrades

nicholaspai: Agreed, but we may have to do something like in Arbitrum_Adapter where we assume the HubPool has a balance of ETH from which to draw its msg.value from

contracts/Scroll_SpokePool.sol
nicholaspai: yeah if hardhat.config specifies 0.8.23 then this will be compiled with that

contracts/chain-adapters/Scroll_Adapter.sol
nicholaspai: Agreed +1

contracts/chain-adapters/Scroll_Adapter.sol
```diff
@@ -0,0 +1,97 @@
...
+    ) {
+        l1GatewayRouter = _l1GatewayRouter;
+        l1ScrollMessenger = _l1ScrollMessenger;
+        l2GasPriceOracle = _l2GasPriceOracle;
+    }
+
+    /**
+     * @notice Generates the relayer fee for a message to be sent to L2.
+     */
+    function _generateRelayerFee() internal view returns (uint256) {
```
nicholaspai, line 43: nit add internal functions under a comment:

```
     *        INTERNAL FUNCTIONS          *
```

contracts/chain-adapters/Scroll_Adapter.sol
james-a-morris: There are comments in the code that give an explainer, but I added an actual link here: https://github.com/across-protocol/contracts-v2/pull/381/commits/272245aeacfc9c7d9387444428a5fbef2e972628

james-a-morris: Yep - the new implementation will need to call `loadEthForL2Calls` but `msg.value` is no longer relied upon

contracts/chain-adapters/Scroll_Adapter.sol
james-a-morris: Done - we resolved this offline @mrice32 but now we are referencing the l2gas oracle on mainnet to get the max fee to send to the contracts.

contracts/chain-adapters/Scroll_Adapter.sol
james-a-morris: Removed in a previous commit: [ref](https://github.com/across-protocol/contracts-v2/pull/381#discussion_r1437140593)

contracts/Scroll_SpokePool.sol
james-a-morris: @mrice32 I was able to update this such that we can call these contracts without payable. We have a separate script now for finalizing scroll txns and that should be able to be integrated into the finalizer.

contracts/chain-adapters/Scroll_Adapter.sol
james-a-morris: Done https://github.com/across-protocol/contracts-v2/pull/381/commits/19ab1c3b1715a45a97b1b0a094f6d2c80327a44b

contracts/Scroll_SpokePool.sol
```diff
@@ -0,0 +1,85 @@
...
+    /**
+     * @notice Verifies that calling method is from the cross domain admin.
+     */
+    function _requireAdminSender() internal view override {
+        // The xdomainMessageSender is set within the Scroll messenger right
+        // before the call to this function (and reset afterwards). This represents
+        // the address that sent the message from L1 to L2. If the calling contract
+        // isn't the Scroll messenger, then the xdomainMessageSender will be the zero
+        // address and *NOT* cross domain admin.
+        address _xDomainSender = l2ScrollMessenger.xDomainMessageSender();
```
nicholaspai, line 82: This is a really bizarre/unique choice by Scroll for validating the message but I think it works? IIUC, [`_executeMessage`](https://github.com/scroll-tech/scroll/blob/develop/contracts/src/L2/L2ScrollMessenger.sol#L142) is called by some bot (Scroll bot? or do we have to build our own?) which then calls into this contract to execute the desired function

contracts/chain-adapters/Scroll_Adapter.sol
```diff
@@ -0,0 +1,106 @@
...
+
+    /**************************************
+     *         INTERNAL FUNCTIONS         *
+     **************************************/
+
+    /**
+     * @notice Generates the relayer fee for a message to be sent to L2.
+     */
+    function _generateRelayerFee() internal view returns (uint256) {
+        return l2GasPriceOracle.estimateCrossDomainMessageFee(l2GasLimit);
```
nicholaspai, line 104: Do we need a buffer on this at all?

contracts/Scroll_SpokePool.sol
james-a-morris: The `relayMessage` is meant to be run by privileged accounts - specifically the counterpart L1Messenger.

```
require(AddressAliasHelper.undoL1ToL2Alias(_msgSender()) == counterpart, "Caller is not L1ScrollMessenger");
```

contracts/chain-adapters/Scroll_Adapter.sol
james-a-morris: Do you mean like a gas padding? We could do that - the L2 price oracle is updated by a scroll bot so it may be less than accurate during high network activity.

Do you think we should add that into the code and initialize it to zero so that if we need it, we can use it

contracts/Scroll_SpokePool.sol
nicholaspai: Wow so we have to finalize L1 to L2 messages for scroll in addition to linea? ugh

contracts/chain-adapters/Scroll_Adapter.sol
nicholaspai: fee padding, not just gas-limit padding. I guess the `gasLimit` value can be used as a buffer, right?

james-a-morris: Sorry, that's what I meant. The value we're propagating in the `{ value: _generateRelayerFee() }`

contracts/Scroll_SpokePool.sol
```diff
@@ -0,0 +1,115 @@
...
+import "@scroll-tech/contracts/L2/gateways/IL2GatewayRouter.sol";
+import "@scroll-tech/contracts/libraries/IScrollMessenger.sol";
+
+/**
+ * @title Scroll_SpokePool
+ * @notice Modified SpokePool contract deployed on Scroll to facilitate token transfers
+ * from Scroll to the HubPool
+ */
+contract Scroll_SpokePool is SpokePool {
+    IL2GatewayRouter public l2GatewayRouter;
```
nicholaspai, line 14: Should add comments to any state variables

contracts/chain-adapters/Scroll_Adapter.sol
```diff
@@ -0,0 +1,106 @@
...
+import "./interfaces/AdapterInterface.sol";
+
+/**
+ * @title Scroll_Adapter
+ * @notice Adapter contract deployed on L1 alongside the HubPool to facilitate token transfers
+ * and arbitrary message relaying from L1 to L2.
+ */
+contract Scroll_Adapter is AdapterInterface {
+    using SafeERC20 for IERC20;
+    uint32 public immutable l2GasLimit = 250_000;
```
nicholaspai, line 18: Same comment as for the SpokePool, need to add comments to state vars

nicholaspai, line 18: nit: why not use `constant` instead of `immutable`..?

contracts/chain-adapters/Scroll_Adapter.sol
nicholaspai, line 104: Can we error earlier if this contract doesn't have enough ETH to pay for the fee? See how we do so with `_contractHasSufficientEthBalance` in Arbitrum_Adapter

deploy/027_deploy_scroll_spokepool.ts
```diff
@@ -0,0 +1,26 @@
...
+  // with deprecated spoke pool.
+  // Set hub pool as cross domain admin since it delegatecalls the Adapter logic.
+  const initArgs = [
+    L2_ADDRESS_MAP[chainId].scrollERC20GatewayRouter,
+    L2_ADDRESS_MAP[chainId].scrollMessenger,
+    1_000_000,
+    hubPool.address,
+    hubPool.address,
+  ];
+  const constructorArgs = [L2_ADDRESS_MAP[chainId].l2Weth, 3600, 32400];
```
nicholaspai, line 21: what are these values? Please add comments

scripts/buildSampleTree.ts
```diff
@@ -14,15 +14,15 @@ const POOL_REBALANCE_LEAF_COUNT = 1;
 const RELAYER_REFUND_LEAF_COUNT = 1;
 const SLOW_RELAY_LEAF_COUNT = 1;
 const POOL_REBALANCE_NET_SEND_AMOUNT = 0.1; // Amount of tokens to send from HubPool to SpokePool
-const RELAYER_REFUND_AMOUNT_TO_RETURN = 0.1; // Amount of tokens to send from SpokePool to HubPool
-const L1_TOKEN = "0x40153ddfad90c49dbe3f5c9f96f2a5b25ec67461";
-const L2_TOKEN = "0x40153ddfad90c49dbe3f5c9f96f2a5b25ec67461";
+const RELAYER_REFUND_AMOUNT_TO_RETURN = 0; // Amount of tokens to send from SpokePool to HubPool
```
nicholaspai, line 17: nit: no need to modify this file, just keep changes local

tasks/finalizeScrollClaims.ts
```diff
@@ -0,0 +1,115 @@
...
+    const l2Operator = String(taskArguments.l2Address);
+    if (!hre_.ethers.utils.isAddress(l2Operator)) {
+      throw new Error("Invalid L2 operator address. Must pass as last argument to script");
+    }
+
+    const signer = (await (hre_ as any).ethers.getSigners())[0] as unknown as Signer;
+    const apiUrl = `https://${String(chainId) === "1" ? "mainnet" : "sepolia"}-api-bridge.scroll.io/api/claimable`;
+    const messengerContract = new Contract(L1_ADDRESS_MAP[chainId].scrollMessengerRelay, relayMessengerAbi, signer);
+    const claimList = (
+      await axios.get<{
```
nicholaspai, line 78: nit: prefer not using `axios` if we don't have to. Doesn't the SDK fetch this APIURL for us?

contracts/test/ERC20Test.sol
```diff
@@ -0,0 +1,11 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import "@uma/core/contracts/common/implementation/TestnetERC20.sol";
+
+/**
+ * @notice Simulated ERC20 for use in testing bridge transfers.
+ */
+contract ERC20Test is TestnetERC20 {
+    constructor() TestnetERC20("ERC20 Test", "ERC20_TEST", 18) {} // solhint-disable-line no-empty-blocks
```
nicholaspai, line 10: Why is this contract being committed to this repo? You could just deploy an ERC20 with:

```
  const erc20 = await (await getContractFactory("ExpandedERC20", deployerWallet)).deploy("ERC20 Test", "ERC20_TEST", 18);
```

james-a-morris: Holdover from testing: removed https://github.com/across-protocol/contracts-v2/pull/381/commits/d5c2afaec63b225bd68c3ae941e1c34d15a29dca

scripts/buildSampleTree.ts
james-a-morris: Holdover from testing: Done https://github.com/across-protocol/contracts-v2/pull/381/commits/9a80eb7b249186281c3abb5514d8b30eadf5f74e

tasks/finalizeScrollClaims.ts
james-a-morris: Which SDK? Scroll doesn't have one and this hasn't been covered by our Across SDK.

deploy/027_deploy_scroll_spokepool.ts
james-a-morris: Done: https://github.com/across-protocol/contracts-v2/pull/381/commits/f3fba07303819bafdb12f5acf4f0d7f552f6f2c2

These are identical to the other spoke pool deployments so I've opened a PR for this here: https://github.com/across-protocol/contracts-v2/pull/389

contracts/chain-adapters/Scroll_Adapter.sol
james-a-morris: This is a great callout. I was following the stylization of these adapters:
- Base
- Boba
- Optimism

That said, I agree this is known at compile time so switching to constant would make more logical sense.

Done: https://github.com/across-protocol/contracts-v2/pull/381/commits/c1bb99414ce0854a8ca478035b5fe7e031647d21

contracts/Scroll_SpokePool.sol
james-a-morris: Done: https://github.com/across-protocol/contracts-v2/pull/381/commits/a44f7569b26cccaba4418aad00cb7d487477a50b

contracts/chain-adapters/Scroll_Adapter.sol
james-a-morris: Done: https://github.com/across-protocol/contracts-v2/pull/381/commits/a44f7569b26cccaba4418aad00cb7d487477a50b

contracts/chain-adapters/Scroll_Adapter.sol
james-a-morris: Added with an additional test: https://github.com/across-protocol/contracts-v2/pull/381/commits/71eac9482d4f5021d0b4e69af573e1ecb9569a0b

contracts/chain-adapters/Scroll_Adapter.sol
```diff
@@ -0,0 +1,126 @@
...
+ * @notice Adapter contract deployed on L1 alongside the HubPool to facilitate token transfers
+ * and arbitrary message relaying from L1 to L2.
+ */
+contract Scroll_Adapter is AdapterInterface {
+    using SafeERC20 for IERC20;
+
+    /**
+     * @notice Used as the gas limit for relaying messages to L2.
+     */
+    uint32 public constant l2GasLimit = 250_000;
```
mrice32, line 22: nit: why not set this to immutable and allow it to be passed in on construction?

The advantage is that we don't have to change the contract code if we ever want to deploy a new version with a different gas limit.

contracts/chain-adapters/Scroll_Adapter.sol
```diff
@@ -0,0 +1,126 @@
...
+     * to the target contract on L2.
+     * @param target L2 address to send message to.
+     * @param message Message to send to `target`.
+     */
+    function relayMessage(address target, bytes calldata message) external payable {
+        // We can specifically send a message with 0 value to the Scroll Bridge
+        // and it will not forward any ETH to the target contract on L2. However,
+        // we need to set the payable value to msg.value to ensure that the Scroll
+        // Bridge has enough gas to forward the message to L2.
+        l1ScrollMessenger.sendMessage{ value: _generateRelayerFee() }(target, 0, message, l2GasLimit);
```
mrice32, line 75: If our gas limit is too high and gas is left unused on the destination, does that money get lost or is it refunded somehow?

mrice32 (approved):
Looks great! Just a few minor comments

contracts/chain-adapters/Scroll_Adapter.sol
james-a-morris: This was originally set to be immutable, but it was updated here: https://github.com/across-protocol/contracts-v2/pull/381#discussion_r1437736300

These adapters use immutable: 
- Base
- Boba
- Optimism

These use const:
- ZKSync
- Arbitrum

---
Should we be looking to normalize this?

contracts/Scroll_SpokePool.sol
```diff
@@ -0,0 +1,115 @@
...
+     *         INTERNAL FUNCTIONS         *
+     **************************************/
+
+    /**
+     * @notice Bridge tokens to the HubPool.
+     * @param amountToReturn Amount of tokens to bridge to the HubPool.
+     * @param l2TokenAddress Address of the token to bridge.
+     */
+    function _bridgeTokensToHubPool(uint256 amountToReturn, address l2TokenAddress) internal virtual override {
+        IL2GatewayRouter _l2GatewayRouter = l2GatewayRouter;
```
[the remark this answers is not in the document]
james-a-morris: I like dereferencing state variables as a point toward clarity. 

This should be optimized out by the compiler - but to be safe: do you recommend that this be updated?

contracts/Scroll_SpokePool.sol
james-a-morris: @nicholaspai just to confirm:

L1 -> L2 is handled entirely by Scroll. It's a one txn operation that is submitted to the L1.
L2 -> L1 requires us to finalize. One txn is submitted against L2. Then, we have to submit the claim txn on L1

contracts/chain-adapters/Scroll_Adapter.sol
nicholaspai: We should but not in this PR. `immutable` should only be used if we want to set at construction time, otherwise constant is fine.

tasks/finalizeScrollClaims.ts
nicholaspai: I just assumed blindly scroll would have an `npmjs` sdk, is that not the case? If so, then this lgtm

contracts/Scroll_SpokePool.sol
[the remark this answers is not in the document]
nicholaspai: I don't usually dereference as its pretty clear what's going on here and there are no gas optimizations

contracts/chain-adapters/Scroll_Adapter.sol
nicholaspai: I'm ok with going immutable if we want to redeploy in the future and change this var without changing contract code. This also makes the unit tests more scalable since we do test specifically for this value

contracts/Scroll_SpokePool.sol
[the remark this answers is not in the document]
james-a-morris: Done: https://github.com/across-protocol/contracts-v2/pull/381/commits/300735d8ec1d9b7ee7e127c79f89594ad57b78e8

contracts/chain-adapters/Scroll_Adapter.sol
james-a-morris: Done: https://github.com/across-protocol/contracts-v2/pull/381/commits/300735d8ec1d9b7ee7e127c79f89594ad57b78e8

james-a-morris:
@nicholaspai @mrice32 @pxrl updated the test txns in this PR's description to the latest txn tests as of https://github.com/across-protocol/contracts-v2/pull/381/commits/300735d8ec1d9b7ee7e127c79f89594ad57b78e8

contracts/Scroll_SpokePool.sol
```diff
@@ -0,0 +1,123 @@
...
+    /**
+     * @notice Verifies that calling method is from the cross domain admin.
+     */
+    function _requireAdminSender() internal view override {
+        // The xdomainMessageSender is set within the Scroll messenger right
+        // before the call to this function (and reset afterwards). This represents
+        // the address that sent the message from L1 to L2. If the calling contract
+        // isn't the Scroll messenger, then the xdomainMessageSender will be the zero
+        // address and *NOT* cross domain admin.
+        address _xDomainSender = l2ScrollMessenger.xDomainMessageSender();
```
nicholaspai, line 108: Can xDomainMessageSender be set to whatever a malicious contract wants it to or is it always the contract that is the `target` of the L1 to L2 message?

nicholaspai (approved):
This looks great. LGTM after we get an answer about excess xchain message fees and I also left two additional comments

contracts/Scroll_SpokePool.sol
```diff
@@ -0,0 +1,123 @@
...
+     * @notice Bridge tokens to the HubPool.
+     * @param amountToReturn Amount of tokens to bridge to the HubPool.
+     * @param l2TokenAddress Address of the token to bridge.
+     */
+    function _bridgeTokensToHubPool(uint256 amountToReturn, address l2TokenAddress) internal virtual override {
+        // The scroll bridge handles arbitrary ERC20 tokens and is mindful of
+        // the official WETH address on-chain. We don't need to do anything specific
+        // to differentiate between WETH and a separate ERC20.
+        // Note: This happens due to the L2GatewayRouter.getERC20Gateway() call
+        l2GatewayRouter.withdrawERC20(l2TokenAddress, hubPool, amountToReturn, 0);
```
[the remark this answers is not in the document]
james-a-morris: Done: https://github.com/across-protocol/contracts-v2/pull/381/commits/3c52c3388d0aaea85c7568aae6ca7bf7d03c1356

It was just a holdover value for forward compatibility reasons. Left an extensive comment including docs back to the calling fn.

contracts/Scroll_SpokePool.sol
james-a-morris: (I'm going to go through each call in the stack so this might be longform, but tl;dr I believe we should be fine as long as we trust the chain)

---

Per master on `scroll`'s official github ( [link](https://github.com/scroll-tech/scroll/blob/develop/contracts/src/L2/L2ScrollMessenger.sol#L142-L160) ), the `xDomainMessenger` is set to be the `_from` sender (sender of the L1 txn) here. The transaction stack is `l2TokenMessenger` calls `relayMessage` -> `_executeMessage`.


### _executeMessage
```
function _executeMessage(
        address _from,
        address _to,
        uint256 _value,
        bytes memory _message,
        bytes32 _xDomainCalldataHash
    ) internal {
        // @note check more `_to` address to avoid attack in the future when we add more gateways.
        require(_to != messageQueue, "Forbid to call message queue");
        _validateTargetAddress(_to);

        // @note This usually will never happen, just in case.
        require(_from != xDomainMessageSender, "Invalid message sender"); 
```
the last line of this snippet always set to be the DEFAULT_XDOMAIN_MESSAGE_SENDER outside of this specific case. This is a call to ensure that doesn't happen

```
        xDomainMessageSender = _from;
        // solhint-disable-next-line avoid-low-level-calls
        (bool success, ) = _to.call{value: _value}(_message);
        // reset value to refund gas.
        xDomainMessageSender = ScrollConstants.DEFAULT_XDOMAIN_MESSAGE_SENDER; <--- Immediately resets
        ...
    }
```
In the above ^ it's only set to be read within the context of `_to.call(...)` and then is immediately reset

### relayMessage

```
    function relayMessage(
        address _from,
        address _to,
        uint256 _value,
        uint256 _nonce,
        bytes memory _message
    ) external override whenNotPaused {
        // It is impossible to deploy a contract with the same address, reentrance is prevented in nature.
        require(AddressAliasHelper.undoL1ToL2Alias(_msgSender()) == counterpart, "Caller is not L1ScrollMessenger");

        bytes32 _xDomainCalldataHash = keccak256(_encodeXDomainCalldata(_from, _to, _value, _nonce, _message));

        require(!isL1MessageExecuted[_xDomainCalldataHash], "Message was already successfully executed");

        _executeMessage(_from, _to, _value, _message, _xDomainCalldataHash);
    }
```

From the above, what we're really looking for is that `_msgSender()` which is the `msg.sender` is the counterpart domain of the `L1ScrollMessenger`. Ergo if we pass that require, we're on the `L2ScrollMessenger`.

### L2ScrollMessenger

We have an implicit trust that the `L2ScrollMessenger` is going to set things right.

contracts/chain-adapters/Scroll_Adapter.sol
james-a-morris: Per the TG group chat with the Scroll team, our ETH may get burned if we overestimate our gas limit. However, we were told in a previous message that 250,000 is an acceptable amount.

james-a-morris:
@nicholaspai @mrice32 @dohaki I've added a second gas limit so that we can differentiate between arbitrary messages & token relays

dohaki (approved):
🔥 LGTM

contracts/chain-adapters/Scroll_Adapter.sol
```diff
@@ -0,0 +1,156 @@
...
+        address to
+    ) external payable {
+        IL1GatewayRouter _l1GatewayRouter = l1GatewayRouter;
+
+        // Confirm that the l2Token that we're trying to send is the correct counterpart
+        // address
+        address _l2Token = _l1GatewayRouter.getL2ERC20Address(l1Token);
+        require(_l2Token == l2Token, "l2Token Mismatch");
+
+        // Bump the allowance
```
nicholaspai, line 118:
```suggestion
```

unneeded comment, code is self explanatory

contracts/chain-adapters/Scroll_Adapter.sol
```diff
@@ -0,0 +1,156 @@
...
+
+    /**************************************
+     *         INTERNAL FUNCTIONS         *
+     **************************************/
+
+    /**
+     * @notice Generates the relayer fee for a message to be sent to L2.
+     * @dev Function will revert if the contract does not have enough ETH to pay the fee.
+     */
+    function _generateRelayerFeeForMessageRelay() internal view returns (uint256 l2Fee) {
```
nicholaspai, line 143: would have implemented this with a single internal function where you pass in `uint256 l2GasLimit`.

Then, within relayTokens and relayMessage you pass in a different constant

nicholaspai (approved):
Looks good to go, I left two comments on making code better

mrice32 (approved):
LGTM!

contracts/chain-adapters/Scroll_Adapter.sol
james-a-morris: Done: https://github.com/across-protocol/contracts-v2/pull/381/commits/e9b44430fb024864fcee3b1abacdefd285221302

PR #383, opened 2023-12-18
fix: use updated deployNewProxy fn

james-a-morris:
The current references to `deployNewProxy` result in compilation errors. We should modify `deployNewProxy` so that it can pass an initializer & constructor args and call that fn correctly.

### DIFF
--- a/deploy/003_deploy_optimism_spokepool.ts
+++ b/deploy/003_deploy_optimism_spokepool.ts
@@ -9,8 +9,9 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
   // with deprecated spoke pool.
   // Set hub pool as cross domain admin since it delegatecalls the Adapter logic.
   const initArgs = [1_000_000, hubPool.address, hubPool.address];
-  await deployNewProxy("Optimism_SpokePool", initArgs, {
+  await deployNewProxy("Optimism_SpokePool", {
     constructorArgs: ["0x4200000000000000000000000000000000000006", 3600, 32400],
+    initArgs,
   });
 };
 module.exports = func;
--- a/deploy/003_deploy_optimism_spokepool.ts
+++ b/deploy/003_deploy_optimism_spokepool.ts
@@ -9,10 +9,7 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
   // with deprecated spoke pool.
   // Set hub pool as cross domain admin since it delegatecalls the Adapter logic.
   const initArgs = [1_000_000, hubPool.address, hubPool.address];
-  await deployNewProxy("Optimism_SpokePool", {
-    constructorArgs: ["0x4200000000000000000000000000000000000006", 3600, 32400],
-    initArgs,
-  });
+  await deployNewProxy("Optimism_SpokePool", ["0x4200000000000000000000000000000000000006", 3600, 32400], initArgs);
 };
 module.exports = func;
 func.tags = ["OptimismSpokePool", "optimism"];
--- a/deploy/003_deploy_optimism_spokepool.ts
+++ b/deploy/003_deploy_optimism_spokepool.ts
@@ -9,7 +9,8 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
   // with deprecated spoke pool.
   // Set hub pool as cross domain admin since it delegatecalls the Adapter logic.
   const initArgs = [1_000_000, hubPool.address, hubPool.address];
-  await deployNewProxy("Optimism_SpokePool", ["0x4200000000000000000000000000000000000006", 3600, 32400], initArgs);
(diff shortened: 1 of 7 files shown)


nicholaspai (approved):

deploy/003_deploy_optimism_spokepool.ts
```diff
@@ -9,9 +9,7 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
   // with deprecated spoke pool.
   // Set hub pool as cross domain admin since it delegatecalls the Adapter logic.
   const initArgs = [1_000_000, hubPool.address, hubPool.address];
-  await deployNewProxy("Optimism_SpokePool", initArgs, {
-    constructorArgs: ["0x4200000000000000000000000000000000000006", 3600, 32400],
-  });
+  await deployNewProxy("Optimism_SpokePool", ["0x4200000000000000000000000000000000000006", 3600, 32400], initArgs);
```
pxrl, line 12: It'd be nice to explicitly call out the constructor args -wdyt?

```suggestion
  const constructorArgs = ["0x4200000000000000000000000000000000000006", 3600, 32400];
  await deployNewProxy("Optimism_SpokePool", constructorArgs, initArgs);
```

utils/utils.hre.ts
```diff
@@ -23,12 +23,15 @@ export async function getSpokePoolDeploymentInfo(
   return { hubPool, hubChainId, spokeChainId };
 }
 
-export async function deployNewProxy(name: string, args: (number | string)[]): Promise<void> {
+type FnArgs = number | string;
+export async function deployNewProxy(name: string, constructorArgs: FnArgs[], initArgs: FnArgs[]): Promise<void> {
```
pxrl, line 27: Nice to name it as it is used in `run("verify:verify": ...);`.

```suggestion
export async function deployNewProxy(name: string, constructorArguments: FnArgs[], initArgs: FnArgs[]): Promise<void> {
```

pxrl (approved):
Couple of minor comments.

deploy/003_deploy_optimism_spokepool.ts
james-a-morris: This was an existing pattern in the code. Do you think it should be changed in a separate PR?

utils/utils.hre.ts
james-a-morris: Fair but this is also used slightly further down as `constructorArgs` in the `deployProxy` fn. 

```
  const proxy = await upgrades.deployProxy(await getContractFactory(name, {}), initArgs, {
    kind: "uups",
    unsafeAllow: ["delegatecall"], // Remove after upgrading openzeppelin-contracts-upgradeable post v4.9.3.
    constructorArgs,
    initializer: "initialize",
  });
```

deploy/003_deploy_optimism_spokepool.ts
pxrl: It's directly related to this change so I think we should just fix it in place :+1:

utils/utils.hre.ts
pxrl: Ah - that's annoying. Ignore this one then :+1:

deploy/003_deploy_optimism_spokepool.ts
james-a-morris: Done

deploy/011_deploy_polygon_spokepool.ts
```diff
@@ -18,9 +18,8 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
     hubPool.address,
     L2_ADDRESS_MAP[spokeChainId].fxChild,
   ];
-  await deployNewProxy("Polygon_SpokePool", initArgs, {
-    constructorArgs: [L2_ADDRESS_MAP[spokeChainId].wMatic, 3600, 32400],
-  });
+  const constructorArgs = [L2_ADDRESS_MAP[spokeChainId].wMatic, 3600, 32400];
+  await deployNewProxy("Polygon_SpokePool", constructorArgs, initArgs);
```
pxrl, line 22: Is the order of `constructorArgs` vs. `initArgs` correct on this one? It looks like they're now reversed vs. previous, and `consutrctorArgs` used to be supplied as part of a larger object.

pxrl: Nevermind - I see that the prototype for deployNewProxy has also been updated :+1:

james-a-morris: The Polygon Spoke's constructor takes 3 items:
```
constructor(
        uint32 _fillDeadlineBuffer
    )
```

The `initialize` function (`initArgs`) takes 5:
```
function initialize(
        uint32 _initialDepositId,
        PolygonTokenBridger _polygonTokenBridger,
        address _crossDomainAdmin,
        address _hubPool,
        address _fxChild
    )
```

PR #384, opened 2023-12-18
feat(SpokePool): Remove maxCount related logic

nicholaspai:
Supporting maxcount adds extra gas cost to soon-to-be deprecated deposit and fill functions. This PR removes them and also removes related functions like `requestRepaymentChainId`.

A maxCount is likely to be added in future to new deposit and fill functions (i.e. USS-functions) but we'll add them in a follow-up PR


contracts/SpokePool.sol
```diff
@@ -447,7 +454,7 @@ abstract contract SpokePool is
         int64 relayerFeePct,
         uint32 quoteTimestamp,
         bytes memory message,
-        uint256 maxCount
+        uint256 // maxCount. Deprecated.
```
james-a-morris, line 457: Should we still keep this as a named parameter - even if we're ignoring it?

james-a-morris (commented):
Left a comment about signatures. I will follow up with more comments in the AM

contracts/SpokePool.sol
nicholaspai: that would trigger the linter since its unused, why keep it named?

pxrl (commented):
No blockers from me but I'm interested to see what other comments James might have.

mrice32 (approved):
LGTM!

PR #385, opened 2023-12-19
feat(SpokePool): Require executeX() caller to be EOA

nicholaspai:
Useful safety feature to protect function that can send xchain messages. Also adds useful scaffolding for USS unit tests

### DIFF
--- a/contracts/Polygon_SpokePool.sol
+++ b/contracts/Polygon_SpokePool.sol
@@ -44,12 +44,16 @@ contract Polygon_SpokePool is IFxMessageProcessor, SpokePool {
     // the caller is the fxChild AND that the fxChild called processMessageFromRoot
     bool private callValidated;
 
-    // Dictionary of function UUID to unique hashes, which we use to mark the last block that a function was called
+    // Dictionary of function lock UUID to unique hashes, which we use to mark the last block that a function was called
     // by a certain caller (e.g. set value equal to keccak256(block.timestamp, tx.origin)). This can be
     // used to prevent certain functions from being called atomically by the same caller.
     // This assumes each block has a different block.timestamp on this network.
-    mapping(string => bytes32) private funcLocks;
+    mapping(bytes32 => bytes32) private funcLocks;
     error CrossFunctionLock();
+
+    // Function lock identifiers used as keys in funcLocks mapping above.
+    bytes32 private constant FILL_LOCK_IDENTIFIER = "Fill";
+    bytes32 private constant EXECUTE_LOCK_IDENTIFIER = "Execute";
 
     event PolygonTokensBridged(address indexed token, address indexed receiver, uint256 amount);
     event SetFxChild(address indexed newFxChild);
@@ -187,6 +191,9 @@ contract Polygon_SpokePool is IFxMessageProcessor, SpokePool {
         // msg.sender is equal to tx.origin which is fine as long as Polygon supports the tx.origin opcode.
         // solhint-disable-next-line avoid-tx-origin
         if (relayerRefundLeaf.amountToReturn > 0 && msg.sender != tx.origin) revert NotEOA();
+        // Prevent calling recipient contract functions atomically with executing relayer refund leaves.
+        _revertIfFunctionCalledAtomically(FILL_LOCK_IDENTIFIER);
+        _setFunctionLock(EXECUTE_LOCK_IDENTIFIER);
         super.executeUSSRelayerRefundLeaf(rootBundleId, relayerRefundLeaf, proof);
     }
 
@@ -201,8 +208,8 @@ contract Polygon_SpokePool is IFxMessageProcessor, SpokePool {
         // solhint-disable-next-line avoid-tx-origin
         if (relayerRefundLeaf.amountToReturn > 0 && msg.sender != tx.origin) revert NotEOA();
         // Prevent calling recipient contract functions atomically with executing relayer refund leaves.
-        _revertIfFunctionCalledAtomically("fill");
-        _setFunctionLock("executeLeaf");
+        _revertIfFunctionCalledAtomically(FILL_LOCK_IDENTIFIER);
+        _setFunctionLock(EXECUTE_LOCK_IDENTIFIER);
         super.executeRelayerRefundLeaf(rootBundleId, relayerRefundLeaf, proof);
     }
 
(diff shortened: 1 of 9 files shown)


contracts/SpokePool.sol
```diff
@@ -1393,6 +1393,8 @@ abstract contract SpokePool is
         // This method by default is a no-op. Different child spoke pools might want to execute functionality here
         // such as wrapping any native tokens owned by the contract into wrapped tokens before proceeding with
         // executing the leaf.
+
+        if (address(msg.sender).isContract()) revert NotEOA();
```
pxrl, line 1397: This will prevent us from bundling HubPool and SpokePool executions together, no? That'd be unfortunate because it's actually really nice to have those executed atomically.

contracts/Polygon_SpokePool.sol
```diff
@@ -164,6 +164,30 @@ contract Polygon_SpokePool is IFxMessageProcessor, SpokePool {
...
+     * whereby someone batches this call with a bunch of other calls and produces a very large L2 burn transaction.
+     * This might make the L2 -> L1 message fail due to exceeding the L1 calldata limit.
+     */
+
+    function executeUSSRelayerRefundLeaf(
+        uint32 rootBundleId,
+        USSRelayerRefundLeaf memory relayerRefundLeaf,
+        bytes32[] memory proof
+    ) public override {
+        if (relayerRefundLeaf.amountToReturn > 0 && AddressLibUpgradeable.isContract(msg.sender)) revert NotEOA();
```
mrice32, line 178: I don't think this is an adequate check. OZ removed this because of how easy it is to misuse.

The issue is that this is checking the code size of the calling address, but if called inside a constructor, the code size is 0, but it can talk to other contracts atomically (which is what we're trying to avoid).

I think there are only two ways to avoid this (we've had this problem previously):
1. Force the msg.sender to pass in a signature, since it's impossible for a contract to generate a cryptographic signature.
2. Verify that tx.origin (if that's supported on polygon) equals msg.sender, since a contract can't be the originator of a transaction.

mrice32 (commented):
One comment

contracts/Polygon_SpokePool.sol
nicholaspai: Good call, done and added tests specifically for this constructor vector

mrice32 (approved):
LGTM!

contracts/Polygon_SpokePool.sol
```diff
@@ -164,10 +171,51 @@ contract Polygon_SpokePool is IFxMessageProcessor, SpokePool {
...
+        SpokePoolInterface.RelayerRefundLeaf memory relayerRefundLeaf,
+        bytes32[] memory proof
+    ) public override {
+        // AddressLibUpgradeable.isContract isn't a sufficient check because it checks the contract code size of
+        // msg.sender which is 0 if called from a constructor function on msg.sender. This is why we check if
+        // msg.sender is equal to tx.origin which is fine as long as Polygon supports the tx.origin opcode.
+        // solhint-disable-next-line avoid-tx-origin
+        if (relayerRefundLeaf.amountToReturn > 0 && msg.sender != tx.origin) revert NotEOA();
+        // Prevent calling recipient contract functions atomically with executing relayer refund leaves.
+        _revertIfFunctionCalledAtomically("fill");
```
mrice32, line 204: I think this implementation works great, but I have a few minor suggestions:
1. We should avoid using strings, since they are dynamically sized and expensive in solidity. Would suggest using bytes32 for all of these, since bytes32 is fixed size and these can easily fit in that data type.
2. To avoid typos, I would make these compile-time constants at the top of the file like so:
```
    bytes32 private constant FILL_IDENTIFIER = "Fill";
```
and then pass it to the functions -- this prevents string typos from causing the logic to fail.

contracts/Polygon_SpokePool.sol
```diff
@@ -164,10 +171,51 @@ contract Polygon_SpokePool is IFxMessageProcessor, SpokePool {
...
+        uint32 rootBundleId,
+        USSRelayerRefundLeaf memory relayerRefundLeaf,
+        bytes32[] memory proof
+    ) public override {
+        // AddressLibUpgradeable.isContract isn't a sufficient check because it checks the contract code size of
+        // msg.sender which is 0 if called from a constructor function on msg.sender. This is why we check if
+        // msg.sender is equal to tx.origin which is fine as long as Polygon supports the tx.origin opcode.
+        // solhint-disable-next-line avoid-tx-origin
+        if (relayerRefundLeaf.amountToReturn > 0 && msg.sender != tx.origin) revert NotEOA();
+        super.executeUSSRelayerRefundLeaf(rootBundleId, relayerRefundLeaf, proof);
```
mrice32, line 190: Why don't we need to do the _revertIfFunctionCalledAtomically/_setFunctionLock logic in this function?

mrice32 (commented):
Looks great! Only two comments/questions!

contracts/Polygon_SpokePool.sol
nicholaspai: Good call out +1

contracts/Polygon_SpokePool.sol
nicholaspai: shoot, we do. I forgot to add

contracts/Polygon_SpokePool.sol
```diff
@@ -188,6 +243,18 @@ contract Polygon_SpokePool is IFxMessageProcessor, SpokePool {
         if (balance > 0) wrappedNativeToken.deposit{ value: balance }();
     }
 
+    function _setFunctionLock(bytes32 funcSig) internal {
```
mrice32, line 246: nit: since we're just using custom string identifiers, we might want to rename this arg (here and in the other function) to `funcIdentifier` or similar, since a function "signature" has a specific meaning.

contracts/Polygon_SpokePool.sol
```diff
@@ -188,6 +243,18 @@ contract Polygon_SpokePool is IFxMessageProcessor, SpokePool {
...
+        // solhint-disable-next-line not-rely-on-time, avoid-tx-origin
+        bytes32 lockValue = keccak256(abi.encodePacked(block.timestamp, tx.origin));
+        if (funcLocks[funcSig] != lockValue) funcLocks[funcSig] = lockValue;
+    }
+
+    // Revert if function was called during this block
```
mrice32, line 252:
```suggestion
    // Revert if function was called during this block with the same tx.origin.
```

mrice32 (approved):
LGTM! Two minor nits

mrice32 (approved):
LGTM!

PR #386, opened 2023-12-19
feat(SpokePool): Remove fillCompleted param from handleAcrossMessage callback in fillUSS function

nicholaspai:


pxrl (approved):

pxrl (approved):

contracts/SpokePool.sol
```diff
@@ -25,6 +25,15 @@ interface AcrossMessageHandler {
         address relayer,
         bytes memory message
     ) external;
+
+    // New function interface to be used with USS functions since fillCompleted no longer has any
+    // meetings now that partial fills are impossible.
```
mrice32, line 30: meetings? Was that the intended word?

contracts/SpokePool.sol
```diff
@@ -25,6 +25,15 @@ interface AcrossMessageHandler {
...
+    function handleUSSAcrossMessage(
```
mrice32, line 31: We should probably loop in Ryan to make sure we feel good about changing this interface on integrators. This means that the moment this change goes into effect, all integrators must have implemented the new function (even if its alongside the old function). If they haven't their relays will revert.

mrice32 (commented):
Looks great! One comment

contracts/SpokePool.sol
nicholaspai: ah, "meaning"

nicholaspai:
Not merging until we confirm that integrators are able to support new interface

PR #387, opened 2023-12-21
feat(SpokePool): Add USS SpeedUp creation and verification methods

nicholaspai:
Fixes ACX-1687

### DIFF
--- a/contracts/SpokePool.sol
+++ b/contracts/SpokePool.sol
@@ -699,6 +699,36 @@ abstract contract SpokePool is
         );
     }
 
+    function speedUpUSSDeposit(
+        address depositor,
+        uint32 depositId,
+        uint256 updatedOutputAmount,
+        address updatedRecipient,
+        bytes memory updatedMessage,
+        bytes memory depositorSignature
+    ) public override nonReentrant {
+        _verifyUpdateUSSDepositMessage(
+            depositor,
+            depositId,
+            chainId(),
+            updatedOutputAmount,
+            updatedRecipient,
+            updatedMessage,
+            depositorSignature
+        );
+
+        // Assuming the above checks passed, a relayer can take the signature and the updated deposit information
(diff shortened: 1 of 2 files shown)


contracts/SpokePool.sol
```diff
@@ -889,13 +919,54 @@ abstract contract SpokePool is
...
+        relayData.destinationChainId = chainId();
+
+        USSRelayExecutionParams memory relayExecution = USSRelayExecutionParams({
+            relay: relayData,
+            relayHash: keccak256(abi.encode(relayData)),
+            updatedOutputAmount: updatedOutputAmount,
+            updatedRecipient: updatedRecipient,
+            updatedMessage: updatedMessage,
+            repaymentChainId: repaymentChainId
+        });
```
nicholaspai, line 959: Up until here the logic is copied from `fillRelay` so it could be refactored

contracts/SpokePool.sol
```diff
@@ -889,13 +919,54 @@ abstract contract SpokePool is
 
         USSRelayExecutionParams memory relayExecution = USSRelayExecutionParams({
             relay: relayData,
-            relayHash: bytes32(0),
+            relayHash: keccak256(abi.encode(relayData)),
```
mrice32, line 922: Was this previously overlooked or something? Surprised that tests passed with all relays having the same hash

mrice32 (approved):
LGTM!

contracts/SpokePool.sol
nicholaspai: Nope if you look at the current `fillRelay` implementation, we first set this prop to 0 and then hash the entire RelayData struct. This is just a readability-motivated refactor since we pass in the `USSRelayData` struct into this function its easy to hash it in-place

mrice32: Oh I see!

mrice32 (approved):
LGTM!

PR #391, opened 2023-12-29
feat(CCTP): ✨ modify adapters to include CCTP bridging of UDSC

james-a-morris:
This PR creates a new CCTP Adapter that can be added to chain adapters to support bridging USDC on L1 -> L2 via the CCTP bridge. This PR also modifies the following chain adapters (both solidity code & deploy script code):

1. Optimism (closes ACX-1795)
2. Polygon (closes ACX-1792)
3. Arbitrum (closes ACX-1794)
4. Base (closes ACX-1793)


contracts/chain-adapters/Arbitrum_Adapter.sol
```diff
@@ -166,12 +168,18 @@ contract Arbitrum_Adapter is AdapterInterface {
      * @param _l1ArbitrumInbox Inbox helper contract to send messages to Arbitrum.
      * @param _l1ERC20GatewayRouter ERC20 gateway router contract to send tokens to Arbitrum.
      * @param _l2RefundL2Address L2 address to receive gas refunds on after a message is relayed.
+     * @param _l1Usdc USDC address on L1.
+     * @param _circleDomain Circle domain set for this chain. NOTE: this is issued by circle and is irrelevant of chain id
```
nicholaspai, line 172: What is an example circle domain? Can/should we make this a constant in the CCTPAdapter?

james-a-morris: It's a chain identifier (agnostic of main/testnet) that is issued by circle. I have docs here that explain/link to the official docs. I'm not sure we should hardcode it because it would add a lookup, but it's essentially circle's version of a chain Id

https://github.com/across-protocol/contracts-v2/blob/7b2a113a5789875972eda7b7f2ed9e2cdfa73500/deploy/consts.ts#L124-L130

nicholaspai: what do you mean it would add a lookup? Couldn't we hardcode the chain ID's that we pass to the CCTP_Adapter in each child adapter?

james-a-morris: Resolved offline - this is a fair point. Done: https://github.com/across-protocol/contracts-v2/pull/391/commits/83c2d22af9e9fdf0faf6b632b4b6f9e1142464be

mrice32 (commented):
Looks great -- I only have one question about variable storage, since it seems like we shouldn't need to store these in the proxy, itself, and we can keep them in the bytecode.

contracts/Ovm_SpokePool.sol
```diff
@@ -43,6 +44,29 @@ contract Ovm_SpokePool is SpokePool {
...
+     * @notice Domain identifier used for Circle's CCTP bridge to L1.
+     * @dev This identifier is assigned by Circle and is not related to a chain ID.
+     * @dev Official domain list can be found here: https://developers.circle.com/stablecoins/docs/supported-domains
+     */
+    uint32 public constant l1CircleDomainId = 0;
+    /**
+     * @notice The official USDC contract address on this chain.
+     * @dev Posted officially here: https://developers.circle.com/stablecoins/docs/usdc-on-main-networks
+     */
+    IERC20 public l2Usdc;
```
nicholaspai, line 57: if we're adding a new storage variable to a base contract like this one we need to decrement this contract's `__gap` variable. This contract is a base contract for Optimism/Base/Boba_SpokePool's. Decrement the `__gap` for each 32 byte slot used up: https://docs.openzeppelin.com/upgrades-plugins/1.x/writing-upgradeable#storage-gaps

contracts/Arbitrum_SpokePool.sol
```diff
@@ -23,6 +24,23 @@ contract Arbitrum_SpokePool is SpokePool {
...
+    /**
+     * @notice The official USDC contract address on this chain.
+     * @dev Posted officially here: https://developers.circle.com/stablecoins/docs/usdc-on-main-networks
+     */
+    IERC20 public l2Usdc;
+    /**
+     * @notice The official Circle CCTP token bridge contract endpoint.
+     * @dev Posted officially here: https://developers.circle.com/stablecoins/docs/evm-smart-contracts
+     */
+    ITokenMessenger public cctpTokenMessenger;
```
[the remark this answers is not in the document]
nicholaspai: I was going to comment the same, the only variable I can see changing here is `cctpTokenMessenger` and we should allow that to be upgraded via cross domain admin

nicholaspai (commented):
Added some comments about storage slots and also doubled up matt's comment about constant state variables

contracts/Arbitrum_SpokePool.sol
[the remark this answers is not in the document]
james-a-morris: Sounds good! I've made adjustments to place all of these state variables in the constructor. I've also made `cctpMessenger` non-immutable and updatable via the `setCCTPTokenMessenger` function which is only callable by a cross domain admin.

contracts/Ovm_SpokePool.sol
```diff
@@ -43,6 +44,29 @@ contract Ovm_SpokePool is SpokePool {
...
+     * @notice The official Circle CCTP token bridge contract endpoint.
+     * @dev Posted officially here: https://developers.circle.com/stablecoins/docs/evm-smart-contracts
+     */
+    ITokenMessenger public cctpTokenMessenger;
+
+    /**
+     * @notice Whether or not this OVM Spoke will attempt to bridge USDC via CCTP.
+     * @dev This is only enabled on Base & Optimism spoke pools.
+     */
+    bool public enableCctpBridge;
```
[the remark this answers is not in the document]
james-a-morris: Thread here: https://github.com/across-protocol/contracts-v2/pull/391#discussion_r1442359151

contracts/Polygon_SpokePool.sol
```diff
@@ -33,6 +34,23 @@ interface IFxMessageProcessor {
...
+     * @notice The official USDC contract address on this chain.
+     * @dev Posted officially here: https://developers.circle.com/stablecoins/docs/usdc-on-main-networks
+     */
+    IERC20 public l2Usdc;
+    /**
+     * @notice The official Circle CCTP token bridge contract endpoint.
+     * @dev Posted officially here: https://developers.circle.com/stablecoins/docs/evm-smart-contracts
+     */
+    ITokenMessenger public cctpTokenMessenger;
+
```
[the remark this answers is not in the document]
james-a-morris: Thread here: https://github.com/across-protocol/contracts-v2/pull/391#discussion_r1442359151

contracts/chain-adapters/Arbitrum_Adapter.sol
```diff
@@ -161,20 +163,45 @@ contract Arbitrum_Adapter is AdapterInterface {
...
+     * @dev Posted officially here: https://developers.circle.com/stablecoins/docs/usdc-on-main-networks
+     */
+    IERC20 public immutable l1Usdc;
+
+    /**
+     * @notice The official Circle CCTP token bridge contract endpoint.
+     * @dev Posted officially here: https://developers.circle.com/stablecoins/docs/evm-smart-contracts
+     */
+    ITokenMessenger public immutable cctpTokenMessenger;
+
```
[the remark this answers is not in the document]
james-a-morris: Thread here: https://github.com/across-protocol/contracts-v2/pull/391#discussion_r1442359151

contracts/chain-adapters/Base_Adapter.sol
```diff
@@ -30,19 +33,44 @@ contract Base_Adapter is CrossDomainEnabled, AdapterInterface {
...
+     * @dev Posted officially here: https://developers.circle.com/stablecoins/docs/usdc-on-main-networks
+     */
+    IERC20 public immutable l1Usdc;
+
+    /**
+     * @notice The official Circle CCTP token bridge contract endpoint.
+     * @dev Posted officially here: https://developers.circle.com/stablecoins/docs/evm-smart-contracts
+     */
+    ITokenMessenger public immutable cctpTokenMessenger;
+
```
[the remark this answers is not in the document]
james-a-morris: Thread here: https://github.com/across-protocol/contracts-v2/pull/391#discussion_r1442359151

contracts/Ovm_SpokePool.sol
james-a-morris: I've decremented the `__gap` by 2 slots to include space for the `CircleCCTPAdapter`'s `recipientCircleDomainId`, `usdcToken`, and `cctpTokenMessenger`. 

This is because: `recipientCircleDomainId` is 4 bytes, and `usdcToken` & `cctpTokenMessenger` should be 20 bytes each. So in total, 44 bytes fit in 2 slots.

contracts/libraries/CircleCCTPAdapter.sol
```diff
@@ -0,0 +1,88 @@
...
+     */
+    function _addressToBytes32(address addr) internal pure returns (bytes32) {
+        return bytes32(uint256(uint160(addr)));
+    }
+
+    /**
+     * @notice Returns whether or not the CCTP bridge is enabled.
+     */
+    function _isCCTPEnabled() private view returns (bool) {
+        return address(cctpTokenMessenger) != address(0) && address(usdcToken) != address(0);
```
nicholaspai, line 63: Prefer only needing to check `address(cctpTokenMessenger) != address(0)` to reduce gas cost. This then assumes we'll always set the `usdcToken` address correctly but we can set `cctpTokenMessenger` to 0x0 in order to "disable" this adapter

contracts/libraries/CircleCCTPAdapter.sol
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    /**
+     * @notice Transfers USDC from the current domain to the given address on the new domain.
+     * @param to Address to receive USDC on the new domain.
+     * @param amount Amount of USDC to transfer.
+     */
+    function _transferUsdc(address to, uint256 amount) internal {
+        // Require that the CCTP bridge is enabled
+        require(_isCCTPEnabled(), "CCTP bridge is disabled");
```
nicholaspai, line 73: In all of the SpokePools, we first check `_isCCTPEnabled()` before calling into here so the check here is redundant and adds gas. Can we remove this check and add a note to devs that this function will revert if `cctpTokenMessenger` is set incorrectly?

contracts/libraries/CircleCCTPAdapter.sol
```diff
@@ -0,0 +1,88 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.0;
+
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import "../external/interfaces/CCTPInterfaces.sol";
+
+abstract contract CircleCCTPAdapter {
```
nicholaspai, line 8: Nit: prefer naming this to CircleCCTP_SpokePool since its extended by SpokePools and not used as an L1 "Adapter" just because this would be confuinsg with the contracts in `chain-adapters`

contracts/test/MockHubPool.sol
```diff
@@ -0,0 +1,86 @@
...
+            )
+        );
+        require(success, "delegatecall failed");
+    }
+
+    /// @notice Queries the balance of a user for a specific L2 token using the adapter contract.
+    /// @dev This function delegates the call to the adapter contract to execute the balance check.
+    /// @param l2Token The address of the L2 token.
+    /// @param user The user address whose balance is being queried.
+    function balanceOf(address l2Token, address user) external {
```
nicholaspai, line 76: why not just implement a more generalized function that takes in `bytes memory l2CallData` so that we can test with any arbitrary function? I think `balanceOf` is an interesting check but I wonder if there are additional test cases we'd want to send with state-modifyin xchain functions

deploy/002_deploy_optimism_adapter.ts
```diff
@@ -18,6 +18,8 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
       L1_ADDRESS_MAP[chainId].weth,
       L1_ADDRESS_MAP[chainId].optimismCrossDomainMessenger,
       L1_ADDRESS_MAP[chainId].optimismStandardBridge,
+      L1_ADDRESS_MAP[chainId].l1UsdcAddress,
```
nicholaspai, line 21: nit: suggest commenting out the `cctpTokenMessenger` line and adding in a zeroAddress for production so we don't accidentally deploy this with full cctp support yet

This comment applies to all deploy scripts as well

contracts/Arbitrum_SpokePool.sol
```diff
@@ -83,16 +92,21 @@ contract Arbitrum_SpokePool is SpokePool {
...
-        require(ethereumTokenToBridge != address(0), "Uninitialized mainnet token");
-        //slither-disable-next-line unused-return
-        StandardBridgeLike(l2GatewayRouter).outboundTransfer(
-            ethereumTokenToBridge, // _l1Token. Address of the L1 token to bridge over.
-            hubPool, // _to. Withdraw, over the bridge, to the l1 hub pool contract.
-            amountToReturn, // _amount.
-            "" // _data. We don't need to send any data for the bridging action.
-        );
+        // If the l2TokenAddress is UDSC, we need to use the CCTP bridge.
+        if (_isCCTPEnabledForToken(l2TokenAddress)) {
```
nicholaspai, line 96: WDYT about calling `_isCCTPEnabled() && l2TokenAddress == usdcToken`? I think its more intuitive what's going on here rather than hiding the second part of that check behind an internal function

nicholaspai (commented):
This makes a lot of sense, I left some suggestions

contracts/libraries/CircleCCTPAdapter.sol
james-a-morris: Hmm agreed but could we use `CircleCCTPConnector` or something similar? The `CircleCCTPAdapter` is currently used by the chain adapters & the spoke pools since the logic is the same

contracts/libraries/CircleCCTPAdapter.sol
james-a-morris: Good call: https://github.com/across-protocol/contracts-v2/pull/391/commits/953e2915227dcfa528541d39e9930021ae8c86f9

contracts/Arbitrum_SpokePool.sol
james-a-morris: I think it's a good call. I'm not sure how much additional readability it would give for someone familiar with the code, but it does bring the boolean check directly into where the bridging is called. 

As a tradeoff - this change requires propagating the `_isCCTPEnabled() && l2TokenAddress == usdcToken` 8 times.

Commit: https://github.com/across-protocol/contracts-v2/pull/391/commits/318afa21b4c84af62b3ac9ea0bce613cb51cfb1d

deploy/002_deploy_optimism_adapter.ts
james-a-morris: Good call - per the sync today I've set all spoke/adapters `cctpTokenMessenger` contract to `ZERO_ADDRESS`.

contracts/test/MockHubPool.sol
james-a-morris: Absolutely. I wrote this test to include something related to arbitrary messaging. I later tested the messaging with a hub pool deployment. 

However, since we're including this in the repo, I'll update it to include arbitrary messaging.

contracts/Ovm_SpokePool.sol
```diff
@@ -171,5 +181,5 @@ contract Ovm_SpokePool is SpokePool {
     // Reserve storage slots for future versions of this base contract to add state variables without
     // affecting the storage layout of child contracts. Decrement the size of __gap whenever state variables
     // are added. This is at bottom of contract to make sure its always at the end of storage.
-    uint256[1000] private __gap;
+    uint256[998] private __gap;
```
mrice32, line 184: What is this change? Did we add variables to the OVM SpokePool that would move storage slots?

contracts/chain-adapters/Arbitrum_Adapter.sol
```diff
@@ -161,17 +163,28 @@ contract Arbitrum_Adapter is AdapterInterface {
 
     ArbitrumL1ERC20GatewayLike public immutable l1ERC20GatewayRouter;
 
+    /**
+     * @notice Domain identifier used for Circle's CCTP bridge on Arbitrum.
+     * @dev This identifier is assigned by Circle and is not related to a chain ID.
+     * @dev Official domain list can be found here: https://developers.circle.com/stablecoins/docs/supported-domains
+     */
+    uint32 public constant circleDomainId = 3;
```
mrice32, line 171: nit: since this is only used to pass up through the constructor stack, I would suggest making it a private variable and name it ARBITRUM_CIRCLE_DOMAIN_ID (I think it's good to include the chain name so it's very clear that the constant is chain specific). It will be exposed via the recipientCircleDomainId in the parent class, so we don't really gain anything from exposing with a public function here.

Same for the other adapters.

mrice32 (commented):
I like this design a lot -- just a few small comments!

contracts/Ovm_SpokePool.sol
james-a-morris: My understanding is that since this class' inherited class was updated (the `CircleCCTPAdapter`) we'd need to modify this. In both cases, we didn't have any upgrade errors as outlined here: https://docs.openzeppelin.com/contracts/3.x/upgradeable#:~:text=You%20may%20notice%20that%20every,storage%20compatibility%20with%20existing%20deployments.

contracts/chain-adapters/Arbitrum_Adapter.sol
james-a-morris: Updated here: https://github.com/across-protocol/contracts-v2/pull/391/commits/dd5e3f8719efe998c811d09831c56fcae71bac4a

contracts/Ovm_SpokePool.sol
nicholaspai: If the variables are constant or immutable then they don't take slots and this shouldn't be decremented

mrice32 (approved):
LGTM!

PR #392, opened 2024-01-02
fix(SpokePool): Address speedup USS depositor signature logic and add unit tests

nicholaspai:
Fixes bug in _verifyUpdateUSSDepositMessage where the incorrect type string is used (i.e. it still includes int64 updatedRelayerFeePct instead of uint256 updatedOutputAmount) which means that the depositor 
signature will never match the passed in "speed up params".

This PR also adds unit tests for both `depositUSS` and `speedUpUSSDeposit`.

### DIFF
--- a/test/SpokePool.Deposit.ts
+++ b/test/SpokePool.Deposit.ts
@@ -598,10 +598,10 @@ describe("SpokePool Depositor Logic", async function () {
       );
     });
   });
-  describe.only("speed up USS deposit", function () {
+  describe("speed up USS deposit", function () {
     const updatedOutputAmount = amountToDeposit.add(1);
     const updatedRecipient = randomAddress();
-    const updatedMessage = "0xrandombytes";
+    const updatedMessage = "0x1234";
     const depositId = 100;
     it("_verifyUpdateUSSDepositMessage", async function () {
       const signature = await getUpdatedUSSDepositSignature(
@@ -621,6 +621,103 @@ describe("SpokePool Depositor Logic", async function () {
         updatedMessage,
         signature
       );
+
+      // Reverts if passed in depositor is the signer or if signature is incorrect
+      await expect(
+        spokePool.verifyUpdateUSSDepositMessage(
+          updatedRecipient,
+          depositId,
+          originChainId,
+          updatedOutputAmount,
+          updatedRecipient,
+          updatedMessage,
+          signature
+        )
+      ).to.be.revertedWith("invalid signature");
+
+      // @dev Creates an invalid signature using different params
+      const invalidSignature = await getUpdatedUSSDepositSignature(
+        depositor,
+        depositId + 1,
+        originChainId,
+        updatedOutputAmount,
+        updatedRecipient,
+        updatedMessage
+      );
+      await expect(
+        spokePool.verifyUpdateUSSDepositMessage(
+          depositor.address,
+          depositId,
+          originChainId,
+          updatedOutputAmount,
+          updatedRecipient,
+          updatedMessage,
+          invalidSignature
+        )
+      ).to.be.revertedWith("invalid signature");
+    });
+    it("passes spoke pool's chainId() as origin chainId", async function () {
+      const spokePoolChainId = await spokePool.chainId();
+
+      const expectedSignature = await getUpdatedUSSDepositSignature(
+        depositor,
+        depositId,
+        spokePoolChainId,
+        updatedOutputAmount,
+        updatedRecipient,
+        updatedMessage
+      );
+      await expect(
+        spokePool.speedUpUSSDeposit(
+          depositor.address,
+          depositId,
+          updatedOutputAmount,
+          updatedRecipient,
+          updatedMessage,
+          expectedSignature
+        )
+      )
+        .to.emit(spokePool, "RequestedSpeedUpUSSDeposit")
+        .withArgs(
+          updatedOutputAmount,
+          depositId,
+          depositor.address,
+          updatedRecipient,
+          updatedMessage,
+          expectedSignature
+        );
+
+      // Can't use a signature for a different chain ID, even if the signature is valid otherwise for the depositor.
+      const otherChainId = spokePoolChainId.add(1);
+      const invalidSignatureForChain = await getUpdatedUSSDepositSignature(
+        depositor,
+        depositId,
+        otherChainId,
+        updatedOutputAmount,
+        updatedRecipient,
+        updatedMessage
+      );
+      await expect(
+        spokePool.verifyUpdateUSSDepositMessage(
+          depositor.address,
+          depositId,
+          otherChainId,
+          updatedOutputAmount,
+          updatedRecipient,
+          updatedMessage,
+          invalidSignatureForChain
+        )
+      ).to.not.be.reverted;
+      await expect(
+        spokePool.speedUpUSSDeposit(
+          depositor.address,
+          depositId,
+          updatedOutputAmount,
+          updatedRecipient,
+          updatedMessage,
+          invalidSignatureForChain
+        )
+      ).to.be.revertedWith("invalid signature");
     });
   });
 });
--- a/test/SpokePool.Deposit.ts
+++ b/test/SpokePool.Deposit.ts
@@ -16,6 +16,7 @@ import {
   getDepositParams,
   RelayData,
   USSRelayData,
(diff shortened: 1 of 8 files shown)


test/fixtures/SpokePool.Fixture.ts
```diff
@@ -410,6 +431,40 @@ export async function modifyRelayHelper(
...
+        { name: "originChainId", type: "uint256" },
+        { name: "updatedOutputAmount", type: "uint256" },
+        { name: "updatedRecipient", type: "address" },
+        { name: "updatedMessage", type: "bytes" },
+      ],
+    },
+    domain: {
+      name: "ACROSS-V2",
+      version: "1.0.0",
+      chainId: Number(originChainId),
```
pxrl, line 455: Minor: originChainId should already be a number.

```suggestion
      chainId: originChainId,
```

test/SpokePool.Deposit.ts
```diff
@@ -402,23 +411,313 @@ describe("SpokePool Depositor Logic", async function () {
...
+      await expect(spokePool.connect(depositor).depositUSS(..._depositArgs)).to.not.be.reverted;
+    });
+    it("invalid quoteTimestamp", async function () {
+      const quoteTimeBuffer = await spokePool.depositQuoteTimeBuffer();
+      const currentTime = await spokePool.getCurrentTime();
+
+      await expect(
+        spokePool.connect(depositor).depositUSS(
+          // quoteTimestamp too far into past (i.e. beyond the buffer)
+          ...getDepositArgsFromRelayData(relayData, currentTime.sub(quoteTimeBuffer).sub(100))
```
pxrl, line 468: Should we test the absolute margin, rather than margin - 100 ?

```suggestion
          ...getDepositArgsFromRelayData(relayData, currentTime.sub(quoteTimeBuffer).sub(1))
```

test/SpokePool.Deposit.ts
```diff
@@ -402,23 +411,313 @@ describe("SpokePool Depositor Logic", async function () {
...
+        );
+      expect(await erc20.balanceOf(depositor.address)).to.equal(balanceBefore.sub(amountToDeposit));
+    });
+    it("deposits are not paused", async function () {
+      await spokePool.pauseDeposits(true);
+      await expect(spokePool.connect(depositor).depositUSS(...depositArgs)).to.be.revertedWith("Paused deposits");
+    });
+    it("reentrancy protected", async function () {
+      // In this test we create a reentrancy attempt by sending a fill with a recipient contract that calls back into
+      // the spoke pool via the deposit function.
```
pxrl, line 585: fwiw I think we should decouple the deposit-side locks from the fill-side locks. Probably not for now, but it could be useful for multi-hop transactions.

test/SpokePool.Relay.ts
```diff
@@ -325,27 +345,345 @@ describe("SpokePool Relayer Logic", async function () {
...
+            relayData.exclusiveRelayer,
+            relayer.address,
+            relayData.depositor,
+            relayData.recipient,
+            relayData.message,
+            [
+              relayExecution.updatedRecipient,
+              relayExecution.updatedMessage,
+              relayExecution.updatedOutputAmount,
+              // Testing that this FillType is "SlowFill"
```
pxrl, line 495:
```suggestion
```

pxrl: Or alternatively `s/SlowFill/FastFill`.

test/SpokePool.Relay.ts
```diff
@@ -325,27 +345,345 @@ describe("SpokePool Relayer Logic", async function () {
...
+        await destErc20.connect(relayer).approve(spokePool.address, _relayExecution.updatedOutputAmount);
+        await expect(() =>
+          spokePool.connect(relayer).fillRelayUSSInternal(
+            _relayExecution,
+            relayer.address,
+            false // isSlowFill
+          )
+        ).to.changeTokenBalance(destErc20, depositor, consts.amountToDeposit.mul(2));
+      });
+      it("unwraps native token if sending to EOA otherwise sends wrapped ERC20", async function () {
```
pxrl, line 535:
```suggestion
      it("unwraps native token if sending to EOA", async function () {
```

test/SpokePool.Relay.ts
```diff
@@ -325,27 +345,345 @@ describe("SpokePool Relayer Logic", async function () {
...
+        const relayExecution = await getRelayExecutionParams(_relayData);
+        await weth.connect(relayer).transfer(spokePool.address, relayExecution.updatedOutputAmount);
+        await expect(() =>
+          spokePool.connect(relayer).fillRelayUSSInternal(
+            relayExecution,
+            relayer.address,
+            true // isSlowFill
+          )
+        ).to.changeEtherBalance(recipient, relayExecution.updatedOutputAmount);
+      });
```
pxrl, line 563: This one doesn't strictly verify that the SpokePool paid for the slow fill, since it only verifies that the recipient received `updatedOutputAmount`.

nb haven't tested the snippet below.

```suggestion
        const initialSpokeBalance = await weth.connect(relayer).balanceOf(spokePool.address);
        await weth.connect(relayer).transfer(spokePool.address, relayExecution.updatedOutputAmount);
        await expect(() =>
          spokePool.connect(relayer).fillRelayUSSInternal(
            relayExecution,
            relayer.address,
            true // isSlowFill
          )
        ).to.changeEtherBalance(recipient, relayExecution.updatedOutputAmount);
       const spokeBalance = await weth.connect(relayer).balanceOf(spokePool.address);
       expect(spokeBalance.eq(initialSpokeBalance)).to.be.true;
      });
```

test/SpokePool.Relay.ts
```diff
@@ -325,27 +345,345 @@ describe("SpokePool Relayer Logic", async function () {
...
+          spokePool.connect(relayer).fillUSSRelay(
+            {
+              ..._relayData,
+              exclusivityDeadline: 0,
+            },
+            consts.repaymentChainId
+          )
+        ).to.not.be.reverted;
+      });
+      it("can't overwrite RelayData destination chain", async function () {
```
pxrl, line 645: I missed the review where this change went in, but ooc, why do we overwrite the destinationChainId instead of just reverting if `destinationChainId != chainId()` ? If the caller has mis-specified that then I'd be concerned about other relayData attributes being wrong.

test/SpokePool.Deposit.ts
nicholaspai: hmm yeah I can see a reason for that but it would add complexity. Should we try to get into this audit?

test/SpokePool.Relay.ts
nicholaspai: > Or alternatively `s/SlowFill/FastFill`.

what do you mean by this?

test/SpokePool.Relay.ts
nicholaspai: can use `changeEtherBalances` instead!

test/SpokePool.Relay.ts
nicholaspai: hmm fair point

nicholaspai: Done 69db9ef7ecdf6f134c9126e0671580f2c8862b48

test/SpokePool.Relay.ts
nicholaspai: 69db9ef7ecdf6f134c9126e0671580f2c8862b48

test/SpokePool.Deposit.ts
pxrl: I think it'd be fine for next cycle - this one is already quite packed.

pxrl (approved):
LGTM - really nice work!

test/SpokePool.Relay.ts
pxrl: Clarified in DMs :+1:

test/SpokePool.Deposit.ts
```diff
@@ -402,23 +411,313 @@ describe("SpokePool Depositor Logic", async function () {
...
+          depositor.address,
+          depositId,
+          originChainId,
+          updatedOutputAmount,
+          updatedRecipient,
+          updatedMessage,
+          invalidSignature
+        )
+      ).to.be.revertedWith("invalid signature");
+    });
```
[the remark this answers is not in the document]
pxrl: I'll follow up once this is merged.

PR #393, opened 2024-01-02
feat: Add SwapAndBridge helper contract

nicholaspai:
implement a contract to atomically swap tokens before submitting for Across deposit.

Fixes ACX-1785

### DIFF
--- a/contracts/SwapAndBridge.sol
+++ b/contracts/SwapAndBridge.sol
@@ -82,6 +82,24 @@ interface IUniswapV3Pool {
     function token1() external view returns (address);
 }
 
+// This is the interface for the Uniswap UniversalRouter that we can use to propose a more complex set of
+// transactions to swap between tokens.
+// - Arbitrum: https://arbiscan.io/address/0xec8b0f7ffe3ae75d7ffab09429e3675bb63503e4
+// - Optimism: https://optimistic.etherscan.io/address/0xeC8B0F7Ffe3ae75d7FfAb09429e3675bb63503e4
+// - Polygon: https://polygonscan.com/address/0x643770e279d5d0733f21d6dc03a8efbabf3255b4
+// - Base: https://basescan.org/address/0xeC8B0F7Ffe3ae75d7FfAb09429e3675bb63503e4
+interface IUniversalRouter {
+    /// @notice Executes encoded commands along with provided inputs. Reverts if deadline has expired.
+    /// @param commands A set of concatenated commands, each 1 byte in length
+    /// @param inputs An array of byte strings containing abi encoded inputs for each command
+    /// @param deadline The deadline by which the transaction must be executed
+    function execute(
+        bytes calldata commands,
+        bytes[] calldata inputs,
+        uint256 deadline
+    ) external payable;
+}
+
 /**
  * @title SwapAndBridge
  * @notice Allows caller to swap a specific on a chain and bridge them via Across atomically.
@@ -97,6 +115,9 @@ contract SwapAndBridge is Lockable, MultiCaller {
 
     // UniswapV3Pool we'll use to swap swapToken for acrossInputToken.
     IUniswapV3Pool public immutable uniswapV3Pool;
+
+    // Uniswap UniversalRouter we can use to express a more complex swapping route.
+    IUniversalRouter public immutable universalRouter;
 
     // The direction of the UniswapV3 swap, true for token0 to token1, false for token1 to token0
     bool public immutable uniswapSwapZeroForOne;
@@ -143,6 +164,7 @@ contract SwapAndBridge is Lockable, MultiCaller {
      * @param _spokePool Address of the SpokePool contract that we'll submit deposits to.
      * @param _oneInchRouter Address of the 1InchAggregationRouterV5 contract that we'll use to swap tokens.
      * @param _uniswapV3Pool Address of the UniswapV3Pool contract that we'll use to swap tokens.
+     * @param _universalRouter Address of the Uniswap UniversalRouter contract that we'll use to swap tokens.
      * @param _swapToken Address of the token that will be swapped for acrossInputToken. Cannot be 0x0
      * @param _acrossInputToken Address of the token that will be bridged via Across as the inputToken.
      */
@@ -150,6 +172,7 @@ contract SwapAndBridge is Lockable, MultiCaller {
         USSSpokePoolInterface _spokePool,
         I1InchAggregationRouterV5 _oneInchRouter,
         IUniswapV3Pool _uniswapV3Pool,
+        IUniversalRouter _universalRouter,
         IERC20 _swapToken,
         IERC20 _acrossInputToken
     ) {
@@ -159,6 +182,9 @@ contract SwapAndBridge is Lockable, MultiCaller {
 
         // Validate 1InchAggregationRouterV5:
         oneInchRouter = _oneInchRouter;
+
+        // Validate UniversalRouter:
+        universalRouter = _universalRouter;
 
         // Validate UniswapV3Pool:
         uniswapV3Pool = _uniswapV3Pool;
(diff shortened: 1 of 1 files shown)


contracts/SwapAndBridge.sol
```diff
@@ -0,0 +1,150 @@
...
+
+/**
+ * @title SwapAndBridge
+ * @notice Allows caller to swap tokens on a chain and bridge them via Across atomically.
+ */
+contract SwapAndBridge is Lockable, MultiCaller {
+    using SafeERC20 for IERC20;
+
+    USSSpokePoolInterface public immutable spokePool;
+    I1InchAggregationRouterV5 public immutable oneInchRouter;
```
mrice32, line 56: Is this interface the same on every chain?

contracts/SwapAndBridge.sol
```diff
@@ -0,0 +1,150 @@
...
+        // Pull tokens from caller into this contract.
+        address swapToken = address(swapDescription.srcToken);
+        uint256 swapTokenAmount = swapDescription.amount;
+        IERC20(swapToken).transferFrom(msg.sender, address(this), swapTokenAmount);
+
+        // Craft 1Inch swap arguments to swap `swapToken` for `acrossInputToken` using this contract as the recipient
+        // of the swap.
+        swapDescription.srcReceiver = payable(address(this));
+        swapDescription.dstReceiver = payable(address(this));
+        swapDescription.flags = 0; // TODO: Change this if we want to support ETH swaps and/or partial fills (i.e.
```
mrice32, line 112: So 1inch guarantees if we set this to 0 that partial fills won't happen (for instance, if one of the trade paths reverts, for instance)?

contracts/SwapAndBridge.sol
```diff
@@ -0,0 +1,150 @@
...
+        // https://optimistic.etherscan.io/tx/0x8a4e77ee1a62e94b42b21e849bcdabd60d43ac49191cd2878f6b47f395f26abc
+        (uint256 returnAmount, uint256 spentAmount) = oneInchRouter.swap(
+            aggregationExecutor,
+            swapDescription,
+            new bytes(0), // TODO: No IERC20Permit.permit needed since we're sending an approval?
+            new bytes(0) // TODO: We don't want to execute any data on swaps but I'm not sure how this is used.
+        );
+
+        // TODO: I don't think we need the following check if we set flags to 0 but just in case:
+        require(spentAmount == swapTokenAmount, "spent amount != swap amount");
```
mrice32, line 127: There's three checks we could do here:
1. spentAmount == swapTokenAmount, which sort of checks that they follow their own rule
2. balanceIncrease >= minExpectedReturn (maybe expectedReturn is passed in separately?)
3. balanceDecrease == swapTokenAmount

I think 2 is probably the only one we _really_ need, since all the user cares about is that at least their min tokens are sent.

Thoughts?

contracts/SwapAndBridge.sol
```diff
@@ -0,0 +1,150 @@
...
+     *     - srcReceiver: Address to receive `srcToken` from msg.sender. Overwritten by this function to be this contract.
+     *     - dstReceiver: Address to receive `dstToken` from 1inch swap. Overwritten by this function to be this contract.
+     *     - amount: Amount of `srcToken` to swap. Return amount is deposited into Across.
+     *     - minReturnAmount: Minimum amount of `dstToken` to receive from 1inch swap.
+     *     - flags: Overwritten by this contract.
+     * @param outputToken Token to receive from Across deposit on destination chain.
+     * @param outputAmount of `outputToken` to receive from Across deposit on
+     * destination chain.
+     */
+    function swap1InchAndBridge(
```
mrice32, line 86: Do you think we should pass in the min amount we expect to get back from 1inch separately so we can check? Or do you think we should fully rely on their contract?

mrice32 (approved):
Looks great! Only minor comments

contracts/SwapAndBridge.sol
nicholaspai: Not sure yet, need to test more outside of Optimism

contracts/SwapAndBridge.sol
nicholaspai: +1 on using point 2 that's a really good check. I don't think check 3 matters as much if we pass check 2 and check 1: however adding it does protect against some weird malicious case where 1inch returns an invalid `spentAmount`, which ultimately leads us to locking swapTokens in the contract forever.

contracts/SwapAndBridge.sol
nicholaspai: yep, +1 d5ba1dae353361f71de766a2acb63cae8b7bd2c9

contracts/SwapAndBridge.sol
nicholaspai: d5ba1dae353361f71de766a2acb63cae8b7bd2c9

contracts/SwapAndBridge.sol
```diff
@@ -0,0 +1,158 @@
...
+            aggregationExecutor,
+            swapDescription,
+            new bytes(0), // TODO: No IERC20Permit.permit needed since we're sending an approval?
+            new bytes(0) // TODO: We don't want to execute any data on swaps but I'm not sure how this is used.
+        );
+
+        // Sanity check that 1inch doesn't partial fill.
+        require(spentAmount == swapTokenAmount, "spent amount != swap amount");
+        // Sanity check that received amount from swap is enough to submit Across deposit with.
+        require(returnAmount >= minExpectedInputTokenAmount, "min expected input amount");
```
mrice32, line 133: Should we check the actual balance rather than depending on the 1inch output?

contracts/SwapAndBridge.sol
```diff
@@ -0,0 +1,158 @@
...
+        // Sanity check that we don't have any leftover swap tokens that would be locked in this contract.
+        require(
+            balanceBefore - IERC20(swapDescription.dstToken).balanceOf(address(this)) == swapTokenAmount,
+            "leftover swap tokens"
+        );
+
+        // Deposit the swapped tokens into Across and bridge them using remainder of input params.
+        spokePool.depositUSS(
```
mrice32, line 141: Do we need to approve the spokePool contract to pull the tokens?

contracts/SwapAndBridge.sol
```diff
@@ -0,0 +1,158 @@
...
+            depositor,
+            recipient,
+            address(swapDescription.dstToken), // input token
+            outputToken, // output token
+            returnAmount, // input amount
```
mrice32, line 146: Similar to the above question -- should we use our own balance computation rather than relying on the 1inch return value?

mrice32 (approved):
Looks good!

contracts/SwapAndBridge.sol
nicholaspai: I'm not sure if they guarantee it but that's how I interpret their contract code

contracts/SwapAndBridge.sol
nicholaspai: We can check `returnAmount` before using it here as a safety check:  3b57ed15e5bc0d851b0277caff304b4f09578191

contracts/SwapAndBridge.sol
nicholaspai: 3b57ed15e5bc0d851b0277caff304b4f09578191

contracts/SwapAndBridge.sol
nicholaspai: 3b57ed15e5bc0d851b0277caff304b4f09578191

contracts/SwapAndBridge.sol
```diff
@@ -0,0 +1,171 @@
...
+        require(returnAmount >= minExpectedInputTokenAmount, "min expected input amount");
+        // Sanity check that we don't have any leftover swap tokens that would be locked in this contract (i.e. check
+        // that we weren't partial filled).
+        require(
+            srcBalanceBefore - IERC20(swapToken).balanceOf(address(this)) == swapTokenAmount,
+            "leftover src tokens"
+        );
+
+        // Deposit the swapped tokens into Across and bridge them using remainder of input params.
+        IERC20(depositData.inputToken).safeApprove(address(spokePool), returnAmount);
```
mrice32, line 152: nit: we probably want to use safeIncreaseAllowance. safeApprove has been removed in the current version of [SafeERC20](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/utils/SafeERC20.sol) because it has some footguns.

contracts/SwapAndBridge.sol
```diff
@@ -0,0 +1,171 @@
...
+        (uint256 returnAmount, ) = oneInchRouter.swap(
+            aggregationExecutor,
+            swapDescription,
+            new bytes(0), // TODO: No IERC20Permit.permit needed since we're sending an approval?
+            new bytes(0) // TODO: We don't want to execute any data on swaps but I'm not sure how this is used.
+        );
+
+        // Sanity check that we received exactly as much as the oneInchRouter said we did.
+        require(
+            returnAmount == dstBalanceBefore - IERC20(depositData.inputToken).balanceOf(address(this)),
```
mrice32, line 139: I think we should have _more_ deposit input tokens than before the swap, right?

```suggestion
            returnAmount == IERC20(depositData.inputToken).balanceOf(address(this)) - dstBalanceBefore,
```

mrice32 (commented):
A few minor comments!

contracts/SwapAndBridge.sol
```diff
@@ -0,0 +1,313 @@
...
+
+        // Validate UniswapV3Pool:
+        uniswapV3Pool = _uniswapV3Pool;
+        // - should have token0/token1 equal to swapToken/acrossInputToken.
+        if (address(_swapToken) == _uniswapV3Pool.token0()) {
+            require(address(_acrossInputToken) == _uniswapV3Pool.token1(), "token1");
+        } else if (address(_swapToken) == _uniswapV3Pool.token1()) {
+            require(address(_acrossInputToken) == _uniswapV3Pool.token0(), "token0");
+        } else {
+            revert("token0/token1");
```
mrice32, line 171: nit: should we try to use an error type rather than a string for these?

mrice32 (approved):
LGTM!

mrice32 (approved):
I think this is great!

One addition that I would like to add in a follow-up for more optionality/flexibility: I'd like to have a variant of this contract where the swap token and input token can be passed in rather than pre-set in the contract.

nicholaspai:
> I think this is great!
> 
> 
> 
> One addition that I would like to add in a follow-up for more optionality/flexibility: I'd like to have a variant of this contract where the swap token and input token can be passed in rather than pre-set in the contract.

Cool I can add that to this file

PR #397, opened 2024-01-04
improve(SpokePool): Replace block.timestamp with getCurrentTIme()

nicholaspai:
Originally I used block.timestamp in new USS functions in order to avoid a potential JUMP opcode when calling `getCurrentTime()` however @pxrl astutely pointed out that downstream tests like those in `relayer-v2` manipulate time via `MockSpokePool.setCurrentTime()` in unit tests so using `block.timestamp` might make it harder for these unit tests to pass when modified to address the USS functions


pxrl (approved):
LGTM

PR #398, opened 2024-01-04
improve(SpokePool): Remove `destinationChainId` from USSRelayData struct

nicholaspai:
# Motivation

1. Reduce the number of params passed into fillUSS-related functions by 1 by removing `destinationChainId` from the passed in `struct`. This also has the added safety benefit of not allowing the caller to set the `destinationChainId` and instead forces them to reference a relay hash using the SpokePool's `chainId()`.

2. This additionally saves gas in these functions because it no longer needs to check the input `relayData.destinationChainId` against the contract's `chainId`.

# Side effects:

Changes implementation of how we construct relay hashes for USS fills by now hashing all of the `USSRelayData` params as a struct (i.e. as a DepositData struct) WITH the `chainId()` so that the destinationChainId is still hashed to protect against replay attacks

### DIFF
--- a/contracts/SpokePool.sol
+++ b/contracts/SpokePool.sol
@@ -91,7 +91,7 @@ abstract contract SpokePool is
     mapping(bytes32 => uint256) private DEPRECATED_relayFills;
 
     // Mapping of USS relay hashes to fill statuses. Distinguished from relayFills
-    // to eliminate any chance of collision between RelayData hashes and USSRelayData hashes.
+    // to eliminate any chance of collision between pre and post USS relay hashes.
     mapping(bytes32 => uint256) public fillStatuses;
 
     // Note: We will likely un-deprecate the fill and deposit counters to implement a better
@@ -912,7 +912,7 @@ abstract contract SpokePool is
      *         USS RELAYER FUNCTIONS          *
      ******************************************/
 
-    function fillUSSRelay(DepositData calldata relayData, uint256 repaymentChainId)
+    function fillUSSRelay(USSRelayData calldata relayData, uint256 repaymentChainId)
         public
         override
         nonReentrant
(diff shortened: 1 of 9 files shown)


contracts/SpokePool.sol
```diff
@@ -924,14 +924,9 @@ abstract contract SpokePool is
...
 
-        // Don't let caller override destination chain ID so they can't attempt a replay attack of an identical
-        // fill that was destined for a different chain. This is required because the relayData
-        // is used to form the relayHash which this contract uses to uniquely ID relays.
-        if (relayData.destinationChainId != chainId()) revert InvalidChainId();
-
         USSRelayExecutionParams memory relayExecution = USSRelayExecutionParams({
             relay: relayData,
-            relayHash: keccak256(abi.encode(relayData)),
+            relayHash: keccak256(abi.encode(relayData, chainId())),
```
pxrl, line 929: We follow the same pattern for computing the relayHash in several places and rely on it always being formatted the same. but abi.encode() won't give us this guarantee (i.e. if we forget to supply `chainId()` in one place). wdyt about a helper function that requires that we pass the chainId?

(nb. this error is tbf something that _should_ be caught in test)

pxrl (approved):
Left one question but otherwise LGTM

contracts/SpokePool.sol
nicholaspai: I was trying to reduce gas costs but I'm not sure it does anything so I'll implement the internal hash helper!

contracts/SpokePool.sol
```diff
@@ -1539,6 +1524,10 @@ abstract contract SpokePool is
         return keccak256(abi.encode(relayData));
     }
 
+    function _getUSSRelayHash(USSRelayData memory relayData) private view returns (bytes32) {
+        return keccak256(abi.encode(relayData, chainId()));
```
mrice32, line 1528: +1 great calldata reduction!!!

mrice32 (approved):
LGTM! Great change.... gas gas gas gas

PR #401, opened 2024-01-04
feat: Add _distributeRelayerRefund unit tests

nicholaspai:
Also refactors the SpokePool.sol code so that the validation of merkle leaf properties are done as close to their usage within the logic as possible

### DIFF
--- a/contracts/SpokePool.sol
+++ b/contracts/SpokePool.sol
@@ -1104,11 +1104,7 @@ abstract contract SpokePool is
     ) public virtual override nonReentrant {
         _preExecuteLeafHook(relayerRefundLeaf.l2TokenAddress);
 
-        _validateRelayerRefundLeaf(
-            relayerRefundLeaf.chainId,
-            relayerRefundLeaf.refundAddresses,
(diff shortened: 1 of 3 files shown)


contracts/SpokePool.sol
```diff
@@ -1376,15 +1370,6 @@ abstract contract SpokePool is
     // Should be overriden by implementing contract depending on how L2 handles sending tokens to L1.
     function _bridgeTokensToHubPool(uint256 amountToReturn, address l2TokenAddress) internal virtual;
 
-    function _validateRelayerRefundLeaf(
-        uint256 _chainId,
-        address[] memory refundAddresses,
-        uint256[] memory refundAmounts
-    ) internal view {
-        if (_chainId != chainId()) revert InvalidChainId();
```
nicholaspai, line 1384: I moved these two checks so that they're closer to the logic that uses the variables checked

mrice32: +1 good for readability

mrice32 (approved):
LGTM!

pxrl (approved):
LGTM

PR #402, opened 2024-01-04
feat: Remove non-USS functionality from SpokePool

nicholaspai:
This PR showcases how we will remove all non-USS functionality in the spoke pools once all non-USS deposits are fills, fills are refunded, and roots are executed.

This PR reduces the bytecode significantly and means we can use the full optimizer runs setting for compiling these new contracts

Fixes ACX-1800


contracts/SpokePool.sol
```diff
@@ -433,218 +317,8 @@ abstract contract SpokePool is
...
-     * @param originToken Token to lock into this contract to initiate deposit.
-     * @param amount Amount of tokens to deposit. Will be amount of tokens to receive less fees.
-     * @param destinationChainId Denotes network where user will receive funds from SpokePool by a relayer.
-     * @param relayerFeePct % of deposit amount taken out to incentivize a fast relayer.
-     * @param quoteTimestamp Timestamp used by relayers to compute this deposit's realizedLPFeePct which is paid
-     * to LP pool on HubPool.
-     * @param message Arbitrary data that can be used to pass additional information to the recipient along with the tokens.
-     * Note: this is intended to be used to pass along instructions for how a contract should use or allocate the tokens.
-     */
-    function deposit(
```
mrice32, line 462: I don't think we want to remove the deposit function, since it will still be used by integrators.

Less sure about other deposit variants and if any of them are used by 3rd parties.

Thoughts?

mrice32 (commented):
Looks great! One comment

contracts/SpokePool.sol
nicholaspai: i am thinking we'll merge and deploy this once we get all integrators moved over to the new function, do you think that will be a big push?

nicholaspai: Keeping them in does allow us to upgrade sooner and get benefits of optimized compilations: 0ce71e98749bf9fa25a1834f79b09556398bb45d

One consideration: deposits sent through legacy functions can still be sped up by `speedUpUSS` right? That's my instinct, just want to double check

test/gas-analytics/BondToken.erc20.ts
pxrl, line 1: ooc, how come we're nuking so much from the gas-analytics directory? This file has no relation to the SpokePool contract, for instance.

test/chain-specific-spokepools/Polygon_SpokePool.ts
```diff
@@ -280,130 +273,6 @@ describe("Polygon Spoke Pool", function () {
       .reverted;
   });
 
-  it("Must be EOA to execute relayer refund leaf with amountToReturn > 0", async function () {
```
pxrl, line 283: This was only recently added and wasn't directly related to USS/non-USS updates - confirmed that we want to remove it?

hardhat.config.ts
pxrl, line 1: Super nice!!

contracts/SpokePool.sol
```diff
@@ -162,56 +139,11 @@ abstract contract SpokePool is
     event SetXDomainAdmin(address indexed newAdmin);
     event SetHubPool(address indexed newHubPool);
     event EnabledDepositRoute(address indexed originToken, uint256 indexed destinationChainId, bool enabled);
-    event RequestedSpeedUpDeposit(
```
pxrl, line 165: I'm wondering if there's value in retaining these event definitions somewhere, so we can continue to reference them in scripts that look back beyond the upgrade bump. Otherwise we'll to define those separately somewhere else, or handle multiple contracts-v2 versions somehow. wdyt?

test/gas-analytics/BondToken.erc20.ts
nicholaspai: Bc it would require too many changes to fix tbh and I'm not sure it even works since it's not run in CI. Do you think it's worth looking into changing--is it useful?

test/chain-specific-spokepools/Polygon_SpokePool.ts
nicholaspai: Yes because we have the same test for the USS execute function, right?

contracts/SpokePool.sol
nicholaspai: Isn't the git history good enough to serve as ground truth? Additionally why not retain these in legacy files in the SDK if we want to query them for some reason. 

I don't love polluting the contracts repo with deprecated code but I'm open to suggestions

pxrl: Yeah to be clear, I wasn't suggesting to retain it in this file, but maybe somewhere else in this repo so we continue to get the benefit of the existing tooling (i.e. typechain). I'm not sure what the best way to handle deprecated contract functionality is :thinking:.

test/gas-analytics/BondToken.erc20.ts
pxrl: I found it useful for testing various scenarios previously. In this specific case the one being removed is BondToken - it shouldn't be broken by any of these changes anyway, right?

PR #403, opened 2024-01-05
feat(SpokePool): Add Natspec comments

nicholaspai:
Fixes ACX-1811

### DIFF
--- a/contracts/SpokePool.sol
+++ b/contracts/SpokePool.sol
@@ -686,8 +686,8 @@ abstract contract SpokePool is
      * where currentTime is block.timestamp on this chain or this transaction will revert.
      * @param exclusivityDeadline The deadline for the exclusive relayer to fill the deposit. After this
      * destination chain timestamp, anyone can fill this deposit on the destination chain.
-     * @param message The message to send to the recipient on the destination chain if the recipient is a contract
-     * that implements the handleUSSAcrossMessage.
+     * @param message The message to send to the recipient on the destination chain if the recipient is a contract.
+     * If the message is not empty, the recipient contract must implement handleUSSAcrossMessage() or the fill will revert.
      */
     function depositUSS(
         address depositor,
--- a/contracts/SpokePool.sol
+++ b/contracts/SpokePool.sol
@@ -652,7 +652,7 @@ abstract contract SpokePool is
      * of output tokens. The fee paid to relayers and the system should be captured in the spread between output
      * amount and input amount when adjusted to be denominated in the input token. A relayer on the destination
      * chain will send outputAmount of outputTokens to the recipient and receive inputTokens on a repayment
-     * chain of their choice plus system fees. Therefore, the fee should account for destination fee transaction costs,
+     * chain of their choice. Therefore, the fee should account for destination fee transaction costs,
      * the relayer's opportunity cost of capital while they wait to be refunded following an optimistic challenge
      * window in the HubPool, and the system fee that they'll be charged.
      * @dev On the destination chain, the hash of the deposit data will be used to uniquely identify this deposit, so
--- a/contracts/SpokePool.sol
+++ b/contracts/SpokePool.sol
@@ -1039,21 +1039,20 @@ abstract contract SpokePool is
 
     /**
      * @notice Identical to fillUSSRelay except that the relayer wants to use a depositor's updated output amount,
-     * recipient, and/or message. The relayer should only use this function if the depositor has signed a message
-     * on the origin chain asking for a modification via speedUpUSSDeposit(). If a RequestedSpeedUpUSSDeposit
-     * event matching this relayData was not emitted, then this fill will not be refunded.
+     * recipient, and/or message. The relayer should only use this function if they can supply a message signed
+     * by the depositor that contains the fill's matching deposit ID along with updated relay parameters.
+     * If the signature can be verified, then this function will emit a FilledUSSEvent that will be used by
+     * the system for refund verification purposes. In otherwords, this function is an alternative way to fill a
+     * a deposit than fillUSSRelay.
      * @dev Subject to same exclusivity deadline rules as fillUSSRelay().
      * @param relayData struct containing all the data needed to identify the deposit to be filled. See fillUSSRelay().
      * @param repaymentChainId Chain of SpokePool where relayer wants to be refunded after the challenge window has
      * passed. See fillUSSRelay().
-     * @param updatedOutputAmount New output amount to use for this deposit. Should be equivalent to value emitted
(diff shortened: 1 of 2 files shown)


contracts/SpokePool.sol
```diff
@@ -647,6 +647,48 @@ abstract contract SpokePool is
...
+     * an EOA or wrapped native token if a contract.
+     * @param inputToken The token pulled from the caller's account and locked into this contract to
+     * initiate the deposit. The equivalent of this token on the relayer's repayment chain of choice will be sent
+     * as a refund. If this is equal to the wrapped native token then the caller can optionally pass in native token as
+     * msg.value, as long as msg.value = inputTokenAmount.
+     * @param outputToken The token that the relayer will send to the recipient on the destination chain. Must be an
+     * ERC20.
+     * @param inputAmount The amount of input tokens to pull from the caller's account and lock into this contract.
+     * This amount will be sent to the relayer on their repayment chain of choice as a refund following an optimistic
+     * challenge window in the HubPool, plus a system fee.
```
pxrl, line 675: Same as above - as I read it, it seems to imply that the relayer receives `inputAmount` of `inputToken` + some additional fee, when instead they should receive `inputAmount - systemFee` of `inputToken`.

contracts/SpokePool.sol
```diff
@@ -647,6 +647,48 @@ abstract contract SpokePool is
...
+     * @param outputAmount The amount of output tokens that the relayer will send to the recipient on the destination.
+     * @param destinationChainId The destination chain identifier. Must be enabled along with the input token
+     * as a valid deposit route from this spoke pool or this transaction will revert.
+     * @param exclusiveRelayer The relayer that will be exclusively allowed to fill this deposit before the
+     * exclusivity deadline timestamp.
+     * @param quoteTimestamp The approximate Ethereum timestamp that the deposit was initiated. This is used to charge
+     * fees for this deposit. This must be set to some time between [currentTime - depositQuoteTimeBuffer, currentTime]
+     * where currentTime is block.timestamp on this chain or this transaction will revert.
```
pxrl, line 683: Integrators might find it confusing that the quoteTimestamp is "approximate".

nb. the text might need to be re-flowed after this - the gh change suggestions UI isn't very good for this :(

```suggestion
     * @param quoteTimestamp The HubPool timestamp that is used to determine the system fee paid by the depositor.
     *  This must be set to some time between [currentTime - depositQuoteTimeBuffer, currentTime]
     * where currentTime is block.timestamp on this chain or this transaction will revert.
```

contracts/SpokePool.sol
```diff
@@ -647,6 +647,48 @@ abstract contract SpokePool is
...
+     * @param fillDeadline The deadline for the relayer to fill the deposit. After this destination chain timestamp,
+     * the fill will revert on the destination chain. Must be set between [currentTime, currentTime + fillDeadlineBuffer]
+     * where currentTime is block.timestamp on this chain or this transaction will revert.
+     * @param exclusivityDeadline The deadline for the exclusive relayer to fill the deposit. After this
+     * destination chain timestamp, anyone can fill this deposit on the destination chain.
+     * @param message The message to send to the recipient on the destination chain if the recipient is a contract
+     * that implements the handleUSSAcrossMessage.
```
pxrl, line 690:
```suggestion
     * @param message The message to send to the recipient on the destination chain if the recipient is a contract.
     * If the message is not empty, the recipient contract must implement handleUSSAcrossMessage() or the fill will revert.
```

pxrl (commented):
Some initial comments; I didn't get a chance to completely review so I will follow up.

contracts/SpokePool.sol
```diff
@@ -1061,8 +1187,23 @@ abstract contract SpokePool is
...
+     * like any other fill sent through fillUSSRelay().
+     * @dev There is no relayer credited with filling this relay since funds are sent directly out of this contract.
+     * @param slowFillLeaf Contains all data necessary to uniquely identify a relay for this chain. This struct is
+     * hashed and included in a merkle root that is relayed to all spoke pools.
+     * - relayData: struct containing all the data needed to identify the original deposit to be slow filled.
+     * - chainId: chain identifier where slow fill leaf should be executed. If this doesn't match this chain's
+     * chainId, then this function will revert.
+     * - updatedOutputAmount: Amount to be sent to recipient out of this contract's balance. Can be set differently
+     * from relayData.outputAmount to charge a different fee because this deposit was "slow" filled. Usually,
+     * this will be set higher to reimburse the recipient for waiting for the slow fill.
```
pxrl, line 1203: So the updatedOutputAmount is basically selected by the proposer and verified via the standard proposal lifecycle? We'll need to track this for an UMIP update.

contracts/SpokePool.sol
```diff
@@ -912,6 +972,48 @@ abstract contract SpokePool is
...
+     * - inputAmount: This amount will be sent to the caller on their repayment chain of choice as a refund
+     * following an optimistic challenge window in the HubPool, plus a system fee.
+     * - outputAmount: The amount of output tokens that the caller will send to the recipient.
+     * - originChainId: The origin chain identifier.
+     * - exclusiveRelayer The relayer that will be exclusively allowed to fill this deposit before the
+     * exclusivity deadline timestamp.
+     * - fillDeadline The deadline for the caller to fill the deposit. After this timestamp,
+     * the fill will revert on the destination chain.
+     * where currentTime is block.timestamp on this chain or this transaction will revert.
+     * - exclusivityDeadline: The deadline for the exclusive relayer to fill the deposit. After this
```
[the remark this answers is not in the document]
nicholaspai: deleted the second line

contracts/SpokePool.sol
```diff
@@ -936,6 +1038,24 @@ abstract contract SpokePool is
         _fillRelayUSS(relayExecution, msg.sender, false);
     }
 
+    /**
+     * @notice Identical to fillUSSRelay except that the relayer wants to use a depositor's updated output amount,
+     * recipient, and/or message. The relayer should only use this function if the depositor has signed a message
+     * on the origin chain asking for a modification via speedUpUSSDeposit(). If a RequestedSpeedUpUSSDeposit
```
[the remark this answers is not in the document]
nicholaspai: the dataworker doesn't rely on the RequestSpeedUp event being scraped so yeah good point i'll generalize this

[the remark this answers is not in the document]
nicholaspai: 3fcd56cbbeb7ec0aefa8f1ff1a84660e12b8a0ad

mrice32 (approved):
LGTM!

PR #404, opened 2024-01-05
fix(SpokePool): Decrement __gap due to new fillStatuses mapping and fix storage slot layout

nicholaspai:
#379 added this new public mapping and we forgot to decrement the storage gap

Additionallly #379 did not move the new mapping to the bottom of the storage slots


mrice32 (approved):

mrice32 (approved):
LGTM!

PR #406, opened 2024-01-06
feat(Scroll): add rescue script for stuck txns

james-a-morris:
How we can replay failed L1 -> L2 messages. 

**NOTE:** per the Scroll docs, we **cannot** replay L2 -> L1 transactions. All tokens are burned.

L1 -> L2 test transactions:
1. Initial bridge request that fails: 
https://sepolia.etherscan.io/tx/0x7e74b85bdaca0358021f886fb4fbab3a72660843add5201e484285bde5ba9f04
2. Visible failed transaction: https://sepolia.scrollscan.com/tx/0x1e384bb0fb42c67234163384ce21ff836e7031210642f4f3764874eb526fc89b
3. Replay Transaction from calling this task: https://sepolia.etherscan.io/tx/0x0042f7b4792d8c08249db10c49748f29cbdd408205d3932812b44e39aa9f4e58
4. Successful landing of funds: https://sepolia.scrollscan.com/tx/0x6cd5273c7a9b16a41eab1c6319a0f36ebf7ea8b8a81ab4be5c5c45179bb130ff

The rescue script was run with the following parameters:
```
yarn hardhat rescue-stuck-scroll-txn --l1-hash 0x7e74b85bdaca0358021f886fb4fbab3a72660843add5201e484285bde5ba9f04 --gas-limit 250000 --network sepolia
```

### DIFF
--- a/tasks/rescueStuckScrollTxn.ts
+++ b/tasks/rescueStuckScrollTxn.ts
@@ -104,12 +104,13 @@ task("rescue-stuck-scroll-txn", "Rescue a failed Scroll transaction")
     if (!["1", "11155111"].includes(String(chainId))) {
       throw new Error("This script can only be run on Sepolia or Ethereum mainnet");
     }
-    const signer = (await (hre_ as any).ethers.getSigners())[0] as unknown as Signer;
+    const signer = (await hre_.ethers.getSigners())[0] as unknown as Signer;
     const messengerContract = new Contract(L1_ADDRESS_MAP[chainId].scrollMessengerRelay, relayMessengerAbi, signer);
 
     const txn = await signer.provider?.getTransactionReceipt(taskArguments.l1Hash);
-    const eventSignature = ethers.utils.id("SentMessage(address,address,uint256,uint256,uint256,bytes)");
-    const relevantEvent = txn?.logs?.find((log) => log.topics[0] === eventSignature);
+    const relevantEvent = txn?.logs?.find(
+      (log) => log.topics[0] === messengerContract.interface.getEventTopic("SentMessage")
+    );
     if (!relevantEvent) {
       throw new Error("No relevant event found. Is this a Scroll bridge transaction?");
     }
--- a/tasks/rescueStuckScrollTxn.ts
+++ b/tasks/rescueStuckScrollTxn.ts
@@ -1,9 +1,7 @@
 /* eslint-disable camelcase */
-import axios from "axios";
 import { task } from "hardhat/config";
 import { Contract, Signer, ethers } from "ethers";
 import { L1_ADDRESS_MAP } from "../deploy/consts";
-import { ZERO_ADDRESS } from "@uma/common";
 
 require("dotenv").config();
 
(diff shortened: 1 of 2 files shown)


pxrl (approved):
A couple of comments, but LGTM

tasks/rescueStuckScrollTxn.ts
```diff
@@ -0,0 +1,143 @@
...
+      value, // _value
+      messageNonce, // _messageNonce
+      message, // _message
+      ethers.BigNumber.from(taskArguments.gasLimit), // _newGasLimit
+      await signer.getAddress(), // _refundAddress
+      {
+        // 0.00001 ETH to be sent to the Scroll relayer (to cover L1 gas costs)
+        // Using recommended value default as described here: https://docs.scroll.io/en/developers/l1-and-l2-bridging/eth-and-erc20-token-bridge/
+        // *Any* leftover ETH will be immediately refunded to the signer - this is just the L1 gas cost
+        value: ethers.utils.parseEther("0.00001"),
```
nicholaspai, line 139: We have to pay value to replay a message? Aren't we paying for the L1 gas cost already by sending this txn, why do we have to send more...?

nicholaspai (commented):
https://github.com/across-protocol/contracts-v2/pull/406/files#r1445102586

tasks/rescueStuckScrollTxn.ts
james-a-morris: Yeah - unfortunately we need to send value. In order to replay the message a value has to be sent (it's required to be larger than the fee) otherwise the txn explicitly reverts.

james-a-morris: ReplayMessage is very similar to `sendMessage` in the `L1TokenMessenger` but with the exception that the parameters provided need to exactly match a previous unsuccessful L1 -> L2 txn

tasks/rescueStuckScrollTxn.ts
```diff
@@ -0,0 +1,146 @@
...
+      _from: sender,
+      _to: target,
+      _value: value.toString(),
+      _messageNonce: messageNonce.toString(),
+      _message: message.toString(),
+      _newGasLimit: taskArguments.gasLimit,
+      _refundAddress: refundAddress,
+    });
+    console.debug("Replaying message (sending with 0.001ETH )...");
+    const resultingTxn = await messengerContract.replayMessage(
```
mrice32, line 130: Wait, if an L2 message fails because of gas, we have to re-send it from _L1_? There's no way to replay it on the L2, right?

mrice32 (commented):
Looks good, just one question

tasks/rescueStuckScrollTxn.ts
james-a-morris: Per the TG chat we have to replay via an L1 transaction. The funds being rescued are on L2

This is a similar process to how the Scroll Bridge UI performs retry attempts for failed bridge attempts

nicholaspai: Does that mean we can "replay" this message by sending an EOA txn or do we need to get the HubPool to resend it?

james-a-morris: Correct - this function is permissionless. The input values to this function are enriched with the internal state of the Scroll messenger before being submitted to the queue. 

The L2 layer will also ensure that only one successful message is transmitted from L1 -> L2. Technically, you could call a "rescue" on successful relay attempts, but that would fail when it landed in Scroll

mrice32 (approved):
LGTM!

PR #407, opened 2024-01-07
feat: Add foundry.toml to support `forge` commands on repo

nicholaspai:
We can use `forge` in many ways, and the immediate way I think this could help is to snapshot and compare storage layouts to validate safe upgrades


foundry.toml
```diff
@@ -0,0 +1,25 @@
...
+    "@openzeppelin/=node_modules/@openzeppelin/",
+    "@uma/=node_modules/@uma/",
+    "@uniswap/=node_modules/@uniswap/",
+    "arb-bridge-eth/=node_modules/arb-bridge-eth/",
+    "arb-bridge-peripherals/=node_modules/arb-bridge-peripherals/",
+    "arbos-precompiles/=node_modules/arbos-precompiles/",
+    "base64-sol/=node_modules/base64-sol/",
+    "eth-gas-reporter/=node_modules/eth-gas-reporter/",
+    "hardhat-deploy/=node_modules/hardhat-deploy/",
+    "hardhat/=node_modules/hardhat/",
```
james-a-morris, line 21: We may need to add the mappings for the @scroll/tech contracts.

It's an NPM package

nicholaspai: done

mrice32 (approved):

PR #408, opened 2024-01-07
feat: Snapshot latest deployed storage layouts

nicholaspai:
Upload latest deployed storage layouts for comparison with potential new contracts

Other potential tools we could use as an alternative to forge are listed here FYI: https://proxies.yacademy.dev/pages/proxies-storage/

Example of comparing latest (commit hash 339bafab237da13c902122fde90f526ea98bc019) storage layout for `Arbitrum_SpokePool` (right file) with the latest deployed storage layout (commit hash b41fceeb54ea85a47ecda26af6a49a1ee1ebd269) (left file). You can see that the new state variable `mapping fillStatuses` is added to the right file while the `__gap` size is decremented such that the Arbitrum_SpokePool specific state variable `l2GatewayRouter` starts at the same slot 3162

Screenshot 2024-01-07 at 08 42 02


mrice32 (approved):
Should we add the tool to regenerate these on demand?

mrice32 (approved):
LGTM!

PR #412, opened 2024-01-25
feat: Deploy PermissionSplitter via forge

nicholaspai:
adds forge deployment script and other setup along with newly deployed contract at [0x0Bf07B2e415F02711fFBB32491f8ec9e5489B2e7](https://etherscan.io/address/0x0Bf07B2e415F02711fFBB32491f8ec9e5489B2e7#code)


script/DeployPermissionSplitterProxy.s.sol
```diff
@@ -0,0 +1,46 @@
...
+// 1. `source .env` where `.env has MNEMONIC="x x x ... x" and ETHERSCAN_API_KEY="x" entries
+// 2. forge script script/DeployPermissionSplitterProxy.s.sol:DeployPermissionSplitterProxy --rpc-url $NODE_URL_1-vvvv
+// 3. Verify the above works in simulation mode.
+// 4. Deploy on mainnet by adding --broadcast --verify flags.
+// 5. forge script script/DeployPermissionSplitterProxy.s.sol:DeployPermissionSplitterProxy --rpc-url $NODE_URL_1 --broadcast --verify -vvvv
+contract DeployPermissionSplitterProxy is Script, Test {
+    PermissionSplitterProxy permissionSplitter;
+
+    address constant defaultAdmin = 0xB524735356985D2f267FA010D681f061DfF03715;
+    address constant hubPool = 0xc186fA914353c44b2E33eBE05f21846F1048bEda;
```
james-a-morris, line 19: Does foundry support shared constants in a similar way to how we handle the hardhat deployments now? I assume the constants would be a solidity file

james-a-morris (commented):
Comment about consolidating some of the hardcoded addresses into a shared solidity file. This will be helpful if we want to deploy this to testnets in the future

script/DeployPermissionSplitterProxy.s.sol
```diff
@@ -0,0 +1,46 @@
...
+
+    bytes4 constant PAUSE_SELECTOR = bytes4(keccak256("setPaused(bool)"));
+    bytes32 constant PAUSE_ROLE = keccak256("PAUSE_ROLE");
+    bytes32 public constant DEFAULT_ADMIN_ROLE = 0x00;
+
+    function run() external {
+        string memory deployerMnemonic = vm.envString("MNEMONIC");
+        uint256 deployerPrivateKey = vm.deriveKey(deployerMnemonic, 0);
+        address deployerPublicKey = vm.addr(deployerPrivateKey);
```
mrice32, line 28:
```suggestion
        address deployer = vm.addr(deployerPrivateKey);
```

mrice32 (approved):
LGTM!

broadcast/DeployPermissionSplitterProxy.s.sol/1/dry-run/run-1706197795.json
```diff
@@ -0,0 +1,97 @@
+{
```
mrice32, line 1: Is this the real deployment? If so, why is it labeled "dry-run"?

nicholaspai: I believe this is the simulation step that happens for [forge scripts](https://book.getfoundry.sh/tutorials/solidity-scripting#high-level-overview)

script/DeployPermissionSplitterProxy.s.sol
nicholaspai: are you proposing moving these to a separate file? I kind of think its overkill

nicholaspai: we will use foundry in all new repos but changing this one over would be pretty difficult because of all the unit tests.

At this point, foundry support is available within the repo so we can use it for new scripts/tests, etc.

PR #417, opened 2024-02-02
feat: Add FundsDeposited event for backwards compatible ABI creation

nicholaspai:
This makes unit tests easier in `relayer-v2` to simulate running against a contract that has emitted both `FundsDeposited` and `V3FundsDeposited` events.


pxrl (approved):
Will make life easier :+1:

PR #420, opened 2024-02-05
improve: Restore TokensBridged and ExecutedRelayerRefundLeaf event signatures

nicholaspai:
We removed these event parameters because they were unused but didn't realize this would inadvertently break clients that query these event signatures. Once contracts are updated, clients wouldn't be able to query historical instances of these events, which is required for the `finalizer` to to finalize remaining withdrawals from deprecated contract versions.


pxrl (approved):
LGTM - I verified that `caller` was the last argument to the event in v2 and that it was not indexed.

dohaki (approved):
lgtm

PR #446, opened 2024-02-21
chore: Bump beta version & sync Base Sepolia ABI

pxrl:


pxrl:
This PR doesn't make any changes to the underlying dependencies sourced. The issues raised by socket security relate to:
 - Deprecation notices for web3 packages.
 - Install scripts that can clean the local directory.
 - husky hooks that access the filesystem.
 - hid/usb packages that inherently interact with the underlying system.

Merging this as-is because I don't see any marginal increase in risk exposure as a result of these warnings.

PR #452, opened 2024-02-26
feat(espresso): add deployments for upcoming demo

james-a-morris:
Adds deployment definitions for the upcoming Espresso demonstration. Two custom rollups (id 10,001 & 10,002) have been created. 

Note: since no canonical bridge exists, the corresponding chain-adapter's `relayMessage` is a no-op & the spoke pools' `_requireAdminSender` is also a no-op. These are _very bad_ candidates for testing the dataworker.

Note: based on the sequencer, these two proxy addresses have the same address. This is intended


hardhat.config.ts
```diff
@@ -125,7 +125,7 @@ const config: HardhatUserConfig = {
       verifyURL: "https://zksync2-mainnet-explorer.zksync.io/contract_verification",
     },
     optimism: {
-      url: getNodeUrl("optimism", true, 10),
+      url: getNodeUrl("optimism-mainnet", true, 10),
```
nicholaspai, line 128: Is this right? How has this worked before?

nicholaspai (approved):
One q

hardhat.config.ts
james-a-morris: Yeah - the first variable is the subdomain. On infura it's `optimism-mainnet` not `optimism`

james-a-morris:
Can close because the demo is complete.

PR #456, opened 2024-02-29
feat: Remove SpokePool v2 functionality

pxrl:
Original PR: https://github.com/across-protocol/contracts-v2/pull/402

Side effects:

Preserve deposit to maintain backwards compatibility with integrators
Preserve depositFor to maintain backwards compatibility with SpokePoolVerifier
Replace depositNow with depositUSSNow to preserve multisig compatibility
Replace depositFor with depositUSS in SpokePoolVerifier
Replace require strings with custom error types to reduce bytecode size
Note/Breaking changes:

Once this PR is merged and the contracts are upgraded on-chain, no more V2 events can be emitted. So the dataworker/sdk can safely remove all logic that is backwards compatible and handles V2 events. Before this PR, the only that V2 events can be emitted is if a very very old deposit is filled, or a V2 fill is sent to try to grief the system, or a very old slow fill V2 leaf is executed.


nicholaspai (approved):
LGTM

PR #457, opened 2024-02-29
fix: remove hardhat from standard import

mrice32:
This PR will remove hardhat from the standard import path. This means that ts/js code that does:
```
import ... from "@across-protocol/contracts-v2"
```

will not need a hardhat config and will not spin up a local hardhat node by default. I've verified that the files in utils don't seem to be used in sdk-v2 or relayer-v2.

I think this was accidentally moved into our production exports in a larger change: https://github.com/across-protocol/contracts-v2/pull/272.

### DIFF
--- a/utils/index.ts
+++ b/utils/index.ts
@@ -1,3 +1,2 @@
-export * from "./utils";
 export * from "./MerkleTree";
 export * from "./constants";


nicholaspai (approved):

james-a-morris (approved):
+1 Thoughts on deploying three packages to NPM?

1. @across-protocol/contracts-v2/contracts (just contracts + typechain)
2. @across-protocol/contracts-v2/test-suite (just the utils that need hardhat)
3. @across-protocol/contracts-v2 (amalgam of 1 & 2)

We'd mostly import (1) but 2 could be useful as a dev resource.

pxrl (approved):

PR #459, opened 2024-03-01
feat: Deploy SpokePoolVerifier that uses depositV3

nicholaspai:


james-a-morris (commented):
We would need to redeploy the FE with this updated deployments.json but this looks great

nicholaspai:
> We would need to redeploy the FE with this updated deployments.json but this looks great

yep exactly

dohaki (approved):
LGTM

deployments/deployments.json
pxrl, line 1: @nicholaspai I noticed that the mainnet deployment wasn't updated as part of this commit, and it still references the pre-v3 version. Was that intentional?

nicholaspai: ah yes i forgot to deploy

mrice32 (approved):
LGTM!

PR #461, opened 2024-03-01
feat: Deploy new SpokePool implementations 

nicholaspai:
- ZkSync: https://explorer.zksync.io/address/0x0DA0a157D40B3348285F1a88a329c9387e51bCF9#contract
- Polygon: https://polygonscan.com/address/0x9f444346FD853084158ce9cfa7e062BaBd51a577#code
- Optimism: https://optimistic.etherscan.io/address/0x5226a5D010F24F0df089963E94ce88E4c87EB53b#code
- Base: https://basescan.org/address/0x053Dc4a08302F11d13d8D2B87D8F3c9c28cEDEb4#code
- Arbitrum: https://arbiscan.io/address/0x53dcC21887d54f41B049724b73a2195CE7343873#code
- Ethereum: https://etherscan.io/address/0xa4D3535f33549749Fb97fA42903AC80F6fb54af6#code


nicholaspai:
> @nicholaspai Do you plan to update `deployments.json` in this PR?
> 
> Aside from that...this is all autogenerated, but do you think there are any specific checks we should make as part of reviewing this change?

Nope, AFAIK the deployments.json only lists the proxy contracts or the SpokePool addresses that any client would be interfacing with

nicholaspai:
> > @nicholaspai Do you plan to update `deployments.json` in this PR?
> > Aside from that...this is all autogenerated, but do you think there are any specific checks we should make as part of reviewing this change?
> 
> Nope, AFAIK the deployments.json only lists the proxy contracts or the SpokePool addresses that any client would be interfacing with

I think checking the forge layouts is important, i can publish those as part of this PR?

nicholaspai:
Consider re-deploying after removing `depositFor` if we upgrade the SpokePoolVerifier in the frontend which IIUC is the only user of this function

pxrl:
> > @nicholaspai Do you plan to update `deployments.json` in this PR?
> > Aside from that...this is all autogenerated, but do you think there are any specific checks we should make as part of reviewing this change?
> 
> Nope, AFAIK the deployments.json only lists the proxy contracts or the SpokePool addresses that any client would be interfacing with

Yeah, good point - no need to update deployments.json :+1:

mrice32 (approved):
Looks good! +1 on the new storage layouts in the diff -- very helpful to take a pass on those before deploying.

If the removal of depositFor was not removed in the audited version, I would suggest we leave it.

I'm assuming it's not causing bytecode issues or anything, right?

nicholaspai:
> Looks good! +1 on the new storage layouts in the diff -- very helpful to take a pass on those before deploying.
> 
> If the removal of depositFor was not removed in the audited version, I would suggest we leave it.
> 
> I'm assuming it's not causing bytecode issues or anything, right?

nope!

PR #478, opened 2024-03-26
fix(Linea): Unwrap correct amount before returning ETH to HubPool

pxrl:
msg.value is unconditionally wrapped by _preExecuteLeafHook(), so it must be unwrapped again before sending ETH back to the HubPool.


mrice32 (approved):
LGTM!

nicholaspai (approved):

PR #481, opened 2024-03-28
feat: Add event in SwapAndBridge

nicholaspai:
This will be important to track for frontends to determine when a non-supported token is swapped into a supported token via this contract before submitting an Across Bridge

### DIFF
--- a/contracts/SwapAndBridge.sol
+++ b/contracts/SwapAndBridge.sol
@@ -53,6 +53,16 @@ abstract contract SwapAndBridgeBase is Lockable, MultiCaller {
         bytes message;
     }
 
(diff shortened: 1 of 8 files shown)


james-a-morris (commented):
Are the storage-layout diffs related to the contracts change?

nicholaspai:
> Are the storage-layout diffs related to the contracts change?

yeah when you update anything in the repo, solidity compiler sometimes updates the metadata of the deployment files. These changes are not storage slot changes which is what we care about

mrice32 (approved):
LGTM!

james-a-morris (approved):
LGTM

PR #485, opened 2024-04-24
feat(MultiCallerUpgradeable): multicall function that allows delegatecall failures

bmzig:
This upgrade should enable parties to multicall functions from a SpokePool without reverting when a single `delegatecall` fails. This is useful since with this upgrade, using `multicall` to batch execute functions like `fillV3Relay` will succeed, even if some of the corresponding deposits were filled by another relayer in the window of time from transaction construction to execution.


bmzig:
There is another very "hacky" way to get this done by encoding booleans in the `bytes[] calldata data` array. We can either prepend the last value with some artifact (e.g. `0xd00d`) and then a boolean string of size `data.length-1` to get precise selection on which delegatecalls can fail/succeed, or (probably a worse idea) check to see if the last element in `data` is strictly less than 32 bits long (so it won't be confused as a function signature). For the former case:

```
// revertLogic is boolean string
...
if iszero(or(and(revertLogic, 0x1), mload(success))) {
    // Do the revert stuff
}
revertLogic := shr(1, revertLogic)
```

contracts/upgradeable/MultiCallerUpgradeable.sol
```diff
@@ -46,6 +46,29 @@ contract MultiCallerUpgradeable {
...
+        for (uint256 i = 0; i < dataLength; i++) {
+            // Typically, implementation contracts used in the upgradeable proxy pattern shouldn't call `delegatecall`
+            // because it could allow a malicious actor to call this implementation contract directly (rather than
+            // through a proxy contract) and then selfdestruct() the contract, thereby freezing the upgradeable
+            // proxy. However, since we're only delegatecall-ing into this contract, then we can consider this
+            // use of delegatecall() safe.
+
+            //slither-disable-start low-level-calls
+            /// @custom:oz-upgrades-unsafe-allow delegatecall
+            (, bytes memory result) = address(this).delegatecall(data[i]);
```
mrice32, line 65: nit: it's probably a good idea to tell the caller which calls succeeded and which errored, so maybe we should return an array of structs with a boolean _and_ bytes?

mrice32 (commented):

contracts/upgradeable/MultiCallerUpgradeable.sol
```diff
@@ -46,6 +52,25 @@ contract MultiCallerUpgradeable {
...
+    function tryMulticall(bytes[] calldata data) external returns (Result[] memory results) {
+        _validateMulticallData(data);
+
+        uint256 dataLength = data.length;
+        results = new Result[](dataLength);
+
+        //slither-disable-start calls-loop
+        for (uint256 i = 0; i < dataLength; i++) {
+            // The delegatecall here is safe for the same reasons outlined in the first multicall function.
+
```
james-a-morris, line 64: Nit

```suggestion
```

contracts/upgradeable/MultiCallerUpgradeable.sol
```diff
@@ -46,6 +52,25 @@ contract MultiCallerUpgradeable {
         //slither-disable-end calls-loop
     }
 
+    function tryMulticall(bytes[] calldata data) external returns (Result[] memory results) {
+        _validateMulticallData(data);
```
james-a-morris, line 56: Since we're modifying both `multicall` & `tryMulticall`, would it make sense to convert `_validateMulticallData` into a modifier of some sort?

james-a-morris (commented):
Two comments (one potentially actionable & one nit)

pxrl:
Discussed in a sync call: the new tests added aren't automatically executed by CI, so we should probably add that in advance.

contracts/upgradeable/MultiCallerUpgradeable.sol
bmzig: That's a good question. To my understanding, if we do end up writing some code in `_validateMulticallData` such as in Polygon_SpokePool.sol, using a modifier instead of the function would cost double the contract bytecode size for the included modifier logic. Apparently some of the SpokePool contracts are pretty close to the max size, and from testing, the gas difference between the two is pretty negligible.

nicholaspai: yeah modifiers cost more bytecode so i think its ok to keep as is

test/foundry/local/MultiCallerUpgradeable.t.sol
```diff
@@ -0,0 +1,107 @@
...
+        }
+    }
+
+    function testTryMulticallWithFailures(uint8 numberOfFunctions, uint256 randomSeed) public {
+        numberOfFunctions = uint8(bound(numberOfFunctions, 1, 255));
+
+        bytes[] memory calls = new bytes[](numberOfFunctions);
+
+        // Set the first call to a success so we can simulate failing on relays which were already taken.
+        calls[0] = abi.encodeWithSelector(SpokePool.fillV3Relay.selector, mockRelayData, 1);
```
nicholaspai, line 83: Do you think we can change this test to not rely on the SpokePool.fillV3Relay internal implementation logic? Its a bit subtle but this test assumes that fillRelay cannot succeed for the same relayData (including its depositId). 

Can we instead define a contract in this test that has a more obvious reverting function? maybe something like even test(bool shouldPass) which reverts if `false` is passed in.

nicholaspai: This way, this unit test won't suddenly break if the fillRelay logic changes drastically for some reaso

nicholaspai (approved):
https://github.com/across-protocol/contracts-v3/pull/485/files#r1586744424

this look great, just one suggestion on the unit tests

nicholaspai:
> Discussed in a sync call: the new tests added aren't automatically executed by CI, so we should probably add that in advance.

also yes let's add this!

test/foundry/local/MultiCallerUpgradeable.t.sol
bmzig: Sounds good. The reverts happen now by calling an admin function.

PR #486, opened 2024-04-25
feat: Change default fillDeadlineBuffer from 8 hours to 6 hours

nicholaspai:
Bundle liveness has been reduced to 60 mins from 90 mins. This parameter can be set no contract construction and is roughly based on bundle challenge window which is the length of time that a deposit can be "unfilled" for before it becomes expired and is due a refund. Therefore, we can reduce the fill deadline buffer, which reduces burden on off-chain infra

### DIFF
--- a/deploy/003_deploy_optimism_spokepool.ts
+++ b/deploy/003_deploy_optimism_spokepool.ts
@@ -25,7 +25,7 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
   const constructorArgs = [
     "0x4200000000000000000000000000000000000006",
     3600,
-    28800,
+    21600,
     L2_ADDRESS_MAP[spokeChainId].l2Usdc,
     // L2_ADDRESS_MAP[spokeChainId].cctpTokenMessenger,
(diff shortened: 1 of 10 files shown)


pxrl (approved):
Thoughts on defining this as a constant somewhere?

nicholaspai:
> Is the plan to redeploy with this value updated?

Yeah. At the very least we'll have to redeploy the spoke pools using CCTP so it'll be good to have this in as well. Do you think there is risk to deploying this in pieces (with Polygon, Optimism, Base, Arbitrum getting deployed first as part of CCTP migration)?

mrice32:
> > Is the plan to redeploy with this value updated?
> 
> Yeah. At the very least we'll have to redeploy the spoke pools using CCTP so it'll be good to have this in as well. Do you think there is risk to deploying this in pieces (with Polygon, Optimism, Base, Arbitrum getting deployed first as part of CCTP migration)?

I think the risk is that the dataworker assumes that this is set to 6 hours everywhere before it's updated everywhere and expirations are missed. If we're not reading this in programmatically, it might be best to try and update it along with CCTP, but redeploy the chains without CCTP as well. Thoughts?

PR #488, opened 2024-04-30
fix(test): include cctp relevant param in mock

james-a-morris:
This is causing some failures in relayer tests where we reference the mock spoke pool.


pxrl:
@james-a-morris Looks like this also requires the storage layouts to be updated. Maybe we can find a way of excluding the ast IDs :thinking:

nicholaspai (approved):

nicholaspai:
> @james-a-morris Looks like this also requires the storage layouts to be updated. Maybe we can find a way of excluding the ast IDs 🤔

I think its just easiest if you run the script to reupload the storage layouts

james-a-morris:
We have a workaround here - will close this out.

PR #489, opened 2024-04-30
feat: add multicall handler

mrice32:
This is a generic handler where the user can perform a series of calls (multicall) after their tokens are received in the contract.

User can specify a fallback address that is:
- Paid all remaining tokens whether the call succeeds or fails.

If the user doesn't specify a fallback address, the call must succeed.

The user can call drainLeftoverTokens on the handler in their calls to drain any remaining amount of any token or ETH in their calls. This is meant as a way to deal with calls where the amount of tokens they take isn't known a priori.


contracts/handlers/AcrossMulticallHandler.sol
```diff
@@ -0,0 +1,79 @@
...
+    ) external nonReentrant {
+        Call[] memory calls = abi.decode(message, (Call[]));
+
+        for (uint256 i = 0; i < calls.length; ) {
+            Call memory call = calls[i];
+            (bool success, ) = call.target.call{ value: call.value }(call.callData);
+            if (!success) revert CallReverted(i);
+            unchecked {
+                ++i;
+            }
```
pxrl, line 50: Is this useful post Solidity 0.8.22?

https://soliditylang.org/blog/2023/10/25/solidity-0.8.22-release-announcement/

contracts/handlers/AcrossMulticallHandler.sol
```diff
@@ -0,0 +1,79 @@
...
+        address,
+        uint256,
+        address,
+        bytes memory message
+    ) external nonReentrant {
+        Call[] memory calls = abi.decode(message, (Call[]));
+
+        for (uint256 i = 0; i < calls.length; ) {
+            Call memory call = calls[i];
+            (bool success, ) = call.target.call{ value: call.value }(call.callData);
```
nicholaspai, line 46: How does this contract get `value`? Doesn't it receive WETH as a contract?

contracts/handlers/AcrossMulticallHandler.sol
```diff
@@ -0,0 +1,79 @@
...
+            if (!success) revert CallReverted(i);
+            unchecked {
```
nicholaspai, line 48: I don't think you need this `unchecked` for gas optimizations anymore as of solidity 0.8.22

contracts/handlers/AcrossMulticallHandler.sol
```diff
@@ -0,0 +1,79 @@
...
+     */
+    function handleV3AcrossMessage(
+        address,
+        uint256,
+        address,
+        bytes memory message
+    ) external nonReentrant {
+        Call[] memory calls = abi.decode(message, (Call[]));
+
+        for (uint256 i = 0; i < calls.length; ) {
```
bmzig, line 44: If we are going for maximum gas efficiency and if we have the stack size, caching `calls.length` into a local variable saves us 3 gas per loop (1 `mload` operation).

mrice32: Good call

contracts/handlers/AcrossMulticallHandler.sol
mrice32: Good pt, glad we can drop these now

contracts/handlers/AcrossMulticallHandler.sol
mrice32: +1, glad we can drop this. I think we're past 0.8.22 in this repo for fresh deployments

contracts/handlers/AcrossMulticallHandler.sol
mrice32: You might unwrap the WETH in one of your calls, right?

contracts/handlers/AcrossMulticallHandler.sol
mrice32: Done.

contracts/handlers/AcrossMulticallHandler.sol
mrice32: Done.

contracts/handlers/AcrossMulticallHandler.sol
mrice32: Done.

contracts/handlers/AcrossMulticallHandler.sol
```diff
@@ -0,0 +1,77 @@
...
+     * @param message abi encoded array of Call structs, containing a target, callData, and value for each call that
+     * the contract should make.
+     */
+    function handleV3AcrossMessage(
+        address,
+        uint256,
+        address,
+        bytes memory message
+    ) external nonReentrant {
+        Call[] memory calls = abi.decode(message, (Call[]));
```
james-a-morris, line 42: Is there any validation we want to do here on converting `message` into `Call[]`

james-a-morris:
> Open question:
> 
> 1. Should we create multiple versions of this contract with slight variants on the `Call` struct? For instance, one could exclude `value` and one could _include_ a `canFail` boolean to allow some calls to fail without reverting. The idea behind multiple versions is that we might not want users who don't need these features to pay for the extra calldata (`value` and `canFail` will add 64 bytes total to the calldata for each call).

 I think this makes sense. OOC could we also have a byte at the start of `message` determine what format we can expect `Call[]` to be? This way, we could unpack accordingly and only need the one contract

bmzig:
> 1. Should we create multiple versions of this contract with slight variants on the `Call` struct? For instance, one could exclude `value` and one could _include_ a `canFail` boolean to allow some calls to fail without reverting. The idea behind multiple versions is that we might not want users who don't need these features to pay for the extra calldata (`value` and `canFail` will add 64 bytes total to the calldata for each call).

From what I found, the price of deployment in gas scales by around $78 \times \text{byte}$, while the price of extra calldata scaled by 4 or 7 gas per byte, so if these contracts will have a lot of volume, which they presumably will, it probably will be worth it to pay the cost of a more expensive deployment for less expensive transactions. 

> 2. Should we create two variants of the contract, one that uses tstore reentrancy guards and one that uses oldschool reentrancy guards? The latter will probably be required to support chains that don't yet have the Cancun EVM upgrade. The former will save gas for chains that do. This might be an overoptimization, though.

If it is a simple implementation, I'd say its a good idea.

contracts/handlers/AcrossMulticallHandler.sol
mrice32: What sort of validation would you suggest?

My understanding is that abi.decode will cause a revert if the format is wrong.

contracts/handlers/AcrossMulticallHandler.sol
nicholaspai: yeah fair

PR #491, opened 2024-05-08
feat: Add Blast contracts

nicholaspai:
This adds a new adapter, spoke pool for blast.

A few considerations:
- For stables, only DAI is currently available on Blast, since they combine all stables into USDB (which maps to DAI). Any other stables would need to be swapped when going to or from blast to get to/from DAI first. For now, Across will only support DAI.
- For ETH and USDB, yield is set to claimable at init time, and then it is claimed whenever those tokens are bridged back to mainnet, generating a surplus in the HubPool.
- Native ETH is handled similarly. When WETH is bridged back, native yield (and gas) is claimed and added to the bridged amount.

One open question: is giving the hub pool the surplus the right call? It may not show up in _fees_ until the surplus outweighs all outstanding liabilities.

Some examples:

- example USDB withdrawal https://blastscan.io/tx/0xfcdd9994a825b27f8f4f7352168399a7f7d782604b31cf206c4445ff85e02a36
- example WETH withdrawal
- example DAI deposit https://etherscan.io/tx/0xda744982971935fb89140265c9e21d203984a2b54dc89c3c37d983a2d183b2e6
- example WETH deposit
- example WBTC deposit https://etherscan.io/tx/0x8226335063c97314d2cd7ae1f84b790b5f0911ad50aae870513a54858f80d9cc

Resources:
- https://docs.blast.io/building/contracts
- https://github.com/blast-io/blast/tree/04ed82f54a627622f82bea2217a090106d1ca2c2/blast-optimism/packages/contracts-bedrock


contracts/Blast_SpokePool.sol
```diff
@@ -0,0 +1,118 @@
...
+        if (claimableAmount != claimedAmount) {
+            revert InvalidClaimedAmount(address(token));
+        }
+    }
+
+    /**
+     * @notice Claims any yield for tokens that accrue yield and then also bridges those
+     */
+    function _bridgeTokensToHubPool(uint256 amountToReturn, address l2TokenAddress) internal override {
+        uint256 amountToWithdraw = amountToReturn;
```
mrice32, line 77: nit: why define a new variable? Why not just increment the value passed in? Note: integers are passed by value (copied), so this won't and can't impact any variables in the caller's context

contracts/chain-adapters/Blast_Adapter.sol
```diff
@@ -0,0 +1,100 @@
...
+/**
+ * @notice Contract containing logic to send messages from L1 to Blast. This is a modified version of the Optimism adapter
+ * that excludes the custom bridging logic. It differs from the Base Adapter in that it uses a special
+ * Blast contract to bridge WETH and DAI, which are yielding rebasing tokens on L2, WETH and USDB.
+ */
+
+// solhint-disable-next-line contract-name-camelcase
+contract Blast_Adapter is CrossDomainEnabled, AdapterInterface, CircleCCTPAdapter {
+    using SafeERC20 for IERC20;
+    uint32 public constant L2_GAS_LIMIT = 200_000;
```
mrice32, line 28: nit: should we make this immutable and pass it in in the constructor, so it can be changed without changing the contract code if we wanted to redeploy with a different value?

contracts/chain-adapters/Blast_Adapter.sol
```diff
@@ -0,0 +1,100 @@
...
+
+    WETH9Interface public immutable L1_WETH;
+
+    IL1StandardBridge public immutable L1_STANDARD_BRIDGE;
+
+    // Bridge used to get yielding version of ERC20's on L2.
+    address private L1_BLAST_BRIDGE = 0x3a05E5d33d7Ab3864D53aaEc93c8301C1Fa49115;
```
mrice32, line 35: nit: should we make this and the DAI address below immutable and passed in the constructor so this contract will work on testnet?

contracts/Blast_SpokePool.sol
```diff
@@ -0,0 +1,118 @@
...
+
+/**
+ * @notice Blast Spoke pool.
+ */
+contract Blast_SpokePool is Ovm_SpokePool {
+    // This is the yield-accruing stablecoin on Blast that USDC/DAI/USDT all bridge into. It can be withdrawn
+    // from L2 into DAI.
+    address private constant USDB = 0x4300000000000000000000000000000000000003;
+    // Token that is received when withdrawing USDB, aka DAI.
+    address private constant L1_USDB = 0x6B175474E89094C44Da98b954EedeAC495271d0F;
```
mrice32, line 24: Same comment here: since this would likely be a different address on testnet, should we make it immutable and pass it in in the constructor?

mrice32 (commented):
Looks great! Only very minor nits and comments

contracts/chain-adapters/Blast_Adapter.sol
```diff
@@ -0,0 +1,106 @@
...
+
+        // If the l1Token is weth then unwrap it to ETH then send the ETH to the blast bridge.
+        if (l1Token == address(L1_WETH)) {
+            L1_WETH.withdraw(amount);
+            // @dev: we can use the standard or the blast bridge to deposit ETH here:
+            L1_STANDARD_BRIDGE.depositETHTo{ value: amount }(to, L2_GAS_LIMIT, "");
+        }
+        // Check if this token is DAI, then use the L1 Blast Bridge
+        else if (l1Token == L1_DAI) {
+            IL1ERC20Bridge(L1_BLAST_BRIDGE).depositERC20To(l1Token, l2Token, to, amount, L2_GAS_LIMIT, "");
```
nicholaspai, line 97: Do we not have to approve for DAI?

nicholaspai (commented):
Looks great  I just have one q

contracts/chain-adapters/Blast_Adapter.sol
mrice32: Good call, we do.

mrice32: Done.

mrice32 (approved):

PR #492, opened 2024-05-08
feat: Add ERC7683 depositor

nicholaspai:
This adds an ERC7683 Depositor contract that takes a signed ERC7683 order and converts it into an AcrossV3 deposit.


contracts/erc7683/ERC7683Depositor.sol
```diff
@@ -0,0 +1,293 @@
...
+        );
+
+        // Verify Permit2 signature and pull user funds into this contract
+        _processPermit2Order(PERMIT2, order, resolvedOrder, signature);
+
+        IERC20(resolvedOrder.swapperInputs[0].token).safeIncreaseAllowance(
+            address(SPOKE_POOL),
+            resolvedOrder.swapperInputs[0].amount
+        );
+        SPOKE_POOL.depositV3(
```
mrice32, line 191: Thoughts on integrating directly into the SpokePool contract rather than making an external call?

Pros: cheaper on gas, people are permitting the spoke pool contract directly
Cons: more complexity in the spoke pool

contracts/erc7683/ERC7683Depositor.sol
```diff
@@ -0,0 +1,293 @@
...
+
+import "@openzeppelin/contracts/utils/math/SafeCast.sol";
+import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+
+import "./Permit2OrderLib.sol";
+
+/// @title CrossChainOrder type
+/// @notice Standard order struct to be signed by swappers, disseminated to fillers, and submitted to settlement contracts
+struct CrossChainOrder {
```
mrice32, line 16: nit: might make sense to move the canonical ERC7683 types into a separate file, just to distinguish them from our Across-specific types.

contracts/erc7683/ERC7683Depositor.sol
```diff
@@ -0,0 +1,293 @@
...
+
+// Data unique to every attempted order fulfillment
+struct AcrossFillerData {
+    // Filler can choose where they want to be repaid
+    uint32 repaymentChainId;
+}
+
+/// @title ISettlementContract
+/// @notice Standard interface for settlement contracts
+interface ISettlementContract {
```
mrice32, line 104: Same for this interface: might be good to put it in a separate file for just "pure" ERC7683 interfaces/structs

contracts/erc7683/ERC7683Depositor.sol
```diff
@@ -0,0 +1,293 @@
...
+    uint32 exclusivityDeadline;
+    address exclusiveRelayer;
+    bytes message;
+}
+
+// Data unique to every attempted order fulfillment
+struct AcrossFillerData {
+    // Filler can choose where they want to be repaid
+    uint32 repaymentChainId;
+}
```
mrice32, line 100: Since we can't accurately assess fees if the filler takes repayment elsewhere and we can't enforce this choice on the fill side, thoughts on just resolving _assuming that they will take repayment on origin_ and not having them specify the repayment chain here?

Then the fees and repayment location change if they input a different repayment chain on the destination?

mrice32 (commented):
Looks good! A few comments

mrice32 (approved):
LGTM!

contracts/erc7683/ERC7683Depositor.sol
nicholaspai: we're also going to run into bytecode limits

contracts/erc7683/ERC7683.sol
```diff
@@ -0,0 +1,96 @@
...
+    Input[] swapperInputs;
+    /// @dev The outputs to be given to the swapper as part of order fulfillment
+    Output[] swapperOutputs;
+    /// @dev The outputs to be given to the filler as part of order settlement
+    Output[] fillerOutputs;
+}
+
+/// @title ISettlementContract
+/// @notice Standard interface for settlement contracts
+interface ISettlementContract {
```
nicholaspai, line 75: should this file be named IERC7683.sol?

nicholaspai (commented):
looks great, thanks for refactoring

mrice32:
Will add tests in a follow-up

PR #494, opened 2024-05-10
feat: Add Mode chain adapter and spoke pool

bmzig:
We need to add a new adapter contract for Mode. Mode is built on the OP stack, so most of this is identical to the Base contract and deployment logic.


deploy/037_deploy_mode_spokepool.ts
```diff
@@ -0,0 +1,35 @@
...
+  //    * A WETH address of the WETH address
+  //    * A depositQuoteTimeBuffer of 1 hour
+  //    * A fillDeadlineBuffer of 6 hours
+  //    * Native USDC address on L2
+  //    * CCTP token messenger address on L2
+  const constructorArgs = [
+    "0x4200000000000000000000000000000000000006",
+    3600,
+    21600,
+    ZERO_ADDRESS,
```
bmzig, line 26: Does this look right? Mode does not have native USDC, only USDC.e.

contracts/chain-adapters/Mode_Adapter.sol
```diff
@@ -0,0 +1,97 @@
...
+     */
+    constructor(
+        WETH9Interface _l1Weth,
+        address _crossDomainMessenger,
+        IL1StandardBridge _l1StandardBridge,
+        IERC20 _l1Usdc,
+        ITokenMessenger _cctpTokenMessenger
+    )
+        CrossDomainEnabled(_crossDomainMessenger)
+        CircleCCTPAdapter(_l1Usdc, _cctpTokenMessenger, CircleDomainIds.Ethereum)
```
james-a-morris, line 52: Is CCTP enabled on this chain?

james-a-morris (commented):
We may want to ask if Circle is planning to add CCTP on Mode. If not, we may not want to include the CCTP adapter in the adapter.

contracts/chain-adapters/Mode_Adapter.sol
bmzig: All the interfaces/addresses for the config is set to the zero address, but yeah, there is no Circle domain ID for Mode, and it also does not look like there is any information about whether or not there will be CCTP support in the near future. @mrice32 and @nicholaspai are much better suited than me at knowing exactly what to do in this case, though.

bmzig: For safety, we are going to keep the CCTP adapter and initialize it to the zero address since all the adapter/spoke pool code has been audited with the CCTP integration.

contracts/chain-adapters/Mode_Adapter.sol
```diff
@@ -0,0 +1,96 @@
...
+     */
+    constructor(
+        WETH9Interface _l1Weth,
+        address _crossDomainMessenger,
+        IL1StandardBridge _l1StandardBridge,
+        IERC20 _l1Usdc,
+        ITokenMessenger _cctpTokenMessenger
+    )
+        CrossDomainEnabled(_crossDomainMessenger)
+        CircleCCTPAdapter(_l1Usdc, _cctpTokenMessenger, CircleDomainIds.Ethereum) // CCTP is not enabled on this chain so Ethereum is a placeholder
```
james-a-morris, line 51: @nicholaspai @mrice32 we're setting this to Ethereum as a placeholder even though it's a no-op since `_cctpTokenMessager` will be zero.

We *need* to ensure that `_cctpTokenMessenger` is empty because this domain will incorrectly route + the `ZERO_ADDRESS` we need to pass for `_l1Usdc` is also invalid.

deploy/037_deploy_mode_spokepool.ts
james-a-morris: Yes we should be able to use the ZERO_ADDRESS as long as we enforce that `cctpTokenMessenger` is the ZERO_ADDRESS

deploy/036_deploy_mode_adapter.ts
```diff
@@ -0,0 +1,33 @@
...
+      // L1_ADDRESS_MAP[chainId].cctpTokenMessenger,
+      // For now, we are not using the CCTP bridge and can disable by setting
+      // the cctpTokenMessenger to the zero address.
+      ZERO_ADDRESS,
+    ],
+  });
+};
+
+module.exports = func;
+func.dependencies = ["HubPool"];
```
nicholaspai, line 32: can we remove this `dependency`?

nicholaspai (approved):
Looks right just one comment

deploy/037_deploy_mode_spokepool.ts
bmzig: Great. Will leave it as is.

deploy/036_deploy_mode_adapter.ts
bmzig: We can. It is done here: https://github.com/across-protocol/contracts-v3/pull/494/commits/ec8fe263ed75eb549c709718c27619144dd3f610

PR #500, opened 2024-05-21
fix(CircleCCTPAdapter): Split messages into parts under the limit

nicholaspai:
Reads limit for token from token minter, splits amount to send into parts under the limit


contracts/libraries/CircleCCTPAdapter.sol
```diff
@@ -81,7 +81,20 @@ abstract contract CircleCCTPAdapter {
     function _transferUsdc(address to, uint256 amount) internal {
         // Only approve the exact amount to be transferred
         usdcToken.safeIncreaseAllowance(address(cctpTokenMessenger), amount);
-        // Submit the amount to be transferred to bridged via the TokenMessenger
-        cctpTokenMessenger.depositForBurn(amount, recipientCircleDomainId, _addressToBytes32(to), address(usdcToken));
+        // Submit the amount to be transferred to bridged via the TokenMessenger.
+        // If the amount to send exceeds the burn limit per message, then split the message into smaller parts.
+        ITokenMinter cctpMinter = cctpTokenMessenger.localMinter();
```
james-a-morris, line 86: We may want to require that the CCTPtoken bridge is enabled otherwise we're going to potentially try to resolve .localMinter() from address(0)

contracts/libraries/CircleCCTPAdapter.sol
```diff
@@ -81,7 +81,20 @@ abstract contract CircleCCTPAdapter {
...
+        uint256 burnLimit = cctpMinter.burnLimitsPerMessage(address(usdcToken));
+        uint256 remainingAmount = amount;
+        do {
+            uint256 partAmount = remainingAmount > burnLimit ? burnLimit : remainingAmount;
+            cctpTokenMessenger.depositForBurn(
+                partAmount,
+                recipientCircleDomainId,
+                _addressToBytes32(to),
```
james-a-morris, line 94: We can probably put this at the top to prevent multiple calls

contracts/libraries/CircleCCTPAdapter.sol
nicholaspai: I don't think that's necessary as `depositForBurn` will revert if this is 0x0 [here](https://github.com/circlefin/evm-cctp-contracts/blob/817397db0a12963accc08ff86065491577bbc0e5/src/TokenMessenger.sol#L441)

contracts/libraries/CircleCCTPAdapter.sol
```diff
@@ -81,7 +81,16 @@ abstract contract CircleCCTPAdapter {
...
+        // If the amount to send exceeds the burn limit per message, then split the message into smaller parts.
+        ITokenMinter cctpMinter = cctpTokenMessenger.localMinter();
+        uint256 burnLimit = cctpMinter.burnLimitsPerMessage(address(usdcToken));
+        uint256 remainingAmount = amount;
+        bytes32 recipient = _addressToBytes32(to);
+        do {
+            uint256 partAmount = remainingAmount > burnLimit ? burnLimit : remainingAmount;
+            cctpTokenMessenger.depositForBurn(partAmount, recipientCircleDomainId, recipient, address(usdcToken));
+            remainingAmount -= partAmount;
+        } while (remainingAmount > 0);
```
mrice32, line 94: Why not just make this a while to make it easier to follow? cc @james-a-morris 

```solidity
        while (remainingAmount > 0) {
            uint256 partAmount = remainingAmount > burnLimit ? burnLimit : remainingAmount;
            cctpTokenMessenger.depositForBurn(partAmount, recipientCircleDomainId, recipient, address(usdcToken));
            remainingAmount -= partAmount;
        }

File: contracts/libraries/CircleCCTPAdapter.sol [hunk 4]
```diff
@@ -81,7 +81,16 @@ abstract contract CircleCCTPAdapter {
...
-        // Submit the amount to be transferred to bridged via the TokenMessenger
-        cctpTokenMessenger.depositForBurn(amount, recipientCircleDomainId, _addressToBytes32(to), address(usdcToken));
+        // Submit the amount to be transferred to bridged via the TokenMessenger.
+        // If the amount to send exceeds the burn limit per message, then split the message into smaller parts.
+        ITokenMinter cctpMinter = cctpTokenMessenger.localMinter();
+        uint256 burnLimit = cctpMinter.burnLimitsPerMessage(address(usdcToken));
+        uint256 remainingAmount = amount;
+        bytes32 recipient = _addressToBytes32(to);
+        do {
+            uint256 partAmount = remainingAmount > burnLimit ? burnLimit : remainingAmount;
```
mrice32, line 91: Just to triple check, you don't need to be _strictly less than_ the burn limit, right? You can be equal to it?

mrice32 (commented):
Looks great! Two small comments

contracts/libraries/CircleCCTPAdapter.sol
mrice32: I checked the contract. Looks like this is right.

contracts/libraries/CircleCCTPAdapter.sol
james-a-morris: I figured this because we always have to run once

mrice32 (approved):
LGTM!

contracts/libraries/CircleCCTPAdapter.sol
mrice32: Yep, I think we save one comparison, but I think the extra check is (IMO) worthwhile to avoid do-while given that it's less readable.

james-a-morris (approved):

PR #513, opened 2024-06-11
improve(MultiCaller): Handle custom errors better

nicholaspai:
## Motivation
The current implementation of the `MultiCallerUpgradable.multicall` function supports only bubbling up custom errors with a string as a parameter (for example, `CustomError("Error occurred")`) and revert messages (for example,` revert("Error occurred"`), or `require(..., "Error occurred")`). 

The current implementation also runs into an issue described below, which is NOT addressed in this PR

> The multicall function extracts the revert message by adding 0x04 to the result's offset and then performs abi.decode on the resulting data. In, short, the result, before the modification will look like this:
   length (32 bytes) | error selector (4 bytes) | error data (x bytes).
By adding 0x04 to result in assembly, we effectively shift this data to the left by 4 bytes. Now, the result will still be interpreted as a bytes variable, hence from the Solidity's point of view, it will look as follows:
   length (32 bytes) | error data (x bytes).
As may be observed, the error's selector becomes the part of the result's length from the compiler's point of view. So, if the initial length was 0xaa and the error selector was 0xbbccddee, then the result will be interpreted as it had the length equal to 0xaabbccddee, which is a very high value.
I have examined the bytecode generated by the Solidity compiler (one of the recent versions) for abi.decode and it actually contains a check verifying that the first argument's length does not exceed type(uint64.max) - if it does, the code will revert. Of course, in our case the length will be lower (as 0xaabbccddee is ~type(uint40.max)), so the code will work for now. But if a future compiler version requires the length to be lower than type(uint32).max, for example (it would still be reasonable as no valid data is that long), the code in the multicall function will revert. It doesn't necessarily have to happen, but just wanted to let you know where to look for a root cause if all of the sudden, after changing the compiler's version, the code reverts without a reaso

## Developer note

Ideally need to get this audited before we merge.

## Credit

All h/t to Bartosz from openzeppelin to this finding


nicholaspai:
No idea why these tests are failing because they don't even call `multicall`

bmzig (approved):

PR #515, opened 2024-06-12
feat(ERC7683Depositor): Initiator of gas-less deposit should specify exclusivity offset, not deadline

nicholaspai:
Similar to #514, this contract is designed to support a user flow where the user and filler communicate off-chain to produce an exclusivity agreement in which a filler gives user's a quote or a commitment that they will fill their deposit, in exchange the user should give the filler exclusive rights to fill for some time.

This PR introduces a way for this contract to set a cap on the exclusivity deadline that any filler can set when they `initiate` a deposit on-chain. In addition, the filler's specified `exclusivityDeadline` is replaced with an `exclusivityDeadlineOffset` which is better UX since they cannot control when the the deposit they `initiate` will be mined.

### DIFF
--- a/contracts/erc7683/ERC7683Depositor.sol
+++ b/contracts/erc7683/ERC7683Depositor.sol
@@ -27,14 +27,24 @@ abstract contract ERC7683OrderDepositor is ISettlementContract {
     // This is a somewhat arbitrary conversion, but order creators need some way to precompute the quote timestamp.
     uint256 public immutable QUOTE_BEFORE_DEADLINE;
 
+    // The maximum offset that can be added to the timestamp that a deposit is mined to get the exclusivityDeadline.
+    // Set this lower to protect depositors from fillers setting this too high.
+    uint256 public immutable MAX_EXCLUSIVITY_DEADLINE_OFFSET;
+
     /**
      * @notice Construct the Permit2Depositor.
      * @param _permit2 Permit2 contract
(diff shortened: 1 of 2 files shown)


contracts/erc7683/ERC7683Depositor.sol
```diff
@@ -79,7 +79,7 @@ abstract contract ERC7683OrderDepositor is ISettlementContract {
             // Note: simplifying assumption to avoid quote timestamps that cause orders to expire before the deadline.
             SafeCast.toUint32(order.initiateDeadline - QUOTE_BEFORE_DEADLINE),
             order.fillDeadline,
-            acrossOrderData.exclusivityDeadline,
+            SafeCast.toUint32(getCurrentTime()) + acrossOrderData.exclusivityDeadlineOffset,
```
pxrl, line 82: Is it worth applying the cast directly in `getCurrentTime()`, so it returns uint32 instead? It seems like we'd normally want to trim it down to a uint32 anyway.

pxrl, line 82: Do we have any tests for this btw?

nicholaspai: No we don't at the moment

bmzig (approved):

pxrl (approved):

PR #520, opened 2024-06-13
improve(OVM_SpokePool): Use bridgeERC20To to bridge native L2 tokens if admin decides to support them

nicholaspai:
bridgeERC20To can be used to bridge native L2 tokens (without L1 counterparts and specifically does not contain a `l1Token()` function in the L2 erc20 contract) if an L1 remote token is identified.

This call require the spoke pool to set an allowance on the standard bridge before calling `bridgeERC20To`


contracts/Ovm_SpokePool.sol
```diff
@@ -154,22 +166,43 @@ contract Ovm_SpokePool is SpokePool, CircleCCTPAdapter {
         else if (_isCCTPEnabled() && l2TokenAddress == address(usdcToken)) {
             _transferUsdc(hubPool, amountToReturn);
         }
-        // Note we'll always use withdrawTo instead of bridgeERC20To here because we can assume
+        // Note we'll default use withdrawTo instead of bridgeERC20To here because we can assume
```
pxrl, line 169:
```suggestion
        // Note we'll default to withdrawTo instead of bridgeERC20To here because we can assume
```

pxrl (approved):

bmzig (approved):

PR #523, opened 2024-06-17
improve(forge-ci): Ignore astId and other non-storage slot related props when diffing storage layouts

nicholaspai:
Remove props like astID and anything in types that are not minimally neccessary to compare storage layouts.

Namely, we care about which slots state variables are using, which are ultimately downstream of `types` and independent of the AST layout


bmzig (approved):

james-a-morris (approved):
Good call at preventing the unnecessary diffs

PR #538, opened 2024-06-24
[WIP] feat: Define in-place alternative to depositId that preserves deposit hash uniqueness while allowing deterministic fill hash computations for honest depositors

nicholaspai:
## Summary

This PR replaces the contract-wide counter of `depositId` with a per-depositor deposit counter.

## Motivation

The motivation for this PR is to allow honest depositors to broadcast ahead of their deposit transaction to fillers that they want to work with and pre-compute their next deposit hash. This would allow depositors to share with fillers off-chain a Permit2 signed ERC7683 intent + Across deposit and then let the filler submit the deposit txn + immediately fill the user because they would know the precomputed relay hash.

The major benefit of this PR is that re-org risk is completely eliminated for any deposits submitted by "honest" depositors who choose not to try to grief fillers by front-running their fill transactions. This PR accompanied with some sort of off-chain infrastructure where depositors can credibly commit to submitting a deposit (e.g. via an account locking system or by staking collateral) would be the ideal end use case.

A nice property of this PR is that certain properties of Across' order flow system are kept intact:
- deposit ID's are always increasing *for a specific depositor address*
- locating a historical deposit using deposit ID's using a binary search is still efficient assuming you know which depositor you are looking up a deposit for
- All deposits produce unique relay hashes, since the combination of {depositor, depositId} is guaranteed unique

## Gas cost implications

Every first time deposit for a depositor incurs a 20,000 SSTORE cost, and every subsequent deposit for that depositor incurs only a 5,000 SSTORE cost

We assume that most depositors re-use `depositor` addresses so the amortized gas cost is only trivially increased by this PR due to one additional ADD operation

### DIFF
--- a/contracts/SpokePool.sol
+++ b/contracts/SpokePool.sol
@@ -63,8 +63,10 @@ abstract contract SpokePool is
     WETH9Interface private DEPRECATED_wrappedNativeToken;
     uint32 private DEPRECATED_depositQuoteTimeBuffer;
 
-    // Count of deposits is used to construct a unique deposit identifier for this spoke pool.
-    uint32 public numberOfDeposits;
+    // This used to be `numberOfDeposits`: The deposit count for this contract and used to construct a
+    // unique deposit identifier for this spoke pool. This was replaced by `depositorDepositCount` which starts
+    // all depositor-specific deposit counters at this value to guarantee unique deposit hashes across all deposits.
+    uint32 private DEPOSIT_COUNTER_START;
 
     // Whether deposits and fills are disabled.
     bool public pausedFills;
@@ -150,6 +152,11 @@ abstract contract SpokePool is
     // token for the input token. By using this magic value, off-chain validators do not have to keep
     // this event in their lookback window when querying for expired deposts.
     uint32 public constant INFINITE_FILL_DEADLINE = type(uint32).max;
+
+    // Storing of deposit count per depositor address. Used to compute the `depositId` for a new deposit and replaces
+    // the `numberOfDeposits` usage in order to allow depositors to precompute their relay data hashes.
+    mapping(address => uint32) public depositorDepositCount;
+
     /****************************************
      *                EVENTS                *
      ****************************************/
@@ -283,7 +290,7 @@ abstract contract SpokePool is
         address _crossDomainAdmin,
         address _hubPool
     ) public onlyInitializing {
-        numberOfDeposits = _initialDepositId;
+        DEPOSIT_COUNTER_START = _initialDepositId;
         __EIP712_init("ACROSS-V2", "1.0.0");
         __UUPSUpgradeable_init();
         __ReentrancyGuard_init();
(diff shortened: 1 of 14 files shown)


centauriAx:
this should significantly reduce reorg risk in general, even for typical relays that are not using a pre-broadcast approach. Most invalid fills occur via the chain-scoped deposit IDs being shuffled around as a result of a reorg while all other deposit details remain the same.

contracts/SpokePool.sol
```diff
@@ -1112,8 +1132,10 @@ abstract contract SpokePool is
         // slither-disable-next-line timestamp
         if (getCurrentTime() - quoteTimestamp > depositQuoteTimeBuffer) revert InvalidQuoteTimestamp();
 
-        // Increment count of deposits so that deposit ID for this spoke pool is unique.
-        uint32 newDepositId = numberOfDeposits++;
+        // Add the last known deposit count to the depositor's ID to ensure that all deposit IDs are unique including
+        // all past deposits which used numberOfDeposits as a counter.
+        uint32 newDepositId = depositCount(depositor);
+        depositorDepositCount[depositor]++;
```
ankitchiplunkar, line 1138: why does the deposit ID need to be an incremental counter? and not be a hash like a transaction hash (eth_sendRawTransaction) or a message hash (userOp)

nicholaspai: It's useful for looking up old deposits if we can assume they always increment. We can implement a binary search on them for example (I'll link to the offchain code we run in production)

But after talking with @mrice32 im not so convinced this property is important to preserve

nicholaspai: But at a minimum I don't want to change the type of depositId emitted in the V3FundsDeposited event because otherwise the upgrade path would be very tricky

contracts/SpokePool.sol
```diff
@@ -1112,8 +1132,10 @@ abstract contract SpokePool is
         // slither-disable-next-line timestamp
         if (getCurrentTime() - quoteTimestamp > depositQuoteTimeBuffer) revert InvalidQuoteTimestamp();
 
-        // Increment count of deposits so that deposit ID for this spoke pool is unique.
-        uint32 newDepositId = numberOfDeposits++;
+        // Add the last known deposit count to the depositor's ID to ensure that all deposit IDs are unique including
+        // all past deposits which used numberOfDeposits as a counter.
+        uint32 newDepositId = depositCount(depositor);
```
nicholaspai, line 1137: something to think about: we don't want other people to be able to spoof the depositor, so maybe do a hash of msg.sender and depositor, to allow this to still be useful for aggregators/smart contracts that deposit on behalf of users while preventing front-running/spoofing to grief fillers

contracts/SpokePool.sol
```diff
@@ -607,14 +614,18 @@ abstract contract SpokePool is
             IERC20Upgradeable(inputToken).safeTransferFrom(msg.sender, address(this), inputAmount);
         }
 
+        // Add the last known deposit count to the depositor's ID to ensure that all deposit IDs are unique including
+        // all past deposits which used numberOfDeposits as a counter.
+        uint32 accountDepositCount = depositCount(depositor);
```
nicholaspai, line 619: An alternative cheaper on gas implementation is to add a deposit entry point where the caller can set `depositId`. This way the gasless deposit receiver like the filler can pre-fill in the depositId using the Permit2 nonce, for example. This function would come with the caveat that the caller shoudl be sure that their deposit ID has not been used yet otherwise the deposit will be unfillable and their deposit will expired

nicholaspai:
I think this is a better alternative implementation: https://github.com/across-protocol/contracts/pull/583

PR #541, opened 2024-06-27
feat(Lisk): add deployment info for Lisk

bmzig:
This commit includes all the information about our mainnet deployments for Lisk.


deployments/deployments.json
```diff
@@ -90,6 +91,11 @@
   },
   "420": { "SpokePool": { "address": "0xeF684C38F94F48775959ECf2012D7E864ffb9dd4", "blockNumber": 17025501 } },
   "919": { "SpokePool": { "address": "0xbd886FC0725Cc459b55BbFEb3E4278610331f83b", "blockNumber": 13999465 } },
+  "1135": {
+    "SpokePool": { "address": "0xeF684C38F94F48775959ECf2012D7E864ffb9dd4", "blockNumber": 2391355 },
+    "SpokePoolVerifier": { "address": "0xB4A8d45647445EA9FC3E1058096142390683dBC2", "blockNumber": 2391565 },
+    "MulticallHandler": { "address": "0xF2720cc7391D237876A17E10914e9b7188009957", "blockNumber": 2391722 }
```
james-a-morris, line 97: OOC are the definitions of the multicaller related to this PR's scope?

james-a-morris (commented):
1 question

deployments/deployments.json
bmzig: I believe so. Going forward we will probably have a multicall handler for each supported chain

pxrl: fwiw we may want to subsequently circle back and redeploy the multicall handler contract, since its implementation was changed slightly and it now deploys to a new address. Given that we deploy with a salt I assume we want the same address on as many chains as possible.

deploy/042_deploy_lisk_adapter.ts
```diff
@@ -20,10 +20,6 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
       L1_ADDRESS_MAP[chainId].liskCrossDomainMessenger,
       L1_ADDRESS_MAP[chainId].liskStandardBridge,
       L1_ADDRESS_MAP[chainId].usdc,
-      // L1_ADDRESS_MAP[chainId].cctpTokenMessenger,
-      // For now, we are not using the CCTP bridge and can disable by setting
-      // the cctpTokenMessenger to the zero address.
-      ZERO_ADDRESS,
```
pxrl, line 26: Might be worth proactively applying this change to the Blast and Mode deployment scripts as well :+1:

pxrl (approved):

deploy/042_deploy_lisk_adapter.ts
bmzig: Good call. Done here: https://github.com/across-protocol/contracts/pull/541/commits/3b8d0d7fe4e7a5a19547eea6f64070ac0b15c612

PR #543, opened 2024-06-27
refactor: Relocate message handler interface

pxrl:
The MulticallHandler contract currently imports the SpokePool contract and experienced a bytecode change after the following unrelated commit:

86593c6

This resulted in its CREATE2 deployment address changing. The MulticallHandler doesn't actually need any of the SpokePool implementation or its dependencies, and it seems like we'd like the MulticallHandler to be deployed to the same address on as many chains as possible, so factor out the interface to a separate file to insulate it from any unintended modifications.


bmzig (approved):
Could we also just do `import { AcrossMessageHandler } from "../SpokePool.sol";` in `MulticallHandler.sol`? It also makes sense to add it to another file, so I'm fine with either.

pxrl:
> Could we also just do `import { AcrossMessageHandler } from "../SpokePool.sol";` in `MulticallHandler.sol`? It also makes sense to add it to another file, so I'm fine with either.

Good call - I hadn't thought of that. I tried it though and I can still get the resulting bytecode to change by making simple changes in `contracts/upgradeable/MultiCallerUpgradeable.sol`.

nicholaspai (approved):
Does this keep the MultiCallHandler create2 address the same?

james-a-morris (approved):
Good call +1

PR #549, opened 2024-07-03
feat(blast): add blast deployments

bmzig:
This commit adds both testnet and mainnet deployments for Blast.


utils/utils.hre.ts
```diff
@@ -37,13 +37,13 @@ export async function deployNewProxy(
...
       kind: "uups",
-      unsafeAllow: ["delegatecall"],
+      unsafeAllow: ["delegatecall", "state-variable-immutable"],
     })) as string;
     console.log(`New ${name} implementation deployed @ ${instance}`);
   } else {
     const proxy = await upgrades.deployProxy(await getContractFactory(name, {}), initArgs, {
       kind: "uups",
-      unsafeAllow: ["delegatecall"], // Remove after upgrading openzeppelin-contracts-upgradeable post v4.9.3.
+      unsafeAllow: ["delegatecall", "state-variable-immutable"], // Remove after upgrading openzeppelin-contracts-upgradeable post v4.9.3.
```
bmzig, line 46: I believe this is OK since we will never actually upgrade our proxies: https://docs.openzeppelin.com/upgrades-plugins/1.x/faq#why-cant-i-use-immutable-variables

james-a-morris (commented):
One comment

dohaki (approved):
LGTM

utils/utils.hre.ts
pxrl: If we keep this in the deployment utility then I think we should append it conditional on the chain being Blast?

I'm also OK with applying the exception to the Blast SpokePool contract, but would prefer to avoid unconditionally suppressing this on other deployments where we might not want this exception.

pxrl (approved):
OK subject to a conclusion on the deployment script updates :+1:

utils/utils.hre.ts
bmzig: I will follow this up with another PR.

PR #553, opened 2024-07-10
feat: redeploy multicall handler contracts to non-zk chains

bmzig:
This commit redeploys all multicall handler contracts to chains where we can reliably use CREATE2 (excluding testnets).


james-a-morris (commented):
fair call +1

pxrl (approved):
I checked each deployment and the deployment block checks out 100% :+1:

PR #554, opened 2024-07-13
fix: fix usdb l2 bridge signature

mrice32:
USDB L2 -> L1 transfers are failing due to an incorrect function signature where the gas type was assumed to be uint256, as it is on other OP Stack chains, rather than uint32, which is what it is on Blast.

### DIFF
--- a/contracts/Blast_SpokePool.sol
+++ b/contracts/Blast_SpokePool.sol
@@ -32,6 +32,17 @@ interface IBlast {
     function claimMaxGas(address contractAddress, address recipientOfGas) external returns (uint256);
 }
 
+interface IUSDBL2Bridge {
+    function bridgeERC20To(
+        address _localToken,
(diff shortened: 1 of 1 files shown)


pxrl (approved):

contracts/Blast_SpokePool.sol
```diff
@@ -32,6 +32,17 @@ interface IBlast {
     function claimMaxGas(address contractAddress, address recipientOfGas) external returns (uint256);
 }
 
+interface IUSDBL2Bridge {
```
pxrl, line 35: For review, the Blast docs are here: https://docs.blast.io/building/contracts.
The relevant contract is the L2BlastBridge, here: https://blastscan.io/address/0x4300000000000000000000000000000000000005#writeProxyContract#F2

bmzig (approved):

PR #562, opened 2024-07-17
fix: fix bridgeERC20To interface for all OP Stack chains 

mrice32:
The interface for bridgeERC20To has been incorrect presumably since Across originally integrated Optimism. Since this function is never used outside of Blast, it has only recently created an issue.

This PR corrects the interface, but may not be deployed until the next major Across upgrade.


nicholaspai:
Is this redundant with https://github.com/across-protocol/contracts/pull/557 ?

mrice32:
> Is this redundant with #557 ?

Yep!

PR #563, opened 2024-07-24
feat(Blast): Add Blast_DaiRetriever and Blast_RescueAdapter

nicholaspai:
We recently found out that USDB L2 to L1 withdrawals are a two step process:
- L2 withdrawal is requested, passes challenge period, and finalized like a normal OPStack withdrawal on the OptimismPortal
- [This function finalizeBridgeErc20](https://github.com/blast-io/blast/blob/706acdde2b3a6614a9a7dbfceb82660789d5e7b8/blast-optimism/packages/contracts-bedrock/src/mainnet-bridge/L1BlastBridge.sol#L187-L210) is called on the L1BlastBridge which calls to the [USDYieldManager](https://etherscan.io/address/0xa230285d5683C74935aD14c446e137c8c8828438) to request the withdrawal of DAI
- This final request needs to be finalized via an admin call which happens every ~12 hours. Once finalized, the original withdrawal's `recipient` can call `claimWithdrawal` on the `USDYieldManager`.

Therefore, this PR adds a Blast_DaiRetriever which should be set as the `recipient` on all Blast withdrawals from the SpokePool going forwards. An EOA can then call `retrieve` with the L1 finalized withdrawal's `requestId` to retrieve DAI following the 12 hour window.

Separately, the `Blast_RescueAdapter` can be used to retrieve any DAI currently available in the USDYieldManager that must be called by the `HubPool`


bmzig (approved):
Looks good.

contracts/Blast_DaiRetriever.sol
```diff
@@ -0,0 +1,70 @@
...
+    /**
+     * @notice Calls USDCYieldManager contract to release DAI and send to the hubPool. Required to use this function
+     * to retrieve DAI since only the L2 withdrawal recipient can make this call.
+     * @notice This can revert if the claim is not ready yet. It takes ~12 hours for a Blast admin to make the DAI
+     * available for retrieval following withdrawal finalization.
+     * @param _requestId L2 withdrawal request ID. Emitted in L1 WithdrawalRequested event when the L2 to L1
+     * withdrawal is first "finalized" but still awaiting the recipient to claim the DAI.
+     */
+    function retrieve(uint256 _requestId) public nonReentrant {
+        uint256 hintId = usdYieldManager.findCheckpointHint(_requestId, 1, usdYieldManager.getLastCheckpointId());
```
mrice32, line 66: Can we compute hintId offchain to save gas since this is on mainnet?

contracts/Blast_DaiRetriever.sol
```diff
@@ -0,0 +1,70 @@
...
+        require(usdYieldManager.claimWithdrawal(_requestId, hintId), "claim failed");
```
mrice32, line 67: In the 2-step process, can both steps be done in the same txn? If so, this setup allow both steps be done in the same txn via makerdao multicall, right?

mrice32 (approved):
Looks great! Just a few minor questions!

contracts/Blast_DaiRetriever.sol
nicholaspai: sure yeah

nicholaspai: 01bd90a8563344831970bc256d74dd67a25dbe75

contracts/Blast_DaiRetriever.sol
nicholaspai: they cannot, because the blast admin is required to submit a transaction to finalize a batch of withdrawal requests. The "finalization" as we know it creates a WithdrawalRequested event which gets finalized manually by the admin

mrice32 (approved):
LGTM!

contracts/Blast_DaiRetriever.sol
```diff
@@ -0,0 +1,63 @@
...
+}
+
+/**
+ * @notice Contract deployed on Ethereum to facilitate DAI transfers from Blast to the HubPool.
+ * @dev Blast USDB withdrawals are a two step process where the L2 to L1 withdrawal must first be finalized via
+ * the typical OP Stack mechanism, and then a claim from the withdrawal's *recipient* must be made against a
+ * USDBYieldManager contract. This means that the Blast_SpokePool must set its recipient to this contract's address
+ * and then an EOA can call this contract to retrieve the DAI.
+ */
+contract Blast_DaiRetriever is Lockable, MultiCaller {
```
mrice32, line 21: Perfect!

mrice32 (approved):
LGTM!

pxrl (approved):

PR #564, opened 2024-07-24
feat(Blast): send tokens to blast retriever

bmzig:
This commit upgrades the `Blast_SpokePool` so that it is compatible with the `Blast_DaiRetriever` contract on mainnet. It just adds a new immutable variable `BLAST_RETRIEVER` which is the new recipient of cross-chain spoke pool rebalances.

### DIFF
--- a/contracts/Blast_SpokePool.sol
+++ b/contracts/Blast_SpokePool.sol
@@ -57,7 +57,7 @@ contract Blast_SpokePool is Ovm_SpokePool {
     // Address that this contract's yield and gas fees accrue to.
     address public immutable YIELD_RECIPIENT;
 
-    // Address which receives USDB on mainnet.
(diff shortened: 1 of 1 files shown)


contracts/Blast_SpokePool.sol
```diff
@@ -56,6 +56,9 @@ contract Blast_SpokePool is Ovm_SpokePool {
 
     // Address that this contract's yield and gas fees accrue to.
     address public immutable YIELD_RECIPIENT;
+
+    // Address which receives DAI from USDB withdrawals on mainnet. Used for spoke pool rebalancing.
```
nicholaspai, line 60:
```suggestion
    // Address which receives DAI from USDB withdrawals on mainnet. USDB withdrawals are a two step finalization process where the recipient of the withdrawals must claim, so we set this address to a contract that claims on behalf of the HubPool
```

james-a-morris (approved):

PR #568, opened 2024-07-26
feat(SpokePoolVerifier): Support calling depositV3 instead of deposit and deploy new contracts

nicholaspai:
We are moving to deprecate all legacy deposit() calls


dohaki (approved):

contracts/SpokePoolVerifier.sol
```diff
@@ -23,41 +23,51 @@ contract SpokePoolVerifier {
...
+        address inputToken,
+        uint256 inputAmount,
+        uint256 outputAmount,
         uint256 destinationChainId,
-        int64 relayerFeePct,
         uint32 quoteTimestamp,
         bytes memory message,
-        uint256 maxCount
+        address exclusiveRelayer,
+        uint32 exclusivityDeadline,
```
bmzig, line 50: Should we reorder the arguments so the calldata is smaller? Might also be nice to use the same structure as `depositV3` just for organization purposes.

bmzig (approved):

pxrl (approved):
lgtm!

mrice32 (approved):
LGTM! This should be tagged as pre-audit.

nicholaspai:
> LGTM! This should be tagged as pre-audit.

How do you suggest we do this?

PR #574, opened 2024-08-06
feat: hardhat task to test message relays and withdrawals

bmzig:
The intended strategy is to deploy two copies of the spoke pool for each chain, but for one spoke pool, set the `crossDomainAdmin` as the L1 chain adapter and the `hubPool` as the `MOCK_ADMIN` address. Ideally, this would allow us to submit an empty relayer refund root with a nonzero `amountToReturn`, since the `onlyAdmin` modifier will succeed in this setup, and calling `_bridgeTokensToHubPool` will send these funds back to our EOA. For completeness, the withdrawal should be triggered by the finalizer.


bmzig:
Example output: 
```
$ yarn hardhat run scripts/sendTokensToHubPool.ts --network base-sepolia
yarn run v1.22.22
secp256k1 unavailable, reverting to browser version
secp256k1 unavailable, reverting to browser version
[+] Successfully constructed all contracts. Determining root bundle Id to use.
[+] Obtained latest root bundle Id 8
[+] Called L1 adapter (0x0B44A7CBC9489a6a881a2397b469D76320aaB90C) to relay refund leaf message to mock spoke pool at 0x55a8EC3Fe18f8074D090Eea8aBd56501A4Ba098E. Txn: 0x32f836936f67b42b11a83511322b8ab260f0468fae3ee26234412bac9b554d1e
[i] Optimistically waiting 5 minutes for L1 message to propagate. If root bundle is not observed, will check spoke every minute thereafter.
[-] Root bundle not found on L2. Waiting another 60 seconds.
[+] Root bundle observed on L2 spoke pool. Attempting to execute.
[+] Executed root bundle with transaction hash 0xa3ebd61d9ded2d9b09033fff143b41b92a813a092fef5a037e6895ec142c2362. You can now test the finalizer in the relayer repository.
Done in 381.96s.
```

bmzig:
Not sure how automated we want this. Some bridges will require the EOA to send a `msg.value` to fund the L2 transaction, and this is tricky since in order to get the exact amount, we need to query that bridge directly, which isn't generalizable. Should we just add an environment variable where the user needs to specify a `msg.value`? We would be the only ones using this script anyways...

tasks/evmRelayMessageWithdrawal.ts
```diff
@@ -0,0 +1,137 @@
...
+    const spokePool = new Contract(taskArguments.spokePool, minimalSpokePoolInterface, l2Signer);
+    const adapter = new Contract(taskArguments.adapter, minimalAdapterInterface, l1Signer);
+
+    console.log("[+] Successfully constructed all contracts. Determining root bundle Id to use.");
+    let rootBundleId = 0;
+    try {
+      while (1) {
+        await spokePool.rootBundles(rootBundleId);
+        rootBundleId++;
+      }
```
pxrl, line 64: This might get rate-limited on a free-tier RPC provider. It's probably OK to assume that we'd test this against a very new SpokePool so hopefully it should resolve very quickly and without many calls.

pxrl (approved):
LGTM! Keen to see this tested on the next deployment.

PR #575, opened 2024-08-08
fix(Scroll): Approve on select ERC20 withdrawals

pxrl:
ERC20s with custom L2 gateways on Scroll require token approvals in order to initiate a withdrawal.


contracts/Scroll_SpokePool.sol
```diff
@@ -4,6 +4,13 @@ pragma solidity ^0.8.0;
 import "./SpokePool.sol";
 import "@scroll-tech/contracts/L2/gateways/IL2GatewayRouter.sol";
 import "@scroll-tech/contracts/libraries/IScrollMessenger.sol";
+import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+
+interface IL2GatewayRouterExtended is IL2GatewayRouter {
+    function defaultERC20Gateway() external view returns (address);
+
```
pxrl, line 11:
```suggestion
```

contracts/Scroll_SpokePool.sol
```diff
@@ -12,11 +19,13 @@ import "@scroll-tech/contracts/libraries/IScrollMessenger.sol";
  * @custom:security-contact bugs@across.to
  */
 contract Scroll_SpokePool is SpokePool {
+    using SafeERC20 for IERC20;
```
pxrl, line 22: Should instead use SafeERC20Upgradeable.

nicholaspai: This is inherited already, no?

contracts/Scroll_SpokePool.sol
```diff
@@ -4,6 +4,12 @@ pragma solidity ^0.8.0;
 import "./SpokePool.sol";
 import "@scroll-tech/contracts/L2/gateways/IL2GatewayRouter.sol";
 import "@scroll-tech/contracts/libraries/IScrollMessenger.sol";
+import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+
+interface IL2GatewayRouterExtended is IL2GatewayRouter {
```
nicholaspai, line 9: Nit: prefer naming this like IL2ERC20GatewayRouter

nicholaspai (approved):
Code looks good but left some comments about names

contracts/Scroll_SpokePool.sol
pxrl: I'm hesitant to use `IL2ERC20GatewayRouter` since `IL2ERC20Gateway` is already defined by the Scroll ABI and is inherited by `IL2GatewayRouter`, so the concern was that the naming ends up extremely similar and it becomes easy to mix them up. Is there another name we could go for?

contracts/Scroll_SpokePool.sol
```diff
@@ -5,18 +5,25 @@ import "./SpokePool.sol";
 import "@scroll-tech/contracts/L2/gateways/IL2GatewayRouter.sol";
 import "@scroll-tech/contracts/libraries/IScrollMessenger.sol";
 
+interface IL2GatewayRouterExtended is IL2GatewayRouter {
+    function defaultERC20Gateway() external view returns (address);
+    function getERC20Gateway(address) external view returns (address);
```
pxrl, line 10: prettier is opinionated about this

```suggestion

    function getERC20Gateway(address) external view returns (address);
```

contracts/Scroll_SpokePool.sol
pxrl: It's doesn't appear to be automatically inherited. We do this in some other contracts, though we're not consistent about the use of SafeERC20Upgradeable vs. SafeERC20. I've followed the Ethereum SpokePool: https://github.com/across-protocol/contracts/blob/1471664a516f0d81632f7228eeb90afcf350236a/contracts/Ethereum_SpokePool.sol#L12

We use SafeERC20 for the Linea and OVM SpokePools. That however requires additional imports, so I followed the Ethereum SpokePool.

https://github.com/across-protocol/contracts/blob/1471664a516f0d81632f7228eeb90afcf350236a/contracts/Linea_SpokePool.sol#L17
https://github.com/across-protocol/contracts/blob/1471664a516f0d81632f7228eeb90afcf350236a/contracts/Ovm_SpokePool.sol#L36

mrice32 (approved):

contracts/Scroll_SpokePool.sol
nicholaspai: I think its fine

deploy/027_deploy_scroll_spokepool.ts
```diff
@@ -24,7 +27,7 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
   //    * A fillDeadlineBuffer of 6 hours
   const constructorArgs = [L2_ADDRESS_MAP[chainId].l2Weth, 3600, 21600];
 
-  await deployNewProxy("Scroll_SpokePool", constructorArgs, initArgs);
+  await deployNewProxy("Scroll_SpokePool", constructorArgs, initArgs, spokeChainId === SCROLL);
```
pxrl, line 30: Reverted per https://github.com/across-protocol/contracts/pull/577

```suggestion
  await deployNewProxy("Scroll_SpokePool", constructorArgs, initArgs);
```

PR #577, opened 2024-08-09
improve(deploy): Improve SpokePool deployment UX

pxrl:
The majority of SpokePool deployment scripts have been updated to support deploying only implementation contracts. This is done with the knowledge that there are already pre-existing deployments on these chains, but it introduces a risk that using one of these scripts as the template for a new chain will inadvertently result in skipping deployment of the proxy contact. Instead, whether to deploy a new proxy or not can be inferred by whether the contract exists in deployments.json.

### DIFF
--- a/deploy/016_deploy_zksync_spokepool.ts
+++ b/deploy/016_deploy_zksync_spokepool.ts
@@ -1,9 +1,10 @@
 import * as zk from "zksync-web3";
 import { Deployer as zkDeployer } from "@matterlabs/hardhat-zksync-deploy";
 import { DeployFunction, DeploymentSubmission } from "hardhat-deploy/types";
-import { L2_ADDRESS_MAP } from "./consts";
 import { HardhatRuntimeEnvironment } from "hardhat/types";
+import { getDeployedAddress } from "../src/DeploymentUtils";
 import { getSpokePoolDeploymentInfo } from "../utils/utils.hre";
+import { L2_ADDRESS_MAP } from "./consts";
 
 const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
   const contractName = "ZkSync_SpokePool";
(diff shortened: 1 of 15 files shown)


deploy/016_deploy_zksync_spokepool.ts
```diff
@@ -29,10 +29,12 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
...
   const constructorArgs = [L2_ADDRESS_MAP[spokeChainId].l2Weth, 3600, 21600];
 
-  let newAddress;
+  let newAddress: string;
   // On production, we'll rarely want to deploy a new proxy contract so we'll default to deploying a new implementation
   // contract.
-  if (spokeChainId === 324) {
+  // If SKIP_PROXY is defined, only deploy an implementation contract.
+  const implementationOnly = process.env.SKIP_PROXY !== undefined;
+  if (implementationOnly) {
```
pxrl, line 37: This is from an earlier iteration; needs to be updated.

pxrl: Updated: https://github.com/across-protocol/contracts/pull/577/commits/d611f4e644c729c2ee4addf647af1f7e5ff92412

nicholaspai (approved):

bmzig (approved):

deploy/016_deploy_zksync_spokepool.ts
```diff
@@ -29,10 +30,14 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
   //    * A fillDeadlineBuffer of 6 hours
   const constructorArgs = [L2_ADDRESS_MAP[spokeChainId].l2Weth, 3600, 21600];
 
-  let newAddress;
+  let newAddress: string;
   // On production, we'll rarely want to deploy a new proxy contract so we'll default to deploying a new implementation
   // contract.
-  if (spokeChainId === 324) {
+  // If SKIP_PROXY is defined, only deploy an implementation contract.
```
pxrl, line 36:
```suggestion
// If a SpokePool can be found in deployments/deployments.json, then only deploy an implementation contract.
```

deploy/016_deploy_zksync_spokepool.ts
```diff
@@ -29,10 +30,14 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
   //    * A fillDeadlineBuffer of 6 hours
   const constructorArgs = [L2_ADDRESS_MAP[spokeChainId].l2Weth, 3600, 21600];
 
-  let newAddress;
+  let newAddress: string;
   // On production, we'll rarely want to deploy a new proxy contract so we'll default to deploying a new implementation
   // contract.
-  if (spokeChainId === 324) {
+// If a SpokePool can be found in deployments/deployments.json, then only deploy an implementation contract.
```
pxrl, line 36:
```suggestion
  // If a SpokePool can be found in deployments/deployments.json, then only deploy an implementation contract.
```

PR #579, opened 2024-08-09
chore(deployments): Remove unsupported deployments

pxrl:
These deployments are all long unsupported, so flush them out to de-clutter.


bmzig (approved):
Very nice

deployments/deployments.json
```diff
@@ -90,13 +57,11 @@
     "UniswapV3_SwapAndBridge": { "address": "0x9220fa27ae680e4e8d9733932128fa73362e0393", "blockNumber": 56675837 },
     "MulticallHandler": { "address": "0x924a9f036260DdD5808007E1AA95f08eD08aA569", "blockNumber": 59194129 }
   },
-  "280": { "SpokePool": { "address": "0x863859ef502F0Ee9676626ED5B418037252eFeb2", "blockNumber": 10263886 } },
```
bmzig, line 93: I think we should verify that this change won't break the relayer since I think that it still needs to construct a boba spoke pool client?

bmzig: Seems to work fine on the relayer side

pxrl: Resolved out of band, but just for reference, 280 is zkSync Görli and 288 is Boba :+1:

nicholaspai (approved):

PR #583, opened 2024-08-20
feat(SpokePool): Introduce opt-in deterministic relay data hashes

nicholaspai:
## Motivation

We should allow a depositor to pre-compute their relay data hash so that they can work with a filler to more quickly fill their deposit while avoiding re-org risk (which could change the ordering of V3FundsDeposited events and change a depositor's expected `depositId`) and race conditions with another innocent or malicious depositor who front-runs their depositId.

The advantage of using the global `depositId` as a nonce today is gas cost savings since we don't need to store a new value in a new slot on each deposit. We should not add gas cost to deposits.

Secondly, incrementing the global `depositId` after each deposit guarantees that each `depositV3()` produces a unique deposit. This means that a depositor does not have to worry about accidentally producing a relay hash collision that means their deposit would be unfillable. Fortunately, unfillable deposits are refunded on the origin chain to the depositor but this possibility is undesirable for many depositors.

Therefore, any implementation of deterministic relay hashes should ideally be opt-in.

## Implementation

Introduces a new function: `unsafeDepositV3()` which gives the msg.sender 12 bytes with which to set a deposit nonce. It is marked unsafe because there is no protection given by the contract that the resultant relay data hash produced will be unique. The 12 byte deposit nonce is combined with the msg.sender's address, allowing each msg.sender to only track their own deposit nonces to avoid collisions. This also protects depositors from being griefed by malicious actors who might try to front-run their unsafe relay data hashes.

`unsafeDepositV3()` is slightly more expensive than `depositV3` from a gas-cost perspective because it adds a keccak256 + abi.encodePacked operation but it removes a SSTORE

The existing `depositV3()` function's implementation is untouched.

## Changes to V3RelayData struct

Changes the `depositId` parameter in the `V3RelayData` struct from uint32 to uint256. In storage, this parameter is stored as a uint256 so at the low-level, this changes nothing. Moreover, in practice until now, depositId's were always set as uint32 meaning that the first 24 bytes of the uint256 `depositId` slot are currently set to 0.

This change is made because this PR introduces a new function: `unsafeDepositV3()` which allows someone to set a `depositId` with non-zero bytes in the first 24 bytes.

## Implications for clients

Any logic that assumes that depositId's will always be monotonically increasing now needs to be changed.  For example, the [`queryHistoricalDepositForFill`](https://github.com/across-protocol/sdk/blob/b8dd3523dd5eac74c4022d943b2c8c16de64ec7d/src/utils/DepositUtils.ts#L38) function performs a binary search on V3FundsDeposited events' depositId values to locate historical fills. This clearly will no longer work following this PR, so this corresponding change in the `relayer` client skips this function in all cases unless the deposit was sent via a legacy function where all deposit ID's are produced "safely": https://github.com/across-protocol/relayer/pull/1769


nicholaspai:
Some off chain work related to this pr:
- https://github.com/across-protocol/relayer/pull/1769


It will be necessary to change all offchain code that tries to query deposits using deposit Id since we can no longer assume that deposit id is monotonically increasing for each deposit

contracts/SpokePool.sol
```diff
@@ -543,75 +543,102 @@ abstract contract SpokePool is
...
+     * which would make this deposit unfillable. In this case, the depositor would subsequently receive a refund
+     * of `inputAmount` of `inputToken` on the origin chain after the fill deadline.
+     * @dev This is slightly more gas efficient than the depositV3() function because it doesn't
+     * increment the global deposit count.
+     * @dev On the destination chain, the hash of the deposit data will be used to uniquely identify this deposit, so
+     * modifying any params in it will result in a different hash and a different deposit. The hash will comprise
+     * all parameters to this function along with this chain's chainId(). Relayers are only refunded for filling
+     * deposits with deposit hashes that map exactly to the one emitted by this contract.
+     * @param depositNonce The nonce that uniquely identifies this deposit. This function will combine this parameter
+     * with the msg.sender address to create a unique uint256 depositNonce and ensure that the msg.sender cannot
```
ankitchiplunkar, line 583: if the deposit is being made via a 4337 bundler, then the msg.sender will be the bundler address?

contracts/SpokePool.sol
```diff
@@ -543,75 +543,102 @@ abstract contract SpokePool is
...
+        // some of this deposit hash's first 24 bytes are not 0 due to the left shift of the msg.sender address, which
+        // won't be the zero address.
+
+        // @dev For some L2's, it might actually be possible that the zero address can be the msg.sender which might
+        // make it possible for an unsafe deposit hash to collide with a safe deposit nonce. To prevent these cases,
+        // we might want to consider explicitly checking that the msg.sender is not the zero address. However,
+        // this is unlikely and we also should consider not checking it and incurring the extra gas cost:
+        // if (msg.sender == address()) revert InvalidUnsafeDepositor();
+
+        uint256 depositHash = uint256((uint160(msg.sender) << 12) | depositNonce);
```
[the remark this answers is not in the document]
nicholaspai: good point

nicholaspai:
I've updated the algorithm to derive the `depositId` in `unsafeDepositV3` by removing the left shift and OR operations in favor of more simply packing together the msg.sender's address and the input `depositNonce`, keccak256 hashing the packed bytes, and converting that bytes32 into a uint256. I believe this preserves the properties we want (every depositor can generate their own unique deposit hashes and not get griefed by other EOA's)

contracts/SpokePool.sol
```diff
@@ -548,59 +548,101 @@ abstract contract SpokePool is
...
+            depositor,
+            recipient,
             inputToken,
             outputToken,
             inputAmount,
             outputAmount,
             destinationChainId,
+            exclusiveRelayer,
             // Increment count of deposits so that deposit ID for this spoke pool is unique.
+            // @dev Implicitly casts from uint32 to uint256 by padding the left-most bytes with zeros. Guarantees
```
nicholaspai, line 561: Something to verify for the reviewer: We want to guarantee that any caller of `depositV3` will never produce a `uint256 depositId` that could collide with one produced by a caller of `unsafeDepositV3`. I think this is true because `numberOfDeposits` is a `uint32` while the way `unsafeDepositV3` produces a `depositId` is by packing the 20 byte address of the `msg.sender` with the `uint96 depositId` which guarantees that the resultant 32 byte value when casted into a uint256 will always be greater than the MAX_UINT32 if the address is non-0

contracts/SpokePool.sol
```diff
@@ -548,59 +548,101 @@ abstract contract SpokePool is
...
+        // that there are no collisions between "unsafe" and "safe" deposit ID's as long as the msg.sender's address
+        // is not the zero address.
+
+        // @dev For some L2's, it could be possible that the zero address can be the msg.sender which might
+        // make it possible for an unsafe deposit hash to collide with a safe deposit nonce. To prevent these cases,
+        // we might want to consider explicitly checking that the msg.sender is not the zero address. However,
+        // this is unlikely and we choose to not check it and incur the extra gas cost:
+        // e.g. if (msg.sender == address(0)) revert InvalidUnsafeDepositor();
+
+        uint256 depositId = uint256(bytes32(abi.encodePacked(msg.sender, depositNonce)));
```
mrice32, line 632: We could flip the ordering and require that depositNonce > 0, making every custom nonce a very large number regardless of address. That would mean that any address would work, which would remove any concerns around precompiles that are at address 0x1, 0x2, 0x3 (not sure if anybody does this, but would prefer to avoid building our noncing around it).

Thoughts?

contracts/SpokePool.sol
```diff
@@ -548,59 +548,101 @@ abstract contract SpokePool is
...
+        // implicitly casting a uint32 to a uint256, where the left-most 24 bytes are set to 0. We guarantee
+        // that there are no collisions between "unsafe" and "safe" deposit ID's as long as the msg.sender's address
+        // is not the zero address.
+
+        // @dev For some L2's, it could be possible that the zero address can be the msg.sender which might
+        // make it possible for an unsafe deposit hash to collide with a safe deposit nonce. To prevent these cases,
+        // we might want to consider explicitly checking that the msg.sender is not the zero address. However,
+        // this is unlikely and we choose to not check it and incur the extra gas cost:
+        // e.g. if (msg.sender == address(0)) revert InvalidUnsafeDepositor();
+
```
mrice32, line 631: OOC, what's the reasoning behind not wrapping all this in a hash? Simplicity for the integrator to predict the deposit hash?

Hash advantages:
- We can use full 256 bit depositNonces (might improve cross compatibility with things like permit2 or other noncing systems that may be used concurrently).
- No matter how we concat the data, there is no concern about creating collisions with normal depositIds.
- We could add more parameters that get baked into the depositId. For instance, this could be called by a contract that passes its msg.sender as the depositor. Adding the depositor to the depositId hash would mean that different depositors wouldn't collide.

contracts/SpokePool.sol
```diff
@@ -548,59 +548,101 @@ abstract contract SpokePool is
...
+        // This guarantees the resultant ID won't collide with a "safe" deposit ID which is set by
+        // implicitly casting a uint32 to a uint256, where the left-most 24 bytes are set to 0. We guarantee
+        // that there are no collisions between "unsafe" and "safe" deposit ID's as long as the msg.sender's address
+        // is not the zero address.
+
+        // @dev For some L2's, it could be possible that the zero address can be the msg.sender which might
+        // make it possible for an unsafe deposit hash to collide with a safe deposit nonce. To prevent these cases,
+        // we might want to consider explicitly checking that the msg.sender is not the zero address. However,
+        // this is unlikely and we choose to not check it and incur the extra gas cost:
+        // e.g. if (msg.sender == address(0)) revert InvalidUnsafeDepositor();
```
mrice32, line 630: Should we provide a method that you can eth_call to get the depositId for a given set of args? Not a strict requirement if the computation is simple, but I could see this being convenient for integrators.

mrice32 (commented):
A few comments! Looks awesome

contracts/SpokePool.sol
nicholaspai: Otherwise I totally agree with all the reasons FOR hashing this

contracts/SpokePool.sol
nicholaspai: > require that depositNonce > 0,

Ideally I was trying to avoid an additional `require` statement here but I agree this would be a good algorithm. This is why I wanted to avoid `require(msg.sender != address(0))`

contracts/SpokePool.sol
mrice32: Did the math on this.

The probability of finding 28 bytes of leading zeros is `1 - (1 - (1/2^224))^num_tries`.

You get to about 0.3% after `1e65` tries (screenshot below). For reference, Bitcoin produces about 2e28 hashes _per year_, meaning it would take about 5e36 years for the entire bitcoin network to crack to have a 0.3% chance of cracking this.


Screenshot 2024-09-26 at 10 26 23 AM

contracts/SpokePool.sol
nicholaspai: I think this comment is irrelevant now that we hash the packed data: 19a33801e31da303616ee12e192589941528eb1c

nicholaspai:
@mrice32 just one question: do you think we should add more params to the hashed struct?

> We could add more parameters that get baked into the depositId. For instance, this could be called by a contract that passes its msg.sender as the depositor. Adding the depositor to the depositId hash would mean that different depositors wouldn't collide.

contracts/SpokePool.sol
```diff
@@ -1065,10 +1100,92 @@ abstract contract SpokePool is
...
+     */
+    function getUnsafeDepositId(address depositor, uint256 depositNonce) public pure returns (uint256) {
+        return uint256(keccak256(abi.encodePacked(depositor, depositNonce)));
+    }
+
     /**************************************
      *         INTERNAL FUNCTIONS         *
      **************************************/
 
+    function _depositV3(
```
nicholaspai, line 1118: This should be copied from the existing depositV3 logic

mrice32:
> @mrice32 just one question: do you think we should add more params to the hashed struct?
> 
> > We could add more parameters that get baked into the depositId. For instance, this could be called by a contract that passes its msg.sender as the depositor. Adding the depositor to the depositId hash would mean that different depositors wouldn't collide.

Yeah, I think msg.sender and depositor should both be in there to handle interactions from both regular users and passthrough contracts.

contracts/SpokePool.sol
```diff
@@ -1065,10 +1098,99 @@ abstract contract SpokePool is
         return block.timestamp; // solhint-disable-line not-rely-on-time
     }
 
+    /**
+     * @notice Returns the deposit ID for an unsafe deposit. This function is used to compute the deposit ID
+     * in unsafeDepositV3 and is provided as a convenience.
+     * @dev msg.sender and depositor are both used as inputs to allow passthrough depositors to create unique
+     * deposit hash spaces for unique depositors.
+     * @param sender The msg.sender address used as input to produce the deposit ID.
```
pxrl, line 1106: This wording is a bit counter-intuitive for me because it's relying on the implementation details of another function (`unsafeDepositV3()`). i.e. it's confusing that `msg.sender` is referenced in the comment, but `sender` is an input argument and can be any address.

Is `relayer` a candidate variable name? It implies that the relayer's address is needed to produce the deterministic deposit ID. In context of a relayer filling via a contract it would be the contract address, not the EOA, so idk whether that introduces a possible additional source of confusion.

nicholaspai: I think msgSender maybe is a better variable name

nicholaspai: Because maybe an aggregator also wants to use the unsafeDeposit method but there would be no natural "relayer" to fill in, right? Maybe i'm over thinking it

nicholaspai: https://github.com/across-protocol/contracts/pull/583/commits/78eb4c03d1954bc2515cbee25f07733d2c1aa1b3

mrice32 (approved):
LGTM!

PR #584, opened 2024-08-20
feat(SpokePool): Specify exclusivityDeadlineOffset by default

nicholaspai:
Better depositor UX if they can specify an offset that is relative to the time that their deposit mines, so they do not have to account for any off-chain latency

### DIFF
--- a/contracts/SpokePool.sol
+++ b/contracts/SpokePool.sol
@@ -831,9 +831,9 @@ abstract contract SpokePool is
         // Exclusivity deadline is inclusive and is the latest timestamp that the exclusive relayer has sole right
         // to fill the relay.
         if (
-            relayData.exclusiveRelayer != address(0) &&
+            relayData.exclusiveRelayer != msg.sender &&
             relayData.exclusivityDeadline >= getCurrentTime() &&
-            relayData.exclusiveRelayer != msg.sender
+            relayData.exclusiveRelayer != address(0)
         ) {
             revert NotExclusiveRelayer();
         }
--- a/contracts/SpokePool.sol
+++ b/contracts/SpokePool.sol
@@ -773,7 +773,11 @@ abstract contract SpokePool is
     {
         // Exclusivity deadline is inclusive and is the latest timestamp that the exclusive relayer has sole right
         // to fill the relay.
-        if (relayData.exclusivityDeadline >= getCurrentTime() && relayData.exclusiveRelayer != msg.sender) {
+        if (
+            relayData.exclusiveRelayer != address(0) &&
+            relayData.exclusivityDeadline >= getCurrentTime() &&
+            relayData.exclusiveRelayer != msg.sender
+        ) {
             revert NotExclusiveRelayer();
         }
 
--- a/contracts/SpokePool.sol
+++ b/contracts/SpokePool.sol
@@ -667,7 +667,7 @@ abstract contract SpokePool is
             exclusiveRelayer,
             uint32(getCurrentTime()),
             uint32(getCurrentTime()) + fillDeadlineOffset,
-            uint32(getCurrentTime()) + exclusivityPeriod,
+            exclusivityPeriod,
             message
         );
     }
@@ -735,7 +735,7 @@ abstract contract SpokePool is
             exclusiveRelayer,
             quoteTimestamp,
             fillDeadline,
-            uint32(getCurrentTime()) + exclusivityPeriod,
+            exclusivityPeriod,
             message
         );
     }
--- a/contracts/SpokePool.sol
+++ b/contracts/SpokePool.sol
@@ -525,8 +525,7 @@ abstract contract SpokePool is
      * where currentTime is block.timestamp on this chain or this transaction will revert.
      * @param exclusivityPeriod Added to the current time to set the exclusive reayer deadline,
      * which is the deadline for the exclusiveRelayer to fill the deposit. After this destination chain timestamp,
-     * anyone can fill the deposit. If exclusiveRelayer is set to address(0), then this also must be set to 0,
-     * (and vice versa), otherwise this must be set > 0.
+     * anyone can fill the deposit.
      * @param message The message to send to the recipient on the destination chain if the recipient is a contract.
      * If the message is not empty, the recipient contract must implement handleV3AcrossMessage() or the fill will revert.
      */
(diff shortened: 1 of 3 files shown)


contracts/SpokePool.sol
```diff
@@ -569,15 +570,15 @@ abstract contract SpokePool is
 
         // As a safety measure, prevent caller from inadvertently locking funds during exclusivity period
         //  by forcing them to specify an exclusive relayer if the exclusivity period
```
pxrl, line 572: Driveby observation.

```suggestion
        // by forcing them to specify an exclusive relayer if the exclusivity period
```

contracts/SpokePool.sol
```diff
@@ -569,15 +570,15 @@ abstract contract SpokePool is
...
-        // is in the future. If this deadline is 0, then the exclusive relayer must also be address(0).
+        // is in the future. If this offset is 0, then the exclusive relayer must also be address(0).
         // @dev Checks if either are > 0 by bitwise or-ing.
-        if (uint256(uint160(exclusiveRelayer)) | exclusivityDeadline != 0) {
+        if (uint256(uint160(exclusiveRelayer)) | exclusivityDeadlineOffset != 0) {
             // Now that we know one is nonzero, we need to perform checks on each.
             // Check that exclusive relayer is nonzero.
             if (exclusiveRelayer == address(0)) revert InvalidExclusiveRelayer();
```
pxrl, line 578: ooc, do we need to retain this check? It prevents a depositor from setting 0x0 as the exclusiveRelayer, but it doesn't prevent the rest of the invalid (i.e. non-relayer) address space from being used. I think the main function would be to catch depositors accidentally setting this field, but that seems like an error that an integrator would quickly identify and fix because none of their testing would succeed.

```suggestion
            // Now that we know one is nonzero, we need to perform checks on each.
            // Check that exclusive relayer is nonzero.
            if (exclusiveRelayer == address(0)) revert InvalidExclusiveRelayer();
```

contracts/SpokePool.sol
```diff
@@ -540,7 +541,7 @@ abstract contract SpokePool is
         address exclusiveRelayer,
         uint32 quoteTimestamp,
         uint32 fillDeadline,
-        uint32 exclusivityDeadline,
+        uint32 exclusivityDeadlineOffset,
```
pxrl, line 544: From the perspective of a depositor I think `exclusivityWindow` or `exclusivityPeriod` might be more self-explanatory than `exclusivityDeadlineOffset`.

contracts/SpokePool.sol
```diff
@@ -642,9 +643,9 @@ abstract contract SpokePool is
      * @param fillDeadlineOffset Added to the current time to set the fill deadline, which is the deadline for the
      * relayer to fill the deposit. After this destination chain timestamp, the fill will revert on the
      * destination chain.
-     * @param exclusivityDeadline The latest timestamp that only the exclusive relayer can fill the deposit. After this
-     * destination chain timestamp, anyone can fill this deposit on the destination chain. Should be 0 if
-     * exclusive relayer is 0.
+     * @param exclusivityDeadlineOffset Added to the current time to set the exclusive relayer deadline,
+     * which is the deadline for the exclusiveRelayer to fill the deposit. After this destination chain timestamp,
+     * anyone can fill the deposit.
```
pxrl, line 648:
```suggestion
     * anyone can fill the deposit up to the fillDeadline timestamp.
```

contracts/SpokePool.sol
```diff
@@ -672,73 +667,6 @@ abstract contract SpokePool is
             exclusiveRelayer,
             uint32(getCurrentTime()),
             uint32(getCurrentTime()) + fillDeadlineOffset,
-            exclusivityDeadline,
```
mrice32, line 675: nit: can we add this back? Just to reduce the diff size? I think we have the bytecode space regardless, no?

contracts/SpokePool.sol
mrice32: I think period is the best, yeah

contracts/SpokePool.sol
mrice32: Maybe not.

Nick and I spoke on this today, and we thought the only check that we may want to leave in is what's there now. Basically to not allow a user to set the exclusive relayer to 0 along with a nonzero exclusivity deadline, since that's clearly wrong. But we could also drop that one too just for simplicity, especially if we use 0x0 exclusive relayer as the on/off switch in the fill function.

nicholaspai: what do you think of the current check now @pxrl ? I'm not sure what your original push for was but the diff has updated

pxrl: This is simpler than what was there before :+1: I do still wonder if we could drop the check entirely, since the probability of funds locking funds is so low and so temporary.

contracts/SpokePool.sol
```diff
@@ -568,16 +569,10 @@ abstract contract SpokePool is
...
-        // is in the future. If this deadline is 0, then the exclusive relayer must also be address(0).
-        // @dev Checks if either are > 0 by bitwise or-ing.
-        if (uint256(uint160(exclusiveRelayer)) | exclusivityDeadline != 0) {
-            // Now that we know one is nonzero, we need to perform checks on each.
-            // Check that exclusive relayer is nonzero.
-            if (exclusiveRelayer == address(0)) revert InvalidExclusiveRelayer();
-
-            // Check that deadline is in the future.
-            if (exclusivityDeadline < currentTime) revert InvalidExclusivityDeadline();
+        //  by forcing them to specify an exclusive relayer if the exclusivity deadline timestamp
```
pxrl, line 572:
```suggestion
        // by forcing them to specify an exclusive relayer if the exclusivity deadline timestamp
```

pxrl (approved):

contracts/SpokePool.sol
mrice32: Yeah, the only thing this check protects against now is someone accidentally setting an exclusivity timestamp, but not a relayer. It's sort of just a bug signal since there is no reason anyone would want to do that.

I think removing is probably better -- saves gas, and this check doesn't really protect against anything.

mrice32 (approved):

contracts/SpokePool.sol
```diff
@@ -845,7 +841,11 @@ abstract contract SpokePool is
     {
         // Exclusivity deadline is inclusive and is the latest timestamp that the exclusive relayer has sole right
         // to fill the relay.
-        if (relayData.exclusivityDeadline >= getCurrentTime() && relayData.exclusiveRelayer != msg.sender) {
+        if (
+            relayData.exclusiveRelayer != address(0) &&
+            relayData.exclusivityDeadline >= getCurrentTime() &&
+            relayData.exclusiveRelayer != msg.sender
```
mrice32, line 847: nit: can we move this to the first check. Small gas optimization.

We think exclusivity will be the most common case, so we probably expect that most fills will be able to exit this if on the `relayData.exclusiveRelayer != msg.sender` check.

mrice32, line 836:
```suggestion
            relayData.exclusiveRelayer != msg.sender &&
            relayData.exclusivityDeadline >= getCurrentTime() &&
            relayData.exclusiveRelayer != address(0)
```

PR #585, opened 2024-08-20
feat(Arbitrum_Adapter): Support custom gas token

nicholaspai:
- Example `createRetryableTicket` paying for submission with custom gas token and also sending custom gas token to L2: https://etherscan.io/tx/0xfbcd1f29a0db9971c885815e74499f3a545e7c0ee0a2d06ee4a94dc08c03f4e0

- Example `outboundTransfer` paying with custom gas token and sending a different ERC20 to L2: https://etherscan.io/tx/0x7f65fff856520cc419c3a9f43a61f3427ad495ddf5dc81276c2767ac74242e39

### DIFF
--- a/contracts/chain-adapters/Arbitrum_Adapter.sol
+++ b/contracts/chain-adapters/Arbitrum_Adapter.sol
@@ -9,10 +9,37 @@ import "../external/interfaces/CCTPInterfaces.sol";
 import "../libraries/CircleCCTPAdapter.sol";
 
 /**
- * @notice Interface for Arbitrum's L1 Inbox contract used to send messages to Arbitrum.
- * @custom:security-contact bugs@across.to
+ * @title Staging ground for incoming and outgoing messages
+ * @notice Unlike the standard Eth bridge, native token bridge escrows the custom ERC20 token which is
+ * used as native currency on L2.
+ * @dev Fees are paid in this token. There are certain restrictions on the native token:
+ *       - The token can't be rebasing or have a transfer fee
+ *       - The token must only be transferrable via a call to the token address itself
+ *       - The token must only be able to set allowance via a call to the token address itself
+ *       - The token must not have a callback on transfer, and more generally a user must not be able to make a transfer to themselves revert
+ *       - The token must have a max of 2^256 - 1 wei total supply unscaled
+ *       - The token must have a max of 2^256 - 1 wei total supply when scaled to 18 decimals
+ */
+interface ArbitrumL1ERC20Bridge {
+    /**
+     * @notice Returns token that is escrowed in bridge on L1 side and minted on L2 as native currency.
+     * @dev This function doesn't exist on the generic Bridge interface.
+     */
+    function nativeToken() external view returns (address);
+}
+
+/**
+ * @title Inbox for user and contract originated messages
+ * @notice Messages created via this inbox are enqueued in the delayed accumulator
+ * to await inclusion in the SequencerInbox
  */
 interface ArbitrumL1InboxLike {
+    /**
+     * @dev we only use this function to check the native token used by the bridge, so we hardcode the interface
+     * to return an ArbitrumL1ERC20Bridge instead of a more generic Bridge interface.
+     */
+    function bridge() external view returns (ArbitrumL1ERC20Bridge);
+
     /**
      * @notice Put a message in the L2 inbox that can be reexecuted for some fixed amount of time if it reverts
      * @dev Gas limit and maxFeePerGas should not be set to 1 as that is used to trigger the RetryableData error
@@ -39,6 +66,36 @@ interface ArbitrumL1InboxLike {
         uint256 maxFeePerGas,
         bytes calldata data
     ) external payable returns (uint256);
+
+    /**
+     * @notice Put a message in the L2 inbox that can be reexecuted for some fixed amount of time if it reverts
+     * @notice Overloads the `createRetryableTicket` function but is not payable, and should only be called when paying
+     * for L1 to L2 message using a custom gas token.
+     * @dev all tokenTotalFeeAmount will be deposited to callValueRefundAddress on L2
+     * @dev Gas limit and maxFeePerGas should not be set to 1 as that is used to trigger the RetryableData error
+     * @dev In case of native token having non-18 decimals: tokenTotalFeeAmount is denominated in native token's decimals. All other value params - l2CallValue, maxSubmissionCost and maxFeePerGas are denominated in child chain's native 18 decimals.
+     * @param to destination L2 contract address
+     * @param l2CallValue call value for retryable L2 message
+     * @param maxSubmissionCost Max gas deducted from user's L2 balance to cover base submission fee
+     * @param excessFeeRefundAddress the address which receives the difference between execution fee paid and the actual execution cost. In case this address is a contract, funds will be received in its alias on L2.
+     * @param callValueRefundAddress l2Callvalue gets credited here on L2 if retryable txn times out or gets cancelled. In case this address is a contract, funds will be received in its alias on L2.
+     * @param gasLimit Max gas deducted from user's L2 balance to cover L2 execution. Should not be set to 1 (magic value used to trigger the RetryableData error)
+     * @param maxFeePerGas price bid for L2 execution. Should not be set to 1 (magic value used to trigger the RetryableData error)
+     * @param tokenTotalFeeAmount amount of fees to be deposited in native token to cover for retryable ticket cost
+     * @param data ABI encoded data of L2 message
+     * @return unique message number of the retryable transaction
+     */
+    function createRetryableTicket(
+        address to,
+        uint256 l2CallValue,
+        uint256 maxSubmissionCost,
+        address excessFeeRefundAddress,
+        address callValueRefundAddress,
+        uint256 gasLimit,
+        uint256 maxFeePerGas,
+        uint256 tokenTotalFeeAmount,
+        bytes calldata data
+    ) external returns (uint256);
 
     /**
      * @notice Put a message in the L2 inbox that can be reexecuted for some fixed amount of time if it reverts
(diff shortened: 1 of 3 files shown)


contracts/chain-adapters/Arbitrum_Adapter.sol
```diff
@@ -233,54 +321,93 @@ contract Arbitrum_Adapter is AdapterInterface, CircleCCTPAdapter {
...
-                    l1Token,
-                    to,
-                    amount,
-                    RELAY_TOKENS_L2_GAS_LIMIT,
-                    L2_GAS_PRICE,
-                    data
-                );
+
+            // If custom gas token, call special functions that handle paying with custom gas tokens.
+            if (address(CUSTOM_GAS_TOKEN) != address(0)) {
```
mrice32, line 327: Instead of these huge if else blocks, do you think it's worthwhile to just make a different adapter for these rather than squeeze them into the existing arb adapter?

contracts/chain-adapters/Arbitrum_Adapter.sol
```diff
@@ -168,6 +225,13 @@ contract Arbitrum_Adapter is AdapterInterface, CircleCCTPAdapter {
 
     ArbitrumL1ERC20GatewayLike public immutable L1_ERC20_GATEWAY_ROUTER;
 
+    // This token can be used to pay for l1 to l2 messages if its configured by an Arbitrum orbit chain.
+    // If this is initialized to be non-zero, then this token will be pulled from the customL2GasTokenFunder contract
+    // to pay for l1 to l2 messages.
+    IERC20 public immutable CUSTOM_GAS_TOKEN;
+
+    address public immutable CUSTOM_GAS_TOKEN_FUNDER;
```
mrice32, line 233: Will this contract be added in a separate PR?

mrice32 (approved):
Looks great! Just two questions

contracts/chain-adapters/Arbitrum_Adapter.sol
nicholaspai: hmm there is a lot of code we can re-use though

contracts/chain-adapters/Arbitrum_Adapter.sol
nicholaspai: yeah let's just assume for now it is some contract that has granted approval to this contract to pull all of its funds. WDYT?

mrice32: Do you think we need to audit that contract?

contracts/chain-adapters/Arbitrum_Adapter.sol
mrice32: So I guess there are three options:
1. Put them all in the same adapter, as we have here.
2. Copy the code that is identical and make a separate custom gas adapter.
3. Create a common base class to share the common code.

mrice32: I think 2 is the approach that makes the adapters the easiest to reason about, but not super strongly opinionated.

mrice32 (commented):
Left follow-up comments

contracts/chain-adapters/Arbitrum_Adapter.sol
nicholaspai: 2 is also slightly cheaper gas cost wise, i'm ok with that. Makes the diff easier to reason about

contracts/chain-adapters/Arbitrum_Adapter.sol
nicholaspai: probably but i didn't want to build it in this PR. It could even be an EOA that we seed with funds, right?

nicholaspai: i think the key thing to center around is whether we should "pull" from the funder via transferFrom or should there be an interface that exposes a function like `pull()` that we can cull to get funds

contracts/chain-adapters/Arbitrum_Adapter.sol
nicholaspai: https://github.com/across-protocol/contracts/pull/585/commits/d6f038d8bf878065cfb9af2821f5428af77cca20

contracts/chain-adapters/Arbitrum_Adapter.sol
nicholaspai: I guess one problem with just calling transferFrom from a set address is that LP's might worry that the address gets set to their EOA, which has approved the HubPool to spend their tokens, and can be used as a vector to take their funds.

nicholaspai: wdyt of this simple funder: edc3fd86b226a7b6d5321f9853c91f28d5a9c8f7

mrice32: Yeah, I like that we're not just doing a transferFrom from an arbitrary address. +1 for simple funder contract.

contracts/chain-adapters/Arbitrum_CustomGasToken_Adapter.sol
```diff
@@ -0,0 +1,291 @@
...
+    function getL1CallValue(uint32 l2GasLimit) public view returns (uint256) {
+        return L2_MAX_SUBMISSION_COST + L2_GAS_PRICE * l2GasLimit;
+    }
+
+    function _pullCustomGas(uint32 l2GasLimit) internal returns (uint256) {
+        uint256 requiredL1CallValue = getL1CallValue(l2GasLimit);
+        CUSTOM_GAS_TOKEN_FUNDER.withdraw(CUSTOM_GAS_TOKEN, requiredL1CallValue);
+        require(CUSTOM_GAS_TOKEN.balanceOf(address(this)) >= requiredL1CallValue, "Insufficient gas balance");
+        return requiredL1CallValue;
+    }
```
mrice32, line 290: Just to be really explicit: the reason we don't want to just pull these tokens from msg.sender is that these executions are often called via the external multicall contract, so this would mean that we could no longer batch these executions with the mainnet spoke pool relayer refund leaf execution, right?

contracts/chain-adapters/Arbitrum_CustomGasToken_Funder.sol
```diff
@@ -0,0 +1,28 @@
...
+    using SafeERC20 for IERC20;
+
+    /**
+     * @notice Deposit tokens into the contract.
+     * @param token Token to deposit.
+     * @param amount Amount of tokens to deposit.
+     */
+    function deposit(IERC20 token, uint256 amount) external {
+        token.safeTransferFrom(msg.sender, address(this), amount);
+    }
```
mrice32, line 18: nit: do we need a deposit function if we can just transfer tokens in?

contracts/chain-adapters/Arbitrum_CustomGasToken_Adapter.sol
```diff
@@ -0,0 +1,291 @@
...
+        FunderInterface _customGasTokenFunder,
+        uint256 _l2MaxSubmissionCost
+    ) CircleCCTPAdapter(_l1Usdc, _cctpTokenMessenger, CircleDomainIds.Arbitrum) {
+        L1_INBOX = _l1ArbitrumInbox;
+        L1_ERC20_GATEWAY_ROUTER = _l1ERC20GatewayRouter;
+        L2_REFUND_L2_ADDRESS = _l2RefundL2Address;
+        CUSTOM_GAS_TOKEN = _customGasToken;
+        L2_MAX_SUBMISSION_COST = _l2MaxSubmissionCost;
+        CUSTOM_GAS_TOKEN_FUNDER = _customGasTokenFunder;
+        require(L1_INBOX.bridge().nativeToken() == address(_customGasToken), "Incorrect custom gas token");
```
mrice32, line 187: nit: why pass _customGasToken and check it vs just reading it from nativeToken()?

contracts/chain-adapters/Arbitrum_CustomGasToken_Adapter.sol
```diff
@@ -0,0 +1,291 @@
...
+                    L2_MAX_SUBMISSION_COST, // maxSubmissionCost Max gas deducted from user's L2 balance to cover base fee
+                    L2_REFUND_L2_ADDRESS, // excessFeeRefundAddress maxgas * gasprice - execution cost gets credited here on L2
+                    L2_REFUND_L2_ADDRESS, // callValueRefundAddress l2Callvalue gets credited here on L2 if retryable txn times out or gets cancelled
+                    RELAY_MESSAGE_L2_GAS_LIMIT, // maxGas Max gas deducted from user's L2 balance to cover L2 execution
+                    L2_GAS_PRICE, // gasPriceBid price bid for L2 execution
+                    amountToBridge, // tokenTotalFeeAmount amount of fees to be deposited in native token.
+                    "0x" // data ABI encoded data of L2 message
+                );
+            } else {
+                CUSTOM_GAS_TOKEN.safeIncreaseAllowance(erc20Gateway, amount);
```
mrice32, line 257: nit: should this be `requiredL1TokenTotalFeeAmount` instead of `amount`?

mrice32 (approved):
Looks great! Only a few minor nits

contracts/chain-adapters/Arbitrum_CustomGasToken_Adapter.sol
nicholaspai: > we could no longer batch these executions with the mainnet spoke pool relayer refund leaf execution

Why not?

contracts/chain-adapters/Arbitrum_CustomGasToken_Funder.sol
nicholaspai: true, will remove

contracts/chain-adapters/Arbitrum_CustomGasToken_Adapter.sol
nicholaspai: Yes great catch. I think I'm also missing an increaseAllowance of the l1Token in this else{} branch: 9674ed0f9fbecb6a1fda87e9d5081a91eb47042c

contracts/chain-adapters/Arbitrum_CustomGasToken_Adapter.sol
mrice32: Sorry, I don't think I said that very clearly. We can batch with this implementation.

If we didn't do the funder contract and just did `transferFrom(msg.sender, ...)`, that would work, but it would break batching since we use the external multicall contract to do batching now, and that contract would be the msg.sender (and contains no tokens).

mrice32 (approved):
LGTM!

PR #587, opened 2024-08-26
feat: redeploy spoke pools to support depositExclusive

bmzig:
This redeploys all spoke pools which were deployed before Lisk to support `depositExclusive`. A nice byproduct is that now all chains also support tryMulticall.


james-a-morris (approved):
lgtm pending at least a second signoff

hardhat.config.ts
```diff
@@ -132,7 +136,7 @@ const config: HardhatUserConfig = {
     },
     arbitrum: {
       chainId: CHAIN_IDs.ARBITRUM,
-      url: getNodeUrl("arbitrum", true, CHAIN_IDs.ARBITRUM),
+      url: getNodeUrl("arbitrum-mainnet", true, CHAIN_IDs.ARBITRUM),
```
nicholaspai, line 139: these were renamed by infura?

bmzig: Yeah they must have been. They now follow a naming convention given by `[network_name]-[mainnet or sepolia].infura.io.`

nicholaspai: huh ok that makes sense, does this just not work currently with the renamed network names?

nicholaspai (approved):

PR #589, opened 2024-08-28
improve(Arbitrum_CustomGasToken_Adapter): Handle non-18 custom gas token precision

nicholaspai:
Based on this [Inbox](https://github.com/OffchainLabs/nitro-contracts/blob/fbbcef09c95f69decabaced3da683f987902f3e2/src/bridge/AbsInbox.sol#L237) code which implies that the `tokenTotalFeeAmount` passed into the [ERC20Inbox](https://github.com/OffchainLabs/nitro-contracts/blob/fbbcef09c95f69decabaced3da683f987902f3e2/src/bridge/ERC20Inbox.sol#L85) should be in the custom gas token's precision while the `maxSubmissionCost`, `gasPrice`, and `gasLimit` should all be same as for normal Arbitrum messages


contracts/chain-adapters/Arbitrum_CustomGasToken_Adapter.sol
```diff
@@ -135,11 +142,10 @@ contract Arbitrum_CustomGasToken_Adapter is AdapterInterface, CircleCCTPAdapter
     // ticket’s calldata in the retry buffer. (current base submission fee is queryable via
     // ArbRetryableTx.getSubmissionPrice). ArbRetryableTicket precompile interface exists at L2 address
     // 0x000000000000000000000000000000000000006E.
-    // @dev Unlike in Arbitrum_Adapter, this is immutable because we don't know what precision the custom gas token has.
     uint256 public immutable L2_MAX_SUBMISSION_COST;
```
nicholaspai, line 145:
```suggestion
    // The Arbitrum Inbox requires that this uses 18 decimal precision.
    uint256 public immutable L2_MAX_SUBMISSION_COST;
```

contracts/chain-adapters/Arbitrum_CustomGasToken_Adapter.sol
```diff
@@ -135,11 +142,11 @@ contract Arbitrum_CustomGasToken_Adapter is AdapterInterface, CircleCCTPAdapter
     // ticket’s calldata in the retry buffer. (current base submission fee is queryable via
     // ArbRetryableTx.getSubmissionPrice). ArbRetryableTicket precompile interface exists at L2 address
     // 0x000000000000000000000000000000000000006E.
-    // @dev Unlike in Arbitrum_Adapter, this is immutable because we don't know what precision the custom gas token has.
+    // The Arbitrum Inbox requires that this uses 18 decimal precision.
     uint256 public immutable L2_MAX_SUBMISSION_COST;
 
     // L2 Gas price bid for immediate L2 execution attempt (queryable via standard eth*gasPrice RPC)
-    uint256 public constant L2_GAS_PRICE = 5e9; // 5 gWei
+    uint256 public immutable L2_GAS_PRICE;
```
nicholaspai, line 149:
```suggestion
    // The Arbitrum Inbox requires that this is specified in gWei (e.g. 1e9 = 1 gWei)
    uint256 public immutable L2_GAS_PRICE;
```

contracts/chain-adapters/Arbitrum_CustomGasToken_Adapter.sol
```diff
@@ -275,10 +290,11 @@ contract Arbitrum_CustomGasToken_Adapter is AdapterInterface, CircleCCTPAdapter
 
     /**
      * @notice Returns required amount of gas token to send a message via the Inbox.
+     * @dev Should return a value in the same precision as the gas token's precision.
      * @return amount of gas token that this contract needs to hold in order for relayMessage to succeed.
      */
     function getL1CallValue(uint32 l2GasLimit) public view returns (uint256) {
-        return L2_MAX_SUBMISSION_COST + L2_GAS_PRICE * l2GasLimit;
+        return _from18ToNativeDecimals(L2_MAX_SUBMISSION_COST + L2_GAS_PRICE * l2GasLimit);
```
nicholaspai, line 297: the benefit of putting the scaling logic into this public function is the deployer can quickly sanity check post-deployment that the L1CallValue will be set correctly with the right precision

mrice32 (approved):
LGTM! Is the L1 representation the only one that has non 18 decimals or would the native representation on the L2 also have non 18 decimals?

contracts/chain-adapters/Arbitrum_CustomGasToken_Adapter.sol
```diff
@@ -296,4 +311,12 @@ contract Arbitrum_CustomGasToken_Adapter is AdapterInterface, CircleCCTPAdapter
         if (CUSTOM_GAS_TOKEN.balanceOf(address(this)) < requiredL1CallValue) revert InsufficientCustomGasToken();
         return requiredL1CallValue;
     }
+
+    function _from18ToNativeDecimals(uint256 amount) internal view returns (uint256) {
+        uint8 nativeTokenDecimals = L1_INBOX.bridge().nativeTokenDecimals();
+        if (nativeTokenDecimals < 18) {
+            return amount / 10**(18 - nativeTokenDecimals);
```
bmzig, line 318: Maybe I'm missing something, but it seems like we will return 0 for all tokens which have decimals < 12? Our L2 gasLimit is 2_000_000, so for example, if `getL1CallValue()` is called and the nativeTokenDecimals() is, for example, 6, this will just be (2_000_000/ 10**12) = 0?

bmzig (approved):

PR #591, opened 2024-09-04
feat: Update @openzeppelin/contracts and /contracts-upgradeable to 5.X

nicholaspai:
- Gets latest major release including security fixes. Full release notes [here](https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/releases/tag/v5.0.0).
- Downsides: changes code in contracts like HubPool that we can't actually upgrade on chain. This is done to make contracts compile.
- Remove @uma/core dependencies. This was useful because updating the @openzeppelin/contracts version in this package means that solidity files that import @uma/core that themselves use @openzeppelin/contracts will use the new @openzeppelin/contracts version (it seems that hardhat doesn't resolve these versions correctly). To support easier updates in the future, move all @uma/core contracts used in this repo directly into this repo

Unfortunately, I don't think we can merge this easily unless we also upgrade `@openzeppelin/upgrades` to [3.0.0](https://github.com/OpenZeppelin/openzeppelin-upgrades/releases/tag/%40openzeppelin%2Fhardhat-upgrades%403.0.0) which supports `contracts>=5`. I've tried updating this locally to lots of errors, so I'm going to park this here.


hardhat.config.ts
```diff
@@ -61,25 +61,7 @@ const config: HardhatUserConfig = {
     compilers: [DEFAULT_CONTRACT_COMPILER_SETTINGS],
     overrides: {
       "contracts/HubPool.sol": LARGE_CONTRACT_COMPILER_SETTINGS,
-      "contracts/Arbitrum_SpokePool.sol": {
```
nicholaspai, line 64: For some reason, without removing this I got this error

```
These files and its dependencies cannot be compiled with your config. This can happen because they have incompatible Solidity pragmas, or don't match any of your configured Solidity compilers.

  * contracts/Linea_SpokePool.sol
  * contracts/Arbitrum_SpokePool.sol
  * contracts/SpokePoolVerifier.sol
```

However I also don't think these lines are needed since we can fix compiler versions for these three contracts. To do so I needed to change the SpokePoolVerifier's `pragma` statement

nicholaspai:
Replaced by https://github.com/across-protocol/contracts/pull/598 which has a much smaller footprint. It adds 5.x contracts as optional aliased contracts to import

PR #598, opened 2024-09-05
feat: Use @openzeppelin/contracts@5.x contracts in contracts we can upgrade

nicholaspai:
Imports an [aliased npm package](https://forum.openzeppelin.com/t/coexist-of-v5-and-v4-contracts/38030/5) so we can use multiple oz libraries in the same solidity contracts.

This way we get access to new and improved contracts. I make changes to all contracts that we can upgrade, such as SpokePools and NOT the HubPool.

For example, I can use the new SafeERC20 method `forceApprove` instead of `safeIncreaseAllowance` which ensures [compatibility with tokens like USDT](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4231) that make sure all approvals are set to 0 before granting a new approval. This hasn't been an issue so far because we always safeIncreaseAllowance to some number and use the complete allowance, but its worth safety checking.

Other changes:
- Moved `MerkleDistributor` out of uma/core into this repo to reduce dependency on this external repo
- Replaced isCode() call with now [recommended](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3945) explicit .code.length check
- Explicitly set SpokePoolVerifier pragma to 0.8.19 and removed the overrides in hardhat.config.ts
- Removed unused import in PolygonERC20Test


contracts/SpokePoolVerifier.sol
```diff
@@ -1,7 +1,8 @@
 // SPDX-License-Identifier: BUSL-1.1
-pragma solidity ^0.8.0;
+// NOTE: Linea and Arbitrum only support 0.8.19, so keep this contract at 19 so create2 addressses for all chains are
+// the same.
+pragma solidity ^0.8.19;
```
bmzig, line 4: Should this be =0.8.19?

nicholaspai: i think that's right

nicholaspai:
I need to revise this PR and remove all the contracts5 imports into SpokePool because Arbitrum/Linea_SpokePool can't use those new contracts because they must be compiled with 8.19 while contracts5 uses 8.20

PR #602, opened 2024-09-10
Add minimal viable Anchor

chrismaree:
This PR re-works this repo to work alongside Solana Anchor. To achieve this, it adds a number of dependencies and re-works a lot of the repo layout, such as moving where some tests are located (hence the large file diff). In broad strokes this PR does:
- Updates the repo to work with node 20. to achieve this two deprecated & redundant packages were removed `arb-bridge-eth` and `arb-bridge-peripherals`. both of these are not supported by Arbitrum anymore. We only used these packages to fetch the interfaces in some tests when creating a fake. to supplement this, the `ArbutrumMocks.sol` was updated to store the same interface (see [here](https://github.com/across-protocol/contracts/pull/602/files#r1752104397)).
- Added required MVP anchor files, including `Anchor.toml, Cargo.lock, Cargo.toml, Programs/x, test/svm/x`. sample program has the simplest posible SVM contract and test to show the tooling working.
- Moved existing EVM tests from `/test/x` to `/test/evm/hardhat` & `/test/evm/foundry` to better separate test directories and not mix and match the way things were before. This required editing all ts files within the test directory, hence the large change in the PR (done with an automated tool).
- Updates the GitHub work flow. This change was pretty extensive. I did the following:
  - Updated to node 20. Added the setup and build for Anchor + solana local key gen + rust compilation steps.
  - Removed the Solhint job and rather placed this within the Lint job. Solhint is just a lint flow. now, we have a Lint job that does: a) lint-js, b) lint-solidity, c) lint-rust.
  - Re-work the Test job to now have: a) test evm-hardhat, b test svm. The test evm-forge is left as a separate job (under the title Forge) to keep the speed of CI up.


contracts/test/ArbitrumMocks.sol
```diff
@@ -29,3 +29,27 @@ contract ArbitrumMockErc20GatewayRouter {
         return address(this);
     }
 }
+
+contract Inbox {
```
chrismaree, line 33: ^

package.json
```diff
@@ -66,8 +71,8 @@
     "@types/node": "^12.0.0",
     "@typescript-eslint/eslint-plugin": "^4.29.1",
     "@typescript-eslint/parser": "^4.29.1",
-    "arb-bridge-eth": "^0.7.4",
```
chrismaree, line 69: packages removed to get node 20 support.

.github/workflows/pr.yml
```diff
@@ -6,53 +6,111 @@ jobs:
...
         uses: actions/checkout@v3
-      - name: Use Node ${{ matrix.node }}
+      - name: "Use Node ${{ matrix.node }}"
         uses: actions/setup-node@v3
         with:
-          node-version: ${{ matrix.node }}
-          cache: "yarn"
+          node-version: "${{ matrix.node }}"
+          cache: yarn
+      - name: Setup Anchor & Build
```
nicholaspai, line 19: Aside: is github CI smart enough to only run certain tasks if certain files are changed? For example, we should only run anchor tasks if solana files are touched right?

.gitignore
```diff
@@ -23,3 +23,11 @@ storage-layouts/proposed*
 
 # IDE
 .idea
+
+
+.anchor
```
nicholaspai, line 28: nit: can you move these into the appropriate commented sections above? maybe make a new one for solana

nicholaspai (approved):

.github/workflows/pr.yml
chrismaree: not easily. right now it will do all parts of the work flow on each run. we can improve this later. it's not too slow by comparison to how it was before.

mrice32: Conditional runs are pretty hard, but I think there are easier ways to make it less imapctful.

Two things on this if it becomes a pain point:
- We can use a docker image with rust tools and/or solana tools preinstalled.
- We can run this setup in parallel while we're doing EVM build, tests, etc, which I suspect will take longer.

.github/workflows/pr.yml
```diff
@@ -6,53 +6,111 @@ jobs:
...
+        uses: metadaoproject/setup-anchor@v2
+        with:
+          anchor-version: 0.30.1
+          solana-cli-version: 1.18.21
+          node-version: 20.17.0
+      - name: Create keypair
+        run: solana-keygen new --no-bip39-passphrase
+        shell: bash
+      - name: Make Anchor.toml compatible with runner
+        run: "sed -i 's:/user/:/runner/:' Anchor.toml"
```
mrice32, line 29: I don't see the string `user` in the Anchor.toml?

.github/workflows/publish.yml
```diff
@@ -24,6 +24,13 @@ jobs:
         uses: manovotny/github-releases-for-automated-package-publishing-action@v2.0.1
 
       # Setup .npmrc file to publish to npm
+      - name: Setup Anchor & Build
```
mrice32, line 27: Maybe move the comment above down to above npm?

.github/workflows/pr.yml
```diff
@@ -6,53 +6,111 @@ jobs:
...
-          node-version: ${{ matrix.node }}
-          cache: "yarn"
+          node-version: "${{ matrix.node }}"
+          cache: yarn
+      - name: Setup Anchor & Build
+        uses: metadaoproject/setup-anchor@v2
+        with:
+          anchor-version: 0.30.1
+          solana-cli-version: 1.18.21
+          node-version: 20.17.0
```
mrice32, line 113: Will this break if the node version varies (i.e. 20 resolves to 20.17.1)?

Can we just specify 20 here or does it have to be specific?

hardhat.config.ts
```diff
@@ -468,6 +468,9 @@ const config: HardhatUserConfig = {
     outDir: "./typechain",
     target: "ethers-v5",
   },
+  paths: {
+    tests: "./test/evm/hardhat",
```
mrice32, line 472: You've double checked that this still runs all the tests, right? Just want to double check that a change like this doesn't accidentally disable all the evm tests because of a minor typo or config issue causes it to point at an empty dir.

mrice32 (approved):
LGTM just a few small questions around ci.

hardhat.config.ts
chrismaree: 2024-09-15 at 19 19 08@2x

chrismaree: 2024-09-15 at 19 19 43@2x

.github/workflows/pr.yml
chrismaree: I think it will work just fine? the version specified above in the config has `20` set, with no sub versions. I'll try changing this here and see if the workflow still runs.

.github/workflows/publish.yml
chrismaree: yup, sorry. I messed that up. have fixed it.

.github/workflows/pr.yml
chrismaree: test worked just fine with the node version simply being "20"

.github/workflows/pr.yml
chrismaree: I might actually be able to delete this. it was trying to get anchor to detect the `solana-keygen new` command from above. the util can in some environments place the key in a location anchor cant find it, hence this command. but it looks like we don't need that after all and the tests run without it.

.github/workflows/pr.yml
mrice32: Perfect

hardhat.config.ts
mrice32: Amazing

mrice32 (approved):
LGTM!

PR #605, opened 2024-09-13
[WIP] refactor: export bridging logic to l2 adapter contracts

bmzig:
There will be shared logic between L2 token retrievers which need to bridge funds from L3 back to L1 and L2 spoke pool logic. We may be able to factor this out for each chain and create another adapter-like contract which can be `delegatecall`-ed by both L2 token retriever and spoke pool contracts.

### DIFF
--- a/contracts/chain-adapters/l2/Arbitrum_WithdrawalAdapter.sol
+++ b/contracts/chain-adapters/l2/Arbitrum_WithdrawalAdapter.sol
@@ -1,0 +1,76 @@
+// SPDX-License-Identifier: BUSL-1.1
+
+// Arbitrum only supports v0.8.19
+// See https://docs.arbitrum.io/for-devs/concepts/differences-between-arbitrum-ethereum/solidity-support#differences-from-solidity-on-ethereum
+pragma solidity ^0.8.19;
+
+import "../../libraries/CircleCCTPAdapter.sol";
+import { StandardBridgeLike } from "../../Arbitrum_SpokePool.sol";
+
+/**
+ * @notice AVM specific bridge adapter. Implements logic to bridge tokens back to mainnet.
+ * @custom:security-contact bugs@across.to
+ */
+
+/*
+ * @notice Interface for the Across Arbitrum_SpokePool contract. Used to access state which
+ * can only be modified by admin functions.
+ */
+interface IArbitrum_SpokePool {
+    function whitelistedTokens(address) external view returns (address);
+}
+
+/**
+ * @title Adapter for interacting with bridges from the Arbitrum One L2 to Ethereum mainnet.
+ * @notice This contract is used to share L2-L1 bridging logic with other L2 Across contracts.
+ */
+contract Arbitrum_WithdrawalAdapter is CircleCCTPAdapter {
+    IArbitrum_SpokePool public immutable spokePool;
+    address public immutable tokenRetriever;
+    address public immutable l2GatewayRouter;
+
+    /*
+     * @notice constructs the withdrawal adapter.
+     * @param _l2Usdc address of native USDC on the L2.
+     * @param _cctpTokenMessenger address of the CCTP token messenger contract on L2.
+     * @param _spokePool address of the spoke pool on L2.
+     * @param _tokenRetriever L1 address of the recipient of withdrawals.
+     * @param _l2GatewayRouter address of the Arbitrum l2 gateway router contract.
+     */
+    constructor(
+        IERC20 _l2Usdc,
+        ITokenMessenger _cctpTokenMessenger,
+        IArbitrum_SpokePool _spokePool,
+        address _tokenRetriever,
+        address _l2GatewayRouter
+    ) CircleCCTPAdapter(_l2Usdc, _cctpTokenMessenger, CircleDomainIds.Ethereum) {
+        spokePool = _spokePool;
+        tokenRetriever = _tokenRetriever;
+        l2GatewayRouter = _l2GatewayRouter;
+    }
+
+    /*
+     * @notice Calls CCTP or the Arbitrum gateway router to withdraw tokens back to the `tokenRetriever`. The
+     * bridge will not be called if the token is not in the Arbitrum_SpokePool's `whitelistedTokens` mapping.
+     * @param amountToReturn amount of l2Token to send back to the token retriever.
+     * @param l2TokenAddress address of the l2Token to send back to the token retriever.
+     */
+    function withdrawToken(uint256 amountToReturn, address l2TokenAddress) external {
+        // If the l2TokenAddress is UDSC, we need to use the CCTP bridge.
+        if (_isCCTPEnabled() && l2TokenAddress == address(usdcToken)) {
+            _transferUsdc(tokenRetriever, amountToReturn);
+        } else {
+            // Check that the Ethereum counterpart of the L2 token is stored on this contract.
+            // Tokens will only be bridged if they are whitelisted by the spoke pool.
+            address ethereumTokenToBridge = spokePool.whitelistedTokens(l2TokenAddress);
+            require(ethereumTokenToBridge != address(0), "Uninitialized mainnet token");
+            //slither-disable-next-line unused-return
+            StandardBridgeLike(l2GatewayRouter).outboundTransfer(
+                ethereumTokenToBridge, // _l1Token. Address of the L1 token to bridge over.
+                tokenRetriever, // _to. Withdraw, over the bridge, to the l1 hub pool contract.
+                amountToReturn, // _amount.
+                "" // _data. We don't need to send any data for the bridging action.
+            );
+        }
+    }
+}
--- a/test/foundry/local/Arbitrum_WithdrawalAdapter.t.sol
+++ b/test/foundry/local/Arbitrum_WithdrawalAdapter.t.sol
@@ -1,0 +1,129 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.0;
+
+import { Test } from "forge-std/Test.sol";
+import { MockERC20 } from "forge-std/mocks/MockERC20.sol";
+import { ERC20, IERC20 } from "@openzeppelin/contracts/token/ERC20/ERC20.sol";
+import { Arbitrum_SpokePool, ITokenMessenger } from "../../../contracts/Arbitrum_SpokePool.sol";
+import { Arbitrum_WithdrawalAdapter, IArbitrum_SpokePool } from "../../../contracts/chain-adapters/l2/Arbitrum_WithdrawalAdapter.sol";
+import { ERC1967Proxy } from "@openzeppelin/contracts/proxy/ERC1967/ERC1967Proxy.sol";
+import "@openzeppelin/contracts-upgradeable/token/ERC20/IERC20Upgradeable.sol";
+
+import "forge-std/console.sol";
+
+contract Token_ERC20 is ERC20 {
+    constructor(string memory name, string memory symbol) ERC20(name, symbol) {}
+
+    function mint(address to, uint256 value) public virtual {
+        _mint(to, value);
+    }
+
+    function burn(address from, uint256 value) public virtual {
+        _burn(from, value);
+    }
+}
+
+contract ArbitrumGatewayRouter {
+    function outboundTransfer(
+        address tokenToBridge,
+        address recipient,
+        uint256 amountToReturn,
+        bytes memory
+    ) external returns (bytes memory) {
+        Token_ERC20(tokenToBridge).burn(msg.sender, amountToReturn);
+        Token_ERC20(tokenToBridge).mint(recipient, amountToReturn);
+        return "";
+    }
+}
+
+contract Arbitrum_WithdrawalAdapterTest is Test {
+    Arbitrum_WithdrawalAdapter arbitrumWithdrawalAdapter;
+    Arbitrum_SpokePool arbitrumSpokePool;
+    Token_ERC20 whitelistedToken;
+    Token_ERC20 usdc;
+    ArbitrumGatewayRouter gatewayRouter;
+
+    // HubPool should receive funds.
+    address hubPool;
+
+    address owner;
+    address aliasedOwner;
+    address wrappedNativeToken;
+
+    // Token messenger is set so CCTP is activated.
+    ITokenMessenger tokenMessenger;
+
+    function setUp() public {
+        whitelistedToken = new Token_ERC20("DAI", "DAI");
+        usdc = new Token_ERC20("USDC", "USDC");
+        gatewayRouter = new ArbitrumGatewayRouter();
+
+        tokenMessenger = ITokenMessenger(vm.addr(1));
+        owner = vm.addr(2);
+        wrappedNativeToken = vm.addr(3);
+        hubPool = vm.addr(4);
+        aliasedOwner = _applyL1ToL2Alias(owner);
+
+        vm.startPrank(owner);
+        Arbitrum_SpokePool implementation = new Arbitrum_SpokePool(wrappedNativeToken, 0, 0, usdc, tokenMessenger);
+        address proxy = address(
+            new ERC1967Proxy(
+                address(implementation),
+                abi.encodeCall(Arbitrum_SpokePool.initialize, (0, address(gatewayRouter), owner, owner))
+            )
+        );
+        vm.stopPrank();
+
+        arbitrumSpokePool = Arbitrum_SpokePool(payable(proxy));
+        arbitrumWithdrawalAdapter = new Arbitrum_WithdrawalAdapter(
+            usdc,
+            tokenMessenger,
+            IArbitrum_SpokePool(proxy),
+            hubPool,
+            address(gatewayRouter)
+        );
+    }
+
+    function testWithdrawWhitelistedTokenNonCCTP() public {
+        assertEq(whitelistedToken.balanceOf(hubPool), 0);
+        assertEq(whitelistedToken.balanceOf(owner), 0);
+        assertEq(whitelistedToken.balanceOf(address(arbitrumWithdrawalAdapter)), 0);
+        uint256 amountToBridge = uint256(block.timestamp + 100);
+
+        // Whitelist tokens in the spoke pool and mint tokens to the withdrawal adapter.
+        vm.startPrank(aliasedOwner);
+        arbitrumSpokePool.whitelistToken(address(whitelistedToken), address(whitelistedToken));
+        whitelistedToken.mint(address(arbitrumWithdrawalAdapter), amountToBridge);
+        vm.stopPrank();
+
+        // Attempt to withdraw token.
+        arbitrumWithdrawalAdapter.withdrawToken(amountToBridge, address(whitelistedToken));
+        assertEq(whitelistedToken.balanceOf(hubPool), amountToBridge);
+        assertEq(whitelistedToken.balanceOf(owner), 0);
+        assertEq(whitelistedToken.balanceOf(address(arbitrumWithdrawalAdapter)), 0);
+    }
+
+    function testWithdrawOtherTokenNonCCTP() public {
+        assertEq(whitelistedToken.balanceOf(hubPool), 0);
+        assertEq(whitelistedToken.balanceOf(owner), 0);
+        assertEq(whitelistedToken.balanceOf(address(arbitrumWithdrawalAdapter)), 0);
+        uint256 amountToBridge = uint256(block.timestamp + 100);
+
+        // Mint tokens to the withdrawal adapter but do not whitelist it in the spoke pool.
+        whitelistedToken.mint(address(arbitrumWithdrawalAdapter), amountToBridge);
+
+        // Attempt to withdraw token.
+        vm.expectRevert();
+        arbitrumWithdrawalAdapter.withdrawToken(amountToBridge, address(whitelistedToken));
+        assertEq(whitelistedToken.balanceOf(hubPool), 0);
+        assertEq(whitelistedToken.balanceOf(owner), 0);
+        assertEq(whitelistedToken.balanceOf(address(arbitrumWithdrawalAdapter)), amountToBridge);
+    }
+
+    function _applyL1ToL2Alias(address l1Address) internal pure returns (address l2Address) {
+        // Allows overflows as explained above.
+        unchecked {
+            l2Address = address(uint160(l1Address) + uint160(0x1111000000000000000000000000000000001111));
+        }
+    }
+}


contracts/Intermediate_TokenRetriever.sol
```diff
@@ -0,0 +1,58 @@
...
+        address l2Token
+    ) external;
+}
+
+/**
+ * @notice Contract deployed on an arbitrary L2 to act as an intermediate contract for withdrawals from L3 to L1.
+ * @dev Since each network has its own bridging requirements, this contract delegates that logic to a bridge adapter contract
+ * which performs the necessary withdraw action.
+ */
+contract Intermediate_TokenRetriever is Lockable, MultiCaller {
```
bmzig, line 22: I don't think we want `MultiCaller` here since we would rather have logic for many withdrawals on the bridge adapter contract and not on the retriever contract, since having it here means `n` delegatecalls for `n` internal transactions vs 1 delegatecall for `n` internal transactions (although removing it complicates the logic a bit).

bmzig: https://github.com/across-protocol/contracts/pull/605/commits/3131eed7f6f65691c99260b5529a43e54c499c21

contracts/L2_TokenRetriever.sol
```diff
@@ -0,0 +1,84 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.19;
+
+import "./Lockable.sol";
+import "@openzeppelin/contracts-upgradeable/token/ERC20/IERC20Upgradeable.sol";
+import "@openzeppelin/contracts-upgradeable/token/ERC20/utils/SafeERC20Upgradeable.sol";
+import "@uma/core/contracts/common/implementation/MultiCaller.sol";
```
pxrl, line 7: Do you still want this import?

contracts/L2_TokenRetriever.sol
```diff
@@ -0,0 +1,84 @@
...
+ * which performs the necessary withdraw action.
+ */
+contract L2_TokenRetriever is Lockable {
+    using SafeERC20Upgradeable for IERC20Upgradeable;
+
+    // Should be set to the bridge adapter which contains the proper logic to withdraw tokens on
+    // the deployed L2
+    address public immutable bridgeAdapter;
+    // Should be set to the L1 address which will receive withdrawn tokens.
+    address public immutable tokenRetriever;
```
pxrl, line 32: If this is the address that should be set as the recipient of tokens on the L1 then `tokenRetriever` is a bit counter-intuitive for me. Would `tokenRecipient` be more descriptive?

```suggestion
    address public immutable tokenRecipient;
```

contracts/chain-adapters/l2/WithdrawalAdapter.sol
```diff
@@ -0,0 +1,62 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.19;
+
+import "../../libraries/CircleCCTPAdapter.sol";
```
pxrl, line 4: Nick's recent commit used explicit imports for CCTP; do we want to replicate that in this file https://github.com/across-protocol/contracts/commit/d197e7d543777bad8725d82ae859122544f5b6f2

bmzig: Yeah. I'll start doing that from now on. https://github.com/across-protocol/contracts/pull/605/commits/d3f624bff2d3b263be3a310a5c9ef36a19075bc0

contracts/L2_TokenRetriever.sol
bmzig: I do think so. I had the blast and polygon contracts on the mind, but it's nice to not repeat retriever when we don't have to. https://github.com/across-protocol/contracts/pull/605/commits/d3f624bff2d3b263be3a310a5c9ef36a19075bc0

contracts/L2_TokenRetriever.sol
bmzig: Removed https://github.com/across-protocol/contracts/pull/605/commits/d3f624bff2d3b263be3a310a5c9ef36a19075bc0

contracts/chain-adapters/l2/Arbitrum_WithdrawalAdapter.sol
```diff
@@ -0,0 +1,79 @@
...
+        uint256 amountToReturn,
+        address l2TokenAddress
+    ) public override {
+        // If the l2TokenAddress is UDSC, we need to use the CCTP bridge.
+        if (_isCCTPEnabled() && l2TokenAddress == address(usdcToken)) {
+            _transferUsdc(recipient, amountToReturn);
+        } else {
+            // Check that the Ethereum counterpart of the L2 token is stored on this contract.
+            // Tokens will only be bridged if they are whitelisted by the spoke pool.
+            address ethereumTokenToBridge = spokePool.whitelistedTokens(l2TokenAddress);
```
mrice32, line 68: I don't love the dependency on the spoke pool, especially since we may end up wanting the spoke pool to call into this contract.

But I can't really think of a way to get this list elsewhere without adding a lot of complexity or additional args that wouldn't apply on non-arbitrum chains... maybe we could add a bytes arg that can be used to supply extra context when needed, but that makes the interface sort of convoluted.

Any ideas?

mrice32 (commented):
Looks great -- just a few comments

contracts/L2_TokenRetriever.sol
```diff
@@ -0,0 +1,83 @@
...
+        if (!success) revert RetrieveFailed(l2Token);
+    }
+
+    /**
+     * @notice delegatecalls the bridge adapter to withdraw multiple different L2 tokens.
+     * @dev this is preferrable to multicalling `retrieve` since instead of `n` delegatecalls for `n`
+     * withdrawal txns, we can have 1 delegatecall for `n` withdrawal transactions.
+     * @param l2Tokens (current network's) contracts addresses of the l2 tokens to be withdrawn.
+     */
+    function retrieveMany(address[] memory l2Tokens) external nonReentrant {
```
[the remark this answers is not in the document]
bmzig: Since it's on L2, it's realistically not going to be too significant. I'm not positive about how much gas it costs to construct the `WithdrawalInformation` struct, but I can't imagine it being too much since we don't really do any expensive operations. I thought about removing the single `retrieve` function, and truthfully, I don't have a good reason to keep it in, so I'm OK with just renaming this function to `retrieve` and taking an array of tokens.

contracts/chain-adapters/l2/Arbitrum_WithdrawalAdapter.sol
```diff
@@ -0,0 +1,66 @@
...
+        address l2TokenAddress,
+        uint256 amountToReturn
+    ) public override {
+        // If the l2TokenAddress is UDSC, we need to use the CCTP bridge.
+        if (_isCCTPEnabled() && l2TokenAddress == address(usdcToken)) {
+            _transferUsdc(recipient, amountToReturn);
+        } else {
+            require(l1TokenAddress != address(0), "Uninitialized mainnet token");
+            ArbitrumL2ERC20GatewayLike tokenBridge = ArbitrumL2ERC20GatewayLike(l2Gateway);
+            require(tokenBridge.calculateL2TokenAddress(l1TokenAddress) == l2TokenAddress, "Invalid token mapping");
```
pxrl, line 56: Nice QoL improvement there!

contracts/chain-adapters/l2/Ovm_WithdrawalAdapter.sol
```diff
@@ -0,0 +1,138 @@
...
+    constructor(
+        IERC20 _l2Usdc,
+        ITokenMessenger _cctpTokenMessenger,
+        address _l2Gateway,
+        IOvm_SpokePool _spokePool
+    ) WithdrawalAdapter(_l2Usdc, _cctpTokenMessenger, _l2Gateway) {
+        spokePool = _spokePool;
+        wrappedNativeToken = spokePool.wrappedNativeToken();
+        l2Eth = spokePool.l2Eth();
+        l1Gas = spokePool.l1Gas();
```
pxrl, line 68: `l1Gas` can be updated administratively in the OVM SpokePool. wdyt about a public function that pulls the latest value into storage here?

contracts/chain-adapters/l2/Ovm_WithdrawalAdapter.sol
```diff
@@ -0,0 +1,138 @@
...
+     * @param amountToReturn amount of l2Token to send back.
+     */
+    function withdrawToken(
+        address recipient,
+        address,
+        address l2TokenAddress,
+        uint256 amountToReturn
+    ) public override {
+        // If the token being bridged is WETH then we need to first unwrap it to ETH and then send ETH over the
+        // canonical bridge. On Optimism, this is address 0xDeadDeAddeAddEAddeadDEaDDEAdDeaDDeAD0000.
```
pxrl, line 84: I think we can drop this comment (potentially also from the upstream `Ovm_SpokePool-sol contract`) `0xDeadDeAddeAddEAddeadDEaDDEAdDeaDDeAD0000` is legacy for Optimism but was deprecated as part of the Bedrock upgrade: https://docs.optimism.io/stack/smart-contracts#deprecated---legacyerc20eth

contracts/L2_TokenRetriever.sol
```diff
@@ -0,0 +1,66 @@
...
+    using SafeERC20Upgradeable for IERC20Upgradeable;
+
+    struct TokenPair {
+        address l1Token;
+        address l2Token;
+    }
+
+    // Should be set to the bridge adapter which contains the proper logic to withdraw tokens on
+    // the deployed L2
+    address public immutable bridgeAdapter;
```
nicholaspai, line 28: is "l2WithdrawalAdapter" a better name? makes it clear which chain it exists on and what its function is

contracts/L2_TokenRetriever.sol
```diff
@@ -0,0 +1,66 @@
...
+interface IBridgeAdapter {
+    function withdrawTokens(WithdrawalAdapter.WithdrawalInformation[] memory) external;
+}
+
+/**
+ * @notice Contract deployed on an arbitrary L2 to act as an intermediate contract for withdrawals from L3 to L1.
+ * @dev Since each network has its own bridging requirements, this contract delegates that logic to a bridge adapter contract
+ * which performs the necessary withdraw action.
+ */
+contract L2_TokenRetriever is Lockable {
```
nicholaspai, line 18: Can we just rename this `L2TokenRetriever`, typically there is no need for the underscore but we have made a bad pattern of this with the L1 adapters

contracts/chain-adapters/l2/Arbitrum_WithdrawalAdapter.sol
```diff
@@ -0,0 +1,66 @@
...
+     * @param l2TokenAddress address of the l2Token to send back.
+     */
+    function withdrawToken(
+        address recipient,
+        address l1TokenAddress,
+        address l2TokenAddress,
+        uint256 amountToReturn
+    ) public override {
+        // If the l2TokenAddress is UDSC, we need to use the CCTP bridge.
+        if (_isCCTPEnabled() && l2TokenAddress == address(usdcToken)) {
```
nicholaspai, line 51: are these copied from the spoke pool or did you modify anything?

contracts/chain-adapters/l2/Ovm_WithdrawalAdapter.sol
```diff
@@ -0,0 +1,138 @@
...
+     * @param _l2Usdc address of native USDC on the L2.
+     * @param _cctpTokenMessenger address of the CCTP token messenger contract on L2.
+     * @param _l2Gateway address of the Optimism ERC20 l2 standard bridge contract.
+     */
+    constructor(
+        IERC20 _l2Usdc,
+        ITokenMessenger _cctpTokenMessenger,
+        address _l2Gateway,
+        IOvm_SpokePool _spokePool
+    ) WithdrawalAdapter(_l2Usdc, _cctpTokenMessenger, _l2Gateway) {
```
nicholaspai, line 64: https://github.com/across-protocol/contracts/pull/605/files#r1773159772

contracts/chain-adapters/l2/WithdrawalAdapter.sol
```diff
@@ -0,0 +1,74 @@
...
+import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+
+/**
+ * @title Adapter for interacting with bridges from a generic L2 to Ethereum mainnet.
+ * @notice This contract is used to share L2-L1 bridging logic with other Across contracts.
+ */
+abstract contract WithdrawalAdapter is CircleCCTPAdapter {
+    using SafeERC20 for IERC20;
+
+    struct WithdrawalInformation {
```
nicholaspai, line 15: nit: prefer naming this Instruction instead of Information

contracts/chain-adapters/l2/WithdrawalAdapter.sol
```diff
@@ -0,0 +1,74 @@
...
+        l2Gateway = _l2Gateway;
+    }
+
+    /*
+     * @notice withdraws tokens to Ethereum given the input parameters.
+     * @param withdrawalInformation array containing information to withdraw a token. Includes the L1 recipient
+     * address, the amount to withdraw, and the token address of the L2 token to withdraw.
+     */
+    function withdrawTokens(WithdrawalInformation[] memory withdrawalInformation) external {
+        uint256 informationLength = withdrawalInformation.length;
```
nicholaspai, line 48: better name would be withdrawals instead of informationLength which sounds awkward

contracts/chain-adapters/l2/WithdrawalAdapter.sol
```diff
@@ -0,0 +1,74 @@
...
+    ) CircleCCTPAdapter(_l2Usdc, _cctpTokenMessenger, CircleDomainIds.Ethereum) {
+        l2Gateway = _l2Gateway;
+    }
+
+    /*
+     * @notice withdraws tokens to Ethereum given the input parameters.
+     * @param withdrawalInformation array containing information to withdraw a token. Includes the L1 recipient
+     * address, the amount to withdraw, and the token address of the L2 token to withdraw.
+     */
+    function withdrawTokens(WithdrawalInformation[] memory withdrawalInformation) external {
```
nicholaspai, line 47:
```suggestion
    function withdrawTokens(WithdrawalInformation[] memory withdrawals) external {
```

contracts/chain-adapters/l2/Ovm_WithdrawalAdapter.sol
```diff
@@ -0,0 +1,138 @@
...
+        // (i.e. non-native L2 tokens have a canonical L1 token). If we should bridge "native L2" tokens then
+        // we'd need to call bridgeERC20To and give allowance to the tokenBridge to spend l2Token from this contract.
+        // Therefore for native tokens we should set ensure that remoteL1Tokens is set for the l2TokenAddress.
+        else {
+            IL2ERC20Bridge tokenBridge = IL2ERC20Bridge(
+                spokePool.tokenBridges(l2TokenAddress) == address(0)
+                    ? Lib_PredeployAddresses.L2_STANDARD_BRIDGE
+                    : spokePool.tokenBridges(l2TokenAddress)
+            );
+            if (spokePool.remoteL1Tokens(l2TokenAddress) != address(0)) {
```
nicholaspai, line 111: prefer using spokePool.tokenBridges(l2Token) in all cases to make it easier to manage all the token mappings.

contracts/chain-adapters/l2/WithdrawalAdapter.sol
```diff
@@ -0,0 +1,74 @@
...
+
+import { CircleCCTPAdapter, ITokenMessenger, CircleDomainIds } from "../../libraries/CircleCCTPAdapter.sol";
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+
+/**
+ * @title Adapter for interacting with bridges from a generic L2 to Ethereum mainnet.
+ * @notice This contract is used to share L2-L1 bridging logic with other Across contracts.
+ */
+abstract contract WithdrawalAdapter is CircleCCTPAdapter {
```
nicholaspai, line 12: Prefer naming this with Base suffix to be consistent

contracts/chain-adapters/l2/Arbitrum_WithdrawalAdapter.sol
```diff
@@ -0,0 +1,66 @@
...
+     * @notice Calls CCTP or the Arbitrum gateway router to withdraw tokens back to the `tokenRetriever`. The
+     * bridge will not be called if the token is not in the Arbitrum_SpokePool's `whitelistedTokens` mapping.
+     * @param recipient L1 address of the recipient.
+     * @param amountToReturn amount of l2Token to send back.
+     * @param l2TokenAddress address of the l2Token to send back.
+     */
+    function withdrawToken(
+        address recipient,
+        address l1TokenAddress,
+        address l2TokenAddress,
```
nicholaspai, line 47: Can you leave a note to revisit the security implication that anyone can pass in an l2/l1/recipient here which is currently not possible when calling _bridgeTokensToHubPool because its internal

bmzig: Yeah. The issue here is that we need to pass in the l1 token to bridge to when withdrawing tokens on Arbitrum, and letting any user pass in the l1 token could possibly mean that an L2 token is withdrawn into an incorrect l1 token. Basically, I'm unsure whether an attacker could make a new Arbitrum gateway which maps an l1 token they created to an l2 token like WETH, and then perform a valid withdrawal with the l2Weth -> fake l1Token. I'm not even sure if this is a valid concern, though (although this could be a concern regardless for other networks in the future).

contracts/chain-adapters/l2/WithdrawalAdapter.sol
bmzig: https://github.com/across-protocol/contracts/pull/605/commits/546f9ffb577b490107e4a514cf076f09642855cb

contracts/chain-adapters/l2/WithdrawalAdapter.sol
bmzig: https://github.com/across-protocol/contracts/pull/605/commits/546f9ffb577b490107e4a514cf076f09642855cb

contracts/chain-adapters/l2/WithdrawalAdapter.sol
bmzig: https://github.com/across-protocol/contracts/pull/605/commits/546f9ffb577b490107e4a514cf076f09642855cb

contracts/chain-adapters/l2/Arbitrum_WithdrawalAdapter.sol
bmzig, line 56: Instead of querying whitelisted tokens from the spoke pool, I do this check.

contracts/chain-adapters/l2/Arbitrum_WithdrawalAdapter.sol
bmzig: https://github.com/across-protocol/contracts/pull/605/files#r1773965189

contracts/chain-adapters/l2/Ovm_WithdrawalAdapter.sol
bmzig: Yes. It's copied from the Spoke Pool.

contracts/chain-adapters/l2/Arbitrum_WithdrawalAdapter.sol
mrice32: The public function would be on the spoke pool? Is that different than what's happening now?

Maybe I'm missing the point.

contracts/chain-adapters/l2/Ovm_WithdrawalAdapter.sol
bmzig: Now it is queried on the spoke pool on every withdrawal https://github.com/across-protocol/contracts/pull/605/commits/4db2572d1a4b4bf7ebb81d87659f57e3afda4f0b

contracts/chain-adapters/l2/Ovm_WithdrawalAdapter.sol
```diff
@@ -0,0 +1,138 @@
...
+            IL2ERC20Bridge(Lib_PredeployAddresses.L2_STANDARD_BRIDGE).withdrawTo{ value: amountToReturn }(
+                l2TokenAddress, // _l2Token. Address of the L2 token to bridge over.
+                recipient, // _to. Withdraw, over the bridge, to the l1 pool contract.
+                amountToReturn, // _amount.
+                l1Gas, // _l1Gas. Unused, but included for potential forward compatibility considerations
+                "" // _data. We don't need to send any data for the bridging action.
+            );
+        }
+        // If the token is USDC && CCTP bridge is enabled, then bridge USDC via CCTP.
+        else if (_isCCTPEnabled() && l2TokenAddress == address(usdcToken)) {
```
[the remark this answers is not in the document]
bmzig: I mainly just copied the exact code from the spoke pool. I  can make the change still if this optimization is wanted.

contracts/L2_TokenRetriever.sol
bmzig: One of the later comments was about merging the token retriever with the adapters, so this is no longer applicable https://github.com/across-protocol/contracts/pull/605/commits/4db2572d1a4b4bf7ebb81d87659f57e3afda4f0b

contracts/chain-adapters/l2/Ovm_WithdrawalAdapter.sol
bmzig: It was factored out here https://github.com/across-protocol/contracts/pull/605/commits/4db2572d1a4b4bf7ebb81d87659f57e3afda4f0b

contracts/L2_TokenRetriever.sol
```diff
@@ -0,0 +1,66 @@
...
+    constructor(address _bridgeAdapter, address _tokenRecipient) {
+        bridgeAdapter = _bridgeAdapter;
+        tokenRecipient = _tokenRecipient;
+    }
+
+    /**
+     * @notice delegatecalls the bridge adapter to withdraw multiple different L2 tokens.
+     * @param tokenPairs l1 and l2 addresses of the token to withdraw.
+     */
+    function retrieve(TokenPair[] memory tokenPairs) external nonReentrant {
```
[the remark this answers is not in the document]
bmzig: Done in this commit https://github.com/across-protocol/contracts/pull/605/commits/4db2572d1a4b4bf7ebb81d87659f57e3afda4f0b

contracts/chain-adapters/l2/WithdrawalAdapter.sol
bmzig: That is better https://github.com/across-protocol/contracts/pull/605/commits/4db2572d1a4b4bf7ebb81d87659f57e3afda4f0b

bmzig:
`WithdrawalAdapter` name is TBD

contracts/chain-adapters/l2/WithdrawalAdapterBase.sol
```diff
@@ -0,0 +1,60 @@
...
+/**
+ * @title WithdrawalAdapterBase
+ * @notice This contract contains general configurations for bridging tokens from an L2 to a single recipient on L1.
+ * @dev This contract should be deployed on L2. It provides an interface to withdraw tokens to some address on L1. The only
+ * function which must be implemented in contracts which inherit this contract is `withdrawToken`. It is up to that function
+ * to determine which bridges to use for an input L2 token. Importantly, that function must also verify that the l2 to l1
+ * token mapping is correct so that the bridge call itself can succeed.
+ */
+abstract contract WithdrawalAdapterBase is CircleCCTPAdapter, MultiCaller {
+    using SafeERC20 for IERC20;
```
nicholaspai, line 18: Prefer naming this contract WithdrawHelper or L3WithdrawHelper as we've typically used Adapter to contain code that gets delegate called and otherwise acts as a library

contracts/chain-adapters/l2/Ovm_WithdrawalAdapter.sol
```diff
@@ -0,0 +1,154 @@
...
+
+/**
+ * @title Ovm_WithdrawalAdapter
+ * @notice This contract interfaces with L2-L1 token bridges and withdraws tokens to a single address on L1.
+ * @dev This contract should be deployed on OpStack L2s which both have a Ovm_SpokePool contract deployed to the L2
+ * network AND only use token bridges defined in the Ovm_SpokePool. A notable exception to this requirement is Optimism,
+ * which has a special SNX bridge (and thus this adapter will NOT work for Optimism).
+ * @custom:security-contact bugs@across.to
+ */
+contract Ovm_WithdrawalAdapter is WithdrawalAdapterBase {
```
nicholaspai, line 42: Should this be upgradeable? If not then is the process for upgrading this and then updating the withdrawTo address in the L3 spoke pool easy?

nicholaspai (commented):
Couple first pass comments

nicholaspai:
> `WithdrawalAdapter` name is TBD

Next time don't open PR for review until names are ready to be merged, otherwise just leave in draft mode

pxrl (commented):
One tiny comment but otherwise no blockers from me.

contracts/chain-adapters/l2/Ovm_WithdrawalAdapter.sol
pxrl: After discussing w/ @bmzig:
- We're probably OK w/o upgradability here. To change this implementation we'd need to deploy a new instance and call `setHubPool()` on the L3 Spoke, which should be fine.
- We are probably lacking a way to rescue any tokens from this contract in case of an implementation problem in `withdrawToken()`. We probably want to support an L1 -> L2 message initiated by the HubPool that instructs this contract to withdraw tokens to an L2 address. We could use this to withdraw tokens to the relevant SpokePool on the same chain and evacuate the tokens via there.

contracts/chain-adapters/l2/Arbitrum_WithdrawalAdapter.sol
```diff
@@ -0,0 +1,79 @@
...
+     * @param l2Token Address of the L2 token to send back.
+     * @param amountToReturn Amount of l2Token to send back.
+     */
+    function withdrawToken(
+        address l1Token,
+        address l2Token,
+        uint256 amountToReturn
+    ) public override {
+        // If the l2TokenAddress is UDSC, we need to use the CCTP bridge.
+        if (_isCCTPEnabled() && l2Token == address(usdcToken)) {
```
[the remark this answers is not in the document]
bmzig: True. Flipped here: https://github.com/across-protocol/contracts/pull/605/commits/da1bc9abb9f9f75e83652b7e9b5be7ba41e1dca3

contracts/chain-adapters/l2/Ovm_WithdrawalAdapter.sol
bmzig: This, along with a basic unit test, was added here https://github.com/across-protocol/contracts/pull/605/commits/da1bc9abb9f9f75e83652b7e9b5be7ba41e1dca3

contracts/chain-adapters/l2/WithdrawalAdapterBase.sol
bmzig: https://github.com/across-protocol/contracts/pull/605/commits/da1bc9abb9f9f75e83652b7e9b5be7ba41e1dca3

contracts/chain-adapters/l2/WithdrawalHelperBase.sol
```diff
@@ -35,10 +44,28 @@ abstract contract WithdrawalAdapterBase is CircleCCTPAdapter, MultiCaller {
         ITokenMessenger _cctpTokenMessenger,
         uint32 _destinationCircleDomainId,
         address _l2TokenGateway,
-        address _tokenRecipient
+        address _tokenRecipient,
+        address _hubPool
```
pxrl, line 48: I think this needs a corresponding param update in the natspec above. Also on the naming, I think the SpokePool contracts typically use `crossDomainAdmin` to specify the owner/admin and `hubPool` to specify the recipient on mainnet for withdrawals, so this seems a bit inconsistent with that in the sense that `hubPool` has permission to execute admin functions here.

pxrl (commented):
One fairly minor comment, otherwise looking good.

contracts/chain-adapters/l2/WithdrawalHelperBase.sol
```diff
@@ -0,0 +1,93 @@
...
+     * @param to Address which should receive the L2 token.
+     * @param amount Amount of the L2 token to send to `to`.
+     * @dev This function should only be callable via a rescue adapter. This function is solely for times when token withdrawals from L2-L1
+     * cannot succeed for whatever reason, so the tokens must be sent to an alternate address.
+     */
+    function sendTokenTo(
+        address l2Token,
+        address to,
+        uint256 amount
+    ) external onlyHubPool {
```
pxrl, line 67: I'm unsure, but does `external` make sense when this would be called from another contract via a delegatecall from this contract? It's a little bit inception-y so I'm not immediately sure. @bmzig do you know?

contracts/chain-adapters/l2/WithdrawalHelperBase.sol
bmzig: Good catch. I added the natspec to reflect the new parameter. https://github.com/across-protocol/contracts/pull/605/commits/2056784034360ccdda33d025d7d86b16be170728. Also, I agree with renaming `hubPool`, so I just changed the functions to match the same ones as the spoke pool (i.e. use `crossDomainAdmin`).

contracts/chain-adapters/l2/WithdrawalHelperBase.sol
bmzig: I just ran a quick test to verify that we can delegatecall these, and it looks to be alright.

contracts/chain-adapters/l2/Arbitrum_WithdrawalHelper.sol
```diff
@@ -23,8 +23,8 @@ contract Arbitrum_WithdrawalHelper is WithdrawalHelperBase {
     // Error which triggers when the supplied L1 token does not match the Arbitrum gateway router's expected L2 token.
     error InvalidTokenMapping();
 
-    modifier onlyFromHubPool() {
-        require(msg.sender == CrossDomainAddressUtils.applyL1ToL2Alias(HUB_POOL), "ONLY_HUB_POOL");
+    modifier onlyCrossDomainAdmin() {
+        require(msg.sender == CrossDomainAddressUtils.applyL1ToL2Alias(crossDomainAdmin), "ONLY_CROSS_DOMAIN_ADMIN");
```
pxrl, line 27: Should this just be relocated down to the `_requireAdminSender()` function below, rather than having it invoke this as a modifier?

contracts/chain-adapters/l2/WithdrawalHelperBase.sol
```diff
@@ -86,8 +99,19 @@ abstract contract WithdrawalHelperBase is CircleCCTPAdapter, MultiCaller {
     ) public virtual;
 
     /*
-     * @notice Checks that the L1 msg.sender is the `HUB_POOL` address.
+     * @notice Checks that the L1 msg.sender is the `CROSS_DOMAIN_ADMIN` address.
```
pxrl, line 102: ooc, why is `CROSS_DOMAIN_ADMIN` all caps?

contracts/chain-adapters/l2/Arbitrum_WithdrawalHelper.sol
bmzig: Yeah that's a good call. I took it out of the spoke pool, but it just seems like extra diff that we can cut out. Done here: https://github.com/across-protocol/contracts/pull/605/commits/167b8a9ddeff09555bb627b7a6419c1cffea36d5

contracts/chain-adapters/l2/WithdrawalHelperBase.sol
bmzig: It shouldn't be... This was an artifact of before it was a proxy. Fixed here https://github.com/across-protocol/contracts/pull/605/commits/167b8a9ddeff09555bb627b7a6419c1cffea36d5

pxrl (approved):

mrice32 (approved):
LGTM!

PR #607, opened 2024-09-16
feat: passthrough L1->L3 adapter to send messages to L3 via an L2 forwarder

bmzig:
The L3 adapter to send messages to Arbitrum-like L3 spoke pools should piggyback off of our existing adapters to first relay tokens/messages to an intermediate contract on L2. These L2 forwarders should then have logic to continue the token bridging or to execute stuck messages.


contracts/chain-adapters/Arbitrum_L3_Adapter.sol
```diff
@@ -0,0 +1,64 @@
...
+    error RelayTokensFailed(address l1Token);
+
+    /**
+     * @notice Constructs new Adapter for sending tokens/messages to Arbitrum-like L3s.
+     * @param _l2Adapter Address of the adapter contract on mainnet which implements message transfers
+     * and token relays.
+     */
+    constructor(address _l2Adapter, address _l2Forwarder) {
+        l2Adapter = _l2Adapter;
+        l2Forwarder = _l2Forwarder;
```
pxrl, line 30: The naming of `l2Adapter` and `l2Forwarder` is a bit counter-intuitive because one is on the L1 and the other is on the L2. In the context of the chain adapters themselves, they're otherwise just called `adapter`; should we maybe follow that?

contracts/chain-adapters/Arbitrum_L3_Adapter.sol
```diff
@@ -0,0 +1,64 @@
...
+    /**
+     * @notice Bridge tokens to an Arbitrum-like L3, using an L2 forwarder.
+     * @param l1Token L1 token to deposit.
+     * @param l2Token L2 token to receive.
+     * @param amount Amount of L1 tokens to deposit and L2 tokens to receive.
+     * @dev we discard the "to" field since tokens are always sent to the l2Forwarder.
+     */
+    function relayTokens(
+        address l1Token,
+        address l2Token, // l2Token is unused for Arbitrum.
```
pxrl, line 55: l2Token is emitted within an event by the Arbitrum adapter, but I'd also just look at it as an implementation detail of a downstream contract that doesn't need to be documented here :+1: 

```suggestion
        address l2Token,
```

bmzig: Very true. https://github.com/across-protocol/contracts/pull/607/commits/2daf57b25a2cff090b1feecbb4eadab1a707a81f

contracts/chain-adapters/Arbitrum_L3_Adapter.sol
bmzig: I think that's reasonable. https://github.com/across-protocol/contracts/pull/607/commits/2daf57b25a2cff090b1feecbb4eadab1a707a81f

contracts/chain-adapters/Arbitrum_L3_Adapter.sol
```diff
@@ -0,0 +1,63 @@
...
+        l2Forwarder = _l2Forwarder;
+    }
+
+    /**
+     * @notice Send cross-chain message to target on L2, which is forwarded to the Arbitrum-like L3.
+     * @dev there is a bijective mapping of L3 adapters (on L1) to L2 forwarders to L3 spoke pools. The
+     * spoke pool address is stored by the L2 forwarder and the L2 forwarder address is stored in this contract.
+     * @param message Data to send to target.
+     */
+    function relayMessage(address, bytes memory message) external payable override {
```
mrice32, line 40: OOC, why did you decide to do this via delegatecall rather than inheritance? To avoid having to have a different contract for an L2 with custom gas vs non custom gas?

mrice32 (commented):
Looks great -- just one comment

contracts/chain-adapters/Arbitrum_L3_Adapter.sol
```diff
@@ -0,0 +1,63 @@
...
+ * @notice Contract containing logic to send messages from L1 to Arbitrum-like L3s using an intermediate L2 message forwarder.
+ * @notice This contract requires an L2 forwarder contract to be deployed, since we overwrite the target field to this new target.
+ * @dev Public functions calling external contracts do not guard against reentrancy because they are expected to be
+ * called via delegatecall, which will execute this contract's logic within the context of the originating contract.
+ * For example, the HubPool will delegatecall these functions, therefore its only necessary that the HubPool's methods
+ * that call this contract's logic guard against reentrancy.
+ */
+
+// solhint-disable-next-line contract-name-camelcase
+contract Arbitrum_L3_Adapter is AdapterInterface {
```
mrice32, line 16: I don't think this contract is specific to arbitrum, right? It's really just an adapter for _any_ forwarder, no?

bmzig: This is likely due to the naming, but I did this to emphasize that the L3 is Arbitrum-like. There is no assumption about the structure of the L2.

pxrl: The naming is still a sticking point for me. I think it makes sense to spend some more time thinking about this. This contract actually functions more like a (re-)router I guess, but maybe that's just my networking bias.

nicholaspai: I agree this naming is not useful right now. Because it has "L3" in it its non-intuitive that its actually meant to be deployed on L1. Let's also remove the Arbitrum mention from the name because its clearly not specific to Arbitrum. Please change all the comments also to make it clear this can be used generically.

I think a rerouter in the name could work, and clearly to be consistent with other L1 contracts it should end with Adapter. What about RerouterAdapter?

@pxrl

contracts/chain-adapters/Arbitrum_L3_Adapter.sol
```diff
@@ -0,0 +1,63 @@
...
+    address public immutable adapter;
```
nicholaspai, line 17: nit: Would prefer explicitly naming this `l1Adapter` since we also name the target the `l2Forwarder`.

bmzig: It's now `l1Adapter` https://github.com/across-protocol/contracts/pull/607/commits/1e984cc52ee30cb74196517e19e36ef7d39abede

contracts/chain-adapters/Arbitrum_L3_Adapter.sol
bmzig: Optimistically changing it to `RerouterAdapter` and updating the comments to emphasize that it can be used in other contexts.

contracts/chain-adapters/Rerouter_Adapter.sol
```diff
@@ -0,0 +1,64 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.0;
+
+import { AdapterInterface } from "./interfaces/AdapterInterface.sol";
+
+/**
+ * @notice Contract containing logic to send messages from L1 to a target (not necessarily a spoke pool) on L2.
```
nicholaspai, line 7: Would be useful to note that one of these adapters should be deployed per L3. Also this adapter be mapped via HubPool to an L3 chain and actually, the L3 spoke pool kind of gets thrown away.

I think we should separately a commit an architecture diagram somewhere in the repo. Maybe something like [mermaid](https://github.blog/developer-skills/github/include-diagrams-markdown-files-mermaid/)

bmzig: Comment updated: https://github.com/across-protocol/contracts/pull/607/commits/cc386a73aeb64319c413c171e677b2465502f431

Also, diagramming something sounds good to me :+1:

contracts/chain-adapters/Rerouter_Adapter.sol
```diff
@@ -0,0 +1,64 @@
...
+ * "relaySpokePoolAdminFunction".
+ * @dev Public functions calling external contracts do not guard against reentrancy because they are expected to be
+ * called via delegatecall, which will execute this contract's logic within the context of the originating contract.
+ * For example, the HubPool will delegatecall these functions, therefore its only necessary that the HubPool's methods
+ * that call this contract's logic guard against reentrancy.
+ * @custom:security-contact bugs@across.to
+ */
+
+// solhint-disable-next-line contract-name-camelcase
+contract Rerouter_Adapter is AdapterInterface {
```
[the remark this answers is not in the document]
bmzig: Should we rename it? I personally think that all the L1/L2/L3 to some extent adds to the naming confusion.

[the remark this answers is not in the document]
nicholaspai: its ok we can leave it for now

contracts/chain-adapters/Rerouter_Adapter.sol
```diff
@@ -0,0 +1,71 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.0;
+
+import { AdapterInterface } from "./interfaces/AdapterInterface.sol";
+
+/**
+ * @notice Contract containing logic to send messages from L1 to a target (not necessarily a spoke pool) on L2.
```
nicholaspai, line 7: what do you mean by "not neccessarily a spoke pool?"

contracts/chain-adapters/Rerouter_Adapter.sol
```diff
@@ -0,0 +1,71 @@
...
+ * @notice Since this adapter is normally called by the hub pool, the target of both `relayMessage` and `relayTokens`
+ * will be the L3 spoke pool due to the constraints of `setCrossChainContracts` outlined in UMIP 157. However, this
+ * contract DOES NOT send anything to the L2 containing info on the target L3 spoke pool. The L3 spoke pool address
+ * must instead be initialized on the `l2Target` contract as the same spoke pool address found in the hub pool's
+ * `crossChainContracts` mapping.
+ * @notice There should be one of these adapters for each L3 spoke pool deployment, or equivalently, each L2
```
nicholaspai, line 13:
```suggestion
 * @dev There should be one of these adapters for each L3 spoke pool deployment, or equivalently, each L2
```

contracts/chain-adapters/Rerouter_Adapter.sol
```diff
@@ -0,0 +1,71 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.0;
+
+import { AdapterInterface } from "./interfaces/AdapterInterface.sol";
+
+/**
+ * @notice Contract containing logic to send messages from L1 to a target (not necessarily a spoke pool) on L2.
+ * @notice Since this adapter is normally called by the hub pool, the target of both `relayMessage` and `relayTokens`
+ * will be the L3 spoke pool due to the constraints of `setCrossChainContracts` outlined in UMIP 157. However, this
+ * contract DOES NOT send anything to the L2 containing info on the target L3 spoke pool. The L3 spoke pool address
```
nicholaspai, line 10:
```suggestion
 * contract DOES NOT send anything directly to the L3 target. Instead, it "re-routes" messages to the L3 via an L2 contract which is set as the `l2Target` in this contract. The L3 spoke pool address
```

contracts/chain-adapters/Rerouter_Adapter.sol
```diff
@@ -0,0 +1,71 @@
...
+ * must instead be initialized on the `l2Target` contract as the same spoke pool address found in the hub pool's
```
nicholaspai, line 11:
```suggestion
 * must be initialized on the `l2Target` contract to the same L3 spoke pool address found in the hub pool's
```

contracts/chain-adapters/Rerouter_Adapter.sol
```diff
@@ -0,0 +1,71 @@
...
+ * `crossChainContracts` mapping.
+ * @notice There should be one of these adapters for each L3 spoke pool deployment, or equivalently, each L2
+ * forwarder/adapter contract.
```
nicholaspai, line 14:
```suggestion
```

contracts/chain-adapters/Rerouter_Adapter.sol
```diff
@@ -0,0 +1,71 @@
...
+ * @notice The contract receiving messages on L2 will be "spoke pool like" functions, e.g. "relayRootBundle" and
```
nicholaspai, line 15:
```suggestion
 * @dev The contract receiving messages on L2 will be "spoke pool like" functions, e.g. "relayRootBundle" and
```

Also, what is a "spoke pool like" function that doesn't make any sense

contracts/chain-adapters/Rerouter_Adapter.sol
```diff
@@ -0,0 +1,71 @@
...
+ * "relaySpokePoolAdminFunction".
+ * @dev Public functions calling external contracts do not guard against reentrancy because they are expected to be
```
nicholaspai, line 17: note: there should not be too many @dev comments. Make sure each @dev comment makes a single point. If two dev comments cover the same concepts, consolidate them.

contracts/chain-adapters/Rerouter_Adapter.sol
```diff
@@ -0,0 +1,71 @@
...
+    address public immutable l1Adapter;
+    address public immutable l2Target;
+
+    error RelayMessageFailed();
+    error RelayTokensFailed(address l1Token);
+
+    /**
+     * @notice Constructs new Adapter for sending tokens/messages to an L2 target.
+     * @param _l1Adapter Address of the adapter contract on mainnet which implements message transfers
+     * and token relays.
```
nicholaspai, line 35:
```suggestion
     * and token relays. This contract will re-route messages to the _l2Target via the _l1Adapter.
```

contracts/chain-adapters/Rerouter_Adapter.sol
nicholaspai: Also, shouldn't this @notice explain why this contract is called a "re-router"? (because it sends messages from L1 to an L3 target by re-routing via an L2 contract.

contracts/chain-adapters/Rerouter_Adapter.sol
```diff
@@ -0,0 +1,71 @@
...
+     * @param _l2Target Address of the L2 contract which receives the token and message relays.
+     */
+    constructor(address _l1Adapter, address _l2Target) {
+        l1Adapter = _l1Adapter;
+        l2Target = _l2Target;
+    }
+
+    /**
+     * @notice Send cross-chain message to a target on L2.
```
nicholaspai, line 44:
```suggestion
     * @notice Send cross-chain message to a target on L2 which will re-route messages to the intended L3 target.
```

test/evm/foundry/local/Arbitrum_L3_Adapter.t.sol
```diff
@@ -0,0 +1,141 @@
...
+    function sendMessage(
+        address target,
+        bytes calldata,
+        uint32
+    ) external {
+        emit MessageSent(target);
+    }
+}
+
+contract StandardBridge {
```
nicholaspai, line 56: None of these test contracts are used in other tests? If so we should to try import them.

test/evm/foundry/local/Arbitrum_L3_Adapter.t.sol
```diff
@@ -0,0 +1,141 @@
...
+            ITokenMessenger(address(0))
+        );
+        l3Adapter = new Mock_Rerouter_Adapter(address(optimismAdapter), l2Target);
+    }
+
+    // Messages should be indiscriminately sent to the l2Forwarder.
+    function testRelayMessage(address target, bytes memory message) public {
+        vm.expectEmit(address(crossDomainMessenger));
+        emit CrossDomainMessenger.MessageSent(l2Target);
+        l3Adapter.relayMessage(target, message);
```
nicholaspai, line 118: can you assert l2Target != target to make sure no false positive?

nicholaspai (commented):
I think the code looks great, but the comments need a lot of cleaning up

test/evm/foundry/local/Arbitrum_L3_Adapter.t.sol
bmzig: Added vm.assume to enforce that they aren't equal. Then the expectEmit should do the checking that the correct argument was passed into the bridge internally. https://github.com/across-protocol/contracts/pull/607/commits/6d8a6339a81368fc47271c89581408d5d8b94f1b

test/evm/foundry/local/Arbitrum_L3_Adapter.t.sol
bmzig: Refactored https://github.com/across-protocol/contracts/pull/607/commits/6d8a6339a81368fc47271c89581408d5d8b94f1b

contracts/chain-adapters/Rerouter_Adapter.sol
bmzig: https://github.com/across-protocol/contracts/pull/607/commits/6d8a6339a81368fc47271c89581408d5d8b94f1b

contracts/chain-adapters/Rerouter_Adapter.sol
bmzig: https://github.com/across-protocol/contracts/pull/607/commits/6d8a6339a81368fc47271c89581408d5d8b94f1b

contracts/chain-adapters/Rerouter_Adapter.sol
bmzig: I meant a spoke pool function. It's removed https://github.com/across-protocol/contracts/pull/607/commits/6d8a6339a81368fc47271c89581408d5d8b94f1b

contracts/chain-adapters/Rerouter_Adapter.sol
bmzig: https://github.com/across-protocol/contracts/pull/607/commits/6d8a6339a81368fc47271c89581408d5d8b94f1b

contracts/chain-adapters/Rerouter_Adapter.sol
bmzig: https://github.com/across-protocol/contracts/pull/607/commits/6d8a6339a81368fc47271c89581408d5d8b94f1b

contracts/chain-adapters/Rerouter_Adapter.sol
bmzig: https://github.com/across-protocol/contracts/pull/607/commits/6d8a6339a81368fc47271c89581408d5d8b94f1b

contracts/chain-adapters/Rerouter_Adapter.sol
bmzig: https://github.com/across-protocol/contracts/pull/607/commits/6d8a6339a81368fc47271c89581408d5d8b94f1b

contracts/chain-adapters/Rerouter_Adapter.sol
bmzig: Addressed by other comments and this commit https://github.com/across-protocol/contracts/pull/607/commits/6d8a6339a81368fc47271c89581408d5d8b94f1b

test/evm/foundry/local/Arbitrum_L3_Adapter.t.sol
```diff
@@ -0,0 +1,113 @@
...
+import { IERC20Upgradeable } from "@openzeppelin/contracts-upgradeable/token/ERC20/IERC20Upgradeable.sol";
+import { IL1StandardBridge } from "@eth-optimism/contracts/L1/messaging/IL1StandardBridge.sol";
+import { Rerouter_Adapter } from "../../../../contracts/chain-adapters/Rerouter_Adapter.sol";
+import { Optimism_Adapter } from "../../../../contracts/chain-adapters/Optimism_Adapter.sol";
+import { WETH9Interface } from "../../../../contracts/external/interfaces/WETH9Interface.sol";
+import { WETH9 } from "../../../../contracts/external/WETH9.sol";
+import { ITokenMessenger } from "../../../../contracts/external/interfaces/CCTPInterfaces.sol";
+import { MockBedrockL1StandardBridge, MockBedrockCrossDomainMessenger } from "../../../../contracts/test/MockBedrockStandardBridge.sol";
+
+import "forge-std/console.sol";
```
nicholaspai, line 16: left over console

bmzig: https://github.com/across-protocol/contracts/pull/607/commits/4076b6bd58ad6f16a303f549118ee03301a6a0b1

contracts/chain-adapters/Rerouter_Adapter.sol
```diff
@@ -0,0 +1,69 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.0;
+
+import { AdapterInterface } from "./interfaces/AdapterInterface.sol";
+
+/**
+ * @notice Contract containing logic to send messages from L1 to a target on L2.
```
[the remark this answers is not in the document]
bmzig: https://github.com/across-protocol/contracts/pull/607/commits/4076b6bd58ad6f16a303f549118ee03301a6a0b1

nicholaspai (approved):

contracts/chain-adapters/Rerouter_Adapter.sol
```diff
@@ -0,0 +1,73 @@
...
+        l2Target = _l2Target;
+    }
+
+    /**
+     * @notice Send cross-chain message to a target on L2 which will re-route messages to the intended L3 target.
+     * @param target Address of the L3 contract which receives `message` after it has been forwarded on L2.
+     * @param message Data to send to target.
+     */
+    function relayMessage(address target, bytes memory message) external payable override {
+        bytes memory updatedMessage = abi.encodeCall(AdapterInterface.relayMessage, (target, message));
```
nicholaspai, line 48: Can you add some comments explaining how this double relayMessage encoding works? Think its a bit confusing and assumes knowledge of the l2Target's interface

bmzig: I added this https://github.com/across-protocol/contracts/pull/607/commits/d87d09c0dc549bd4451bae0557c0e09bca1d95d6. It definitely assumes knowledge of the l2Target's interface.

nicholaspai (approved):

contracts/chain-adapters/Rerouter_Adapter.sol
```diff
@@ -0,0 +1,87 @@
...
+ * @dev Public functions calling external contracts do not guard against reentrancy because they are expected to be
+ * called via delegatecall, which will execute this contract's logic within the context of the originating contract.
+ * For example, the HubPool will delegatecall these functions, therefore its only necessary that the HubPool's methods
+ * that call this contract's logic guard against reentrancy.
+ * @custom:security-contact bugs@across.to
+ */
+
+// solhint-disable-next-line contract-name-camelcase
+contract Rerouter_Adapter is AdapterInterface {
+    address public immutable l1Adapter;
```
nicholaspai, line 26: nit: immutable and constant variables should be in ALL_CAPS

bmzig: https://github.com/across-protocol/contracts/pull/607/commits/1274920a219caf9f9fc492f6dac6e067a979c67d

contracts/chain-adapters/Rerouter_Adapter.sol
```diff
@@ -0,0 +1,89 @@
...
+
+import { AdapterInterface } from "./interfaces/AdapterInterface.sol";
+
+/**
+ * @notice Contract containing logic to send messages from L1 to non-spoke pool L2 targets via re-routing message recipients
+ * before submission. This ultimately enables sending messages to targets on arbitrary layers via forwarding through intermediate
+ * contracts.
+ * @dev Since this adapter is normally called by the hub pool, the target of both `relayMessage` and `relayTokens`
+ * will be the remote spoke pool due to the constraints of `setCrossChainContracts` outlined in UMIP 157. However, this
+ * contract cannot send anything directly to the this target, since it does not exist on L2. Instead, it "re-routes"
```
nicholaspai, line 12:
```suggestion
 * contract cannot send anything directly to this target since it does not exist on L2. Instead, it "re-routes"
```

nicholaspai (approved):
Looks good from my end

contracts/chain-adapters/Rerouter_Adapter.sol
```diff
@@ -0,0 +1,90 @@
...
+        address target
+    ) external payable override {
+        // Relay tokens to the forwarder.
+        (bool success, ) = L1_ADAPTER.delegatecall(
+            abi.encodeCall(AdapterInterface.relayTokens, (l1Token, l2Token, amount, L2_TARGET))
+        );
+        if (!success) revert RelayTokensFailed(l1Token);
+
+        // Follow-up token relay with a message to continue the token relay on L2.
+        bytes memory message = abi.encodeCall(AdapterInterface.relayTokens, (l1Token, l2Token, amount, target));
```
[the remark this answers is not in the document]
bmzig: The reason I pass in l2 token and l3 token and not something else is so that each forwarder has the same logic. Suppose we were relaying messages to l4. If we passed in l2 and l3, then the l2 forwarder only wouldn't need to calculate the l3 token (since it was given to it), but then the l3 forwarder _would_ need to calculate the l4 token. So not storing this makes the behavior of the forwarders more general.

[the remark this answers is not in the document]
bmzig: I would imagine doing it this way would be helpful in scenarios where we can determine the next layer's token by calling a bridge contract, for example. Passing in the l2 and l3 token is useful if this adapter is only being used for L3 spoke pools.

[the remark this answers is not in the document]
bmzig: Updated here: https://github.com/across-protocol/contracts/pull/607/commits/1b1ceb5a2a2edfbd42e6614dd77f13353978405e

nicholaspai (approved):

pxrl (approved):
Very minor nit on naming: "Rerouter" is a mouthfull; "Router" is easier to communicate.

PR #609, opened 2024-09-17
feat: generic L2 forwarder base contract

bmzig:
This PR adds the abstract contract `ForwarderBase` which is meant to be a base for forwarders. Forwarders should relay messages/tokens from an L2 to an L3. A forwarder which inherits `ForwarderBase` will have to be constructed for each chain to account for an L2's message validation logic, but since `ForwarderBase` makes use of L2 adapters, it shouldn't matter what the architecture of the L3 is. To be clear, only `_requireAdminSender`, an internal function which ensures that the caller is the aliased hub pool, must be implemented for each new forwarder.


contracts/chain-adapters/interfaces/ArbitrumForwarderInterface.sol
```diff
@@ -0,0 +1,254 @@
...
+    /**
+     * @notice This function can only be called via a rescue adapter, and is used to recover potentially stuck
+     * funds on this contract.
+     */
+    function rescue(
+        address target,
+        uint256 value,
+        bytes memory message
+    ) external onlyAdmin {
+        (bool success, ) = target.call{ value: value }(message);
```
bmzig, line 217: There are a bunch of ways to rescue tokens. Should this be kept like this, or should we delegatecall to a target, or should we just withdraw the tokens back to the specified address, or something else?

contracts/chain-adapters/interfaces/ArbitrumForwarderInterface.sol
```diff
@@ -0,0 +1,254 @@
...
+    // Amount of gas token allocated to pay for the base submission fee. The base submission fee is a parameter unique to
+    // retryable transactions; the user is charged the base submission fee to cover the storage costs of keeping their
+    // ticket’s calldata in the retry buffer. (current base submission fee is queryable via
+    // ArbRetryableTx.getSubmissionPrice). ArbRetryableTicket precompile interface exists at L2 address
+    // 0x000000000000000000000000000000000000006E.
+    // @dev This is immutable because we don't know what precision the custom gas token has.
+    uint256 public immutable L3_MAX_SUBMISSION_COST;
+
+    // L3 Gas price bid for immediate L3 execution attempt (queryable via standard eth*gasPrice RPC)
+    uint256 public constant L3_GAS_PRICE = 5e9; // 5 gWei
```
bmzig, line 127: Leaving a comment here since I will have to check whether gas prices across L3s are consistent.

contracts/chain-adapters/interfaces/ArbitrumForwarderInterface.sol
bmzig, line 1: For the most part, the first parts of this file are interfaces/definitions shared by both the Arbitrum_CustomGasToken_Adapter and the Arbitrum_Adapter contracts.

contracts/chain-adapters/interfaces/ArbitrumForwarderInterface.sol
```diff
@@ -0,0 +1,285 @@
...
+ * used as native currency on L3.
+ * @dev Fees are paid in this token. There are certain restrictions on the native token:
+ *       - The token can't be rebasing or have a transfer fee
+ *       - The token must only be transferrable via a call to the token address itself
+ *       - The token must only be able to set allowance via a call to the token address itself
+ *       - The token must not have a callback on transfer, and more generally a user must not be able to make a transfer to themselves revert
+ *       - The token must have a max of 2^256 - 1 wei total supply unscaled
+ *       - The token must have a max of 2^256 - 1 wei total supply when scaled to 18 decimals
+ */
+interface ArbitrumERC20Bridge {
```
mrice32, line 16: nit: we probably want to move these arb interfaces to contracts/external/interfaces, since they are used in the Arbitrum adapter as well. Keeps files succinct, reduces chances of typos, etc

contracts/chain-adapters/interfaces/ArbitrumForwarderInterface.sol
mrice32: Probably a good idea to make these immutable and set in the constructor, so we can change their values on a chain-by-chain basis without any solidity modifications.

We can still leave the suggested default as a comment.

contracts/chain-adapters/interfaces/ArbitrumForwarderInterface.sol
mrice32: The function, as built, can do far more than rescue tokens. It can perform any action on behalf of the contract. We should probably be explicit about that in the naming. Something more like `adminCall` or similar.

Relatedly: there are reasons we might want a function that gives the admin broad power like this. For instance, if the L2->L3 canonical bridge interface changes underneath us and this contract is non-upgradable (something we should discuss), it could brick the L3 SpokePool without a function like this that can be used as an escape-hatch to make an arbitrary call.

We should decide whether we want the escape hatch to be upgradability OR an arbitrary call admin method like this.

Thoughts?

mrice32 (commented):
Looks great! A few comments

contracts/chain-adapters/interfaces/ArbitrumForwarderInterface.sol
```diff
@@ -0,0 +1,285 @@
...
+
+/**
+ * @notice Contract containing logic to send messages from L2 to Arbitrum-like L3s.
+ * @dev This contract is meant to share code for Arbitrum L2 forwarder contracts deployed to various
+ * different L2 architectures (e.g. Base, Arbitrum, ZkSync, etc.). It assumes that the L3 conforms
+ * to an Arbitrum-like interface.
+ */
+
+// solhint-disable-next-line contract-name-camelcase
+abstract contract ArbitrumForwarderInterface {
```
mrice32, line 144: nit: `Interface` typically implies that no functions are implemented.

`ArbitrumForwarderBase` or just `ArbitrumForwarder` might be better.

bmzig: I like `ArbitrumForwarderBase` https://github.com/across-protocol/contracts/pull/609/commits/0de93033f590bd7886fae198bf05dc4b36e9732e

contracts/chain-adapters/interfaces/ArbitrumForwarderInterface.sol
bmzig: Done here. https://github.com/across-protocol/contracts/pull/609/commits/0de93033f590bd7886fae198bf05dc4b36e9732e

contracts/chain-adapters/interfaces/ArbitrumForwarderInterface.sol
bmzig: That'd definitely better https://github.com/across-protocol/contracts/pull/609/commits/0de93033f590bd7886fae198bf05dc4b36e9732e

contracts/chain-adapters/interfaces/ArbitrumForwarderInterface.sol
bmzig: Made into a proxy here: https://github.com/across-protocol/contracts/pull/609/commits/f2e3c41f3a0f78ba94d2ee2dc749a782e54512e0

contracts/chain-adapters/ForwarderBase.sol
```diff
@@ -0,0 +1,119 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.0;
+
+import { UUPSUpgradeable } from "@openzeppelin/contracts-upgradeable/proxy/utils/UUPSUpgradeable.sol";
+import { EIP712CrossChainUpgradeable } from "../upgradeable/EIP712CrossChainUpgradeable.sol";
+
+abstract contract ForwarderBase is UUPSUpgradeable, EIP712CrossChainUpgradeable {
```
nicholaspai, line 7: How will this contract be used? I'm not sure I understand. I can't really review this contract without understanding its context

contracts/chain-adapters/ForwarderBase.sol
```diff
@@ -0,0 +1,119 @@
...
+    /**
+     * @notice Initializes the forwarder contract.
+     * @param _l3SpokePool L3 address of the contract which will receive messages and tokens which are temporarily
+     * stored in this contract on L2.
+     * @param _crossDomainAdmin L1 address of the contract which can send root bundles/messages to this forwarder contract.
+     * In practice, this is the hub pool.
+     */
+    function __Forwarder_init(address _l3SpokePool, address _crossDomainAdmin) public onlyInitializing {
+        __UUPSUpgradeable_init();
+        __EIP712_init("ACROSS-V2", "1.0.0");
```
pxrl, line 49: What's the use of EIP-712 here? We have it in the SpokePool to permit depositors to update their deposit details but I don't see how it's related to transport of tokens and messages to an L3.

contracts/chain-adapters/Lisk_Adapter.sol
pxrl, line 1: I've seen this change pop up in a couple of PRs; should we just pull it out to a standalone PR to address it in one change?

contracts/interfaces/ArbitrumBridgeInterfaces.sol
pxrl, line 1: This is a sensible change to make in its own right, since it shows that we have multiple interface definitions for the same bridge contracts. IMO it could also be pulled out to a separate PR to reduce the scope of this PR, which is really intended to add new functionality for L3s.

contracts/chain-adapters/ForwarderBase.sol
bmzig: I created it so it could act like an interface to be implemented by specific L2 forwarders (i.e. every forwarder should have the `relayTokens`/`_requireAdminSender` implemented depending on the L2/L3 format). There are also some functions which should be common in all forwarders, like the fallback function and the proxy access control, so it made sense to factor those out into another (abstract) contract.

contracts/chain-adapters/ForwarderBase.sol
bmzig: Ah I didn't realize that is all it was used for: https://github.com/across-protocol/contracts/pull/609/commits/fe66b780fa9c9e63e79f15fe7827bd2c6b86e3bb

contracts/chain-adapters/ForwarderBase.sol
```diff
@@ -0,0 +1,117 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.0;
+
+import { UUPSUpgradeable } from "@openzeppelin/contracts-upgradeable/proxy/utils/UUPSUpgradeable.sol";
+
+abstract contract ForwarderBase is UUPSUpgradeable {
```
nicholaspai, line 6: Needs natspec comments to help reviewer understand how this contract will be used

contracts/chain-adapters/ForwarderBase.sol
```diff
@@ -0,0 +1,117 @@
...
+    // L3 address of the recipient of fallback messages and tokens.
```
nicholaspai, line 7: the l3 address is the recipient of fallback messages? do you mean relayed l1 to l2 messages?

bmzig: Yes. I changed the comment https://github.com/across-protocol/contracts/pull/609/commits/814498b0feb17b83088e27eaba26306cd0ac0108

contracts/chain-adapters/ForwarderBase.sol
bmzig: https://github.com/across-protocol/contracts/pull/609/commits/814498b0feb17b83088e27eaba26306cd0ac0108

contracts/chain-adapters/ForwarderBase.sol
```diff
@@ -0,0 +1,125 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.0;
+
+import { UUPSUpgradeable } from "@openzeppelin/contracts-upgradeable/proxy/utils/UUPSUpgradeable.sol";
+
+/**
+ * @title ForwarderBase
+ * @notice Base contract deployed some network between L1 and the network housing a spoke pool contract. If a message is sent
```
nicholaspai, line 8: Might be helpful to note that messages received on this contract are expected to be sent from the `rerouter_adapter` or w.e. we decide to name it

contracts/chain-adapters/ForwarderBase.sol
```diff
@@ -0,0 +1,125 @@
...
+ * to this contract which came from the cross domain admin, then it is routed to the next network via the canonical messaging
+ * bridge (which may be done automatically by the network's finalization infrastructure). Tokens sent from L1 are stored temporarily
+ * on this contract. Any EOA can initiate a bridge of these tokens to the target `l3SpokePool`.
+ * @custom:security-contact bugs@across.to
+ */
+abstract contract ForwarderBase is UUPSUpgradeable {
+    // L3 address of the recipient of L1 messages and tokens.
+    address public l3SpokePool;
```
nicholaspai, line 16: We might want to note either here or in the rerouter_adapter that this l3 spoke pool address i only used in the contracts IN THIS CONTRACT, and its essentially useless in the `rerouter_adapter`. However, the developer/admin should still set the CrossChainContracts correctly including the correct L3 spoke pool address because the UMIP prescribes that this event contains the canonical mapping of chain Id's to spoke pool addresses

bmzig: I think it makes more sense to put that in the adapter contract: https://github.com/across-protocol/contracts/pull/607/commits/6cbbfd48618fff81f6027f9a88cc632757c1879c

contracts/chain-adapters/ForwarderBase.sol
```diff
@@ -0,0 +1,127 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.0;
+
+import { UUPSUpgradeable } from "@openzeppelin/contracts-upgradeable/proxy/utils/UUPSUpgradeable.sol";
+
+/**
+ * @title ForwarderBase
+ * @notice This contract expects to receive messages and tokens from the Hub Pool on L1. It rejects messages which do not originate
+ * from the hub pool.
```
nicholaspai, line 9:
```suggestion
 * from the cross domain admin, which should be set to the hub pool.
```

contracts/chain-adapters/ForwarderBase.sol
```diff
@@ -0,0 +1,127 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.0;
+
+import { UUPSUpgradeable } from "@openzeppelin/contracts-upgradeable/proxy/utils/UUPSUpgradeable.sol";
+
+/**
+ * @title ForwarderBase
+ * @notice This contract expects to receive messages and tokens from the Hub Pool on L1. It rejects messages which do not originate
```
nicholaspai, line 8:
```suggestion
 * @notice This contract expects to receive messages and tokens from the Hub Pool on L1 and forwards them to targets on L3. Messages are intended to originate from the Rerouter adapter on L1 which will send messages here to re-route them to their expected targets on L3. It rejects messages which do not originate
```

contracts/chain-adapters/ForwarderBase.sol
```diff
@@ -0,0 +1,127 @@
...
+ * from the hub pool.
+ * @notice Base contract deployed some network between L1 and the network housing a spoke pool contract. If a message is sent
```
nicholaspai, line 10:
```suggestion
 * @dev Base contract designed to be deployed on L2 to re-route messages from L1 to L3. If a message is sent
```

contracts/chain-adapters/ForwarderBase.sol
```diff
@@ -0,0 +1,127 @@
...
+ * to this contract which came from the cross domain admin, then it is routed to the next network via the canonical messaging
```
nicholaspai, line 11:
```suggestion
 * to this contract which came from the L1 cross domain admin, then it is routed to L3 via the canonical messaging
```

contracts/chain-adapters/ForwarderBase.sol
```diff
@@ -0,0 +1,127 @@
...
+ * bridge (which may be done automatically by the network's finalization infrastructure). Tokens sent from L1 are stored temporarily
```
nicholaspai, line 12:
```suggestion
 * bridge. Tokens sent from L1 are stored temporarily
```

contracts/chain-adapters/ForwarderBase.sol
```diff
@@ -0,0 +1,127 @@
...
+ * on this contract. Any EOA can initiate a bridge of these tokens to the target `l3SpokePool`.
+ * @custom:security-contact bugs@across.to
+ */
+abstract contract ForwarderBase is UUPSUpgradeable {
+    // L3 address of the recipient of L1 messages and tokens.
```
nicholaspai, line 17:
```suggestion
    // L3 address of the recipient of L1 messages sent via the re-router adapter on L1 and tokens.
```

contracts/chain-adapters/ForwarderBase.sol
```diff
@@ -0,0 +1,127 @@
...
+
+    /**
+     * @notice Initializes the forwarder contract.
+     * @param _l3SpokePool L3 address of the contract which will receive messages and tokens which are temporarily
+     * stored in this contract on L2.
+     * @param _crossDomainAdmin L1 address of the contract which can send root bundles/messages to this forwarder contract.
+     * In practice, this is the hub pool.
+     */
+    function __Forwarder_init(address _l3SpokePool, address _crossDomainAdmin) public onlyInitializing {
+        __UUPSUpgradeable_init();
```
nicholaspai, line 57: Should this contract be upgradeable? I think so as it could be helpful to re-deploy updates and not have to update the RerouterAdapter to point to new forwarder contracts, but I want to make sure we've carefully considered why we are making this upgradeable. I assume you made it upgradeable by default because you copied the SpokePool's code.

nicholaspai (commented):
comments

contracts/chain-adapters/ForwarderBase.sol
bmzig: designed to be deployed on L2 to re-route messages from L1 to L3

contracts/chain-adapters/ForwarderBase.sol
bmzig: https://github.com/across-protocol/contracts/pull/609/commits/110a8e93ce65176a786469002300dd5b87fdaf45

contracts/chain-adapters/ForwarderBase.sol
bmzig: https://github.com/across-protocol/contracts/pull/609/commits/110a8e93ce65176a786469002300dd5b87fdaf45

contracts/chain-adapters/ForwarderBase.sol
bmzig: https://github.com/across-protocol/contracts/pull/609/commits/110a8e93ce65176a786469002300dd5b87fdaf45

contracts/chain-adapters/ForwarderBase.sol
bmzig: https://github.com/across-protocol/contracts/pull/609#discussion_r1763825724

contracts/chain-adapters/ForwarderBase.sol
```diff
@@ -0,0 +1,128 @@
...
+        _setL3SpokePool(_l3SpokePool);
+        _setCrossDomainAdmin(_crossDomainAdmin);
+    }
+
+    // Added so that this contract may receive ETH to fund transaction gas fees.
+    receive() external payable {}
+
+    /**
+     * @dev When called by the cross domain admin (i.e. the hub pool), the msg.data should be some function
```
nicholaspai, line 66:
```suggestion
     * @dev When called by the cross domain admin (i.e. the hub pool), the calldata should be reference function
```

contracts/chain-adapters/ForwarderBase.sol
```diff
@@ -0,0 +1,128 @@
...
+     * recognizable by the L3 spoke pool, such as "relayRootBundle" or "upgradeTo". Therefore, we forward
+     * this message to the L3 spoke pool using the implemented messaging logic of the L2 forwarder.
```
nicholaspai, line 68: Is this always the case? Doesn't the hub pool sometimes intend to call this contract, not the L3 contract? For example, the rerouter adapter will call relayTokens/relayMessage() on this contract, right?

Also, how does the hubpool call `setCrossDomainAdmin` on the L3 spoke pool? In this case it wouldn't hit the `fallback` because this function is implemented here, right?

bmzig: That's very true. I removed these functions here https://github.com/across-protocol/contracts/pull/609/commits/19813c6c0a7f304969f64249b9a8ea213c804815. Now all messages destined for L3 should be called by `relayMessage` (although this will be reflected in this PR https://github.com/across-protocol/contracts/pull/610).

contracts/chain-adapters/ForwarderBase.sol
```diff
@@ -0,0 +1,95 @@
...
+ * @title ForwarderBase
+ * @notice This contract expects to receive messages and tokens from the Hub Pool on L1 and forwards them to targets on L3. Messages
+ * are intended to originate from the re-router adapter on L1 which will send messages here to re-route them to the corresponding L3
+ * spoke pool. It rejects messages which do not originate from the cross domain admin, which should be set to the hub pool.
+ * @dev Base contract designed to be deployed on L2 to re-route messages from L1 to L3. If a message is sent to this contract which
+ * came from the L1 cross domain admin, then it is routed to the L3 via the canonical messaging bridge. Tokens sent from L1 are stored
+ * temporarily on this contract. Any EOA can initiate a bridge of these tokens to the target `l3SpokePool`.
+ * @custom:security-contact bugs@across.to
+ */
+abstract contract ForwarderBase is UUPSUpgradeable {
```
nicholaspai, line 16: Do you think this contract should inherit AdapterInterface?

It kind of makes sense to me that this contract should enforce that all child contracts implement relayMessage/relayTokens. All of the comments about this contract  suggest this contract should forward messages from L1 to L3, so if a child contract extended ForwarderBase but didn't extend AdapterInterface, it'd be weird

bmzig: Yeah that's a good refactor https://github.com/across-protocol/contracts/pull/609/commits/d78ad6fdd7e6b09cd7ffd4ca20d51df534ac7ba1

contracts/chain-adapters/ForwarderBase.sol
bmzig: removed https://github.com/across-protocol/contracts/pull/609/commits/19813c6c0a7f304969f64249b9a8ea213c804815

nicholaspai (approved):

pxrl:
@bmzig If this contract is now stateful then can we re-consolidate on having one per L3? That would ease the management overheads vs. one per L3, which is what the pre-existing approach was (though it might have changed since).

contracts/chain-adapters/ForwarderBase.sol
```diff
@@ -0,0 +1,179 @@
...
+ */
+abstract contract ForwarderBase is UUPSUpgradeable, AdapterInterface {
+    /**
+     * @notice Struct containing all information needed in order to send a message to the next layer.
+     * @dev We need to know three different addresses:
+     *   - The address of the adapter to delegatecall, so that we may propagate message and token relays.
+     *   - The address of the next layer's forwarder, so that we can identify the intended recipient of message and token relays.
+     *   - The address of the next layer's spoke pool, so that we can determine whether the next chain contains the target spoke pool.
+     */
+    struct RouteIndex {
```
nicholaspai, line 25: Why call this an index?

contracts/chain-adapters/ForwarderBase.sol
```diff
@@ -0,0 +1,182 @@
...
+            (success, ) = route.adapter.delegatecall(
+                abi.encodeCall(AdapterInterface.relayTokens, (baseToken, remoteToken, amount, target))
+            );
+            if (!success) revert RelayTokensFailed(baseToken);
+        } else {
+            // Case 2: We are not immediately before the target spoke pool, so we send the tokens to the next
+            // forwarder and accompany it with a message containing information about its intended destination.
+
+            // Send tokens to the forwarder.
+            (success, ) = route.adapter.delegatecall(
```
nicholaspai, line 153: Is this code chunk the same as in the Rerouter adapter?

nicholaspai (changes requested):

nicholaspai (approved):
Meant to approve last time

contracts/chain-adapters/ForwarderBase.sol
bmzig: Yes exactly.

contracts/chain-adapters/ForwarderBase.sol
bmzig: I'm trying to find the right word, but I am envisioning each of these contracts as a node in a graph/tree, and in order for each node to connect to subsequent layers, they need to know the info here, so I'm almost "indexing" in the graph to determine what that edge looks like.
... it's a stretch.

nicholaspai: hmm i like the visual of the graph/tree but I don't understand index. Naming is important, think on it some more.

contracts/chain-adapters/ForwarderBase.sol
```diff
@@ -0,0 +1,216 @@
...
+        address _destinationAddress,
+        address _baseToken,
+        address _remoteToken
+    ) external onlyAdmin {
+        remoteTokens[_destinationAddress][_baseToken] = _remoteToken;
+        emit RemoteTokensUpdated(_destinationAddress, _baseToken, _remoteToken);
+    }
+
+    /**
+     * @notice Relays a specified message to a target on an arbitrary layer.
```
nicholaspai, line 128: This is not forwarding to an "arbitrary" layer, though, is it? There is a hardcoded next layer to target messages to and depending on the `target` address, this contract changes its behavior

contracts/chain-adapters/ForwarderBase.sol
```diff
@@ -0,0 +1,216 @@
...
+        address adapter;
+        address forwarder;
+        address spokePool;
+    }
+
+    // Address of the contract which can relay messages to the subsequent forwarders/spoke pools and update this proxy contract.
+    address public crossDomainAdmin;
+    // Map from a destination address to the next token/message bridge route to reach the destination address. The destination
+    // address can either be a spoke pool or a forwarder. This mapping requires that no destination addresses (i.e. spoke pool or
+    // forwarder addresses) collide.
```
nicholaspai, line 40:
```suggestion
    // forwarder addresses) collide. We can't use a chain ID as the key here because we don't have this information
    // from the call into `relayMessage` or `relayTokens`, which is where we'll load this data.
```

contracts/chain-adapters/ForwarderBase.sol
```diff
@@ -0,0 +1,216 @@
...
+        address target
+    ) internal {
+        RouteIndex storage route = availableRoutes[target];
+        if (route.adapter == address(0)) revert UninitializedRoute();
+        bool success;
+        // Case 1: We are immediately before the target spoke pool, so we send the tokens to the spoke pool
+        // and do NOT follow it up with a message.
+        if (target == route.spokePool) {
+            (success, ) = route.adapter.delegatecall(
+                abi.encodeCall(AdapterInterface.relayTokens, (baseToken, remoteToken, amount, target))
```
nicholaspai, line 178: If you're going to use `remote` here to reference the "next layer" then that's confusing because you should use the same word to reference possible target contracts in the next layer, but instead you name that list `availableRoutes`. I think you should use the same word to reference either {tokens} or {possible target} contracts on the next layer when naming those `mappings`

contracts/chain-adapters/ForwarderBase.sol
```diff
@@ -0,0 +1,216 @@
...
+    /**
+     * @notice Relays a specified message to a target on an arbitrary layer.
+     * @param target The address of the contract that will receive the input message.
+     * @param message The data to execute on the contract.
+     * @dev Each forwarder will only know if they are on the layer directly before the layer which contains the `target` contract.
+     * If the next layer contains the target, we send `message` cross-chain normally. Otherwise, we wrap `message` into a `relayMessage`
+     * call to the next layer's forwarder contract.
+     */
+    function relayMessage(address target, bytes memory message) external payable override onlyAdmin {
+        RouteIndex storage route = availableRoutes[target];
```
nicholaspai, line 136: I know we are constrained here but its really confusing that we grab information about the next layer to forward the message to using the `target` as the key---and then we subsequently check if the `target` is equal to any of hte values in the route information.

- `route` is an inaccurate name. We are loading information of known contracts we could target on the next layer. We are comparing the `target`--the end target of the `message`--with contracts from `availableRoutes[target]`.
- `availableRoutes`: this is really list of possible target contracts. Could it be renamed `destinationChainTargets`?
- I think `RouteIndex` can be better named. Perhaps `PossibleTargetContracts` or `TargetContracts`
- See comment below about `remoteTokens` but a consistent name with these suggestions would be `destinationChainTokens`

nicholaspai (commented):
Some comments on naming. You shouldn't take naming lightly, it's the difference between code that is easy to review and code that's not, and that's a major difference in code quality.

contracts/chain-adapters/ForwarderBase.sol
bmzig: https://github.com/across-protocol/contracts/pull/609/commits/d33099343f18b48c9db005412529ced8980b7c86

contracts/chain-adapters/ForwarderBase.sol
bmzig: https://github.com/across-protocol/contracts/pull/609/commits/d33099343f18b48c9db005412529ced8980b7c86

contracts/chain-adapters/ForwarderBase.sol
```diff
@@ -0,0 +1,214 @@
...
+     */
+    struct Route {
+        address adapter;
+        address forwarder;
+        address spokePool;
+    }
+
+    // Map from a destination address to the next token/message bridge route to reach the destination address. The destination address can either be a
+    // spoke pool or a forwarder. Destination addresses are used as the key in this mapping because the `target` address is the only information
+    // we receive when `relay{Message/Token}` is called on this contract, so we assume that these destination contracts on all the possible next
```
mrice32, line 42: I don't really see why we can't just include the chainId? The only thing we'd need is that each chain has a unique adapter on mainnet, right?

nicholaspai: do you mean add `chainId` to the `Route` struct?

bmzig: The issue with using the chain id was `relayTokens`. Ideally, we send tokens to the target contract and then call `relayMessage`, which calls `relayTokens` on the forwarder, but since `relayTokens` (as we define it in `AdapterInterface`) has no way of adding extra data, we would probably have to make a special `relayTokens` function and therefore a new interface. Far from untenable, but it was simpler to use the addresses instead.

nicholaspai: > Far from untenable

How would we do it? Just curious, maybe its worth doing since its a much more natural key than `destinationAddress`

bmzig: (Perhaps far from optimal) but a simple way is if the chain id is stored on the L1 adapter, then we just send a message to the forwarder which also includes this chain id (like `abi.encode(chainId, message)`) which can be processed by all subsequent forwarders. If we made a new relayTokens function, it could look something like `relayTokens(address baseToken, address remoteToken, uint256 to, address to, uint256 destinationChainId)` and then for each forwarder keep encoding this message on every token relay.

contracts/chain-adapters/ForwarderBase.sol
```diff
@@ -0,0 +1,240 @@
...
+    function _setCrossDomainAdmin(address _newCrossDomainAdmin) internal {
+        if (_newCrossDomainAdmin == address(0)) revert InvalidCrossDomainAdmin();
+        crossDomainAdmin = _newCrossDomainAdmin;
+        emit SetXDomainAdmin(_newCrossDomainAdmin);
+    }
+
+    // Reserve storage slots for future versions of this base contract to add state variables without
+    // affecting the storage layout of child contracts. Decrement the size of __gap whenever state variables
+    // are added. This is at bottom of contract to make sure it's always at the end of storage.
+    uint256[1000] private __gap;
```
pxrl, line 239: Should this be relocated above the function and event definitions? It seems more likely that any subsequent additions would be added before them and might overlook the existence of this gap.

contracts/chain-adapters/ForwarderBase.sol
pxrl: > a simple way is if the chain id is stored on the L1 adapter, then we just send a message to the forwarder which also includes this chain id (like `abi.encode(chainId, message)`) which can be processed by all subsequent forwarders.

This makes sense to me. The router contract on the L1 would need to know this (set in the constructor arguments) and then each subsequent recipient would keep propagating that same message until the next-hop bridge is owned by that destination chain ID. I think it could simplify the logic quite a bit.

contracts/chain-adapters/interfaces/ForwarderInterface.sol
bmzig, line 1: This is mostly a copy of `AdapterInterface` but with `destinationChainId` as a field in the relay functions and updated event signatures. It's an interface so we can do `abi.encodeCall` on the rerouter adapters and benefit from Solidity type checking.

contracts/chain-adapters/ForwarderBase.sol
nicholaspai: I think you raise a good point but for consistency let's leave it at bottom of contract. Any developer familiar with our SpokePool will know to look for it down here

contracts/chain-adapters/ForwarderBase.sol
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    /**
+     * @notice Maps a new destination chain ID to an adapter contract which facilitates bridging to that chain.
+     * @param _destinationChainId The chain ID of the target network.
+     * @param _l2Adapter Contract address of the adapter which interfaces with the L2-L3 bridge.
+     * @dev Actual bridging logic is delegated to the adapter contract so that the forwarder can function irrespective of the "flavor" of
+     * L3 (e.g. ArbitrumOrbit, OpStack, etc.).
+     */
+    function updateRoute(uint256 _destinationChainId, address _l2Adapter) external onlyAdmin {
```
nicholaspai, line 84: nit should this be renamed to updateAdapter?

nicholaspai (approved):

contracts/chain-adapters/ForwarderBase.sol
bmzig: Yes, it should have been. Fixed here https://github.com/across-protocol/contracts/pull/609/commits/6298cc69fd08459d75735bb53cb6f1af66f95e1b

contracts/chain-adapters/ForwarderBase.sol
pxrl: Ah sorry; I was mixed up and thought it was the opposite. Agree 100%:

pxrl:
> This PR adds the contract `ArbitrumForwarderInterface` which is used to send messages/tokens from an arbitrary L2 to an Arbitrum-like L3. The functions which will need to be implemented for the forwarder contracts will be:
> 
> * `relayTokens`: an external function callable by anybody which bridges tokens to the L3 spoke pool.
> * `_relayMessage`: an internal function which sends a message to a specified target on L3.
> * `_requireAdminSender`: an internal function which ensures that the caller is the aliased hub pool.
> 
> In general, `relayTokens` and `_relayMessage` will be taken from the current `Arbitrum_*_Adapter` code. Under the assumption that the L2 does aliasing differently than adding 0x11...000...11 to the `msg.sender`, `_requireAdminSender` will have to be modified accordingly.

@bmzig This description probably needs an update to reflect the changes that were made during review.

pxrl (approved):

contracts/chain-adapters/ForwarderBase.sol
```diff
@@ -0,0 +1,154 @@
...
+    // relay functions of `AdapterInterface`.
+    mapping(uint256 => address) chainAdapters;
+
+    event ChainAdaptersUpdated(uint256 indexed destinationChainId, address l2Adapter);
+    event SetXDomainAdmin(address indexed crossDomainAdmin);
+    event DestinationChainTokensUpdated(
+        address indexed target,
+        address indexed baseToken,
+        address indexed destinationChainToken
+    );
```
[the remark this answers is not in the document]
bmzig: Good catch. That shouldn't have been in there. I double checked events and errors so now they should all be used. https://github.com/across-protocol/contracts/pull/609/commits/7e2adf1fc098f317d4789d2dd3a00be423c0b932

contracts/chain-adapters/ForwarderBase.sol
bmzig: Done here https://github.com/across-protocol/contracts/pull/609/commits/41b4856a3bf93ba93296db1a38ea69b899e9f926

contracts/chain-adapters/ForwarderBase.sol
bmzig: This (and the comment below) was addressed here https://github.com/across-protocol/contracts/pull/609/commits/d33099343f18b48c9db005412529ced8980b7c86

pxrl (approved):
OK from me!

PR #611, opened 2024-09-18
Add initial SVM-Spoke MVP from closed repo

chrismaree:
this PR brings over the initial SVM implementation for the spoke pool. 

two worthy callouts:
1. the svm ts is not yet type safe and failed on ts build. it's ignored for now
2. the svm build process is a bit janky (see the comment in this PR) and will be improved soon.


package.json
```diff
@@ -27,7 +27,10 @@
     "prettier": "prettier .",
     "clean-fast": "for dir in node_modules cache cache-zk artifacts artifacts-zk dist typechain; do mv \"${dir}\" \"_${dir}\"; rm -rf \"_${dir}\" &; done",
     "clean": "rm -rf node_modules cache cache-zk artifacts artifacts-zk dist typechain",
-    "build": "hardhat compile && anchor build && tsc && rsync -a --include '*/' --include '*.d.ts' --exclude '*' ./typechain ./dist/",
+    "build-evm": "hardhat compile",
+    "build-svm": "echo 'Generating IDLs...' && anchor build > /dev/null 2>&1 || true && anchor run generateExternalTypes && anchor build",
```
chrismaree, line 31: note this `build-svm` command right now is super hacky. in order for `anchor build` to run correctly first time we require that first `anchor run generateEternalTypes` is run. however, this second command requires the IDLs and types to be generated, which are first built from `anchor build`. i.e we run the build command, which generates the IDLs and types, but then fails to build, which is then used in the `generateExternalTypes` command, which is then used again in the anchor build. 

we will improve this in a follow on PR but for now it works as expected, despite being a bit round about.

tsconfig.json
```diff
@@ -18,5 +18,9 @@
     "tasks/enableL1TokenAcrossEcosystem.ts",
     "utils/utils.ts"
   ],
+  "exclude": [
```
chrismaree, line 21: the new ts associated with sim stuff is not yet totally type safe! we are simply ignoring this for now and will improve it soon

Anchor.toml
```diff
@@ -5,14 +5,58 @@ resolution = true
...
 
 [provider]
-cluster = "Localnet"
-wallet = "~/.config/solana/id.json"
+cluster = "localnet"
+wallet = "test/svm/keys/localnet-wallet.json"
 
 [scripts]
-test = "yarn run ts-mocha -p ./tsconfig.json -t 1000000 test/svm/**/*.ts"
+test = "anchor run generateExternalTypes && yarn run ts-mocha -p ./tsconfig.json -t 1000000 test/**/*.ts"
```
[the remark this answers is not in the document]
chrismaree: good catch. I dont think this made any diff but I'll add it.

Reinis-FRP (approved):

PR #614, opened 2024-09-20
fix(svm): override chain id in executing slow fill

Reinis-FRP:
Adds overriding to local `chain_id` in `execute_v3_slow_relay_leaf`. This makes SVM behavior consistent with EVM SpokePool: if someone was trying to replay slow fill targeted at another chain, this should fail the proof verification due to mismatched chain id.


chrismaree (approved):
really good check. the base slow fill implementation was missing many of these.

PR #615, opened 2024-09-20
Add missing events and re-factor event location

chrismaree:
PR add an additional events and pulls events into a common defined location.

### DIFF
--- a/programs/svm-spoke/src/event.rs
+++ b/programs/svm-spoke/src/event.rs
@@ -1,0 +1,108 @@
+use anchor_lang::prelude::*;
+
+// Admin events
+#[event]
+pub struct SetXDomainAdmin {
+    pub new_admin: Pubkey,
+}
+
+#[event]
+pub struct PausedDeposits {
+    pub is_paused: bool,
+}
+
+#[event]
+pub struct PausedFills {
+    pub is_paused: bool,
+}
+
+#[event]
+pub struct EnabledDepositRoute {
+    pub origin_token: Pubkey,
+    pub destination_chain_id: u64,
+    pub enabled: bool,
+}
+
+// Deposit events
+#[event]
+pub struct V3FundsDeposited {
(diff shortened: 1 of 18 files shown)


programs/svm-spoke/src/event.rs
```diff
@@ -0,0 +1,107 @@
...
+
+// Deposit events
+#[event]
+pub struct V3FundsDeposited {
+    pub input_token: Pubkey,
+    pub output_token: Pubkey,
+    pub input_amount: u64,
+    pub output_amount: u64,
+    pub destination_chain_id: u64,
+    pub deposit_id: u64,
```
Reinis-FRP, line 34: `depositId` is `uint32` in EVM SpokePool, and we also use `u32` in fills

programs/svm-spoke/src/event.rs
```diff
@@ -0,0 +1,107 @@
...
+
+// Deposit events
+#[event]
+pub struct V3FundsDeposited {
+    pub input_token: Pubkey,
+    pub output_token: Pubkey,
+    pub input_amount: u64,
+    pub output_amount: u64,
+    pub destination_chain_id: u64,
+    pub deposit_id: u32,
```
Reinis-FRP, line 34: I think we also might need to change `number_of_deposits` counter in state to u32

programs/svm-spoke/src/state/state.rs
```diff
@@ -7,7 +7,7 @@ pub struct State {
     pub paused_fills: bool,
     pub owner: Pubkey,
     pub seed: u64, // Add a seed to the state to enable multiple deployments.
-    pub number_of_deposits: u64,
+    pub number_of_deposits: u32,
```
Reinis-FRP, line 10: also need u32 in initial_number_of_deposits if initialize both in lib and admin.

Reinis-FRP (approved):

PR #618, opened 2024-09-23
Improve events & make consistant with EVM

chrismaree:


programs/svm-spoke/src/instructions/deposit.rs
```diff
@@ -88,7 +88,24 @@ pub fn deposit_v3(
...
     // verify that this check is sufficient or update accordingly.
-    require!(ctx.accounts.route.enabled, CustomError::RouteNotEnabled);
+    require!(ctx.accounts.route.enabled, CustomError::DisabledRoute);
+
+    let current_timestamp = if state.current_time != 0 {
+        state.current_time
+    } else {
+        Clock::get()?.unix_timestamp as u32
+    };
+    // TODO: in solidity these are if then revert. here they are requires. i.e the assertion goes in the opposite direction to get the same effect. verify we want <= vs < for these assertions as this can change things for inclusive/exclusive checks.
```
Reinis-FRP, line 98: It might be easier to keep consistency with Solidity implementation by putting inverted (!) if condition in the require statement

chrismaree: I used a different syntax where we can use an `if` and `return Err` to match the solidity logic.

Reinis-FRP (approved):

PR #619, opened 2024-09-23
feat(svm-routes): add additional param to block non-enabled routes

md0x:
Changes proposed in this PR:
- Fix routes seed to use state key to prevent passing fake enabled routes


chrismaree (approved):
this is great! really good check. 

there might be other kinds of derivation issue we have as a result of not using the state within parts of the PDA key seed...we should explore every bit of state to ensure they are sufficiently locked like this.

PR #620, opened 2024-09-24
test: use address lookup table in max refund test

Reinis-FRP:
Increases max refund count by using Address Lookup Table (ALT). This is capped at 19 refunds as higher count causes out of memory panic when looping through refund transfers. To overcome this limit we would need to split individual refund claims in separate transaction. And if we split this, then we don't really need to use ALT.

Edit: limit is now increased to 21 by dropping expensive debug msg of input data. Now the bottleneck is calldata size, but that could be resolved by passing it from a buffer account.

### DIFF
--- a/programs/svm-spoke/src/instructions/bundle.rs
+++ b/programs/svm-spoke/src/instructions/bundle.rs
@@ -92,7 +92,6 @@ impl RelayerRefundLeaf {
 
     pub fn to_keccak_hash(&self) -> [u8; 32] {
         let input = self.to_bytes();
-        msg!("input: {:?}", input);
         keccak::hash(&input).0
     }
 }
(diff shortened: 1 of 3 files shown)


md0x (approved):
This is very nice, and without even having to touch the Rust code! I wonder what the limit would be if we got rid of CPI events and thus passed one fewer account.

Reinis-FRP:
> This is very nice, and without even having to touch the Rust code! I wonder what the limit would be if we got rid of CPI events and thus passed one fewer account.

Turns out the out of memory panic was caused by debug `msg!("input: {:?}", input);` within `to_keccak_hash` which is quite memory expensive due to parsing raw bytes as array in strings. I'm now pushing update here that allows reaching 21 refunds, and now the message size limit is hit due to refund accounts also being present in calldata. In a follow-up I plan to check how to work around this passing calldata from buffer account.

PR #623, opened 2024-09-26
fix(svm): pass instruction data from prefilled account

Reinis-FRP:
Increases maximum relayer repayment count to 28 by passing instruction arguments from a pre-filled account. Higher refund count hits inner instruction size limit when doing `emit_cpi` on public devnet. On localnet this would hit out of memory panic above 31 refunds. Though currently Across protocol config has set `MAX_RELAYER_REPAYMENT_LEAF_SIZE` to 25, which should be sufficient.

In addition, Execute Max Refunds test also now includes EVM chain leaves, so that the proofs is non-empty array and we can check this does not hit memory limits when processing them.


test/svm/SvmSpoke.InstructionParams.ts
```diff
@@ -0,0 +1,132 @@
...
+    try {
+      await program.methods.writeInstructionParamsFragment(0, inputData).accounts({ instructionParams }).rpc();
+      assert.fail("Write instruction params should have failed");
+    } catch (err) {
+      assert.instanceOf(err, anchor.AnchorError);
+      assertSE(err.error.errorCode.code, "ConstraintSeeds", "Expected error code ConstraintSeeds");
+    }
+  });
+
+  it("Cannot close another caller's instruction params", async () => {
```
chrismaree, line 104: These are great.

programs/svm-spoke/src/instructions/bundle.rs
```diff
@@ -101,10 +108,13 @@ impl RelayerRefundLeaf {
 
 pub fn execute_relayer_refund_leaf<'info>(
     ctx: Context<'_, '_, '_, 'info, ExecuteRelayerRefundLeaf<'info>>,
-    root_bundle_id: u32,
-    relayer_refund_leaf: RelayerRefundLeaf,
-    proof: Vec<[u8; 32]>,
 ) -> Result<()> {
```
md0x, line 111: Do you think it makes sense to have 2 versions of this function? one using the pre-loaded instructions and a classic (backup) version?

programs/svm-spoke/src/state/instruction_params.rs
```diff
@@ -0,0 +1,12 @@
+use anchor_lang::prelude::*;
+
+use crate::RelayerRefundLeaf;
+
+#[account]
+#[derive(InitSpace)]
+pub struct ExecuteRelayerRefundLeafParams {
+    pub root_bundle_id: u32,
+    pub relayer_refund_leaf: RelayerRefundLeaf,
+    #[max_len(0)]
```
md0x, line 10: Why do we need this constrain?

md0x, line 10: Are they actually empty when we pre-load the instruction params account?

programs/svm-spoke/src/instructions/bundle.rs
```diff
@@ -101,10 +108,13 @@ impl RelayerRefundLeaf {
...
+    // Get pre-loaded instruction parameters.
+    let instruction_params = &ctx.accounts.instruction_params;
+    let root_bundle_id = instruction_params.root_bundle_id;
+    let relayer_refund_leaf = instruction_params.relayer_refund_leaf.to_owned();
+    let proof = instruction_params.proof.to_owned();
+
```
md0x, line 117: Very cool how we don't need to handle any decoding!

test/svm/SvmSpoke.InstructionParams.ts
```diff
@@ -0,0 +1,132 @@
...
+    const ix = await program.methods
+      .initializeInstructionParams(totalSize)
+      .accounts({ signer: caller.publicKey, instructionParams })
+      .instruction();
+    await anchor.web3.sendAndConfirmTransaction(connection, new anchor.web3.Transaction().add(ix), [caller]);
+  };
+
+  const writeInstructionParams = async (data: Buffer) => {
+    // Should not exceed message size limit when writing to the data account
+    const maxInstructionParamsFragment = 900;
```
md0x, line 27: This limit needs to be taken into consideration when calculating the rootBundles off-chain otherwise we might have non-claimable bundles in solana right?

md0x (approved):
Great improvement, @Reinis-FRP! I’ve just added a few questions regarding some of the constraints, particularly the max_length(0) ones, but otherwise, this looks good to me.

programs/svm-spoke/src/instructions/bundle.rs
Reinis-FRP: I have not found yet a clean way to implement 2 external methods that use the same internal implementation in solana. This would require maintaining duplicate versions of most of the code.

programs/svm-spoke/src/state/instruction_params.rs
Reinis-FRP: its a declaration to help calculate initial space when doing account initialization. I'll add required comments

Reinis-FRP: This only is helper to calculate total account space when initializing. Initially its all zeroed, but it is up to the user to correctly encode raw data when writing fragments to this account.

chrismaree (approved):

PR #624, opened 2024-09-26
refactor: move all Arbitrum interfaces to a single location

bmzig:
To reduce the diff of https://github.com/across-protocol/contracts/pull/609


contracts/interfaces/ArbitrumBridgeInterfaces.sol
```diff
@@ -0,0 +1,220 @@
...
+ * used as native currency on upper layer.
+ * @dev Fees are paid in this token. There are certain restrictions on the native token:
+ *       - The token can't be rebasing or have a transfer fee
+ *       - The token must only be transferrable via a call to the token address itself
+ *       - The token must only be able to set allowance via a call to the token address itself
+ *       - The token must not have a callback on transfer, and more generally a user must not be able to make a transfer to themselves revert
+ *       - The token must have a max of 2^256 - 1 wei total supply unscaled
+ *       - The token must have a max of 2^256 - 1 wei total supply when scaled to 18 decimals
+ */
+interface ArbitrumERC20Bridge {
```
nicholaspai, line 16: nit: i don't think this file should end with "Interfaces" if its in an /interfaces path. Maybe rename to ArbitrumBridge

nicholaspai (approved):
One comment then LGTM

contracts/interfaces/ArbitrumBridgeInterfaces.sol
bmzig: Done here https://github.com/across-protocol/contracts/pull/624/commits/1da144998a5a7ea50a56bb1314383786d0b0311b

PR #610, opened 2024-09-17
feat: add Arbitrum forwarder contracts

bmzig:
This contains L2 forwarder contracts based on the interface https://github.com/across-protocol/contracts/pull/609, under the assumption that the L2 on which it is deployed is Arbitrum. It mostly follows the same format as the Arbitrum_CustomGasToken_Adapter and Arbitrum_Adapter contracts.

PR #625, opened 2024-09-27
feat: forwarders for AVM and OVM networks

bmzig:
This implements the newer version of `ForwarderBase` and does not make any refactors to the adapters, since the forwarders now assume that the appropriate adapters have been deployed to LX.


contracts/chain-adapters/Arbitrum_Forwarder.sol
```diff
@@ -0,0 +1,52 @@
...
+     * address.
+     */
+    function relayTokens(
+        address,
+        address currentLayerToken,
+        uint256 amount,
+        address target
+    ) external payable override onlyAdmin {
+        address remoteToken = remoteTokens[target][currentLayerToken] == address(0)
+            ? arbitrumGatewayRouter.calculateL2TokenAddress(currentLayerToken)
```
bmzig, line 46: The gateway router should return the correct address if the token in question is actually routed through an Arbitrum gateway. The most notable exception we know of now is USDC and CCTP, which is an example of when we would instead need to set `remoteTokens`. I think if we go with this, we will need to be careful and ensure that the remote token we are attempting to bridge to is indeed created through a `bridgeMint` call.

contracts/chain-adapters/Arbitrum_Forwarder.sol
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * @title Arbitrum_Forwarder
+ * @notice This contract expects to receive messages and tokens from an authorized contract on a previous layer and forwards them to
+ * contracts on the subsequent layer. It rejects messages which do not originate from a cross domain admin.
+ * @custom:security-contact bugs@across.to
+ */
+contract Arbitrum_Forwarder is ForwarderBase {
+    ArbitrumL1ERC20GatewayLike public immutable arbitrumGatewayRouter;
+
+    modifier onlyFromCrossDomainAdmin() {
```
bmzig, line 17: For all the access control functions, there is an exception on L2 only where we have two possible cross domain admins: the hub pool and an upgrade contract. For all forwarders beyond L2, there is only 1 cross domain admin (the previous forwarder).

contracts/Arbitrum_SpokePool.sol
```diff
@@ -54,7 +55,7 @@ contract Arbitrum_SpokePool is SpokePool, CircleCCTPAdapter {
     }
 
     modifier onlyFromCrossDomainAdmin() {
-        require(msg.sender == _applyL1ToL2Alias(crossDomainAdmin), "ONLY_COUNTERPART_GATEWAY");
+        require(msg.sender == AddressUtils._applyL1ToL2Alias(crossDomainAdmin), "ONLY_COUNTERPART_GATEWAY");
```
nicholaspai, line 58: I think this library should be called CrossDomainAddressUtils as AddressUtils overlaps with the openzeppelin utility file that has utilities for any arbitrary EVM. This aliasing is very OVM/AVM specific

contracts/chain-adapters/Arbitrum_Forwarder.sol
nicholaspai: how would cctp work? cctp from L1 to L2, and then canonical bridge from there? can we support that?

nicholaspai (commented):
I know its in draft mode so just left some comments upon first pass

contracts/chain-adapters/Arbitrum_Forwarder.sol
bmzig: For reference, there was this comment https://github.com/across-protocol/contracts/pull/610#discussion_r1765177183 on the previous PR which inherited the adapter code, and the idea was that there may be a CCTP bridge from L2-L3 which bridges native to native. This should work fine if we use CCTP for L1 -> L2, have a native on L2, and then bridge native on L2 -> bridged on L3 via the canonical bridge, since that uses an Arbitrum gateway. TLDR CCTP is just speculation. It's mainly an example if there was another non-arbitrum bridge from L2-L3, and also `calculateL2TokenAddress` should return the remote token which is minted on L2 given the input L1 token.

contracts/Arbitrum_SpokePool.sol
bmzig: https://github.com/across-protocol/contracts/pull/625/commits/24c1836a2a6c0db8ba04c13b8939cd738c4c9eb7

pxrl (approved):

contracts/libraries/CrossDomainAddressUtils.sol
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * @title AddressUtils
+ * @notice This library contains internal functions for manipulating addresses.
+ */
+library CrossDomainAddressUtils {
+    // L1 addresses are transformed during l1->l2 calls.
+    // This cannot be pulled directly from Arbitrum contracts because their contracts are not 0.8.X compatible and
+    // this operation takes advantage of overflows, whose behavior changed in 0.8.0.
+    function _applyL1ToL2Alias(address l1Address) internal pure returns (address l2Address) {
```
nicholaspai, line 12: nit: I prefer not having an underscore prefix on library functions even if they are internal

nicholaspai (approved):
https://github.com/across-protocol/contracts/pull/625/files#r1784701027

contracts/libraries/CrossDomainAddressUtils.sol
bmzig: Done here https://github.com/across-protocol/contracts/pull/625/commits/dfee0a11d506eba84be91c76b7a4bcdc78fa4248

contracts/ZkSync_SpokePool.sol
```diff
@@ -118,14 +119,5 @@ contract ZkSync_SpokePool is SpokePool {
         emit SetZkBridge(address(_zkErc20Bridge), oldErc20Bridge);
     }
 
-    // L1 addresses are transformed during l1->l2 calls.
-    // See https://github.com/matter-labs/era-contracts/blob/main/docs/Overview.md#mailboxfacet for more information.
-    // Another source: https://github.com/matter-labs/era-contracts/blob/41c25aa16d182f757c3fed1463c78a81896f65e6/ethereum/contracts/vendor/AddressAliasHelper.sol#L28
-    function _applyL1ToL2Alias(address l1Address) internal pure returns (address l2Address) {
```
mrice32, line 124: Just to triple check, arbitrum and zksync do this the exact same way, right? Code looks the same, but just want to sanity check that I read this right.

mrice32 (approved):
LGTM!

contracts/ZkSync_SpokePool.sol
bmzig: Yeah. (Just to enumerate) Arbitrum and ZkSync both have the same aliasing method, and all other chains query the messenger contract.

mrice32: Perfect.

PR #627, opened 2024-10-02
fix: svm ts issues

md0x:
Changes proposed in this PR:
- Fix typescript issues in SVM tests


test/svm/SvmSpoke.Bundle.ts
```diff
@@ -317,36 +332,29 @@ describe("svm_spoke.bundle", () => {
...
+        .accounts(executeRelayerRefundLeafAccounts)
         .remainingAccounts(wrongRemainingAccounts)
         .rpc();
-    } catch (err) {
+    } catch (err: any) {
       assert.include(err.toString(), "Account not found");
     }
 
     // Verify valid leaf
+    executeRelayerRefundLeafAccounts = {
```
chrismaree, line 344: this block is repeated many times in this test file. we could re-use it at the top to decrease repetition in these tests & re-assign the values that change.

can save it for a follow on RP though.

chrismaree (approved):
This looks good. there are still a few small places where we have `any` type, which we can try to eliminate later.

PR #628, opened 2024-10-03
feat: ZkStack Adapters

mrice32:
This PR adds two adapters: one for ZkStack with custom gas tokens and one for ZkStack with ETH gas tokens. Note: we were previously told that our existing adapter would work for ZkStack chains that used ETH, but after reading the code, I found restrictions on using legacy methods for any chain that isn't ZkSync.

### DIFF
--- a/contracts/chain-adapters/ZkStack_Adapter.sol
+++ b/contracts/chain-adapters/ZkStack_Adapter.sol
@@ -232,7 +232,7 @@ contract ZkStack_Adapter is AdapterInterface {
                     l2GasPerPubdataByteLimit: L1_GAS_TO_L2_GAS_PER_PUB_DATA_LIMIT,
                     refundRecipient: L2_REFUND_ADDRESS,
                     secondBridgeAddress: BRIDGE_HUB.sharedBridge(),
-                    secondBridgeValue: amount,
+                    secondBridgeValue: 0,
                     secondBridgeCalldata: _secondBridgeCalldata(to, l1Token, amount)
                 })
             );
--- a/contracts/chain-adapters/ZkStack_Adapter.sol
+++ b/contracts/chain-adapters/ZkStack_Adapter.sol
@@ -7,7 +7,7 @@ import "../external/interfaces/WETH9Interface.sol";
 import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
 
-// The BridgeHub is the main interaction point for briding into ZkStack chains.
+// The BridgeHub is the main interaction point for bridging into ZkStack chains.
 interface BridgeHubInterface {
     struct L2TransactionRequestDirect {
         uint256 chainId;
--- a/contracts/chain-adapters/ZkStack_Adapter.sol
+++ b/contracts/chain-adapters/ZkStack_Adapter.sol
@@ -87,22 +87,7 @@ interface BridgeHubInterface {
 }
 
 /**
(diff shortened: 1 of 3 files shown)


contracts/chain-adapters/ZkStack_Adapter.sol
```diff
@@ -0,0 +1,269 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.0;
+
+import "./interfaces/AdapterInterface.sol";
+import "../external/interfaces/WETH9Interface.sol";
+
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+
+// The BridgeHub is the main interaction point for briding into ZkStack chains.
```
nicholaspai, line 10:
```suggestion
// The BridgeHub is the main interaction point for bridging into ZkStack chains.
```

contracts/chain-adapters/ZkStack_Adapter.sol
```diff
@@ -0,0 +1,269 @@
...
+    function l2TransactionBaseCost(
+        uint256 _chainId,
+        uint256 _gasPrice,
+        uint256 _l2GasLimit,
+        uint256 _l2GasPerPubdataByteLimit
+    ) external view returns (uint256);
+}
+
+/**
+ * @notice Contract containing logic to send messages from L1 to ZkStack with ETH as the gas token.
```
nicholaspai, line 90: So this CAN'T be used to bridge to ZkSync? Which seems non-intuitive

contracts/chain-adapters/ZkStack_Adapter.sol
```diff
@@ -0,0 +1,269 @@
...
+     * @notice Send cross-chain message to target on ZkStack.
+     * @dev The HubPool must hold enough ETH to pay for the L2 txn.
+     * @param target Contract on L2 that will receive message.
+     * @param message Data to send to target.
+     */
+    function relayMessage(address target, bytes memory message) external payable override {
+        uint256 txBaseCost = _computeETHTxCost(L2_GAS_LIMIT);
+
+        // Returns the hash of the requested L2 transaction. This hash can be used to follow the transaction status.
+        bytes32 canonicalTxHash = BRIDGE_HUB.requestL2TransactionDirect{ value: txBaseCost }(
```
nicholaspai, line 169: You note in the interface for `requestL2TransactionDirect` that:

> This means this is not ideal for contract calls, as the contract would have to handle token allowance of the base Token.

And yet you're making a contract call here, right?

nicholaspai (approved):
Hard to verify these work but they seem sensible

contracts/chain-adapters/ZkStack_Adapter.sol
mrice32: This natspec is copied directly from their docs. This was also confusing to me, but I copied it just for posterity.

I used [this example](https://etherscan.io/address/0x3766Eb5F07DBc60d39a2059A9A29fD9b7D3C356D#code) as a guide to how these functions are used for both token and native asset bridging.

contracts/chain-adapters/ZkStack_Adapter.sol
mrice32: They can be used to bridge to ZkSync, but I wanted to remain unopinionated about whether we do that or not.

pxrl (approved):
I didn't spot any issues, LGTM :+1:

bmzig (approved):

PR #629, opened 2024-10-03
improve: rename Spoke Pool's hubPool variable to withdrawalRecipient

bmzig:
This is meant to generalize the `hubPool` variable in all spoke pools. The `hubPool` variable is only used in `_bridgeTokensToHubPool`, where it is set as the recipient. For spoke pools deployed on L3s, or for spoke pools which may not want to withdraw token to the hub pool (e.g. the `Blast_SpokePool`), the `hubPool` variable is either a misnomer or unused. This changes the variable to be called `withdrawalRecipient` to better reflect the purpose of the `hubPool` variable.


mrice32 (approved):
LGTM!

nicholaspai (commented):
This looks good to me as I have checked that there are no calls to `SpokePool.hubPool()` in production scripts. There are several non-production scripts using this though

nicholaspai (approved):

PR #632, opened 2024-10-07
feat: add updated implementation of ERC-7683

mrice32:
This PR implements the updated version of ERC-7683 as described [here](https://github.com/across-protocol/ERCs/pull/1). This change ended up being more involved than expected.

There are only a few open questions remaining:
1. Should this handle pre-fills? Right now, it assumes the incrementing deposit id. Would require #583 to be merged first to support them.
2. Quote timestamp is assumed to be the deposit timestamp for Onchain deposits. I think this is probably a reasonable initial assumption to keep the AcrossOrderData struct consistent between the onchain and gasless versions, but we could add a param for it if necessary.
3. The external depositor contract (which we rely on since we don't have room to merge the bulk of this into the spoke) now needs a lookup table of destination settler contracts (spoke pools until we allow relayer refunds to be redirected). I designated this as a simple ownable contract to avoid having to bake in chain-specific authentication logic, but I don't think we will have a good way to administer this (since we don't want to use chain-specific multisigs). This is becoming a common issue -- I wonder if it would make sense to have a lib deployed to every chain or a per-chain admin passthrough contract to help manage all the downstream contracts on that chain. Simplest solution I could think of for now to avoid a lot of code bloat was to just make it ownable (meaning we could configure at construction and discard ownership or administer with a chain-specific multisig).

### DIFF
--- a/contracts/erc7683/ERC7683OrderDepositor.sol
+++ b/contracts/erc7683/ERC7683OrderDepositor.sol
@@ -4,9 +4,10 @@ pragma solidity ^0.8.0;
 import "../external/interfaces/IPermit2.sol";
 import "@openzeppelin/contracts/utils/math/SafeCast.sol";
 import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
 
-import { Input, Output, CrossChainOrder, ResolvedCrossChainOrder, ISettlementContract } from "./ERC7683.sol";
-import { AcrossOrderData, AcrossFillerData, ERC7683Permit2Lib } from "./ERC7683Across.sol";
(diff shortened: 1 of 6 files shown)


contracts/erc7683/ERC7683OrderDepositor.sol
```diff
@@ -36,134 +41,271 @@ abstract contract ERC7683OrderDepositor is ISettlementContract {
...
+        _callDeposit(
+            msg.sender,
+            acrossOrderData.recipient,
+            acrossOrderData.inputToken,
+            acrossOrderData.outputToken,
+            acrossOrderData.inputAmount,
+            acrossOrderData.outputAmount,
+            acrossOrderData.destinationChainId,
+            acrossOrderData.exclusiveRelayer,
+            // Note: simplifying assumption to avoid the order type having to bake in the quote timestamp.
```
nicholaspai, line 103: I think this is a fair UX trade off. Not sure why depositor would ever want a different timestamp anyways

nicholaspai (approved):
This is a lot of code but it's very clean. LGTM

PR #633, opened 2024-10-07
fix: separate refund claim when there is no token account

Reinis-FRP:
This fix protects against a relayer passing invalid refund token account to block the whole relayer leaf execution. If a relayer had provided invalid ATA for refunds, the leaf executor can replace it with a claim account PDA that is derived from mint and relayer's token account to only accrue given relayer's refund claims. If this was a valid token address, but was closed at the time of leaf execution, all accrued claims can be refunded separately whenever the given refund account is created.

For the sake of code simplicity, this does not attempt to initialize a claim account when executing the relayer refund leaf. The operator is expected to initialize it in advance, but this is required only when a given relayer's token account is invalid.


programs/svm-spoke/src/state/refund_account.rs
```diff
@@ -0,0 +1,131 @@
...
+
+use crate::error::CustomError;
+
+#[account]
+#[derive(InitSpace)]
+pub struct ClaimAccount {
+    pub amount: u64,
+}
+
+// When executing relayer refund leaf, refund accounts are passed as remaining accounts and can hold either a regular
```
chrismaree, line 12: does this mean that, in theory, a relayer could close their ATA right before the bundle execution tx lands on-chain, thereby making the execution of the bundle in-valid as the remaining account type is wrong? so when the executor looks at the tx to send they build TokenAccount, expecting that the relayer has an ATA but before the tx lands on-chain the relayer can front run the tx and delete their ATA, which then requires, for this method to work, that the executor passes in the ClaimAccount, not token account, which breaks the bundle execution and the whole thing reverts?

In theory, this means that a relayer who can front run the execution can always "toggle" the status of their ATA (opening it or closing it), thereby blocking the bundle execution?

if the executor sends these execution transactions via private mempool they might be able to protect themselves (you cant close/open the account before the execution) but this is not going to be failsafe?

thoughts on this?

programs/svm-spoke/src/state/refund_account.rs
```diff
@@ -0,0 +1,132 @@
+use anchor_lang::prelude::*;
```
chrismaree, line 1: nit on this file naming: we call the instruction `refund_claims` and the instruction `refund_accounts`. we should think about it makes sense to re-name this state to `claim_account`? equally, for other files between instruction & state having a consistent name between the two.

programs/svm-spoke/src/state/refund_account.rs
```diff
@@ -0,0 +1,132 @@
...
+        Some(token_account)
+    }
+
+    // This implements the following Anchor account constraints when parsing remaining account as a claim account:
+    // #[account(
+    //     mut,
+    //     seeds = [b"claim_account", mint.key().as_ref(), token_account.key().as_ref()],
+    //     bump
+    // )]
+    // pub claim_account: Account<'info, ClaimAccount>,
```
chrismaree, line 106: is this meant to be commented out?

test/svm/SvmSpoke.Bundle.ts
```diff
@@ -643,167 +656,215 @@ describe("svm_spoke.bundle", () => {
     }
   });
 
-  it("Execute Max Refunds", async () => {
-    const relayerRefundLeaves: RelayerRefundLeafType[] = [];
+  describe("Execute Max Refunds", () => {
+    const executeMaxRefunds = async (refundType: RefundType) => {
+      const relayerRefundLeaves: RelayerRefundLeafType[] = [];
+      // Higher refund count hits inner instruction size limit when doing `emit_cpi` on public devnet. On localnet this is
```
chrismaree, line 662: to test this on testnet did you use the [script](https://github.com/across-protocol/contracts/blob/master/scripts/svm/simpleFakeRelayerRepayment.ts)?

chrismaree: it would be wise I think for us to run this same test on mainnet to ensure that there are no other memory issues that we don't know about on this. does mean deploying the whole program, but that's ok as we can always close it afterwards.

test/svm/SvmSpoke.Bundle.ts
```diff
@@ -643,167 +656,215 @@ describe("svm_spoke.bundle", () => {
...
+              : (await program.account.claimAccount.fetch(account)).amount.toString();
+          }
+        })
+      );
+      refundBalances.forEach((balance, i) => {
+        assertSE(balance, refundAmounts[i].toString(), `Refund account ${i} balance should match refund amount`);
+      });
+    };
+
+    it("Execute Max Refunds to Token Accounts", async () => {
```
chrismaree, line 858: this is cool. nice structure within the describe. these tests are now...complex, but that's ok.

programs/svm-spoke/src/state/refund_account.rs
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    // This implements the following Anchor account constraints when parsing remaining account as a token account:
+    // #[account(
+    //     mut,
+    //     address = expected_token_account @ CustomError::InvalidRefund,
+    //     token::mint = expected_mint,
+    //     token::token_program = token_program
+    // )]
+    // pub token_account: InterfaceAccount<'info, TokenAccount>,
```
chrismaree, line 65: is this mean to be commented out?

programs/svm-spoke/src/instructions/bundle.rs
```diff
@@ -106,9 +106,12 @@ impl RelayerRefundLeaf {
...
 
-pub fn execute_relayer_refund_leaf<'info>(
-    ctx: Context<'_, '_, '_, 'info, ExecuteRelayerRefundLeaf<'info>>,
-) -> Result<()> {
+pub fn execute_relayer_refund_leaf<'c, 'info>(
+    ctx: Context<'_, '_, 'c, 'info, ExecuteRelayerRefundLeaf<'info>>,
+) -> Result<()>
+where
+    'c: 'info,
+{
```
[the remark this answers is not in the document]
Reinis-FRP: Added the constraint that lifetime for remaining accounts (`'c`) is at least as long as `'info` for accounts that are referenced by this account array. I think its needed because of manual remaining account processing in `try_from_remaining_account` method. Without these lifetime constraints the compiler errors.

programs/svm-spoke/src/instructions/bundle.rs
```diff
@@ -188,5 +209,16 @@ pub fn execute_relayer_refund_leaf<'info>(
         caller: ctx.accounts.signer.key(),
     });
 
+    // Emit the DeferredRelayerRefunds event if any claim accounts were passed instead of token accounts. We cannot emit
+    // individual accounts and amounts as that would run out of memory for larger leaves.
+    if deferred_refunds {
+        emit_cpi!(DeferredRelayerRefunds {
```
chrismaree, line 215: I think it's ok to emit all this but it's not really needed, right? we really just want to know that it happened, but care less about all the other props give we have `ExecutedRelayerRefundRoot`.

an alternative structure could simply be to add a bool to `ExecutedRelayerRefundRoot` that indicates if `deferred_refunds` is set so we don't need a. new event?

programs/svm-spoke/src/state/refund_account.rs
Reinis-FRP: `RefundAccount` can hold either `TokenAccount` or `ClaimAccount` and `state/refund_account.rs` has methods for handling both, so would not want to name it as `claim_account`. Also the wrapped `RefundAccount` is used in `bundle` instruction while individual `ClaimAccount` is used in `refund_claims` instruction, so there is no 1-1 relationship between state and instructions like we have in other places.

programs/svm-spoke/src/state/refund_account.rs
```diff
@@ -0,0 +1,132 @@
...
+            return None;
+        }
+
+        // Checks the token address matches.
+        if account_info.key != expected_token_account {
+            return None;
+        }
+
+        // Checks if the token account is associated with the expected mint.
+        if &token_account.mint != expected_mint {
```
chrismaree, line 88: this is a meta-comment but this check has a `&` (borrowing) and the others don't. `expected_token_account` is passed in `&` but is not checked against the account_info with borrowing. 

point is: is there away to do this more consistently in this file and beyond?

programs/svm-spoke/src/state/refund_account.rs
Reinis-FRP: Yes, it is meant to document what Anchor macros this method is reimplementing, so its easier to audit knowing the intention of the implementation.

programs/svm-spoke/src/instructions/bundle.rs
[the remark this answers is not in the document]
chrismaree: o that's interesting. ye, I've had other lifetime issues while implementing other things and was able to address it through other syntax but this is more complex here.

test/svm/SvmSpoke.Bundle.ts
Reinis-FRP: Yes, [here](https://explorer.solana.com/tx/2Smpe9DHWfLxS49gbnbEjeoGYhSYeRMfUcmDLyqwt6z6GVcGza1daJ7NHxKpfKKe5pU5cp7rJqXybXfm6nA3qsgb?cluster=devnet) is the refund execution for 28 relayers. At 29 count would get this error: `Instruction passed to inner instruction is too large (1306 > 1280)` when doing `emit_cpi`.

Though, we might need to improve the script as in some runs it closed the instruction parameter account before leaf execution tx got mined.

programs/svm-spoke/src/state/refund_account.rs
Reinis-FRP: yes, left them for auditing purposes as we reimplement what Anchor macro would have done on static account.

programs/svm-spoke/src/instructions/bundle.rs
Reinis-FRP: yes, if we can change interface of `ExecutedRelayerRefundRoot` this would be much easier to avoid separate `emit_cpi`

programs/svm-spoke/src/state/refund_account.rs
Reinis-FRP: but you're right that its not fully consistent with comparison in `try_claim_account_from_account_info` below, so I just pushed commit to use `account_info.key != &pda_address` instead of `account_info.key() != pda_address` on line 121.

programs/svm-spoke/src/instructions/bundle.rs
Reinis-FRP: pushed the commit to include the bool in `ExecutedRelayerRefundRoot`

chrismaree (approved):

PR #641, opened 2024-10-09
feat(chain-adapters): add solana adapter

Reinis-FRP:
Adds adapter for HubPool being able to relay messages and tokens to Solana SpokePool.

Solana adapter maps Solana bytes32 addresses to EVM address representation by trimming it to keep its lowest 20 byte part. This same conversion must be done by the HubPool owner when adding Solana spoke pool and setting the corresponding pool rebalance and deposit routes, so that the adapter can translate them back correctly.

When setting deposit route, this also translates setEnableRoute call to use bytes32 token address and uint64 chainId as expected by SVM spoke pool program.


contracts/chain-adapters/Solana_Adapter.sol
```diff
@@ -0,0 +1,193 @@
...
+    // Custom errors for relayMessage validation.
+    error InvalidRelayMessageTarget(address target);
+    error InvalidOriginToken(address originToken);
+    error InvalidDestinationChainId(uint256 destinationChainId);
+
+    // Custom errors for relayTokens validation.
+    error InvalidL1Token(address l1Token);
+    error InvalidL2Token(address l2Token);
+    error InvalidAmount(uint256 amount);
+    error InvalidTokenRecipient(address to);
```
Reinis-FRP, line 57: I used these for debugging, though HubPool does not bubble up revert data on delegatecall. Shall we still keep custom errors or revert with empty data at the adapter?

contracts/chain-adapters/Solana_Adapter.sol
```diff
@@ -0,0 +1,193 @@
...
+ */
+
+// solhint-disable-next-line contract-name-camelcase
+contract Solana_Adapter is AdapterInterface, CircleCCTPAdapter {
+    /**
+     * @notice The official Circle CCTP MessageTransmitter contract endpoint.
+     * @dev Posted officially here: https://developers.circle.com/stablecoins/docs/evm-smart-contracts
+     */
+    // solhint-disable-next-line immutable-vars-naming
+    IMessageTransmitter public immutable cctpMessageTransmitter;
```
Reinis-FRP, line 27: This more belongs to CircleCCTPAdapter, though that would require passing it in constructor for all other adapters that don't use CCTP for general message transmission.

contracts/chain-adapters/Solana_Adapter.sol
```diff
@@ -0,0 +1,193 @@
...
+
+import { IMessageTransmitter, ITokenMessenger } from "../external/interfaces/CCTPInterfaces.sol";
+import { SpokePoolInterface } from "../interfaces/SpokePoolInterface.sol";
+import { AdapterInterface } from "./interfaces/AdapterInterface.sol";
+import { CircleCCTPAdapter, CircleDomainIds } from "../libraries/CircleCCTPAdapter.sol";
+
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+
+/**
+ * @notice Contract containing logic to send messages from L1 to Solana.
```
chrismaree, line 12:
```suggestion
 * @notice Contract containing logic to send messages from L1 to Solana via CCTP.
```

contracts/chain-adapters/Solana_Adapter.sol
```diff
@@ -0,0 +1,193 @@
...
+ * @dev Public functions calling external contracts do not guard against reentrancy because they are expected to be
+ * called via delegatecall, which will execute this contract's logic within the context of the originating contract.
+ * For example, the HubPool will delegatecall these functions, therefore it's only necessary that the HubPool's methods
+ * that call this contract's logic guard against reentrancy.
+ * @custom:security-contact bugs@across.to
+ */
+
+// solhint-disable-next-line contract-name-camelcase
+contract Solana_Adapter is AdapterInterface, CircleCCTPAdapter {
```
chrismaree, line 21: nit: is this a "solana_adapter" or it is better called "Cctp_Adapter"?

contracts/chain-adapters/Solana_Adapter.sol
```diff
@@ -0,0 +1,193 @@
...
+     * @param solanaUsdc USDC mint address on Solana, decoded from Base58 to bytes32.
+     * @param solanaSpokePoolUsdcVault USDC token address on Solana for the spoke pool, decoded from Base58 to bytes32.
+     */
+    constructor(
+        IERC20 _l1Usdc,
+        ITokenMessenger _cctpTokenMessenger,
+        IMessageTransmitter _cctpMessageTransmitter,
+        bytes32 solanaSpokePool,
+        bytes32 solanaUsdc,
+        bytes32 solanaSpokePoolUsdcVault
```
chrismaree, line 74: nit: it might be easier/cleaner/simpler/more consistant to just have these hard coded as constants, as done in some of the other [adapters](https://github.com/across-protocol/contracts/blob/5f08c27dd09f8c674ccc4af39d8b6b7c87a0444e/contracts/chain-adapters/Optimism_Adapter.sol#L52).

contracts/chain-adapters/Solana_Adapter.sol
```diff
@@ -0,0 +1,193 @@
...
+    }
+
+    /**
+     * @notice Send cross-chain message to target on Solana.
+     * @dev Only allows sending messages to the Solana spoke pool.
+     * @param target Program on Solana (translated as EVM address) that will receive message.
+     * @param message Data to send to target.
+     */
+    function relayMessage(address target, bytes calldata message) external payable override {
+        if (target != SOLANA_SPOKE_POOL_ADDRESS) {
```
chrismaree, line 102: what is the downside with letting this target non spoke pool address?

contracts/chain-adapters/Solana_Adapter.sol
```diff
@@ -0,0 +1,193 @@
...
+            revert InvalidRelayMessageTarget(target);
+        }
+
+        bytes4 selector = bytes4(message[:4]);
+        if (selector == SpokePoolInterface.setEnableRoute.selector) {
```
chrismaree, line 107: this is great. it's nice that we only need to worry about one selector as the others are bool.

contracts/chain-adapters/Solana_Adapter.sol
```diff
@@ -0,0 +1,193 @@
...
+        if (l1Token != address(usdcToken)) {
+            revert InvalidL1Token(l1Token);
+        }
+        if (l2Token != SOLANA_USDC_ADDRESS) {
+            revert InvalidL2Token(l2Token);
+        }
+        if (amount > type(uint64).max) {
+            revert InvalidAmount(amount);
+        }
+        if (to != SOLANA_SPOKE_POOL_ADDRESS) {
```
chrismaree, line 144: similar comment to before: if the `to` is _not_ the spoke address, it in theory is ok?

I agree constraining the execution flow is properly better but in theory there is no issue not having this require, right?

test/evm/hardhat/chain-adapters/Solana_Adapter.ts
```diff
@@ -0,0 +1,140 @@
...
+let solanaSpokePoolBytes32: string,
+  solanaUsdcBytes32: string,
+  solanaSpokePoolUsdcVaultBytes32: string,
+  solanaSpokePoolAddress: string,
+  solanaUsdcAddress: string;
+
+const solanaChainId = 1234567890; // TODO: Decide how to represent Solana in Across as it does not have a chainId.
+const solanaDomainId = 5;
+
+describe("Solana Chain Adapter", function () {
```
chrismaree, line 43: this is great

contracts/chain-adapters/Solana_Adapter.sol
md0x: Having it hardcoded potentially limits flexibility for future updates or use cases. Maybe we can add a setter or remove the check if not strictly required as @chrismaree is exploring.

contracts/chain-adapters/Solana_Adapter.sol
```diff
@@ -0,0 +1,193 @@
...
+        SOLANA_SPOKE_POOL_USDC_VAULT = solanaSpokePoolUsdcVault;
+    }
+
+    /**
+     * @notice Send cross-chain message to target on Solana.
+     * @dev Only allows sending messages to the Solana spoke pool.
+     * @param target Program on Solana (translated as EVM address) that will receive message.
+     * @param message Data to send to target.
+     */
+    function relayMessage(address target, bytes calldata message) external payable override {
```
md0x, line 101: Do we need permission checks here?

contracts/chain-adapters/Solana_Adapter.sol
```diff
@@ -0,0 +1,193 @@
...
+    /**
+     * @notice Helper to map a Solana address to an Ethereum address representation.
+     * @dev The Ethereum address is derived from the Solana address by hashing it and then truncating to its lowest 20
+     * bytes. This same conversion must be done by the HubPool owner when adding Solana spoke pool and setting the
+     * corresponding pool rebalance and deposit routes.
+     * @param solanaAddress Solana address (Base58 decoded to bytes32) to map to its Ethereum address representation.
+     * @return Ethereum address representation of the Solana address.
+     */
+    function _mapSolanaAddress(bytes32 solanaAddress) internal pure returns (address) {
+        return address(uint160(uint256(keccak256(abi.encodePacked(solanaAddress)))));
```
nicholaspai, line 163: Why do you need to hash the bytes32 string?

nicholaspai (commented):
One driveby comment

contracts/chain-adapters/Solana_Adapter.sol
Reinis-FRP: This includes logic specific to Solana, e.g. how `setEnableRoute` is translated. I was though considering to move `cctpMessageTransmitter` to base `CircleCCTPAdapter`, but that would have required adding it to constructor for all other adapters that don't use CCTP for general message transmission.

contracts/chain-adapters/Solana_Adapter.sol
Reinis-FRP: I left them as immutables so that its easier to test integration flow on testnets that would have different USDC and we don't have a proper authority for the spoke yet.

contracts/chain-adapters/Solana_Adapter.sol
Reinis-FRP: This accepts target address that must be translated to bytes32, but since we don't have translation mapping for any other target then obviously the caller has been misconfigured.

As for updates, I understand the general pattern for these adapters is to have them immutable and upgrade by redeploying and updating the instance in HubPool.

contracts/chain-adapters/Solana_Adapter.sol
Reinis-FRP: We don't have recipient ATA for any other than spoke. If this does not match then the caller has been misconfigured, so better to block it here.

test/evm/hardhat/chain-adapters/Solana_Adapter.ts
```diff
@@ -0,0 +1,140 @@
...
+let hubPool: Contract, solanaAdapter: Contract, weth: Contract, usdc: Contract, timer: Contract, mockSpoke: Contract;
+let owner: SignerWithAddress, dataWorker: SignerWithAddress, liquidityProvider: SignerWithAddress;
+let cctpTokenMessenger: FakeContract, cctpMessageTransmitter: FakeContract, cctpTokenMinter: FakeContract;
+let solanaSpokePoolBytes32: string,
+  solanaUsdcBytes32: string,
+  solanaSpokePoolUsdcVaultBytes32: string,
+  solanaSpokePoolAddress: string,
+  solanaUsdcAddress: string;
+
+const solanaChainId = 1234567890; // TODO: Decide how to represent Solana in Across as it does not have a chainId.
```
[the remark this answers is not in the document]
Reinis-FRP: This should not clash with any future EVM we might need to support and should fit in u64 as max supported on Solana. So one option is to use 8 bytes from the hashed cluster name. E.g. `uint64(uint256(keccak256(abi.encodePacked("solana-mainnet"))))` is `11938232455623604379`

contracts/chain-adapters/Solana_Adapter.sol
Reinis-FRP: this is delegatecalled, so we check CCTP sender on Solana side matches HubPool

contracts/chain-adapters/Solana_Adapter.sol
chrismaree: @Reinis-FRP that's a good point. this is _not_ a generic adapter as a result of the encoding so there is no situation where you should ever be passing in anything other than a known target address. if you are, you're doing something wrong and it should error. I agree with leaving it like this.

test/evm/hardhat/chain-adapters/Solana_Adapter.ts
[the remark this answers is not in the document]
chrismaree: `uint64(uint256(keccak256(abi.encodePacked("solana-mainnet")))` is great. this is pretty much exactly what we've wanted from the start and I think is a good structure to use. 

it's a bit irritating that we've got two notions of "chainId" now; the "across-standard" which is EVM first and then our own definition of non-evm chains (such as this solana chainId) and then internal CCTP chain IDs, where solana is 5.

contracts/chain-adapters/Solana_Adapter.sol
Reinis-FRP: Initially I added hash just to be on the safe side to avoid collisions, but we only here have couple of addresses and they already should be pretty homogeneous. The only case where this would matter is native system programs that have zeros in their trailing bytes. But we need to represent only key or seed derived addresses on EVM, so their bytes32 representation should be pretty homogeneous.

Pushed a commit to drop the hashing and only trim off the first 12 bytes.

contracts/chain-adapters/Solana_Adapter.sol
chrismaree: I think they are ok to have adapter specific errors? if there are other errors that other adapters re-use we should cleary use them but things like `InvalidCctpTokenMessenger` only apply in this context.

chrismaree (approved):

PR #644, opened 2024-10-09
feat: world chain spoke pool

bmzig:
This makes a slightly modified world chain spoke pool, which is largely based on `Optimism_SpokePool`. The contract of the USDC bridge can be found here: https://worldchain-mainnet.explorer.alchemy.com/address/0xbD80b06d3dbD0801132c6689429aC09Ca6D27f82?tab=contract_code. We need to approve this contract due to this line of code, https://github.com/defi-wonderland/opUSDC/blob/ef22e5731f1655bf5249b2160452cce9aa06ff3f/src/contracts/L2OpUSDCBridgeAdapter.sol#L304. Finally, I make the assumption that USDC is now passed in as a constructor argument and NOT set as the zero address. This required a change to the deploy script (and will require a change to the constants repository). This should be OK, since we check whether CCTP is activated by looking at the token messenger contract, which should still be set to the zero address.

### DIFF
--- a/contracts/WorldChain_SpokePool.sol
+++ b/contracts/WorldChain_SpokePool.sol
@@ -5,11 +5,25 @@ import "@eth-optimism/contracts/libraries/constants/Lib_PredeployAddresses.sol";
 import "./Ovm_SpokePool.sol";
 import "./external/interfaces/CCTPInterfaces.sol";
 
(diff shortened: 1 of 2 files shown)


pxrl:
@bmzig Oops, that was a mistake. Can you re-open?

pxrl:
Reopening.

PR #643, opened 2024-10-09
fix: Support OP bridged USDC in WorldChain adapter

pxrl:
World Chain implements Circle's bridged USDC standard. Deposits & withdrawals cannot be initiated via the L1StandardBridge and L2StandardBridge contracts respectively. Instead they must be sent to World Chain's own implementations of Circle's bridged USDC standard.

Circle's standard is described here: https://github.com/circlefin/stablecoin-evm/blob/master/doc/bridged_USDC_standard.md

The process for bridged USDC deposits & withdrawals is here:
https://github.com/defi-wonderland/opUSDC/tree/main?tab=readme-ov-file#l1--l2-usdc-canonical-bridging

PR #646, opened 2024-10-09
fix: Support OP bridged USDC in WorldChain adapter

pxrl:
World Chain implements Circle's bridged USDC standard. Deposits &
withdrawals cannot be initiated via the L1StandardBridge and
L2StandardBridge contracts respectively. Instead they must be sent to
World Chain's own implementations of Circle's bridged USDC standard.

In recognition of the fact that this will likely be a repeating pattern
for new deployments, restyle the World Chain adapter as a generic OP
adapter and instead rely on deployment scripts to configure it with the 
correct bridges.

Circle's standard is described here:
https://github.com/circlefin/stablecoin-evm/blob/master/doc/bridged_USDC_standard.md

The process for bridged USDC deposits & withdrawals is here:
https://github.com/defi-wonderland/opUSDC/tree/main?tab=readme-ov-file#l1--l2-usdc-canonical-bridging

### DIFF
--- a/contracts/chain-adapters/WorldChain_Adapter.sol
+++ b/contracts/chain-adapters/WorldChain_Adapter.sol
@@ -1,19 +1,19 @@
 // SPDX-License-Identifier: BUSL-1.1
 pragma solidity ^0.8.0;
 
-import "./interfaces/AdapterInterface.sol";
-import "../external/interfaces/WETH9Interface.sol";
-
-// @dev Use local modified CrossDomainEnabled contract instead of one exported by eth-optimism because we need
-// this contract's state variables to be `immutable` because of the delegateCall call.
-import "./CrossDomainEnabled.sol";
 import "@eth-optimism/contracts/L1/messaging/IL1StandardBridge.sol";
-
 import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
 
-import "../libraries/CircleCCTPAdapter.sol";
-import "../external/interfaces/CCTPInterfaces.sol";
+import { CircleCCTPAdapter, CircleDomainIds } from "../libraries/CircleCCTPAdapter.sol";
+import { ITokenMessenger } from "../external/interfaces/CCTPInterfaces.sol";
+import { IOpUSDCBridgeAdapter } from "../external/interfaces/IOpUSDCBridgeAdapter.sol";
+import { WETH9Interface } from "../external/interfaces/WETH9Interface.sol";
+import { AdapterInterface } from "./interfaces/AdapterInterface.sol";
+
+// @dev Use local modified CrossDomainEnabled contract instead of one exported by eth-optimism because we need
+// this contract's state variables to be `immutable` because of the delegateCall call.
+import { CrossDomainEnabled } from "./CrossDomainEnabled.sol";
 
 /**
  * @notice Contract containing logic to send messages from L1 to World Chain. This is a clone of the Base/Mode adapter
@@ -32,6 +32,7 @@ contract WorldChain_Adapter is CrossDomainEnabled, AdapterInterface, CircleCCTPA
     WETH9Interface public immutable L1_WETH;
 
     IL1StandardBridge public immutable L1_STANDARD_BRIDGE;
+    IOpUSDCBridgeAdapter public immutable L1_OP_USDC_BRIDGE;
 
     /**
      * @notice Constructs new Adapter.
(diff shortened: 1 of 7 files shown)


pxrl:
nb. This is a resubmission of https://github.com/across-protocol/contracts/pull/643, with the target/base set to `audit-latest` instead of `master`. This avoids all the subsequent changes that exist in `master` but have not been audited.

contracts/chain-adapters/WorldChain_Adapter.sol
```diff
@@ -85,15 +88,17 @@ contract WorldChain_Adapter is CrossDomainEnabled, AdapterInterface, CircleCCTPA
...
-            _transferUsdc(to, amount);
+        } else if (l1Token == address(usdcToken)) {
+            if (_isCCTPEnabled()) {
+                _transferUsdc(to, amount);
+            } else {
+                // Use WorldChain's OP USDC bridge to received bridged USDC on L2.
+                IERC20(l1Token).safeIncreaseAllowance(address(L1_OP_USDC_BRIDGE), amount);
+                L1_OP_USDC_BRIDGE.sendMessage(to, amount, L2_GAS_LIMIT);
+            }
         } else {
```
pxrl, line 99: There's talk of wanting to reuse this on future adapters so we don't need audit on those. Are we OK w/ a copy/paste of this contract on future OP stack chains? That's what we have done previously for Mode, Lisk, Redstone & Zora.

I'm tempted to factor it out somehow though, such that it's more modular and we can more clearly identify it as pre-audited code.

contracts/chain-adapters/WorldChain_Adapter.sol
```diff
@@ -42,9 +43,10 @@ contract WorldChain_Adapter is CrossDomainEnabled, AdapterInterface, CircleCCTPA
      */
     constructor(
         WETH9Interface _l1Weth,
+        IERC20 _l1Usdc,
         address _crossDomainMessenger,
         IL1StandardBridge _l1StandardBridge,
-        IERC20 _l1Usdc
```
nicholaspai, line 47: Why move this param around? Seems like it could cause errors

contracts/chain-adapters/WorldChain_Adapter.sol
```diff
@@ -32,6 +32,7 @@ contract WorldChain_Adapter is CrossDomainEnabled, AdapterInterface, CircleCCTPA
     WETH9Interface public immutable L1_WETH;
 
     IL1StandardBridge public immutable L1_STANDARD_BRIDGE;
+    IOpUSDCBridgeAdapter public immutable L1_OP_USDC_BRIDGE;
```
nicholaspai, line 35: Should this be name have Cctp in it somehow since it's a cctp system contract not an op stack one?

contracts/chain-adapters/WorldChain_Adapter.sol
nicholaspai: Can we not put this in the ovm adapter?

nicholaspai (commented):
Looks good

contracts/chain-adapters/WorldChain_Adapter.sol
pxrl: I did it to group WETH and USDC together, and to then keep the OP contract addresses together. I've updated the deployment script so that it follows along, but can revert this if it's too risky.

contracts/chain-adapters/WorldChain_Adapter.sol
pxrl: It's not actually a CCTP contract, it's a third-party implementation of a Circle interface that will support a migration to CCTP, but from what I can see the actual deposit/withdrawal methods are implementation-defined. So I believe this interface is actually hard-coupled to OP chains that choose to use the interface.

contracts/chain-adapters/WorldChain_Adapter.sol
pxrl: Yeah; we don't have an OVM adapter but I agree it'd be best to implement one - then we get maximum reuse here.

contracts/chain-adapters/WorldChain_Adapter.sol
nicholaspai: It's fine

contracts/chain-adapters/WorldChain_Adapter.sol
nicholaspai: But is this change required for any other chains today?

pxrl: I'm not aware of any that don't use L1StandardBridge for bridged USDC, so this is a new requirement. I think we should anticipate it popping up on subsequent new deployments, though.

nicholaspai: Understood so let's just not rename this contract to `OpStack_Adapter` or something more generic for the purpose of keeping this PR diff small

bmzig (approved):

PR #648, opened 2024-10-09
fix: add receive() to withdrawal helpers

bmzig:
The issue:
> When handling WETH withdrawals the withdrawToken function of the Ovm_WithdrawalHelper attempts to unwrap WETH and bridge the ether received to the recipient. However, the Ovm_WithdrawalHelper contract does not expose a receive() function to accept ether transfers. This means that calls to the WETH contract to unwrap the ether will fail. As a consequence any call to withdrawToken function of the Ovm_WithdrawalHelper contract which specifies the L2 wrappedNativeToken will fail.
As there is no other way to withdraw tokens from the contract the L2 wrapped native token sent to this contract cannot be retrieved without a contract upgrade.


bmzig:
Closing in favor of https://github.com/across-protocol/contracts/pull/725

PR #649, opened 2024-10-09
fix(SpokePool): Enforce exclusivity consistently

pxrl:
The new relative exclusivity check has not been propagated to fillV3RelayWithUpdatedDeposit() or requestV3SlowFill(). Identified via test case failures in the relayer.

This is a very-low criticality issue. It's very unlikely that any relayer would be trying to call `fillV3RelayWithUpdatedDeposit()` within the exclusivity window. There's a slightly higher chance of hitting this issue on slow fill requests because the RL relayer is fast enough to hit the probable exclusivity windows on some routes for larger transfers, which is where a token shortfall is more likely.

### DIFF
--- a/contracts/SpokePool.sol
+++ b/contracts/SpokePool.sol
@@ -936,7 +936,6 @@ abstract contract SpokePool is
         // this deadline, not slow filled. As a simplifying assumption, we will not allow slow fills to be requested
         // during this exclusivity period.
         if (
-            relayData.exclusiveRelayer != msg.sender &&
             relayData.exclusivityDeadline >= getCurrentTime() &&
             relayData.exclusiveRelayer != address(0)
         ) {
--- a/contracts/SpokePool.sol
+++ b/contracts/SpokePool.sol
@@ -935,7 +935,13 @@ abstract contract SpokePool is
         // fast fill within this deadline. Moreover, the depositor should expect to get *fast* filled within
         // this deadline, not slow filled. As a simplifying assumption, we will not allow slow fills to be requested
         // during this exclusivity period.
-        if (relayData.exclusivityDeadline >= getCurrentTime()) revert NoSlowFillsInExclusivityWindow();
+        if (
+            relayData.exclusiveRelayer != msg.sender &&
+            relayData.exclusivityDeadline >= getCurrentTime() &&
+            relayData.exclusiveRelayer != address(0)
+        ) {
+            revert NotExclusiveRelayer();
+        }
         if (relayData.fillDeadline < getCurrentTime()) revert ExpiredFillDeadline();
 
         bytes32 relayHash = _getV3RelayHash(relayData);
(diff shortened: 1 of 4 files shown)


contracts/SpokePool.sol
```diff
@@ -885,7 +885,11 @@ abstract contract SpokePool is
...
         // Exclusivity deadline is inclusive and is the latest timestamp that the exclusive relayer has sole right
         // to fill the relay.
-        if (relayData.exclusivityDeadline >= getCurrentTime() && relayData.exclusiveRelayer != msg.sender) {
+        if (
+            relayData.exclusiveRelayer != msg.sender &&
+            relayData.exclusivityDeadline >= getCurrentTime() &&
+            relayData.exclusiveRelayer != address(0)
+        ) {
             revert NotExclusiveRelayer();
         }
```
pxrl, line 894: We now have this check in 3 places; worth factoring out to a function?

contracts/SpokePool.sol
```diff
@@ -931,7 +935,13 @@ abstract contract SpokePool is
         // fast fill within this deadline. Moreover, the depositor should expect to get *fast* filled within
         // this deadline, not slow filled. As a simplifying assumption, we will not allow slow fills to be requested
         // during this exclusivity period.
-        if (relayData.exclusivityDeadline >= getCurrentTime()) revert NoSlowFillsInExclusivityWindow();
+        if (
+            relayData.exclusiveRelayer != msg.sender &&
```
pxrl, line 939:
```suggestion
```

nicholaspai (approved):
Dang good catch. I think refactoring to shared internal function could make sense

nicholaspai (changes requested):
Can you add unit tests to these functions similar to these so we can prevent regressions? https://github.com/across-protocol/contracts/blob/master/test/evm/hardhat/SpokePool.Relay.ts#L330

contracts/SpokePool.sol
pxrl: I factored out the common logic to a `_fillIsExclusive()` function. I think this is probably the right call but could be persuaded otherwise.

contracts/interfaces/V3SpokePoolInterface.sol
```diff
@@ -222,7 +222,6 @@ interface V3SpokePoolInterface {
     error DisabledRoute();
     error InvalidQuoteTimestamp();
     error InvalidFillDeadline();
-    error InvalidExclusiveRelayer();
```
pxrl, line 225: Driveby change - this error is unused. It was dropped with the recent changes to permit relative exclusivity.

pxrl:
> Can you add unit tests to these functions similar to these so we can prevent regressions? https://github.com/across-protocol/contracts/blob/master/test/evm/hardhat/SpokePool.Relay.ts#L330

Added tests here: https://github.com/across-protocol/contracts/pull/649/commits/6b5f5f9f65a927bb272b59e1a5fe481b22c06742

nicholaspai (approved):

PR #650, opened 2024-10-10
feat: address to bytes32 contract changes

md0x:
Changes proposed in this PR:
- Move addresses to bytes32 to support SVM


contracts/SpokePool.sol
```diff
@@ -535,22 +561,22 @@ abstract contract SpokePool is
...
+        bytes32 exclusiveRelayer,
         uint32 quoteTimestamp,
         uint32 fillDeadline,
         uint32 exclusivityPeriod,
         bytes calldata message
     ) public payable override nonReentrant unpausedDeposits {
         // Check that deposit route is enabled for the input token. There are no checks required for the output token
         // which is pulled from the relayer at fill time and passed through this contract atomically to the recipient.
-        if (!enabledDepositRoutes[inputToken][destinationChainId]) revert DisabledRoute();
+        if (!enabledDepositRoutes[inputToken.toAddress()][destinationChainId]) revert DisabledRoute();
```
md0x, line 579: Should we changed the data structure to:

`mapping(bytes32 => mapping(uint256 => bool)) public enabledDepositRoutes;`

Reinis-FRP: I think yes, as that would avoid unnecessary conversion, and it would still use the same storage slot for the mapped values.

contracts/SpokePool.sol
```diff
@@ -753,19 +779,19 @@ abstract contract SpokePool is
...
-        address depositor,
+        bytes32 depositor,
         uint32 depositId,
         uint256 updatedOutputAmount,
-        address updatedRecipient,
+        bytes32 updatedRecipient,
         bytes calldata updatedMessage,
         bytes calldata depositorSignature
     ) public override nonReentrant {
         _verifyUpdateV3DepositMessage(
```
Reinis-FRP, line 789: maybe better to change interface of _verifyUpdateV3DepositMessage

contracts/SpokePool.sol
```diff
@@ -753,19 +779,19 @@ abstract contract SpokePool is
...
-            depositor,
+            depositor.toAddress(),
```
Reinis-FRP, line 790: maybe also update _verifyUpdateV3DepositMessage interface so we don't need to translate to address here

contracts/SpokePool.sol
```diff
@@ -836,9 +864,9 @@ abstract contract SpokePool is
         // Exclusivity deadline is inclusive and is the latest timestamp that the exclusive relayer has sole right
         // to fill the relay.
         if (
-            relayData.exclusiveRelayer != msg.sender &&
+            relayData.exclusiveRelayer.toAddress() != msg.sender &&
             relayData.exclusivityDeadline >= getCurrentTime() &&
-            relayData.exclusiveRelayer != address(0)
+            relayData.exclusiveRelayer.toAddress() != address(0)
```
Reinis-FRP, line 869:
```suggestion
            relayData.exclusiveRelayer != bytes32(0)
```

contracts/libraries/AddressConverters.sol
```diff
@@ -0,0 +1,14 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.0;
+
+library Bytes32ToAddress {
+    function toAddress(bytes32 _bytes32) internal pure returns (address) {
```
Reinis-FRP, line 5: should this also check that highest 12 bytes are all 0? Also, we should avoid using this method in comparisons where possible and convert the other value to bytes32 first

contracts/SpokePool.sol
```diff
@@ -979,7 +1009,7 @@ abstract contract SpokePool is
     ) public override nonReentrant {
         V3RelayData memory relayData = slowFillLeaf.relayData;
 
-        _preExecuteLeafHook(relayData.outputToken);
+        _preExecuteLeafHook(relayData.outputToken.toAddress());
```
Reinis-FRP, line 1012: shall we also change this _preExecuteLeafHook

contracts/SpokePool.sol
```diff
@@ -141,9 +146,12 @@ abstract contract SpokePool is
     // token for the input token. By using this magic value, off-chain validators do not have to keep
     // this event in their lookback window when querying for expired deposts.
     uint32 public constant INFINITE_FILL_DEADLINE = type(uint32).max;
-    /****************************************
```
chrismaree, line 144: you might not have meant to modify bits like this?

contracts/libraries/AddressConverters.sol
```diff
@@ -0,0 +1,14 @@
...
+        return address(uint160(uint256(_bytes32)));
+    }
+}
+
+library AddressToBytes32 {
+    function toBytes32(address _address) internal pure returns (bytes32) {
```
chrismaree, line 11: I like this. we should re-use it here as well: https://github.com/across-protocol/contracts/blob/5f08c27dd09f8c674ccc4af39d8b6b7c87a0444e/contracts/libraries/CircleCCTPAdapter.sol#L71

contracts/SpokePool.sol
```diff
@@ -37,6 +38,8 @@ abstract contract SpokePool is
 {
     using SafeERC20Upgradeable for IERC20Upgradeable;
     using AddressLibUpgradeable for address;
+    using Bytes32ToAddress for bytes32;
```
chrismaree, line 41: this is actually pretty clean. good job.

contracts/SpokePool.sol
md0x: No, I have some issues with the formatter, I will try to fix my formatter or manually revert this changes

contracts/libraries/AddressConverters.sol
nicholaspai: do you think that will get too expensive gas-cost wise? Maybe we can have safeToAddress and unsafeToAddress and use them depending on the context

contracts/SpokePool.sol
```diff
@@ -836,9 +864,9 @@ abstract contract SpokePool is
         // Exclusivity deadline is inclusive and is the latest timestamp that the exclusive relayer has sole right
         // to fill the relay.
         if (
```
pxrl, line 866: I have a PR [here](https://github.com/across-protocol/contracts/pull/649) that factors this out to a function. Seems like that will go in, so it might be worth considering that here (it should permit a slight simplification of your change).

test/evm/hardhat/fixtures/SpokePool.Fixture.ts
```diff
@@ -449,7 +450,7 @@ export async function getUpdatedV3DepositSignature(
         { name: "depositId", type: "uint32" },
         { name: "originChainId", type: "uint256" },
         { name: "updatedOutputAmount", type: "uint256" },
-        { name: "updatedRecipient", type: "address" },
+        { name: "updatedRecipient", type: "bytes32" },
```
chrismaree, line 453: ^

contracts/Polygon_SpokePool.sol
```diff
@@ -146,7 +146,8 @@ contract Polygon_SpokePool is IFxMessageProcessor, SpokePool, CircleCCTPAdapter
      * @param data ABI encoded function call to execute on this contract.
      */
     function processMessageFromRoot(
-        uint256, /*stateId*/
```
chrismaree, line 149: ^

contracts/SpokePool.sol
```diff
@@ -1221,37 +1421,75 @@ abstract contract SpokePool is
...
+    function _verifyUpdateV3DepositMessage(
+        address depositor,
+        uint32 depositId,
+        uint256 originChainId,
+        uint256 updatedOutputAmount,
+        address updatedRecipient,
+        bytes memory updatedMessage,
+        bytes memory depositorSignature
+    ) internal view {
+        _verifyUpdateV3DepositMessageHelper(
```
Reinis-FRP, line 1483: do we really need this extra helper? might be simpler to change the interface of _verifyUpdateV3DepositMessage to accept additional hashType parameter (see comment above).

contracts/erc7683/ERC7683.sol
```diff
@@ -6,10 +6,10 @@ pragma solidity ^0.8.0;
 struct GaslessCrossChainOrder {
     /// @dev The contract address that the order is meant to be settled by.
     /// Fillers send this order to this contract address on the origin chain
-    address originSettler;
+    bytes32 originSettler;
```
Reinis-FRP, line 9: Does this require any changes in ERC proposal?

contracts/interfaces/V3SpokePoolInterface.sol
```diff
@@ -97,67 +97,82 @@ interface V3SpokePoolInterface {
...
+        uint256 destinationChainId,
+        bytes32 exclusiveRelayer,
+        uint32 quoteTimestamp,
+        uint32 fillDeadline,
+        uint32 exclusivityDeadline,
+        bytes calldata message
+    ) external payable;
+
     function depositV3(
         address depositor,
```
Reinis-FRP, line 177: Do we need to support backwards overload in the interface?

contracts/libraries/CircleCCTPAdapter.sol
```diff
@@ -88,7 +81,7 @@ abstract contract CircleCCTPAdapter {
      * @param amount Amount of USDC to transfer.
      */
     function _transferUsdc(address to, uint256 amount) internal {
-        _transferUsdc(_addressToBytes32(to), amount);
+        _transferUsdc(to.toBytes32(), amount);
```
Reinis-FRP, line 84: neat that we can reuse the lib here :)

contracts/SpokePool.sol
```diff
@@ -785,6 +930,59 @@ abstract contract SpokePool is
...
+     * otherwise the relayer has no incentive to use this updated value.
+     * @param updatedRecipient The new recipient for this deposit. Can be modified if the original recipient is a contract that
+     * expects to receive a message from the relay and needs to be changed.
+     * @param updatedMessage The new message for this deposit. Can be modified if the recipient is a contract that expects
+     * to receive a message from the relay and needs to be updated.
+     * @param depositorSignature The signed EIP712 hashstruct containing the deposit ID. Should be signed by the depositor account.
+     * If the depositor is a contract, it should implement EIP1271 to sign as a contract. See `_verifyUpdateV3DepositMessage()`
+     * for more details on how the signature should be constructed.
+     */
+    function speedUpV3Deposit(
```
chrismaree, line 956: question: is it useful to have both this `address` and `bytes` versions? do we expect intergrators (UIs) to call the `address ` or `bytes` version?  the context of my question is that seeing we wont support speedup on svm domain (requiring bytes32) is it useful to have this? 

I think yes: we should have it, as we can then support it going forward and be consistent with the interface changes , but just through I'd bring it up to see what you think.

contracts/SwapAndBridge.sol
```diff
@@ -131,39 +133,42 @@ abstract contract SwapAndBridgeBase is Lockable, MultiCaller {
...
+     */
     function _checkSwapOutputAndDeposit(
         uint256 swapTokenAmount,
         uint256 swapTokenBalanceBefore,
         uint256 inputTokenBalanceBefore,
         uint256 minExpectedInputTokenAmount,
         DepositData calldata depositData,
-        IERC20 _swapToken,
-        IERC20 _acrossInputToken
+        bytes32 _swapToken,
```
chrismaree, line 144: I know this is not your change, but its funny to me that these two variables are `_` and the others are not.

contracts/erc7683/ERC7683.sol
chrismaree: yes. the ERC itself. is being updated by Matt and others.

contracts/SpokePool.sol
md0x: `_verifyUpdateV3DepositMessage` is also overloaded with 2 implementations, one for address and one for `bytes32`. I added the helper to reduce code duplication. Without it each `_verifyUpdateV3DepositMessage` would need to implement `_hashTypedDataV4(keccak256(abi.encode...` does it make sense?

contracts/SpokePool.sol
```diff
@@ -1221,37 +1421,75 @@ abstract contract SpokePool is
...
+            originChainId,
+            updatedOutputAmount,
+            updatedRecipient,
+            updatedMessage,
+            UPDATE_V3_DEPOSIT_DETAILS_HASH,
+            depositorSignature
+        );
+    }
+
+    function _verifyUpdateV3DepositMessage(
```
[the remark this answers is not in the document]
md0x: See my answer here: https://github.com/across-protocol/contracts/pull/650/files#r1808446450

contracts/Linea_SpokePool.sol
```diff
@@ -15,6 +15,7 @@ import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
  */
 contract Linea_SpokePool is SpokePool {
     using SafeERC20 for IERC20;
+    using AddressToBytes32 for address;
```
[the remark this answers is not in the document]
md0x: Sorry @Reinis-FRP What do you mean here?
These are imported in the SpokePool.sol and used here

contracts/SpokePool.sol
```diff
@@ -160,15 +168,15 @@ abstract contract SpokePool is
         uint256[] refundAmounts,
         uint32 indexed rootBundleId,
         uint32 indexed leafId,
-        address l2TokenAddress,
-        address[] refundAddresses,
+        bytes32 l2TokenAddress,
+        bytes32[] refundAddresses,
         address caller
```
md0x, line 173: Should we make `caller` a `bytes32` here for consistency with other events?

contracts/SpokePool.sol
```diff
@@ -160,15 +168,15 @@ abstract contract SpokePool is
...
     );
     event TokensBridged(
         uint256 amountToReturn,
         uint256 indexed chainId,
         uint32 indexed leafId,
-        address indexed l2TokenAddress,
+        bytes32 indexed l2TokenAddress,
         address caller
```
md0x, line 180: Should we make `caller` a `bytes32` here for consistency with other events?

contracts/SpokePool.sol
```diff
@@ -1039,16 +1237,17 @@ abstract contract SpokePool is
         SpokePoolInterface.RelayerRefundLeaf memory relayerRefundLeaf,
         bytes32[] memory proof
     ) public payable virtual override nonReentrant {
-        _preExecuteLeafHook(relayerRefundLeaf.l2TokenAddress);
+        _preExecuteLeafHook(relayerRefundLeaf.l2TokenAddress.toBytes32());
```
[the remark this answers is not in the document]
md0x: I kept it to prevent extra castings in the [_distributeRelayerFunds](https://github.com/across-protocol/contracts/blob/48e402374969662ea666192a7926d7f8c8b08949/contracts/SpokePool.sol#L1366) function but you are right, I think these are negligible so I've updated the interface now.

contracts/interfaces/V3SpokePoolInterface.sol
md0x: Good point. I've removed them from the interface 👍

contracts/SpokePool.sol
md0x: I think so! But open to remove them if someone disagree.

contracts/Linea_SpokePool.sol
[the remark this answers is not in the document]
Reinis-FRP: sorry, meant using curly braces to explicitly state what contracts/libraries from source files we are importing

contracts/SpokePool.sol
Reinis-FRP: I meant avoiding overload in _verifyUpdateV3DepositMessage and let its caller in the external function to do address->bytes32 conversion when needed, as well as pass the relevant hashType from constants.

contracts/SpokePool.sol
Reinis-FRP: I think yes, that would be more consistent

PR #651, opened 2024-10-10
feat: Add simple leaf testnet script & LUT closer script

chrismaree:
This PR adds two new scripts for SVM things:
1. script to show, on testnet, how a batch of up to 25 relayer repayments can be executed. its meant for testing only.
2. script to enable an account to close all LUTs they've opened. We'll use logic similar to this when we implement the finalizer eventually for solana logic.


Anchor.toml
```diff
@@ -28,7 +28,9 @@ simpleDeposit = "NODE_NO_WARNINGS=1 yarn run ts-node ./scripts/svm/simpleDeposit
 queryDeposits = "NODE_NO_WARNINGS=1 yarn run ts-node ./scripts/svm/queryDeposits.ts"
 queryFills = "NODE_NO_WARNINGS=1 yarn run ts-node ./scripts/svm/queryFills.ts"
 simpleFill = "NODE_NO_WARNINGS=1 yarn run ts-node ./scripts/svm/simpleFill.ts"
+simpleFakeRelayerRepayment = "NODE_NO_WARNINGS=1 yarn run ts-node ./scripts/svm/simpleFakeRelayerRepayment.ts"
 closeRelayerPdas = "NODE_NO_WARNINGS=1 yarn run ts-node ./scripts/svm/closeRelayerPdas.ts"
+closeDataworkerLookUpTables = "NODE_NO_WARNINGS=1 yarn run ts-node ./scripts/svm/closeDataworkerLookUpTables.ts"
```
Reinis-FRP, line 33:
```suggestion
closeDataWorkerLookUpTables = "NODE_NO_WARNINGS=1 yarn run ts-node ./scripts/svm/closeDataWorkerLookUpTables.ts"
```
This did not resolve on Linux where filenames are case sensitive

scripts/svm/closeDataWorkerLookUpTables.ts
```diff
@@ -0,0 +1,71 @@
...
+
+// Set up the provider
+const provider = anchor.AnchorProvider.env();
+anchor.setProvider(provider);
+const connection = provider.connection;
+const payer = provider.wallet.publicKey;
+
+// Parse arguments
+// const argv = yargs(hideBin(process.argv))
+//   .option("payer", { type: "string", default: payer.toBase58(), describe: "Payer public key" }).argv;
```
Reinis-FRP, line 15: are these redundant?

scripts/svm/closeDataWorkerLookUpTables.ts
```diff
@@ -0,0 +1,71 @@
...
+        const deactivateSignature = await provider.sendAndConfirm(deactivateTransaction, [
+          (anchor.AnchorProvider.env().wallet as anchor.Wallet).payer,
+        ]);
+        console.log(`Deactivated: ${lookupTableAddress.toBase58()} | Tx Hash: ${deactivateSignature}`);
+      } catch {
+        console.error(`Pending deactivation: ${lookupTableAddress.toBase58()}`);
+      }
+
+      try {
+        // Attempt to close the lookup table
```
Reinis-FRP, line 49: Is this expected behavior I got that on the first run all my lookup tables get deactivated, but account closure fails. Once I rerun the script, deactivation fails, but account gets closed successfully?

Reinis-FRP: maybe there should be some await before trying to close account after the deactivation?

chrismaree: there is a cooldown period after deactivation. after they are de-activated, only then can they be closed.

it's not posible, at least wit the limited research that I did, to detect this middle phase. we might be able to catch the errors better and enfer the status but for now the above logic achieved this without over engineering the flow.

scripts/svm/closeDataWorkerLookUpTables.ts
chrismaree: yes, as only the payer (which is the unlocked account) can run this script as we need to send transactions from their address

Reinis-FRP (approved):

PR #653, opened 2024-10-11
feat: Add relayer repayment address

chrismaree:
This PR changes the interface of two existing EVM functions: `fillV3Relay` & `fillV3RelayWithUpdatedDeposit`. They are extended to now contain `repaymentAddress` as the address the relayer wants to get repaid on on chain `repaymentChainId`.

one open question: do we want to support the old `fillV3Relay` interface, without this new prop, that just internally maps this address to `msg.sender` on EVM? this will enable current relayers to continue to work without requiring them to upgrade. main reason against having this backwards compatibility is we are adding breaking changes in other places (eg here: https://github.com/across-protocol/contracts/pull/650) that will also force relayes to upgrade so it could be argued that we should simple upgrade all the things at once without backwards compatibility in any location.


pxrl:
This change looks clean to me; I don't have any comments to the changes. Will leave opportunity for others to comment before approving.

> one open question: do we want to support the old `fillV3Relay` interface, without this new prop, that just internally maps this address to `msg.sender` on EVM? this will enable current relayers to continue to work without requiring them to upgrade. main reason against having this backwards compatibility is we are adding breaking changes in other places (eg here: #650) that will also force relayes to upgrade so it could be argued that we should simple upgrade all the things at once without backwards compatibility in any location.

I think relayers will just deal with this but it's a headache for other integrators who are running infra against Across. Aggregators are an example - Socket and Li.Fi both have their own explorer implementations and they have specific knowledge about Across events. Wallet integrations is probably another stakeholder - Rabby can for instance decode Across transactions, so it'll probably impose some subtle breakage for them.

It's unfortunate they we'll be breaking some integrations and will be imposing re-work, though I should note it's not the first time that this will be happening because we already have an ABI break for events in the current audit (RE deterministic deposit IDs). I don't think there's necessarily anything to do differently here - we should just be extra persuasive in encouraging integrators to migrate to our new app SDK, which will permit us to handle the breakage internally, such that others can make the transition by upgrading their package.

contracts/SpokePool.sol
```diff
@@ -1311,7 +1311,7 @@ abstract contract SpokePool is
     // exclusiveRelayer if passed exclusivityDeadline or if slow fill.
     function _fillRelayV3(
```
chrismaree, line 1312: ^

contracts/SpokePool.sol
```diff
@@ -1311,7 +1311,7 @@ abstract contract SpokePool is
...
         V3RelayExecutionParams memory relayExecution,
-        address relayer,
+        bytes32 relayer,
```
mrice32, line 1314: Should we change this to `repaymentAddress` for consistency?

mrice32 (commented):
Only one comment

mrice32:
> This change looks clean to me; I don't have any comments to the changes. Will leave opportunity for others to comment before approving.
> 
> > one open question: do we want to support the old `fillV3Relay` interface, without this new prop, that just internally maps this address to `msg.sender` on EVM? this will enable current relayers to continue to work without requiring them to upgrade. main reason against having this backwards compatibility is we are adding breaking changes in other places (eg here: #650) that will also force relayes to upgrade so it could be argued that we should simple upgrade all the things at once without backwards compatibility in any location.
> 
> I think relayers will just deal with this but it's a headache for other integrators who are running infra against Across. Aggregators are an example - Socket and Li.Fi both have their own explorer implementations and they have specific knowledge about Across events. Wallet integrations is probably another stakeholder - Rabby can for instance decode Across transactions, so it'll probably impose some subtle breakage for them.
> 
> It's unfortunate they we'll be breaking some integrations and will be imposing re-work, though I should note it's not the first time that this will be happening because we already have an ABI break for events in the current audit (RE deterministic deposit IDs). I don't think there's necessarily anything to do differently here - we should just be extra persuasive in encouraging integrators to migrate to our new app SDK, which will permit us to handle the breakage internally, such that others can make the transition by upgrading their package.

I agree. I think we should batch any updates to interfaces/abis.

chrismaree:
> One open question is whether we should keep a backwards compatible fill interface.
> 
> This ties into how we're planning to orchestrate the migration. My straw man:
> 
> * We keep RelayData hash abi compatible (I think we can do this), so we know that old deposits and new deposits will collide if they have the same info.
> * Both fill functions can still work with both the old deposit event and new deposit event.
> * We keep both fill functions even after we do the deposit event upgrade, so relayers must start watching the new event, but can migrate the fill function they call at their own leisure. Old fill function is considered deprecated.
> * We remove the old fill function in a later upgrade.
> 
> Thoughts?

I like this and it leans into why I wanted to keep the event structure the same and only change the relayer field :)

contracts/SpokePool.sol
chrismaree: we _could_ BUT it's called `relayer` in the event and we don't really want to change the name there, do we?

programs/svm-spoke/src/instructions/slow_fill.rs
```diff
@@ -256,7 +256,7 @@ pub fn execute_v3_slow_relay_leaf(
         fill_deadline: relay_data.fill_deadline,
         exclusivity_deadline: relay_data.exclusivity_deadline,
         exclusive_relayer: relay_data.exclusive_relayer,
-        relayer: *ctx.accounts.signer.key,
+        relayer: Pubkey::default(), // There is no repayment address for slow
```
md0x, line 259: Question: Why are we moving from the relayer public key to address_zero here?

md0x (approved):
Looks good to me! Just had a question

programs/svm-spoke/src/instructions/slow_fill.rs
Reinis-FRP: I think we had it wrong and EVM implementation passes empty address as well

Reinis-FRP (approved):
looking good

hardhat.config.ts
```diff
@@ -52,7 +52,7 @@ const DEFAULT_CONTRACT_COMPILER_SETTINGS = {
     optimizer: { enabled: true, runs: 1000000 },
     viaIR: true,
     // Only strip revert strings if not testing or in ci.
-    debug: { revertStrings: isTest ? "default" : "strip" },
+    debug: { revertStrings: isTest ? "debug" : "strip" },
```
chrismaree, line 55: ^

pxrl: Very odd :thinking:

PR #660, opened 2024-10-14
fix: allow the forwarder to receive wrapped native tokens

bmzig:
> When tokens are being bridged from L1 to L3, the Router_Adapter contract bridges the token to the forwarder on L2 and sends an additional transaction, which causes the forwarder to relay the tokens to L3. The L2 token passed to that transaction is taken from the poolRebalanceRoute function of the HubPool. The problem is that it doesn't always correspond to the real token being bridged. For example, for WETH on Ethereum, the poolRebalanceRoute returns the WETH address on Optimism, but whenever WETH is bridged to Optimism by the Optimism_Adapter contract, ETH is in fact bridged. It means that assets will be correctly transferred to L2, but they will not be transferred to L3 by the forwarder as it will receive a different token than the one it will attempt to transfer. 

The fix is similar to the one [here](https://github.com/across-protocol/contracts/pull/648). We automatically wrap incoming native token transfers except for when we are calling the native token contract to withdraw. This should allow the contract to keep all of its native token supply in a wrapped balance, thereby also keeping the token addresses passed in on L1 correct.

### DIFF
--- a/contracts/chain-adapters/Arbitrum_Forwarder.sol
+++ b/contracts/chain-adapters/Arbitrum_Forwarder.sol
@@ -3,6 +3,7 @@ pragma solidity ^0.8.0;
 
 import { ForwarderBase } from "./ForwarderBase.sol";
 import { CrossDomainAddressUtils } from "../libraries/CrossDomainAddressUtils.sol";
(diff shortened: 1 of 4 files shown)


bmzig:
Closing in favor of https://github.com/across-protocol/contracts/pull/725

PR #663, opened 2024-10-15
chore(SpokePool): Remove pre-v3 types and events

pxrl:
These are no longer needed and would potentially otherwise be clobbered by upcoming svm changes. Remove them now to simplify that work.


contracts/test/MockSpokePool.sol
```diff
@@ -145,247 +145,6 @@ contract MockSpokePool is SpokePool, MockV2SpokePoolInterface, OwnableUpgradeabl
...
-            destinationChainId,
-            relayerFeePct,
-            newDepositId,
-            quoteTimestamp,
-            originToken,
-            recipient,
-            msg.sender,
-            message
-        );
-    }
```
pxrl, line 178: We might want to retain this, tbf.

pxrl:
@mrice32 @nicholaspai Are we OK to move on this cleanup? We have a lot of WIP work that would ideally come after it, but those PRs are instead internalising some overlapping changes.

md0x (approved):
Looks good to me

nicholaspai (approved):
I think we can merge this

contracts/test/MockSpokePool.sol
pxrl: Restored in a test capacity here: https://github.com/across-protocol/contracts/pull/663/commits/0186b0714467640f93f16882d5ebbf44f393d731

mrice32 (approved):

PR #664, opened 2024-10-15
fix: M-01 `relayTokens` Calls Made by Forwarders May Fail

bmzig:
> The Router_Adapter contract sends assets from L1 to L3 in two steps: first it bridges the tokens to the forwarder contract and then it sends a message instructing the forwarder to bridge the tokens to L3. The problem with this approach is that although the transaction bridging the tokens is scheduled first, there is no guarantee that it will be executed first as the tokens are bridged using different channels than the messages are being sent. In particular, in case of Arbitrum, the USDC token will be bridged through the CCTP protocol, but the messages will be passed through the Arbitrum Inbox contract, which is completely independent of CCTP. So, it may happen that the forwarder may attempt to bridge tokens to L3 before it receives them.
Although failed messages on L2s can be retried, we believe that it would be better to cache such failed messages inside forwarders, so that they could be reexectued by anyone in the future (possibly in a batch).

This PR applies the suggestion to the `ForwarderBase` contract. The router adapter on L1 will follow up a token relay with a call to `relayTokens`. Now, `relayTokens` will save the information about this message in the `tokenRelays` array so that it may be executed later via a call to `executeRelayTokens`.


contracts/chain-adapters/ForwarderBase.sol
```diff
@@ -107,30 +110,60 @@ abstract contract ForwarderBase is UUPSUpgradeable, ForwarderInterface {
...
+        // Save the token relay to state.
+        tokenRelays.push(tokenRelay);
+        emit TokenRelayReceived(tokenRelayId, tokenRelay);
+    }
+
+    /**
+     * @notice Sends a stored TokenRelay to L3.
+     * @param tokenRelayId Index of the relay to send in the `tokenRelays` array.
+     */
+    function sendTokens(uint32 tokenRelayId) external payable {
```
nicholaspai, line 150: maybe rename to finalizeRelayTokens or executeRelayTokens?

nicholaspai, line 150: You should make this reentrancy guarded as it likely calls an external contract

contracts/chain-adapters/ForwarderBase.sol
```diff
@@ -107,30 +110,60 @@ abstract contract ForwarderBase is UUPSUpgradeable, ForwarderInterface {
...
-            abi.encodeCall(AdapterInterface.relayTokens, (baseToken, destinationChainToken, amount, target))
+            abi.encodeCall(
+                AdapterInterface.relayTokens,
+                (tokenRelay.l2Token, tokenRelay.l3Token, tokenRelay.amount, tokenRelay.to)
+            )
         );
-        if (!success) revert RelayTokensFailed(baseToken);
-        emit TokensForwarded(baseToken, destinationChainToken, amount, destinationChainId, target);
+        if (!success) revert RelayTokensFailed(tokenRelay.l2Token);
+        tokenRelay.executed = true;
```
nicholaspai, line 164: I would set this `executed` value before the external delegatecall to follow the check-effects-interaction pattern

contracts/chain-adapters/ForwarderBase.sol
bmzig: I changed it to `executeRelayTokens` since it will most-likely be called by the executor https://github.com/across-protocol/contracts/pull/664/commits/6b6ebecfd50acd3f1f7d6ff743cdea5f4c12edb7

bmzig: Good call https://github.com/across-protocol/contracts/pull/664/commits/6b6ebecfd50acd3f1f7d6ff743cdea5f4c12edb7

contracts/chain-adapters/ForwarderBase.sol
bmzig: Did that here https://github.com/across-protocol/contracts/pull/664/commits/6b6ebecfd50acd3f1f7d6ff743cdea5f4c12edb7

nicholaspai (approved):

PR #667, opened 2024-10-16
fix(svm): use consistent imports

Reinis-FRP:
This proposes to standardize formatting of Solana program imports grouping them in 3 blocks separated by newline:
- standard library imports
- external imports
- current crate imports
Also avoids replicated use statements and has nested grouping when imports share the same root.

Fixes: https://linear.app/uma/issue/ACX-2911/instructionsadminrs-standardize-imports-across-all-files

### DIFF
--- a/programs/svm-spoke/src/instructions/admin.rs
+++ b/programs/svm-spoke/src/instructions/admin.rs
@@ -1,14 +1,13 @@
 use anchor_lang::prelude::*;
 
-// TODO: standardize imports across all files
-use anchor_spl::associated_token::AssociatedToken;
-use anchor_spl::token_interface::{Mint, TokenAccount, TokenInterface};
-
-// TODO: check that the discriminator size is used everywhere
-use crate::constants::DISCRIMINATOR_SIZE;
-use crate::constraints::is_local_or_remote_owner;
+use anchor_spl::{
+    associated_token::AssociatedToken,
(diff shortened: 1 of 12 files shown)


programs/svm-spoke/src/instructions/admin.rs
```diff
@@ -20,7 +18,7 @@ use crate::{
 pub struct Initialize<'info> {
     #[account(init, // Use init, not init_if_needed to prevent re-initialization.
               payer = signer,
-              space = DISCRIMINATOR_SIZE + State::INIT_SPACE, // TODO: check that INIT_SPACE is used everywhere
```
Reinis-FRP, line 23: checked this, no changes needed, so removed the comment just to avoid separate PR for that.

programs/svm-spoke/src/instructions/admin.rs
```diff
@@ -1,14 +1,12 @@
 use anchor_lang::prelude::*;
-
-// TODO: standardize imports across all files
-use anchor_spl::associated_token::AssociatedToken;
-use anchor_spl::token_interface::{Mint, TokenAccount, TokenInterface};
-
-// TODO: check that the discriminator size is used everywhere
```
Reinis-FRP, line 7: checked this, no changes needed, so removed the comment just to avoid separate PR for that.

chrismaree (approved):

PR #672, opened 2024-10-18
feat: EVM changes required to support Solana

chrismaree:
This is a feature branch where we will place all changes needed to be applied to the EVM to support Solana within across.

This branch contains 4 main categories of changes (each of which was originally a separate PR before moving them into this feature branch):
1. [Change input props & internal data structures within SpokePool.sol to be bytes32](https://github.com/across-protocol/contracts/pull/650). This is a somewhat opinionated change & was done to support bytes32 address types. We also changed internal data structures as it make things easier to reason about.
2. [CCTP <> Solana Adapter](https://github.com/across-protocol/contracts/pull/641). This change facilitates EVM <> SVM communication over CCTP. It contains an amount of wiring to enable the hub (which does not support bytes32, for example) to interface with SVM.
3. [Relayer repayment address as part of `fill` method](https://github.com/across-protocol/contracts/pull/653). As a relayer will now have a separate address on EVM & SVM, they need the ability to inform the data worker _where_ to be re-paid on the new domain.
4. [Improve how relayer repayments work, supporting blacklisted relayer addresses to not block bundles](https://github.com/across-protocol/contracts/pull/675). Due to change 3, it is now trivial to set a relayer repayment address to a recipient that is blacklisted by the repayment token. If a relayer did this, they would be able to effectively block the whole bundle from being executed. To work around this, a new data structure `relayerRefund` is added that stores refunds that could not be executed due to reverting transfers. Relayers can then claim these separately, provided they can prove they own the original blocked address.


contracts/SpokePool.sol
```diff
@@ -1012,10 +1216,27 @@ abstract contract SpokePool is
...
     }
 
+    /**
+     * @notice Enables a relayer to claim outstanding repayments. Should virtually never be used, unless for some reason
+     * relayer repayment transfer fails for reasons such as token transfer reverts due to blacklisting. In this case,
+     * the relayer can still call this method and claim the tokens to a new address.
+     * @param l2TokenAddress Address of the L2 token to claim refunds for.
+     * @param refundAddress Address to send the refund to.
+     */
+    function claimRelayerRefund(address l2TokenAddress, address refundAddress) public {
```
chrismaree, line 1231: ^

contracts/SpokePool.sol
```diff
@@ -768,17 +970,16 @@ abstract contract SpokePool is
...
-    function fillV3Relay(V3RelayData calldata relayData, uint256 repaymentChainId)
-        public
-        override
-        nonReentrant
-        unpausedFills
-    {
+    function fillV3Relay(
+        V3RelayData calldata relayData,
+        uint256 repaymentChainId,
+        bytes32 repaymentAddress
```
nicholaspai, line 976: Should we add a `fillV3Relay` overloaded function where the default repayment address is `msg.sender`? Could this avoid some breaking changes for existing fillers to give them time to migrate to new functions? Or will they already be broken by all the changes in the `V3RelayData` struct, for example the address -> bytes32 changes.

contracts/SpokePool.sol
```diff
@@ -1111,28 +1337,73 @@ abstract contract SpokePool is
...
+            for (uint256 i = 0; i < numRefunds; ++i) {
+                if (refundAmounts[i] > 0) {
+                    totalRefundedAmount += refundAmounts[i];
+
+                    // Only if the total refunded amount exceeds the spoke starting balance, should we revert. This
+                    // ensures that bundles are atomic, if we have sufficient balance to refund all relayers and
+                    // prevents can only re-pay some of the relayers.
+                    if (totalRefundedAmount > spokeStartBalance) revert InsufficientSpokePoolBalanceToExecuteLeaf();
+
+                    bool success = _noRevertTransfer(
```
nicholaspai, line 1361: @pxrl how much lift do you think there will be for the executor to deal with this change where refunds can now fail?

contracts/SpokePool.sol
```diff
@@ -723,6 +871,60 @@ abstract contract SpokePool is
...
+    ) public {
+        _verifyUpdateV3DepositMessage(
+            depositor.toBytes32(),
+            depositId,
+            chainId(),
+            updatedOutputAmount,
+            updatedRecipient.toBytes32(),
+            updatedMessage,
+            depositorSignature,
+            UPDATE_V3_DEPOSIT_ADDRESS_OVERLOAD_DETAILS_HASH
```
nicholaspai, line 913: Why does the hash need to be different?

contracts/SpokePool.sol
nicholaspai, line 1231: So I see `refundAddress` is type `address`, and the `relayerRefund` mapping also uses an address=>address key. This is because these variables are only used on the chain that this spoke pool would be deployed to, which will always be an EVM?

contracts/libraries/AddressConverters.sol
```diff
@@ -0,0 +1,15 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.0;
+
+library Bytes32ToAddress {
+    function toAddress(bytes32 _bytes32) internal pure returns (address) {
+        require(uint256(_bytes32) >> 192 == 0, "Invalid bytes32: highest 12 bytes must be 0");
```
nicholaspai, line 6: nit: prefer using [Custom errors](https://soliditylang.org/blog/2021/04/21/custom-errors/) to save on bytecode, whose limits we're running up against in some of the `SpokePool`s.

test/evm/foundry/fork/BlacklistedRelayerRecipient.t.sol
```diff
@@ -0,0 +1,164 @@
...
+        vm.prank(largeUSDTAccount);
+        usdt.transfer(address(spokePool), seedAmount);
+        assertEq(usdt.balanceOf(address(spokePool)), seedAmount, "Seed transfer failed");
+
+        vm.prank(largeUSDCAccount);
+        usdc.transfer(address(spokePool), seedAmount);
+        assertEq(usdc.balanceOf(address(spokePool)), seedAmount, "USDC seed transfer failed");
+    }
+
+    function testStandardRefundsWorks() public {
```
nicholaspai, line 74: This is a really cool usage of foundry fork tests to verify that we're compatible with USDT in production! This doesn't run in CI, right?

contracts/chain-adapters/Solana_Adapter.sol
```diff
@@ -0,0 +1,193 @@
...
+
+    /**
+     * @notice Helper to map a Solana address to an Ethereum address representation.
+     * @dev The Ethereum address is derived from the Solana address by truncating it to its lowest 20 bytes. This same
+     * conversion must be done by the HubPool owner when adding Solana spoke pool and setting the corresponding pool
+     * rebalance and deposit routes.
+     * @param solanaAddress Solana address (Base58 decoded to bytes32) to map to its Ethereum address representation.
+     * @return Ethereum address representation of the Solana address.
+     */
+    function _trimSolanaAddress(bytes32 solanaAddress) internal pure returns (address) {
```
nicholaspai, line 162: Why did you not add this to the `AddressConverters` library? Do you not think other contracts in the future might use this or is this logic only useful when sending messages via this adapter?

contracts/chain-adapters/Solana_Adapter.sol
```diff
@@ -0,0 +1,193 @@
...
+            revert InvalidAmount(amount);
+        }
+        if (to != SOLANA_SPOKE_POOL_ADDRESS) {
+            revert InvalidTokenRecipient(to);
+        }
+
+        _transferUsdc(SOLANA_SPOKE_POOL_USDC_VAULT, amount);
+
+        // TODO: consider if we need also to emit the translated addresses.
+        emit TokensRelayed(l1Token, l2Token, amount, to);
```
nicholaspai, line 151: I think it could be useful to emit the translated addresses at very litttle additional marginal cost

contracts/chain-adapters/Solana_Adapter.sol
```diff
@@ -0,0 +1,193 @@
...
+     * @param target Program on Solana (translated as EVM address) that will receive message.
+     * @param message Data to send to target.
+     */
+    function relayMessage(address target, bytes calldata message) external payable override {
+        if (target != SOLANA_SPOKE_POOL_ADDRESS) {
+            revert InvalidRelayMessageTarget(target);
+        }
+
+        bytes4 selector = bytes4(message[:4]);
+        if (selector == SpokePoolInterface.setEnableRoute.selector) {
```
nicholaspai, line 107: Why is `setEnableRoute` a specific case here? Is it the only function we'll be calling on the Solana spoke pool? For example, does this prevent us from pausing deposits/fills, emergency deleting root bundles, or forcing through admin root bundles? These haven't really happened. much in the past but the latter two have happened at least once

nicholaspai (commented):
This code is very clean, amazing work that you all should be very happy with. I left some questions and will take a second pass once we go through them all

contracts/SpokePool.sol
```diff
@@ -67,7 +70,7 @@ abstract contract SpokePool is
     RootBundle[] public rootBundles;
 
     // Origin token to destination token routings can be turned on or off, which can enable or disable deposits.
-    mapping(address => mapping(uint256 => bool)) public enabledDepositRoutes;
+    mapping(bytes32 => mapping(uint256 => bool)) public enabledDepositRoutes;
```
mrice32, line 73: Why do we need to convert this to `bytes32`? Isn't this mapping from an _origin_ chain token (which will always be a 20 byte `address`)?

contracts/SpokePool.sol
```diff
@@ -578,6 +664,67 @@ abstract contract SpokePool is
      * @param message The message to send to the recipient on the destination chain if the recipient is a contract.
      * If the message is not empty, the recipient contract must implement handleV3AcrossMessage() or the fill will revert.
      */
+    function depositV3Now(
+        bytes32 depositor,
+        bytes32 recipient,
+        bytes32 inputToken,
```
mrice32, line 670: Interface design question: if we _know_ this token is local (and therefore an address), is there a point in making it bytes32?

Same question for depositor.

There may be an advantage in keeping them consistent, but there is a disadvantage in validation/error cases: it's possible to pass in `bytes32` addresses that do not correspond to a valid local address.

contracts/SpokePool.sol
```diff
@@ -546,6 +566,72 @@ abstract contract SpokePool is
...
+            inputAmount,
+            outputAmount,
+            destinationChainId,
+            exclusiveRelayer.toBytes32(),
+            quoteTimestamp,
+            fillDeadline,
+            exclusivityPeriod,
+            message
+        );
+    }
```
mrice32, line 633: I still think we may want to consider just forwarding the raw msg.data to the other function via a self delegatecall. I think this would be cheaper and less bytecode (although I could be wrong).

contracts/chain-adapters/Solana_Adapter.sol
```diff
@@ -0,0 +1,193 @@
...
+ * @notice Contract containing logic to send messages from L1 to Solana via CCTP.
+ * @dev Public functions calling external contracts do not guard against reentrancy because they are expected to be
+ * called via delegatecall, which will execute this contract's logic within the context of the originating contract.
+ * For example, the HubPool will delegatecall these functions, therefore it's only necessary that the HubPool's methods
+ * that call this contract's logic guard against reentrancy.
+ * @custom:security-contact bugs@across.to
+ */
+
+// solhint-disable-next-line contract-name-camelcase
+contract Solana_Adapter is AdapterInterface, CircleCCTPAdapter {
```
mrice32, line 21: How much of this contract is specific to Solana vs could be used with any chain we connect to via CCTP?

I haven't gone deep on the specifics, but, for instance, can we build an adapter that could be used just as well with an EVM chain like BSC?

contracts/erc7683/ERC7683.sol
```diff
@@ -6,10 +6,10 @@ pragma solidity ^0.8.0;
 struct GaslessCrossChainOrder {
     /// @dev The contract address that the order is meant to be settled by.
     /// Fillers send this order to this contract address on the origin chain
-    address originSettler;
+    bytes32 originSettler;
     /// @dev The address of the user who is initiating the swap,
     /// whose input tokens will be taken and escrowed
-    address user;
+    bytes32 user;
```
mrice32, line 12: Aren't these local addresses? Why would they need to be bytes32?

contracts/erc7683/ERC7683.sol
```diff
@@ -45,7 +45,7 @@ struct OnchainCrossChainOrder {
 /// @dev Intended to improve integration generalization by allowing fillers to compute the exact input and output information of any order
 struct ResolvedCrossChainOrder {
     /// @dev The address of the user who is initiating the transfer
-    address user;
+    bytes32 user;
```
mrice32, line 48: Same question here.

contracts/erc7683/ERC7683Across.sol
```diff
@@ -6,19 +6,19 @@ import { GaslessCrossChainOrder } from "./ERC7683.sol";
 
 // Data unique to every CrossChainOrder settled on Across
 struct AcrossOrderData {
-    address inputToken;
+    bytes32 inputToken;
```
mrice32, line 9: Same question here

contracts/chain-adapters/Solana_Adapter.sol
```diff
@@ -0,0 +1,193 @@
...
+            revert InvalidDestinationChainId(destinationChainId);
+        }
+
+        return
+            abi.encodeWithSignature(
+                "setEnableRoute(bytes32,uint64,bool)",
+                SOLANA_USDC_BYTES32,
+                uint64(destinationChainId),
+                enable
+            );
```
mrice32, line 191: OOC, why not do this mapping on the Solana side?

In theory, could we not have a mapping (from truncated address to full address) exist on the Solana side that could be modified by an admin call. Then, the flow to add a new token would require no contract changes:
1. Send admin call to update the mapping with a new element.
2. Add the route to use the newly added element.

Side note: this is somewhat similar to something we have to do on [arbitrum](https://github.com/across-protocol/contracts/blob/98c761ee9ca3a496b4a368bdf12743d727644138/contracts/Arbitrum_SpokePool.sol#L80-L82).

mrice32 (commented):
Looks awesome! Left a few questions to start. Still working my way through

contracts/SpokePool.sol
chrismaree: note that they dont fail. the behaviour is _more_ kind to the executor. in a very rare event where you set a blacklisted recipient in the transfer then the recipient gets a credit they can claim. the design here should require _no additional changes to the executor_. we should, however, listen to events that indicate if this has happened.

contracts/SpokePool.sol
md0x: Yes, the changes in `V3RelayData` would also be breaking changes, so an update would still be necessary. I’m leaning towards only adding backward compatibility to functions that are called by other contracts, like `depositV3`.

contracts/SpokePool.sol
md0x: Because, when adding backwards compatibility to `speedUpV3Deposit` with the address version, we still need to support an `updatedMessage` constructed with the old data type. Maybe this is unnecessary?

I was assuming that there might be cases were contracts are calling this `speedUpV3Deposit` with the old signature and thus passing updatedMessage built with the [old data structure](https://github.com/across-protocol/contracts/blob/231b5097bed59f17068b4f6fde916f940b423571/contracts/SpokePool.sol#L138-L141)

However if `speedUpV3Deposit` is only expected to be called off-chain we probably don't need backwards compatibility and can drop `UPDATE_V3_DEPOSIT_ADDRESS_OVERLOAD_DETAILS_HASH`.

Can you pls clarify how is `speedUpV3Deposit` is used @nicholaspai ?

contracts/SpokePool.sol
md0x: I think we can migrate this to `bytes32` as well to be consistent with the rest of updates. What do you think @chrismaree?

contracts/libraries/AddressConverters.sol
md0x: Thanks, yes, I will do that!

contracts/chain-adapters/Solana_Adapter.sol
md0x: Good point. Fixed here: https://github.com/across-protocol/contracts/pull/727

contracts/chain-adapters/Solana_Adapter.sol
md0x: We need to keep the address version of [setEnableRoute](https://github.com/across-protocol/contracts/blob/d97325003a1a38828bb7b87b52be67c3f25ab65a/contracts/SpokePool.sol#L313-L320) for backward compatibility with the hub pool, and that specific handler translates the message to the bytes32 version.

@Reinis-FRP, do you remember why we decided to handle the translation here, which led us to do the casting here instead of handling it in SVM?

contracts/SpokePool.sol
md0x: Given that all the arguments have been moved to bytes32-- a decision that you can challenge as well--this made sense from a consistency perspective. We also decided to emit all addresses in their bytes32 form, so keeping the address representation here would require extra casting. Does this make sense @mrice32 ?

contracts/chain-adapters/Solana_Adapter.sol
Reinis-FRP: The main reason is that `setEnableRoute` is invoked via `setDepositRoute` on HubPool that we are not upgrading and we need different types on Solana. For other admin methods they are crafted manually and bridged via generic `relaySpokePoolAdminFunction` method, so the operator can adjust the interface types for all other methods to Solana.

contracts/SpokePool.sol
md0x: The main reason was consistency with the rest of the data. We still have the [overloaded](https://github.com/across-protocol/contracts/blob/d97325003a1a38828bb7b87b52be67c3f25ab65a/contracts/SpokePool.sol#L728-L740) versions of these functions that can help with the validation/casting of addresses.

We have tried to move everything to bytes32 in this contract. While in the short term there might be a bit of friction, I think in the medium term it’s better to have everything in either bytes32 or address, but not mixed, as mixing is more error-prone. However, we have made these changes somewhat radically and are looking for your feedback.

We do have bytes32 address encoding checks https://github.com/across-protocol/contracts/blob/d97325003a1a38828bb7b87b52be67c3f25ab65a/contracts/libraries/AddressConverters.sol#L11-L13 to partially help with the address to bytes32 conversion.

contracts/chain-adapters/Solana_Adapter.sol
md0x: But @Reinis-FRP this conversion could also be done in SVM side instead right? (related to https://github.com/across-protocol/contracts/pull/672#discussion_r1831856338)

contracts/erc7683/ERC7683.sol
md0x: Mainly, consistency with the rest of the data, as this would require extra casting at some point. Your point is that everything that is exclusively local to the evm should remain an address and be cast when needed (for example, in events?), so having a mixed model where we have both addresses and bytes32 seems inconsistent. Here, we have intentionally converted everything to bytes32 for greater consistency. I think we should discuss a standard criteria with the team for these changes.

contracts/SpokePool.sol
chrismaree: I agree for consistency it should be moved over. doing it now & updating the tests.

chrismaree: have updated this accordingly.

test/evm/foundry/fork/BlacklistedRelayerRecipient.t.sol
chrismaree: nope, you need to run it sepertly.

contracts/SpokePool.sol
chrismaree: ye, there is an overall design here to make _everything_ bytes32 internally this makes reasoning about data structures and props easier and more consistent as its _always_ bytes32. we could in theory change this accordingly such that we treat all evm data as address type and type cast as soon as possible when entering methods with bytes32. very much opinonated, very much can be changed if you'll prefer the other option.

contracts/SpokePool.sol
chrismaree: I'll look into this and report back.

chrismaree: I played with this for a fiew hours and think that the msg.data + self delegate call is overly complex and not worth it. I did something like this:
```
fallback() external payable {
        // Extract the function selector
        bytes4 selector;
        assembly {
            selector := calldataload(0)
        }

        // Check if selector matches `depositV3(address, address, address, address, uint256, uint256, uint256, address, uint32, uint32, uint32, bytes)`
        if (
            selector ==
            bytes4(
                keccak256(
                    "depositV3(address,address,address,address,uint256,uint256,uint256,address,uint32,uint32,uint32,bytes)"
                )
            )
        ) {
            // Extract each address parameter and cast to bytes32
            bytes32 depositor;
            bytes32 recipient;
            bytes32 inputToken;
            bytes32 outputToken;
            bytes32 exclusiveRelayer;

            // Using assembly for efficient calldata extraction and casting
            // Using assembly for efficient calldata extraction and conversion
            assembly {
                depositor := shl(96, calldataload(4)) // address -> bytes32
                recipient := shl(96, calldataload(36)) // address -> bytes32
                inputToken := shl(96, calldataload(68)) // address -> bytes32
                outputToken := shl(96, calldataload(100)) // address -> bytes32
                exclusiveRelayer := shl(96, calldataload(228)) // address -> bytes32
            }

            // Load other uint256 and uint32 values from calldata directly
            uint256 inputAmount;
            uint256 outputAmount;
            uint256 destinationChainId;
            uint32 quoteTimestamp;
            uint32 fillDeadline;
            uint32 exclusivityPeriod;
            bytes memory message;

            // Directly load the rest of the parameters from calldata
            assembly {
                inputAmount := calldataload(132)
                outputAmount := calldataload(164)
                destinationChainId := calldataload(196)
                quoteTimestamp := calldataload(260)
                fillDeadline := calldataload(292)
                exclusivityPeriod := calldataload(324)
            }

            // uint256 messageStart = 356; // Correct starting position for message
            // uint256 messageLength = msg.data.length > messageStart ? msg.data.length - messageStart : 0;

            // Copy `message` data directly
            // message = new bytes(messageLength);
            // if (messageLength > 0) {
            //     for (uint i = 0; i < messageLength; i++) {
            //         message[i] = msg.data[messageStart + i];
            //     }
            // }
            message = new bytes(0);

            // Forward the call to `depositV3(bytes32, bytes32, bytes32, bytes32, uint256, uint256, uint256, bytes32, uint32, uint32, uint32, bytes)
            (bool success, ) = address(this).delegatecall(
                abi.encodeWithSelector(
                    this.depositV3.selector,
                    depositor,
                    recipient,
                    inputToken,
                    outputToken,
                    inputAmount,
                    outputAmount,
                    destinationChainId,
                    exclusiveRelayer,
                    quoteTimestamp,
                    fillDeadline,
                    exclusivityPeriod,
                    message
                )
            );
            require(success, "Delegatecall to depositV3 failed");
        } else {
            // Handle unexpected calls
            revert("Function selector not supported");
        }
    }
```
(note the `message` decoding is broken so I hard coded it to bytes 0).

this works. but:
1.  it's more expensive than the wrapped version (marginally, but still more).
2. it was super irritating to implement, 
3. it makes our ABIs a massive pain to work with (we need a custom interface in the build process to append this to the ABI as we no longer have public functions exposing them

is this the kind of thing you were envisioning? I think if this is the pattern we need to go for then I'm a strong proponent of the much easier version that the PR has rn.

contracts/chain-adapters/Solana_Adapter.sol
Reinis-FRP: It is quite specific in a sense that it must be aware of Solana token model where recipient SpokePool address needs to be translated to its corresponding vault token address.

contracts/chain-adapters/Solana_Adapter.sol
Reinis-FRP: This would be bit tricky as we would need to pass mapping account that stores address translation when receiving the message on Solana side. It does not really belong under `HandleReceiveMessage` which translates all other supported admin calls statelessly. And if we move the address translation account down to self-invoked `SetEnableRoute` then we would end up changing the interface of `set_enable_route` to use `[u8; 20]` as its `origin_token` parameter, which is not ideal as we also support calling it from Solana admin (or need to implement both flavors). And this also requires additional admin method to update the mapping account with correct address translation.

Not saying its impossible to do, but it seemed to have less code complexity if we move the translation to the adapter on EVM side, especially, since it already needs to translate trimmed Solana USDC address to its vault associated token account anyway.

contracts/SpokePool.sol
nicholaspai: I think the changes in V3Relay might not be as breaking necessarily since address can be implictly converted to bytes32. But I take your point.

@pxrl wdyt about this?

contracts/SpokePool.sol
nicholaspai: > However if speedUpV3Deposit is only expected to be called off-chain 

What do you mean by off-chain?

> Can you pls clarify how is speedUpV3Deposit is used @nicholaspai ?

The way this is used is a depositor sees a speed up button on the FE which prompts them to sign a message which they then submit to speedUpV3Deposit, code [here](https://github.com/across-protocol/frontend/blob/74a687cff46be5aad6f4a3d4fc65bef6dee130b8/src/views/Transactions/hooks/useSpeedUp.tsx#L86)

contracts/SpokePool.sol
nicholaspai: are there any storage/gas cost considerations to think about with this change?

contracts/chain-adapters/Solana_Adapter.sol
nicholaspai: So it sounds like the tradeoff here is (less code complexity, special case for `setEnableDepositRoute` on the EVM side, generic SVM side logic) versus (more code complexity, but totally generic EVM logic and bespoke SVM side logic), right? What about gas costs?

contracts/erc7683/ERC7683.sol
nicholaspai: yeah this is a really good point @mrice32 , and it's how we designed ERC7683 to work too: where code that is expected to run on EVM doesn't need to convert addresses into bytes32. If we undid all these address=>bytes32 changes, would we remove some breaking changes? That very well might be worth it

nicholaspai (approved):

nicholaspai (dismissed):

nicholaspai (approved):

mrice32 (dismissed):

nicholaspai (approved):

PR #673, opened 2024-10-18
fix(svm): check invalid refund leaf

Reinis-FRP:
SVM implementation of SpokePool lacked the check on the matching length for refund account and amount arrays when executing relayer refund leaf.


md0x (approved):
+1 This check was missing

chrismaree (approved):

PR #675, opened 2024-10-21
feat: update spokepool relayer refund to handle blocked transfers

chrismaree:


contracts/SpokePool.sol
```diff
@@ -1373,24 +1397,41 @@ abstract contract SpokePool is
...
+        uint256 totalRefundedAmount = 0;
 
         // Send each relayer refund address the associated refundAmount for the L2 token address.
         // Note: Even if the L2 token is not enabled on this spoke pool, we should still refund relayers.
-        uint256 length = refundAmounts.length;
-        for (uint256 i = 0; i < length; ++i) {
-            uint256 amount = refundAmounts[i];
-            if (amount > 0)
-                IERC20Upgradeable(l2TokenAddressParsed).safeTransfer(refundAddresses[i].toAddress(), amount);
+        for (uint256 i = 0; i < refundAmounts.length; ++i) {
```
md0x, line 1408: nit:  we can cache refundAmounts.length by storing it in memory

contracts/SpokePool.sol
```diff
@@ -1373,24 +1397,41 @@ abstract contract SpokePool is
...
+            if (totalRefundedAmount > spokeStartBalance) revert InsufficientSpokePoolBalanceToExecuteLeaf();
+            if (refundAmounts[i] > 0) {
+                bool success;
+                try
+                    IERC20Upgradeable(l2TokenAddress.toAddress()).transfer(
+                        refundAddresses[i].toAddress(),
+                        refundAmounts[i]
+                    )
+                returns (bool result) {
+                    success = result;
```
md0x, line 1420: should we also check that the balance has decreased as expected before considering it successful?

contracts/SpokePool.sol
```diff
@@ -1373,24 +1397,41 @@ abstract contract SpokePool is
...
+                    deferredRefunds = true;
+                }
+            }
         }
-
         // If leaf's amountToReturn is positive, then send L2 --> L1 message to bridge tokens back via
         // chain-specific bridging method.
         if (amountToReturn > 0) {
-            _bridgeTokensToHubPool(amountToReturn, l2TokenAddressParsed);
+            _bridgeTokensToHubPool(amountToReturn, l2TokenAddress.toAddress());
```
md0x, line 1434: We don't need to check here that `amountToReturn <= erc20.balanceOf(address(this))` because this case would be handled by the revert?

contracts/SpokePool.sol
chrismaree: why? if the transfer succeeded it will be set as the bool returned from the call. if it fails it either reverts or returns false.

contracts/SpokePool.sol
chrismaree: I cant add any extra state/memory variables due to yul issues beyond what I have now.

contracts/SpokePool.sol
```diff
@@ -1373,24 +1397,41 @@ abstract contract SpokePool is
...
+
+            if (totalRefundedAmount > spokeStartBalance) revert InsufficientSpokePoolBalanceToExecuteLeaf();
+            if (refundAmounts[i] > 0) {
+                bool success;
+                try
+                    IERC20Upgradeable(l2TokenAddress.toAddress()).transfer(
+                        refundAddresses[i].toAddress(),
+                        refundAmounts[i]
+                    )
+                returns (bool result) {
```
Reinis-FRP, line 1419: This might fail on successful transfers for tokens that do not return bool on success (e.g. USDT). So instead of try/catch pattern we might need to reimplement OZ `safeTransfer` without bubbling up the reverts.

chrismaree: yup, I agree. we should just catch on reverts. what do you think of this implementation?

else, you're right: we can implement [this](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/c343ee3768cc61a76a14d6f8402b1b19be1497fc/contracts/token/ERC20/utils/SafeERC20.sol#L35) kind of structure internally

chrismaree: the safe erc20 assumes `non-reverting calls are assumed to be successful`, which is what the new implementation now assumes.

Reinis-FRP: `IERC20Upgradeable` expects bool to be returned from the transfer, so USDT would always hit the catch block. Since try/catch works only on external calls or contract creation, the only viable option is to replicate the logic of OZ `_callOptionalReturn` and replacing all reverts with returning bool that we can track in our logic.

contracts/SpokePool.sol
```diff
@@ -1373,29 +1397,71 @@ abstract contract SpokePool is
...
-            _bridgeTokensToHubPool(amountToReturn, l2TokenAddressParsed);
+            _bridgeTokensToHubPool(amountToReturn, l2TokenAddress.toAddress());
 
             emit TokensBridged(amountToReturn, _chainId, leafId, l2TokenAddress, msg.sender);
         }
     }
 
+    // Re-implementation of OZ _callOptionalReturnBool to use private logic. Function executes a transfer and returns a
+    // bool indicating if the external call was successful, rather than reverting. Original method:
+    // https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/utils/SafeERC20.sol#L188
```
mrice32, line 1447: Use permalinks, so the code doesn't change when their repo gets updated.

```suggestion
    // https://github.com/OpenZeppelin/openzeppelin-contracts/blob/28aed34dc5e025e61ea0390c18cac875bfde1a78/contracts/token/ERC20/utils/SafeERC20.sol#L188
```

contracts/SpokePool.sol
```diff
@@ -1373,29 +1397,71 @@ abstract contract SpokePool is
...
+    function noRevertTransfer(
```
mrice32, line 1448:
```suggestion
    function _noRevertTransfer(
```

contracts/SpokePool.sol
```diff
@@ -1373,29 +1397,71 @@ abstract contract SpokePool is
...
+        address token,
+        address to,
+        uint256 amount
+    ) internal returns (bool) {
```
mrice32, line 1452: +1, this function is exactly what we need to work with old tokens, like USDT.

contracts/SpokePool.sol
```diff
@@ -1429,6 +1442,26 @@ abstract contract SpokePool is
         }
     }
 
+    // Re-implementation of OZ _callOptionalReturnBool to use private logic. Function executes a transfer and returns a
+    // bool indicating if the external call was successful, rather than reverting. Original method:
+    // https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/utils/SafeERC20.sol#L188
+    function noRevertTransfer(
```
Reinis-FRP, line 1448: nit: use underscore in internal methods. Or alt, move this to library

contracts/SpokePool.sol
```diff
@@ -102,6 +102,8 @@ abstract contract SpokePool is
     // to eliminate any chance of collision between pre and post V3 relay hashes.
     mapping(bytes32 => uint256) public fillStatuses;
 
+    mapping(address => mapping(address => uint256)) public relayerRefund;
```
mrice32, line 105: Make sure you decrement the __gap array size at the bottom of the file to keep storage aligned.

mrice32 (commented):
Looks good! Only a few minor comments

contracts/SpokePool.sol
```diff
@@ -1429,6 +1442,26 @@ abstract contract SpokePool is
...
+    function noRevertTransfer(
+        address token,
+        address to,
+        uint256 amount
+    ) internal returns (bool) {
+        bool success;
+        uint256 returnSize;
+        uint256 returnValue;
+        bytes memory data = abi.encodeCall(IERC20Upgradeable.transfer, (to, amount));
+        assembly {
```
Reinis-FRP, line 1457: should this be annotated as memory-safe, like in OZ?

chrismaree: I think the solidity version we are using does not support this. it gave me an error when trying to do that.

contracts/SpokePool.sol
chrismaree: just decrement by 1 for a 2d mapping? I'm not totally sure how this gap variable works.

contracts/SpokePool.sol
```diff
@@ -1271,10 +1274,27 @@ abstract contract SpokePool is
...
+    /**
+     * @notice Enables a relayer to claim outstanding repayments. Should virtually never be used, unless for some reason
+     * relayer repayment transfer fails for reasons such as token transfer reverts due to blacklisting. In this case,
+     * the relayer can still call this method and claim the tokens to a new address.
+     * @param l2TokenAddress Address of the L2 token to claim refunds for.
+     * @param refundAddress Address to send the refund to.
+     */
+    function claimRelayerRefund(address l2TokenAddress, address refundAddress) public {
+        uint256 liability = relayerRefund[l2TokenAddress][msg.sender];
+        require(liability > 0, "No liability to claim");
```
Reinis-FRP, line 1291: nit: rest of the contract uses if/revert pattern instead of reverts. Also in other places we use custom errors

contracts/SpokePool.sol
Reinis-FRP: Spoke implementations for each chain inherit this SpokePool, which is also upgradable, so we need to ensure the storage layout on derived contracts stays the same

contracts/SpokePool.sol
chrismaree: we use `require` syntax in many other files but you're right that in this file we don't. I've updated it accordingly.

PR #677, opened 2024-10-22
fix(svm): todos from the 2nd reading session

Reinis-FRP:


chrismaree (approved):
might make sense to just merge these in so we can close them against master?

PR #678, opened 2024-10-22
fix(svm): no transfer on self-relay

Reinis-FRP:
EVM logic skips the fill transfer if relayer is the same as recipient. This adds the same logic in SVM SpokePool.

Fixes: https://linear.app/uma/issue/ACX-2929/instructionsfillrs-check-what-happens-if-the-relayer-and-recipient-are


chrismaree (approved):
great!

md0x (approved):
LGTM

PR #679, opened 2024-10-22
fix(svm): remove msg logs

Reinis-FRP:
msg logs are memory expensive, especially when formatting variables.

fixes: https://linear.app/uma/issue/ACX-2930/instructionsfillrs-remove-msg-everywhere


chrismaree (approved):
✨🧹🚿

md0x (approved):

PR #680, opened 2024-10-22
feat(svm): add emergency delete bundle

md0x:
Changes proposed in this PR:
- Add emergency_delete_root_bundle

### DIFF
--- a/programs/svm-spoke/src/event.rs
+++ b/programs/svm-spoke/src/event.rs
@@ -114,6 +114,7 @@ pub struct ClaimedRelayerRefund {
     pub refund_address: Pubkey,
 }
 
+// TODO: rename EVM event to match this name
 #[event]
 pub struct EmergencyDeletedRootBundle {
     pub root_bundle_id: u32,
--- a/programs/svm-spoke/src/event.rs
+++ b/programs/svm-spoke/src/event.rs
@@ -113,3 +113,8 @@ pub struct ClaimedRelayerRefund {
     pub claim_amount: u64,
     pub refund_address: Pubkey,
 }
+
+#[event]
+pub struct EmergencyDeletedRootBundle {
+    pub root_bundle_id: u32,
+}
--- a/programs/svm-spoke/src/instructions/admin.rs
+++ b/programs/svm-spoke/src/instructions/admin.rs
@@ -261,7 +261,7 @@ pub struct EmergencyDeleteRootBundle<'info> {
     pub signer: Signer<'info>,
 
     // TODO: standardize usage of state.seed vs state.key()
-    #[account(mut, seeds = [b"state", state.seed.to_le_bytes().as_ref()], bump)]
+    #[account(seeds = [b"state", state.seed.to_le_bytes().as_ref()], bump)]
     pub state: Account<'info, State>,
 
     // TODO: consider deriving seed from state.seed instead of state.key() as this could be cheaper (need to verify).
@@ -276,12 +276,6 @@ pub fn emergency_delete_root_bundle(
     ctx: Context<EmergencyDeleteRootBundle>,
     root_bundle_id: u32,
 ) -> Result<()> {
-    let root_bundle = &mut ctx.accounts.root_bundle;
-
-    root_bundle.claimed_bitmap.clear();
-    root_bundle.relayer_refund_root = [0; 32];
-    root_bundle.slow_relay_root = [0; 32];
-
     emit_cpi!(EmergencyDeletedRootBundle { root_bundle_id });
 
     Ok(())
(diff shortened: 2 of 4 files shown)


test/svm/SvmSpoke.Bundle.ts
```diff
@@ -155,6 +155,61 @@ describe("svm_spoke.bundle", () => {
...
+      assert.fail("Non-owner should not be able to delete the root bundle");
+    } catch (err: any) {
+      assert.include(err.toString(), "NotOwner", "Expected error for non-owner trying to delete root bundle");
+    }
+
+    // Execute the emergency delete
+    const emergencyDeleteRootBundleAccounts = { state, rootBundle, signer: owner, program: program.programId };
+    await program.methods.emergencyDeleteRootBundle(rootBundleId).accounts(emergencyDeleteRootBundleAccounts).rpc();
+
+    // Verify that the root bundle has been deleted
```
chrismaree, line 204: might want to slightly extend this test to show that post deletion we can continue to add new bundles without the deleted bundle affecting the subsequent bundles.

chrismaree (approved):
looks good! just one additional test should be added but past that this is great

programs/svm-spoke/src/event.rs
```diff
@@ -113,3 +113,8 @@ pub struct ClaimedRelayerRefund {
     pub claim_amount: u64,
     pub refund_address: Pubkey,
 }
+
+#[event]
+pub struct EmergencyDeletedRootBundle {
```
Reinis-FRP, line 118: EVM implementation has `EmergencyDeleteRootBundle` event, though that would conflict with instruction context struct. So maybe add comment here to later fix EVM event naming.

programs/svm-spoke/src/instructions/admin.rs
```diff
@@ -247,4 +250,39 @@ pub fn relay_root_bundle(
...
 }
 
-// TODO: add emergency_delete_root_bundle
+#[event_cpi]
+#[derive(Accounts)]
+#[instruction(root_bundle_id: u32)]
+pub struct EmergencyDeleteRootBundle<'info> {
+    #[account(
+        mut,
+        constraint = is_local_or_remote_owner(&signer, &state) @ CustomError::NotOwner
```
Reinis-FRP, line 259: We also need to add translation in handle_receive_message, so that we are able to invoke this from mainnet

programs/svm-spoke/src/instructions/admin.rs
```diff
@@ -247,4 +250,39 @@ pub fn relay_root_bundle(
...
+    )]
+    pub signer: Signer<'info>,
+
+    // TODO: standardize usage of state.seed vs state.key()
+    #[account(mut, seeds = [b"state", state.seed.to_le_bytes().as_ref()], bump)]
```
Reinis-FRP, line 264: no need for mutable state here

programs/svm-spoke/src/instructions/admin.rs
```diff
@@ -247,4 +250,39 @@ pub fn relay_root_bundle(
...
+    pub state: Account<'info, State>,
+
+    // TODO: consider deriving seed from state.seed instead of state.key() as this could be cheaper (need to verify).
+    #[account(mut,
+        seeds =[b"root_bundle", state.key().as_ref(), root_bundle_id.to_le_bytes().as_ref()],
+        close = signer,
```
Reinis-FRP, line 270: Not sure how this work for remote calls as the signer would be PDA. Have to add tests and see if PDA would hold returned lamports.

programs/svm-spoke/src/instructions/admin.rs
```diff
@@ -247,4 +250,39 @@ pub fn relay_root_bundle(
...
+}
+
+pub fn emergency_delete_root_bundle(
+    ctx: Context<EmergencyDeleteRootBundle>,
+    root_bundle_id: u32,
+) -> Result<()> {
+    let root_bundle = &mut ctx.accounts.root_bundle;
+
+    root_bundle.claimed_bitmap.clear();
+    root_bundle.relayer_refund_root = [0; 32];
```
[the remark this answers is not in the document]
Reinis-FRP: not sure we really need to clear these structures as the account gets closed and we don't have any path to reinit the same account as rood bundle id is always incrementing

Reinis-FRP (approved):

programs/svm-spoke/src/instructions/admin.rs
Reinis-FRP: I'll add this together with remote call for relayRootBundle in separate PR

programs/svm-spoke/src/instructions/admin.rs
Reinis-FRP: I will modify RelayRootBundle to include dedicated payer as CCTP requires this (same as in SetEnableRoute), so I will change account close recipient to the same payer when I add CCTP support for these calls

PR #681, opened 2024-10-23
fix(svm): use pubkey in origin token

Reinis-FRP:
It is easier to use Pubkey as origin_token parameter in set_enable_route.

Fixes: https://linear.app/uma/issue/ACX-2917/instructionsadminrs-is-it-possible-to-replace-origin-token-with-pubkey


chrismaree (approved):
ye, this is great. we should use the native types like this where we can.

md0x (approved):
✅

PR #683, opened 2024-10-23
feat(svm): relay root bundle event

md0x:
Proposed changes in this PR:
- Add relay root bundle event

### DIFF
--- a/programs/svm-spoke/src/event.rs
+++ b/programs/svm-spoke/src/event.rs
@@ -24,10 +24,10 @@ pub struct EnabledDepositRoute {
 }
 
 #[event]
-pub struct RelayRootBundleEvent {
+pub struct RelayedRootBundle {
+    pub root_bundle_id: u32,
     pub relayer_refund_root: [u8; 32],
     pub slow_relay_root: [u8; 32],
-    pub root_bundle_id: u32,
 }
 
 // Deposit events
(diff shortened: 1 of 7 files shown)


programs/svm-spoke/src/event.rs
```diff
@@ -23,6 +23,13 @@ pub struct EnabledDepositRoute {
     pub enabled: bool,
 }
 
+#[event]
+pub struct RelayRootBundleEvent {
```
Reinis-FRP, line 27: this is named `RelayedRootBundle` in EVM. Also the ordering is different (rootBundleId, relayerRefundRoot, slowRelayRoot)

programs/svm-spoke/src/instructions/admin.rs
```diff
@@ -241,10 +244,15 @@ pub fn relay_root_bundle(
...
     root_bundle.relayer_refund_root = relayer_refund_root;
     root_bundle.slow_relay_root = slow_relay_root;
+
+    emit_cpi!(RelayRootBundleEvent {
+        relayer_refund_root,
+        slow_relay_root,
+        root_bundle_id: state.root_bundle_id,
+    });
+
+    // Finally, increment the root bundle id
```
Reinis-FRP, line 254: not directly related to this PR, but I think its confusing to have prop named `root_bundle_id` in state account. Instead it should be something to resemble number of relayed bundles or `next_root_bundle_id` - Maybe add this as TODO here

Reinis-FRP (approved):
nice!

PR #686, opened 2024-10-24
fix(svm): close claim account manually

Reinis-FRP:
Though claim accounts are being closed automatically when claiming the refund, there might be a scenario where relayer refunds were executed with ATA after initializing the claim account. In such cases, the initializer should be able to close the claim account manually.

Fixes: https://linear.app/uma/issue/ACX-3024/instructionsrefund-claimsrs-add-manual-instruction-to-close-claim


chrismaree (approved):

md0x (approved):
Looks good

PR #687, opened 2024-10-24
fix(svm): root bundles over cctp

Reinis-FRP:
We require ability to relay and emergency delete root bundles over CCTP.

Fixes: https://linear.app/uma/issue/ACX-3029/instructionsadminrs-remote-relay-emergency-delete-root-bundle

### DIFF
--- a/programs/svm-spoke/src/instructions/admin.rs
+++ b/programs/svm-spoke/src/instructions/admin.rs
@@ -259,7 +259,6 @@ pub fn relay_root_bundle(
 #[instruction(root_bundle_id: u32)]
 pub struct EmergencyDeleteRootBundle<'info> {
     #[account(
-        mut,
         constraint = is_local_or_remote_owner(&signer, &state) @ CustomError::NotOwner
     )]
     pub signer: Signer<'info>,
--- a/programs/svm-spoke/src/instructions/admin.rs
+++ b/programs/svm-spoke/src/instructions/admin.rs
@@ -212,10 +212,12 @@ pub fn set_enable_route(
 #[derive(Accounts)]
 pub struct RelayRootBundle<'info> {
     #[account(
-        mut,
         constraint = is_local_or_remote_owner(&signer, &state) @ CustomError::NotOwner
     )]
     pub signer: Signer<'info>,
+
+    #[account(mut)]
+    pub payer: Signer<'info>,
 
     // TODO: standardize usage of state.seed vs state.key()
     #[account(mut, seeds = [b"state", state.seed.to_le_bytes().as_ref()], bump)]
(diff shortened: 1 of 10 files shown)


programs/svm-spoke/src/instructions/admin.rs
```diff
@@ -257,17 +259,21 @@ pub fn relay_root_bundle(
 #[instruction(root_bundle_id: u32)]
 pub struct EmergencyDeleteRootBundle<'info> {
     #[account(
-        mut,
         constraint = is_local_or_remote_owner(&signer, &state) @ CustomError::NotOwner
     )]
     pub signer: Signer<'info>,
 
+    #[account(mut)]
+    // We do not restrict who can receive lamports from closing root_bundle account as that would require storing the
```
chrismaree, line 267: makes sense, as long as it's authenticated with `is_local_or_remote_owner` we don't really mind.

programs/svm-spoke/src/instructions/handle_receive_message.rs
```diff
@@ -82,6 +82,18 @@ fn translate_message(data: &Vec<u8>) -> Result<Vec<u8>> {
             (origin_token, destination_chain_id, enabled)
                 .encode_instruction_data("global:set_enable_route")
         }
+        s if s == utils::encode_solidity_selector("relayRootBundle(bytes32,bytes32)") => {
```
chrismaree, line 85: great! we can now test this on testnet as well over cctp

programs/svm-spoke/src/instructions/admin.rs
Reinis-FRP: In remote authentication this checks only that this is coming from HubPool via CCTP, but we do not use permissioned caller to receive the message on Solana side (though CCTP has that option). So there might be a case that someone frontruns receiving message on Solana and gets all lamports from the closing of root bundle account. But I think this is rare enough when we would use this emergency method and it should be no concern that someone else could gain some extra lamports.

md0x: Makes sense. Anyone can "finalize" this messages

test/svm/SvmSpoke.HandleReceiveMessage.ts
```diff
@@ -406,4 +409,168 @@ describe("svm_spoke.handle_receive_message", () => {
...
+    // Remaining accounts specific to RelayRootBundle.
+    const rootBundleId = (await program.account.state.fetch(state)).rootBundleId;
+    const rootBundleIdBuffer = Buffer.alloc(4);
+    rootBundleIdBuffer.writeUInt32LE(rootBundleId);
+    const seeds = [Buffer.from("root_bundle"), state.toBuffer(), rootBundleIdBuffer];
+    const [rootBundle] = web3.PublicKey.findProgramAddressSync(seeds, program.programId);
+    // Same 3 remaining accounts passed for HandleReceiveMessage context.
+    const relayRootBundleRemainingAccounts = remainingAccounts.slice(0, 3);
+    // payer in self-invoked SetEnableRoute.
+    relayRootBundleRemainingAccounts.push({
```
md0x, line 439: Does order matter here?

md0x (approved):
Looks good!

test/svm/SvmSpoke.HandleReceiveMessage.ts
Reinis-FRP: Regular accounts have named props, so the client can reorder them as needed, but remaining accounts is plain array, so I believe the ordering here matters.

chrismaree (approved):

PR #688, opened 2024-10-24
test: multi-chain verifiable merkle tree

md0x:
Changes proposed in this PR:
- Add tests in hardhat to verify that a tree with mixed leaves (sorted by chain and mixed together) using different hashing functions is fully verifiable on the EVM.
- Add equivalent tests for the SVM.

### DIFF
--- a/test/evm/hardhat/SpokePool.ExecuteRootBundle.ts
+++ b/test/evm/hardhat/SpokePool.ExecuteRootBundle.ts
@@ -1,4 +1,5 @@
 import { SignerWithAddress, seedContract, toBN, expect, Contract, ethers, BigNumber } from "../../../utils/utils";
+import { buildRelayerRefundMerkleTree } from "../../svm/utils";
 import * as consts from "./constants";
 import { spokePoolFixture } from "./fixtures/SpokePool.Fixture";
 import { buildRelayerRefundTree, buildRelayerRefundLeaves } from "./MerkleLib.utils";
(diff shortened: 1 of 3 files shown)


test/svm/utils.ts
```diff
@@ -103,7 +179,7 @@ export const relayerRefundHashFn = (input: RelayerRefundLeaf | RelayerRefundLeaf
     const abiCoder = new ethers.utils.AbiCoder();
     const encodedData = abiCoder.encode(
       [
-        "tuple(uint256 leafId, uint256 chainId, uint256 amountToReturn, address l2TokenAddress, address[] refundAddresses, uint256[] refundAmounts)",
+        "tuple( uint256 amountToReturn, uint256 chainId, uint256[] refundAmounts, uint256 leafId, address l2TokenAddress, address[] refundAddresses)",
```
md0x, line 182: The order wasn't matching the one in EVM tests.

chrismaree: ye, this is the thing I was trying to say in the call on Tuesday: we should make them match within the rust/solidity code! there is no reason from them not to match, its just extra complexity. I would suggest we update the rust to match the solidity ordering, even with differently named fields

test/evm/hardhat/SpokePool.ExecuteRootBundle.ts
```diff
@@ -1,4 +1,5 @@
 import { SignerWithAddress, seedContract, toBN, expect, Contract, ethers, BigNumber } from "../../../utils/utils";
+import { buildRelayerRefundMerkleTree } from "../../svm/utils";
```
chrismaree, line 2: this is great, I really like this pattern that you have here being able to mix the EVM/SVM tests via the tree building logic.

test/svm/utils.ts
```diff
@@ -74,6 +77,79 @@ export function convertLeafIdToNumber(leaf: RelayerRefundLeafSolana) {
...
+
+  const createEvmLeaf = (index: number) =>
+    ({
+      isSolana: false,
+      leafId: BigNumber.from(index),
+      chainId: BigNumber.from(chainId),
+      amountToReturn: BigNumber.from(0),
+      l2TokenAddress: evmTokenAddress ?? randomAddress(),
+      refundAddresses: evmRelayers || [randomAddress(), randomAddress()],
+      refundAmounts: evmRefundAmounts || [BigNumber.from(randomBigInt()), BigNumber.from(randomBigInt())],
```
chrismaree, line 123: it's generally a code smell to have random numbers within unit tests but I think for this it's ok.

test/svm/SvmSpoke.Bundle.ts
```diff
@@ -9,26 +15,19 @@ import {
   TransactionMessage,
   VersionedTransaction,
 } from "@solana/web3.js";
+import { MerkleTree } from "@uma/common/dist/MerkleTree";
```
Reinis-FRP, line 18: nit: better import from `utils/MerkleTree.ts`. Though under the hood its the same uma common package import, it might be better to import it consistently from one source

Reinis-FRP (approved):

test/svm/utils.ts
md0x: Yes, makes sense @chrismaree. I will work in the svm changes in a subsequent PR 👍

chrismaree (approved):

PR #689, opened 2024-10-24
fix(svm): remove redundant accounts

Reinis-FRP:
This removes redundant accounts from fill and slow fill instructions.

Fixes: https://linear.app/uma/issue/ACX-3020/instructionsfillrs-recipient-systemaccount-might-be-redundant
Fixes: https://linear.app/uma/issue/ACX-3019/instructionsfillrs-should-relayer-be-the-same-as-signer

### DIFF
--- a/test/svm/SvmSpoke.SlowFill.ts
+++ b/test/svm/SvmSpoke.SlowFill.ts
@@ -324,7 +324,6 @@ describe("svm_spoke.slow_fill", () => {
       vault: vault,
       tokenProgram: TOKEN_PROGRAM_ID,
       mint: mint,
-      recipient,
       recipientTokenAccount: recipientTA,
       program: program.programId,
     };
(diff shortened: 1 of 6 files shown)


md0x (approved):
LGTM

test/svm/SvmSpoke.SlowFill.ts
```diff
@@ -65,11 +65,9 @@ describe("svm_spoke.slow_fill", () => {
     fillAccounts = {
       state,
       signer: relayer.publicKey,
-      relayer: relayer.publicKey,
-      recipient: relayData.recipient, // This could be different from global recipient.
       mintAccount: mint,
-      relayerTA: relayerTA,
-      recipientTA: recipientTA,
+      relayerTokenAccount: relayerTA,
```
chrismaree, line 69: feels a bit messay that we mix TA and TokenAccount in names of things but we can make it more consistant later.

chrismaree (approved):

PR #690, opened 2024-10-24
fix: supply decimals and CCTP domain in custom token adapter

bmzig:
Three changes needed to be made:
1. The Arbitrum "default ERC20 bridge" contract does not have a `nativeTokenDecimals` function we can call, so we instead define the decimals of the native token it in the constructor. We do not resolve this dynamically since decimals is not required to implement ERC20, so we can't rely on IERC20 to call `decimals()`. 
2. We need to swap the CCTP domain ID from Arbitrum to some arbitrary number, which can be defined in the constructor.
3. When calling `createRetryableTicket` to bridge the native gas token, `l2CallValue` must actually be the amount of the native gas token to send, not zero. This is because `l2CallValue` is with respect to the L2's native token, not ETH.

### DIFF
--- a/contracts/chain-adapters/Arbitrum_CustomGasToken_Adapter.sol
+++ b/contracts/chain-adapters/Arbitrum_CustomGasToken_Adapter.sol
@@ -51,9 +51,9 @@ contract Arbitrum_CustomGasToken_Adapter is AdapterInterface, CircleCCTPAdapter 
     // The Arbitrum Inbox requires that this is specified in gWei (e.g. 1e9 = 1 gWei)
     uint256 public immutable L2_GAS_PRICE;
 
-    // Native token expected to be sent in L2 message. Should be 0 for all use cases of this constant, which
-    // includes sending messages from L1 to L2 and sending Custom gas token ERC20's, which won't be the native token
-    // on the L2 by definition.
+    // Native token expected to be sent in L2 message. Should be 0 for most use cases of this constant. This
+    // constant is unused when sending the native gas token over the inbox since the inbox interprets `l2CallValue`
+    // as the amount of the L2 native token to send.
     uint256 public constant L2_CALL_VALUE = 0;
 
     // Gas limit for L2 execution of a cross chain token transfer sent via the inbox.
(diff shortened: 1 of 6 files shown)


deploy/consts.ts
```diff
@@ -48,6 +51,9 @@ export const L1_ADDRESS_MAP: { [key: number]: { [contractName: string]: string }
     zoraStandardBridge: "0x3e2Ea9B92B7E48A52296fD261dc26fd995284631",
     worldChainCrossDomainMessenger: "0xf931a81D18B1766d15695ffc7c1920a62b7e710a",
     worldChainStandardBridge: "0x470458C91978D2d929704489Ad730DC3E3001113",
+    l1AlephZeroInbox: "0x56D8EC76a421063e1907503aDd3794c395256AEb",
+    l1AlephZeroERC20GatewayRouter: "0xeBb17f398ed30d02F2e8733e7c1e5cf566e17812",
+    donationBox: "0x5e40a88E302Da80089Dd13178b8DD6b15534419c",
```
bmzig, line 56: Needs to be swapped out once we deploy an official adapter.

contracts/chain-adapters/Arbitrum_CustomGasToken_Adapter.sol
```diff
@@ -116,6 +122,7 @@ contract Arbitrum_CustomGasToken_Adapter is AdapterInterface, CircleCCTPAdapter
         L2_MAX_SUBMISSION_COST = _l2MaxSubmissionCost;
         L2_GAS_PRICE = _l2GasPrice;
         CUSTOM_GAS_TOKEN_FUNDER = _customGasTokenFunder;
+        NATIVE_TOKEN_DECIMALS = _nativeTokenDecimals;
```
pxrl, line 125: Is it easier to set this by calling `CUSTOM_GAS_TOKEN.decimals()` instead? It'd save a constructor argument so it's one less thing we can get wrong.

deploy/052_deploy_donation_box.ts
```diff
@@ -0,0 +1,19 @@
+import { DeployFunction } from "hardhat-deploy/types";
+import { HardhatRuntimeEnvironment } from "hardhat/types";
+
+const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
+  const { deployer } = await hre.getNamedAccounts();
+
+  // @note if deploying this contract on a chain like Linea that only supports up to
+  // solc 0.8.19, the hardhat.config solc version needs to be overridden and this
```
pxrl, line 8: This comment is outdated - Linea supports later versions of solidity but the evm version must be set to london. See https://docs.linea.build/developers/quickstart/ethereum-differences#evm-opcodes

contracts/chain-adapters/Arbitrum_CustomGasToken_Adapter.sol
bmzig: The ERC20 standard actually [doesn't guarantee](https://eips.ethereum.org/EIPS/eip-20) a `decimals` function/variable per the EIP, so OZ's `IERC20` won't let us call a `IERC20().decimals()` function due to type checking. We could do a `staticcall` here, but that just seems messy, especially if some ERC20 were to actually NOT implement decimals().

deploy/052_deploy_donation_box.ts
bmzig: I didn't know this. Nice. https://github.com/across-protocol/contracts/pull/690/commits/8e2018374154198f9fdcc56bef5416af7eaf6992

contracts/chain-adapters/Arbitrum_CustomGasToken_Adapter.sol
pxrl: TIL :eyes:

nicholaspai (approved):

PR #694, opened 2024-10-25
fix(svm): consistent anchor token constraints

Reinis-FRP:
We were unnecessarily restricting relayer token account in the fill instruction to be canonically derived from the relayer and mint addresses. By replacing associated_token constraints with token constraints this would allow the relayer to pass any token account as long as its mint and authority matches.

Overall, we want to use canonical token accounts for vault and fill/slow fill recipients (hence, associated_token constraints) and any valid token accounts for all others.

Also, some of mint accounts were incorrectly using token constraints for their program account checks. Event the checks are implemented the same way (account being owned by the token program), Anchor error message is slightly different.

Fixes: https://linear.app/uma/issue/ACX-2925/instructionsfillrs-consistent-token-imports

### DIFF
--- a/test/svm/SvmSpoke.Fill.ts
+++ b/test/svm/SvmSpoke.Fill.ts
@@ -3,6 +3,7 @@ import { BN } from "@coral-xyz/anchor";
 import {
   ASSOCIATED_TOKEN_PROGRAM_ID,
   TOKEN_PROGRAM_ID,
+  createAccount,
   createMint,
   getOrCreateAssociatedTokenAccount,
   mintTo,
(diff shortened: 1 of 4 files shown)


md0x (approved):

test/svm/SvmSpoke.Fill.ts
```diff
@@ -390,4 +391,34 @@ describe("svm_spoke.fill", () => {
       "Token Program should not be invoked"
     );
   });
+
+  it("Fills a V3 relay from custom relayer token account", async () => {
```
chrismaree, line 395: great!

chrismaree (approved):

PR #695, opened 2024-10-25
feat: token bridge cpi event 

md0x:
Changes proposed in this PR:
- Emit event from `bridge_tokens_to_hub_pool` function

### DIFF
--- a/programs/svm-spoke/src/lib.rs
+++ b/programs/svm-spoke/src/lib.rs
@@ -194,11 +194,13 @@ pub mod svm_spoke {
             proof,
         )
     }
-    pub fn bridge_tokens_to_hub_pool(
-        ctx: Context<BridgeTokensToHubPool>,
+    pub fn bridge_tokens_to_hub_pool<'info>(
+        ctx: Context<'_, '_, '_, 'info, BridgeTokensToHubPool<'info>>,
         amount: u64,
     ) -> Result<()> {
-        ctx.accounts.bridge_tokens_to_hub_pool(amount, &ctx.bumps)?;
+        ctx.accounts
+            .clone()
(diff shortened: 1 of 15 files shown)


programs/svm-spoke/src/lib.rs
```diff
@@ -194,11 +194,18 @@ pub mod svm_spoke {
...
         )
     }
-    pub fn bridge_tokens_to_hub_pool(
-        ctx: Context<BridgeTokensToHubPool>,
+    pub fn bridge_tokens_to_hub_pool<'info>(
+        ctx: Context<'_, '_, '_, 'info, BridgeTokensToHubPool<'info>>,
         amount: u64,
     ) -> Result<()> {
-        ctx.accounts.bridge_tokens_to_hub_pool(amount, &ctx.bumps)?;
+        let context = Context {
```
md0x, line 201: Workaround to be able to pass the context required by `emit_cpi!`

Reinis-FRP: I think it would be more consistent with other instructions to use free functions and pass ctx as parameter.

Reinis-FRP: that should also hopefully avoid account cloning below

programs/svm-spoke/src/lib.rs
```diff
@@ -194,11 +194,11 @@ pub mod svm_spoke {
             proof,
         )
     }
-    pub fn bridge_tokens_to_hub_pool(
+    pub fn bridge_tokens_to_hub_pool<'info>(
```
Reinis-FRP, line 197: do we strictly need this info lifetime here? its not being passed in `Context<BridgeTokensToHubPool>` below

Reinis-FRP (approved):

PR #698, opened 2024-10-28
Improve shared state locations and other stylistic improvements

chrismaree:
- WIP
- WIP


programs/svm-spoke/src/instructions/fill.rs
```diff
@@ -179,27 +161,18 @@ pub struct CloseFillPda<'info> {
...
-pub fn close_fill_pda(
-    ctx: Context<CloseFillPda>,
-    _relay_hash: [u8; 32], // TODO: check if we can use underscore in functions while in context macro leave as is?
-    relay_data: V3RelayData
-) -> Result<()> {
+pub fn close_fill_pda(ctx: Context<CloseFillPda>, _: [u8; 32], relay_data: V3RelayData) -> Result<()> {
     let state = &ctx.accounts.state;
     let current_time = get_current_time(state)?;
 
-    // Check if the fill status is filled
```
chrismaree, line 195: note that I changed the requires here! I think this check is redundant. the ONLY thing we care about is if the deposit is expired, not if it was filled (or not) to let us close the PDA. LMK if you'll agree.

programs/svm-spoke/src/instructions/fill.rs
```diff
@@ -65,27 +65,9 @@ pub struct FillV3Relay<'info> {
...
-    pub deposit_id: u32,
-    pub fill_deadline: u32,
-    pub exclusivity_deadline: u32,
-    pub message: Vec<u8>,
-}
-
 pub fn fill_v3_relay(
     ctx: Context<FillV3Relay>,
-    relay_hash: [u8; 32], // include in props, while not using it, to enable us to access it from the #Instruction Attribute within the accounts. This enables us to pass in the relay_hash PDA.
+    _: [u8; 32], // include in props, while not using it, to enable us to access it from the #Instruction Attribute within the accounts. This enables us to pass in the relay_hash PDA.
```
Reinis-FRP, line 70: we can delete this prop altogether in the inner instruction call as we don't necessarily need to pass it from the outer function in `svm-spoke/src/lib.rs`. Since instruction args in `FillV3Relay` accounts is passed from the outer instruction call, not the inner function. So only need to rename `relay_hash` in the outer function to `_relay_hash` and not pass it to this inner instruction function.

programs/svm-spoke/src/instructions/fill.rs
```diff
@@ -179,27 +161,18 @@ pub struct CloseFillPda<'info> {
...
     )]
     pub fill_status: Account<'info, FillStatusAccount>,
 }
 
-pub fn close_fill_pda(
-    ctx: Context<CloseFillPda>,
-    _relay_hash: [u8; 32], // TODO: check if we can use underscore in functions while in context macro leave as is?
-    relay_data: V3RelayData
-) -> Result<()> {
+pub fn close_fill_pda(ctx: Context<CloseFillPda>, _: [u8; 32], relay_data: V3RelayData) -> Result<()> {
```
Reinis-FRP, line 169: we can delete the unused prop here and rename to `_relay_hash` in the outer function.

programs/svm-spoke/src/instructions/refund_claims.rs
```diff
@@ -26,11 +26,7 @@ pub struct InitializeClaimAccount<'info> {
     pub system_program: Program<'info, System>,
 }
 
-pub fn initialize_claim_account(
-    ctx: Context<InitializeClaimAccount>,
-    mint: Pubkey,
-    token_account: Pubkey
-) -> Result<()> {
+pub fn initialize_claim_account(ctx: Context<InitializeClaimAccount>, _: Pubkey, _: Pubkey) -> Result<()> {
```
Reinis-FRP, line 29: we can delete the unused props here and rename them to `_mint` and  `_token_account` in the outer function.

programs/svm-spoke/src/instructions/slow_fill.rs
```diff
@@ -35,15 +33,15 @@ pub struct SlowFillV3Relay<'info> {
...
+        constraint = is_relay_hash_valid(&relay_hash, &relay_data, &state) @ SvmError::InvalidRelayHash
     )]
     pub fill_status: Account<'info, FillStatusAccount>,
     pub system_program: Program<'info, System>,
 }
 
 pub fn request_v3_slow_fill(
     ctx: Context<SlowFillV3Relay>,
-    relay_hash: [u8; 32], // include in props, while not using it, to enable us to access it from the #Instruction Attribute within the accounts. This enables us to pass in the relay_hash PDA.
+    _: [u8; 32], // include in props, while not using it, to enable us to access it from the #Instruction Attribute within the accounts. This enables us to pass in the relay_hash PDA.
```
Reinis-FRP, line 44: same comment as in others, rename it to `_relay_hash` in the outer call and not pass it to this inner function

programs/svm-spoke/src/instructions/slow_fill.rs
```diff
@@ -173,9 +171,9 @@ pub struct ExecuteV3SlowRelayLeaf<'info> {
 
 pub fn execute_v3_slow_relay_leaf(
     ctx: Context<ExecuteV3SlowRelayLeaf>,
-    relay_hash: [u8; 32],
+    _token_account: [u8; 32],
     slow_fill_leaf: V3SlowFill,
-    root_bundle_id: u32,
+    _: u32,
```
Reinis-FRP, line 176: better to skip passing unused params from the outer call

programs/svm-spoke/src/state/fill.rs
```diff
@@ -13,3 +13,19 @@ pub struct FillStatusAccount {
     pub status: FillStatus,
     pub relayer: Pubkey,
 }
+
+#[derive(AnchorSerialize, AnchorDeserialize, Clone)]
+pub struct V3RelayData {
```
Reinis-FRP, line 18: maybe better to have this under something like `src/common/v3_relay_data`? then we could also move `RefundAccount` to `src/common/refund_account` as it also does not strictly belong to the state

programs/svm-spoke/src/instructions/testable.rs
```diff
@@ -1,6 +1,6 @@
 use anchor_lang::prelude::*;
 
-use crate::{ error::CustomError, state::State };
+use crate::{ error::SvmError, state::State };
```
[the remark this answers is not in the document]
chrismaree: ye, that's great. or conditionally import the file. I've changed it to just use `crate::error::SvmError` when referencing the error.

programs/svm-spoke/src/instructions/bundle.rs
```diff
@@ -33,8 +33,7 @@ pub struct ExecuteRelayerRefundLeaf<'info> {
     )]
     pub root_bundle: Account<'info, RootBundle>,
 
-    #[account(
-        mut,
+    #[account(mut,
```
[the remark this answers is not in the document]
chrismaree: yes, you're right. I tried to see if I could fit all on one line & could not but messed up the formatting after the fact.

programs/svm-spoke/src/instructions/fill.rs
chrismaree: ye, that's great! works as you explained. that's actually a fair bit cleaner. we still "expose" the method with this prop but the internal function does not need it as it's used by the instructions.

programs/svm-spoke/src/state/fill.rs
chrismaree: ye, I think that's better. it's not really part of fill, and it's also not "state" in that it's not initialized. I'll update the way you've recommended.

I did not pull `refund_account`, though, as I think we might still end up changing this structure given the conversation we had last week where by we don't let the account be either TokenAccount OR ClaimAccount but rather pass in both of them to prevent relayers front running this set of methods.

test/svm/SvmSpoke.Fill.ts
```diff
@@ -15,7 +15,7 @@ import { common } from "./SvmSpoke.common";
 const { provider, connection, program, owner, chainId, seedBalance } = common;
 const { recipient, initializeState, setCurrentTime, assertSE, assert } = common;
 
-describe("svm_spoke.fill", () => {
+describe.only("svm_spoke.fill", () => {
```
Reinis-FRP, line 18:
```suggestion
describe("svm_spoke.fill", () => {
```

Reinis-FRP (approved):
looking great!. just remove only mode in fill tests

PR #699, opened 2024-10-28
fix: modify Arbitrum_CustomGasToken_Adapter for Aleph Zero

bmzig:
# Summary

## Adding backwards and forwards compatibility
When attempting to deploy the Arbitrum_CustomGasTokens_Adapter to support Aleph Zero, we realized that the deployed ERC20BridgeContract does not actually contain the `nativeTokenDecimals()` function as the [contracts](https://github.com/OffchainLabs/nitro-contracts/blob/b140ed63acdb53cb906ffd1fa3c36fdbd474364e/src/bridge/ERC20Bridge.sol#L42) would suggest. More critically, the amount to be bridged is [not scaled](https://github.com/OffchainLabs/nitro-contracts/blob/7c46876077c6353c7ebdf9cd364710d357fa3914/src/bridge/AbsInbox.sol#L235) in the legacy AbsInbox but it is in the [new one](https://github.com/OffchainLabs/nitro-contracts/blob/fbbcef09c95f69decabaced3da683f987902f3e2/src/bridge/AbsInbox.sol#L237). 

It turns out the issue is that Aleph Zero contracts were deployed using an outdated version of Arbitrum contracts. The contract has been modified to work with legacy and new versions of the Inbox contract:
- We add a new constructor argument to set the native token decimals, rather than reading it from the bridge contract’s nativeTokenDecimals() function which won’t exist on legacy deployments.
- We have confirmed that all legacy deployments like Aleph Zero's will use 18 decimals of precision, so we can keep the `_from18ToNativeDecimals()` conversion function in the relayTokens/relayMessage logic

## Other changes
The following opportunistic fixes were also made to the Arbitrum_CustomGasTokens_Adapter contract in this PR:
- Pass the Circle CCTP Domain ID in the constructor of the adapter instead of hardcoding the CCTP domain ID. This is just a stylistic refactor.
- Modify the l2CallValueField of our call to createRetryableTicket when bridging the custom gas token to specify the amount of native token to bridge.

#### Notes

Re-implementation of #690


james-a-morris:
Is this ready to go in? cc @melisaguevara @bmzig @pxrl

deploy/053_deploy_alephzero_adapter.ts
```diff
@@ -0,0 +1,35 @@
...
+  // set to the Risk Labs relayer address. The deployer should change this if necessary.
+  const l2RefundAddress = "0x07aE8551Be970cB1cCa11Dd7a11F47Ae82e70E67";
+
+  const args = [
+    L1_ADDRESS_MAP[chainId].l1AlephZeroInbox,
+    L1_ADDRESS_MAP[chainId].l1AlephZeroERC20GatewayRouter,
+    l2RefundAddress,
+    ZERO_ADDRESS,
+    ZERO_ADDRESS,
+    0, // No Circle CCTP domain ID.
```
bmzig, line 19: This shouldn't actually be zero since that is Ethereum's domain ID. We can set it to type(uint32).max though

contracts/chain-adapters/Arbitrum_LegacyCustomGasToken_Adapter.sol
```diff
@@ -0,0 +1,345 @@
...
+// solhint-disable-next-line contract-name-camelcase
+contract Arbitrum_LegacyCustomGasToken_Adapter is AdapterInterface, CircleCCTPAdapter {
+    using SafeERC20 for IERC20;
+
+    // Amount of gas token allocated to pay for the base submission fee. The base submission fee is a parameter unique to
+    // retryable transactions; the user is charged the base submission fee to cover the storage costs of keeping their
+    // ticket’s calldata in the retry buffer. (current base submission fee is queryable via
+    // ArbRetryableTx.getSubmissionPrice). ArbRetryableTicket precompile interface exists at L2 address
+    // 0x000000000000000000000000000000000000006E.
+    // The Arbitrum Inbox requires that the precision of this variable is equal to the precision of the native token on the L2.
```
bmzig, line 158: Modified this comment.

contracts/chain-adapters/Arbitrum_LegacyCustomGasToken_Adapter.sol
```diff
@@ -0,0 +1,339 @@
...
+    }
+
+    /**
+     * @notice Returns required amount of gas token to send a message via the Inbox.
+     * @dev Should return a value in the same precision as the gas token's precision.
+     * @param l2GasLimit L2 gas limit for the message.
+     * @return amount of gas token that this contract needs to hold in order for relayMessage to succeed.
+     */
+    function getL1CallValue(uint32 l2GasLimit) public view returns (uint256) {
+        return L2_MAX_SUBMISSION_COST + L2_GAS_PRICE * l2GasLimit;
```
bmzig, line 330: Changed this to not use `_from18ToNativeTokenDecimals` under the assumption that when the deployer deploys this adapter, they will correctly set `L2_MAX_SUBMISSION_COST` and `L2_GAS_PRICE` to the correct precision.

contracts/chain-adapters/Arbitrum_LegacyCustomGasToken_Adapter.sol
bmzig, line 1: Everything in this file is identical to that of `Arbitrum_CustomGasToken_Adapter.sol` in this PR (so it also has the same changes to `createRetryableTicket` and the circle domain IDs with the exceptions of where I left comments.

contracts/chain-adapters/Arbitrum_LegacyCustomGasToken_Adapter.sol
```diff
@@ -0,0 +1,339 @@
...
+ * used as native currency on L2.
+ * @dev Fees are paid in this token. There are certain restrictions on the native token:
+ *       - The token can't be rebasing or have a transfer fee
+ *       - The token must only be transferrable via a call to the token address itself
+ *       - The token must only be able to set allowance via a call to the token address itself
+ *       - The token must not have a callback on transfer, and more generally a user must not be able to make a transfer to themselves revert
+ *       - The token must have a max of 2^256 - 1 wei total supply unscaled
+ *       - The token must have a max of 2^256 - 1 wei total supply when scaled to 18 decimals
+ */
+interface ArbitrumL1ERC20Bridge {
```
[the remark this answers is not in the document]
bmzig: Nice. Yeah it should: https://github.com/across-protocol/contracts/pull/699/commits/897a339fb3581b4cd596d3993d3ef54f5b33cdc4

contracts/chain-adapters/Arbitrum_LegacyCustomGasToken_Adapter.sol
```diff
@@ -0,0 +1,213 @@
...
+        emit TokensRelayed(l1Token, l2Token, amount, to);
+    }
+
+    /**
+     * @notice Returns required amount of gas token to send a message via the Inbox.
+     * @dev Should return a value in the same precision as the gas token's precision.
+     * @param l2GasLimit L2 gas limit for the message.
+     * @return amount of gas token that this contract needs to hold in order for relayMessage to succeed.
+     */
+    function getL1CallValue(uint32 l2GasLimit) public view returns (uint256) {
```
nicholaspai, line 203: btw @bmzig can't we make this contract a lot simpler if we make this function virtual and just override it in this function? If everything else is the same

bmzig: Yeah that's true. https://github.com/across-protocol/contracts/pull/699/commits/b0bc90b01d30030c7b0abb41534378182f98027e

contracts/chain-adapters/Arbitrum_CustomGasToken_Adapter.sol
```diff
@@ -335,23 +336,4 @@ contract Arbitrum_CustomGasToken_Adapter is AdapterInterface, CircleCCTPAdapter
         if (CUSTOM_GAS_TOKEN.balanceOf(address(this)) < requiredL1CallValue) revert InsufficientCustomGasToken();
         return requiredL1CallValue;
     }
-
-    function _from18ToNativeDecimals(uint256 amount) internal view returns (uint256) {
```
nicholaspai, line 339: I think we can keep this if we assume that all legacy inboxes, like the one we call with AlephZero, use 18 decimal precision

nicholaspai: And then we pass in the precision number

contracts/chain-adapters/Arbitrum_CustomGasToken_Adapter.sol
```diff
@@ -155,16 +155,16 @@ contract Arbitrum_CustomGasToken_Adapter is AdapterInterface, CircleCCTPAdapter
...
     // ArbRetryableTx.getSubmissionPrice). ArbRetryableTicket precompile interface exists at L2 address
     // 0x000000000000000000000000000000000000006E.
-    // The Arbitrum Inbox requires that this uses 18 decimal precision.
+    // The Arbitrum Inbox requires that this uses the same precision as the L2's native gas token.
     uint256 public immutable L2_MAX_SUBMISSION_COST;
 
-    // L2 Gas price bid for immediate L2 execution attempt (queryable via standard eth*gasPrice RPC)
-    // The Arbitrum Inbox requires that this is specified in gWei (e.g. 1e9 = 1 gWei)
+    // L2 Gas price bid for immediate L2 execution attempt (queryable via standard eth_gasPrice RPC)
+    // Since this price fluctuates, it should be set to a conservative value. The gas price must be specified in Wei.
```
nicholaspai, line 162: i don't think we should change this comment

contracts/chain-adapters/Arbitrum_CustomGasToken_Adapter.sol
```diff
@@ -215,10 +215,11 @@ contract Arbitrum_CustomGasToken_Adapter is AdapterInterface, CircleCCTPAdapter
         address _l2RefundL2Address,
         IERC20 _l1Usdc,
         ICCTPTokenMessenger _cctpTokenMessenger,
+        uint32 _cctpDomainId,
         FunderInterface _customGasTokenFunder,
         uint256 _l2MaxSubmissionCost,
         uint256 _l2GasPrice
-    ) CircleCCTPAdapter(_l1Usdc, _cctpTokenMessenger, CircleDomainIds.Arbitrum) {
```
nicholaspai, line 221: I think we should pass in nativeTokenDecimals here and use it in _from18ToNativeTokenDecimals()

nicholaspai (commented):
Two suggestions

contracts/chain-adapters/Arbitrum_CustomGasToken_Adapter.sol
bmzig: Done https://github.com/across-protocol/contracts/pull/699/commits/6437f7d55cc9c274576350617b28cbb08bb65c10

contracts/chain-adapters/Arbitrum_CustomGasToken_Adapter.sol
bmzig: It's been switched back https://github.com/across-protocol/contracts/pull/699/commits/6437f7d55cc9c274576350617b28cbb08bb65c10

contracts/chain-adapters/Arbitrum_CustomGasToken_Adapter.sol
bmzig: Sounds good https://github.com/across-protocol/contracts/pull/699/commits/6437f7d55cc9c274576350617b28cbb08bb65c10

contracts/chain-adapters/Arbitrum_CustomGasToken_Adapter.sol
```diff
@@ -337,12 +344,11 @@ contract Arbitrum_CustomGasToken_Adapter is AdapterInterface, CircleCCTPAdapter
     }
 
     function _from18ToNativeDecimals(uint256 amount) internal view returns (uint256) {
-        uint8 nativeTokenDecimals = L1_INBOX.bridge().nativeTokenDecimals();
```
nicholaspai, line 340: Can you remove `nativeTokenDecimals` from the interface now

bmzig: https://github.com/across-protocol/contracts/pull/699/commits/a6f7e3fc0c9e33a5c296127211c38a3edc32b266

nicholaspai (approved):

contracts/chain-adapters/Arbitrum_CustomGasToken_Adapter.sol
```diff
@@ -240,7 +250,17 @@ contract Arbitrum_CustomGasToken_Adapter is AdapterInterface, CircleCCTPAdapter
                 return divFloor;
             }
         } else {
-            return amount * 10**(nativeTokenDecimals - 18);
+            return amount * 10**(NATIVE_TOKEN_DECIMALS - 18);
+        }
+    }
+
+    function _fromNativeTo18Decimals(uint256 amount) internal view returns (uint256) {
```
bmzig, line 257: This was taken from [here](https://github.com/OffchainLabs/nitro-contracts/blob/fbbcef09c95f69decabaced3da683f987902f3e2/src/bridge/ERC20Inbox.sol#L152). I removed the check for uint256 overflow, since this shouldn't happen assuming the constructors are correct, which we can easily verify before enabling anything in the hub pool. Also, the scaling is taken from [here](https://github.com/OffchainLabs/nitro-contracts/blob/fbbcef09c95f69decabaced3da683f987902f3e2/src/libraries/DecimalsConverterHelper.sol#L20).

contracts/chain-adapters/Arbitrum_CustomGasToken_Adapter.sol
```diff
@@ -177,7 +187,7 @@ contract Arbitrum_CustomGasToken_Adapter is AdapterInterface, CircleCCTPAdapter
                 CUSTOM_GAS_TOKEN.safeIncreaseAllowance(address(L1_INBOX), amountToBridge);
                 L1_INBOX.createRetryableTicket(
                     to, // destAddr destination L2 contract address
-                    L2_CALL_VALUE, // l2CallValue call value for retryable L2 message
+                    _fromNativeTo18Decimals(amount), // l2CallValue call value for retryable L2 message
```
mrice32, line 190: This isn't just a decimal change, this seems to be a deeper functional change.

Previously, the second arg looks like it was _always_ set to 0. Now it appears that we're passing in the amount (decimal converted).

Was the previous code wrong in setting this field to a 0 constant? Or is the new code doing something more subtle that I'm failing to understand?

bmzig: For (hopefully) full context: 
Setting this to 0 was incorrect. This function is called only when trying to bridge the `AZERO` token. When we set this to zero, the aliased L1 sender address will receive `amountToBridge`, i.e. amount+bridge fees, and then call our EOA with 0 value. It then _only_ refunds excess bridge gas costs to the L2 target. This means that `amount` is stuck in the aliased address on L2 (for example, we can see this happened with the aliased [dev wallet](https://evm-explorer.alephzero.org/address/0xddC021f002Eaf230c9fa810B34837a3739B7208c) and the aliased [adapter](https://evm-explorer.alephzero.org/address/0xaBA092A830a5cb89a3816E3d267cB7791C16c15E)). While it's not loss of funds, since we effectively have control over the aliased address by nudging the bridge from L1, we obviously didn't want this.

The first change was to just put in `amount` as the `l2CallValue` field here, which works for `AZERO`, since `AZERO` happens to have 18 decimals, but for a token which has non-18 decimal precision, we would be passing in a value with native token precision, when the contracts [assume](https://github.com/OffchainLabs/nitro-contracts/blob/fbbcef09c95f69decabaced3da683f987902f3e2/src/bridge/AbsInbox.sol#L238) `l2Callvalue` is in 18 decimal precision. This lead to the second change, seen above, where we convert `l2CallValue` from native token decimals to 18.

contracts/chain-adapters/Arbitrum_CustomGasToken_Adapter.sol
nicholaspai: So just to note, in the case where `NATIVE_TOKEN_DECIMALS > 18`, we'll be dividing and effectively rounding down the result. Is that OK when passing in this rounded down value into the Inbox.createRetryableTicket?

mrice32 (approved):
LGTM!

PR #669, opened 2024-10-16
feat: hardhat task to test chain adapter token bridging

pxrl:
This has some limitations around USDC native/bridged coincidence, but should otherwise be useful for testing chain adapters for new deployments (where CCTP is very unlikely to be enabled anyway).

PR #701, opened 2024-10-28
feat: hardhat task to test chain adapter token bridging (#669)

pxrl:
This has some limitations around USDC native/bridged coincidence, but should otherwise be useful for testing chain adapters for new deployments (where CCTP is very unlikely to be enabled anyway).


pxrl:
nb. This is just cherry-picking a commit that was previously made to `master`. This task can be used to test chain adapters after deployment.

bmzig (approved):

PR #703, opened 2024-10-29
fix: rust linting

Reinis-FRP:
We introduced prettier plugin to lint rust code, mostly to format Anchor attribute macros that rustfmt does not handle well. This though, brought in quite a lot of opinionated formatting for the rest of the rust code from prettier. In order to be more aligned with default rust formatting, this PR introduces wrapper pre-commit hook that first runs prettier-quick and then applies rustfmt on top for any staged rust files.

The only deviation from default rust formatting this PR adds is extending maximum line with to 120 characters to be consistent with other parts of this repo.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,8 +22,8 @@
     "lint": "yarn lint-solidity && yarn lint-js && yarn lint-rust",
     "lint-solidity": "yarn solhint ./contracts/**/*.sol",
     "lint-js": "yarn prettier --list-different **/*.js **/*.ts",
-    "lint-rust": "yarn prettier --list-different ./programs/**/*.rs",
-    "lint-fix": "yarn prettier --write **/*.js **/*.ts ./programs/**/*.rs ./contracts**/*.sol",
+    "lint-rust": "cargo fmt --all -- --check && cargo clippy",
+    "lint-fix": "yarn prettier --write **/*.js **/*.ts ./programs/**/*.rs ./contracts**/*.sol && cargo fmt --all",
     "clean-fast": "for dir in node_modules cache cache-zk artifacts artifacts-zk dist typechain; do mv \"${dir}\" \"_${dir}\"; rm -rf \"_${dir}\" &; done",
     "clean": "rm -rf node_modules cache cache-zk artifacts artifacts-zk dist typechain",
     "build-evm": "hardhat compile",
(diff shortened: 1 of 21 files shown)


package.json
```diff
@@ -22,8 +22,8 @@
     "lint": "yarn lint-solidity && yarn lint-js && yarn lint-rust",
     "lint-solidity": "yarn solhint ./contracts/**/*.sol",
     "lint-js": "yarn prettier --list-different **/*.js **/*.ts",
-    "lint-rust": "yarn prettier --list-different ./programs/**/*.rs",
-    "lint-fix": "yarn prettier --write **/*.js **/*.ts ./programs/**/*.rs ./contracts**/*.sol",
+    "lint-rust": "cargo fmt --all -- --check && cargo clippy",
```
chrismaree, line 25: so this is the command that is run in [CI](https://github.com/across-protocol/contracts/blob/3cfd778c1bd6d805b10cbe91f794bb24b060c26a/.github/workflows/pr.yml#L44). as a result, we might want `lint-rust` to be:
```
yarn prettier --list-different ./programs/**/*.rs && cargo fmt --all -- --check && cargo clippy
```

i.e it mirrors what is done in `lint-fix` except errors if the linting is not applied correctly.

Reinis-FRP: I'm not sure on this. These are conflicting formats, so prettier would error if cargo fmt was run before. I think its more appropriate to use prettier only in pre-commit hook and `lint-fix` script before rustfmt, but we don't want to break CI if code does not confirm to prettier rules (as they might conflict with rustfmt)

Reinis-FRP: The pre-commit hook or `lint-fix` applies these in order. First, it runs prettier, but once it runs `rustfmt` this changes the formatting. As prettier is opinionated and differs from standard `rustfmt` settings then having `prettier` within `lint-rust` would error.

chrismaree: I see. hhmm. that's a bit of a pain then that they've diverged like that. I guess if we want both prettier and rustfmt this is the path we need to take.

This means in some ways that the prettier formatter is optional but the rustfmt is compulsory (its the only linter enforced by CI). I think this is ok

chrismaree (approved):

package.json
Reinis-FRP: yes, we really need prettier to deal with anchor macros here

PR #705, opened 2024-10-29
improve(ERC7683): Adjust for updated ERC specs

nicholaspai:
Standard will likely be updated with changes [here](https://github.com/ethereum/ERCs/pull/689/files) so this PR adjusts for those changes:
- `originChainId` of intent is now uint256
- `orderId` is a new parameter in the `ResolvedCrossChainOrder` struct, which in Across' system can be the hash of the `V3RelayData`

### DIFF
--- a/contracts/erc7683/ERC7683OrderDepositor.sol
+++ b/contracts/erc7683/ERC7683OrderDepositor.sol
@@ -2,6 +2,7 @@
 pragma solidity ^0.8.0;
 
 import "../external/interfaces/IPermit2.sol";
+import { V3SpokePoolInterface } from "../interfaces/V3SpokePoolInterface.sol";
 import "@openzeppelin/contracts/utils/math/SafeCast.sol";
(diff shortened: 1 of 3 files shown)


contracts/erc7683/ERC7683OrderDepositor.sol
```diff
@@ -237,7 +238,8 @@ abstract contract ERC7683OrderDepositor is IOriginSettler {
             fillDeadline: order.fillDeadline,
             minReceived: minReceived,
             maxSpent: maxSpent,
-            fillInstructions: fillInstructions
+            fillInstructions: fillInstructions,
+            orderId: keccak256(abi.encode(relayData, block.chainid))
```
mrice32, line 242: Wouldn't this be the destination chainId?

nicholaspai: oh true

PR #706, opened 2024-10-29
improve(verifySpokePool): handling custom native tokens

melisaguevara:


pxrl (approved):

tasks/verifySpokePool.ts
```diff
@@ -64,7 +64,11 @@ task("verify-spokepool", "Verify the configuration of a deployed SpokePool")
     console.log("SpokePool.fillDeadlineBuffer()".padEnd(TEXT_PADDING) + ": " + fillDeadlineBuffer);
 
     const wrappedNative = bytes32ToAddress(results.wrappedNativeToken);
-    const wrappedNativeSymbol = spokeChainId === POLYGON ? "WMATIC" : "WETH";
+    const customWrappedNativeSymbols = {
+      [POLYGON]: "WMATIC",
+      [ALEPH_ZERO]: "WAZERO",
+    };
```
pxrl, line 70: @melisaguevara We can btw souce this directly from the `PRODUCTION_NETWORKS` object exported by `@across-protocol/constants`. It can be resolved by `PRODUCTION_NETWORKS[ALEPH_ZERO].nativeToken`. It might be worth doing that so we don't need to update this with future tokens.

melisaguevara: yes that would be definitely better! Let me update this.

melisaguevara: @pxrl The `PRODUCTION_NETWORKS` object currently only provides the native token, but we actually need the wrapped symbol. Do you think it would make sense to add that in as well?

pxrl: Ah, that's a good point. For now I'd probably be inclined to just assume that any wrapped variant would be prepended with "W" and add that locally, and reserve the right to be wrong about it in future. wdyt?

melisaguevara: Yes, I think that will work! I implemented it here [96f5c43](https://github.com/across-protocol/contracts/pull/706/commits/96f5c43e595b3763133dee8d4b7f389d44ea5df5)

bmzig (approved):

tasks/verifySpokePool.ts
```diff
@@ -64,7 +64,9 @@ task("verify-spokepool", "Verify the configuration of a deployed SpokePool")
     console.log("SpokePool.fillDeadlineBuffer()".padEnd(TEXT_PADDING) + ": " + fillDeadlineBuffer);
 
     const wrappedNative = bytes32ToAddress(results.wrappedNativeToken);
-    const wrappedNativeSymbol = spokeChainId === POLYGON ? "WMATIC" : "WETH";
+    const nativeTokenSymbol = PRODUCTION_NETWORKS[spokeChainId].nativeToken;
+    const wrappedPrefix = "W";
+    const wrappedNativeSymbol = `${wrappedPrefix}${nativeTokenSymbol}`;
```
james-a-morris, line 69: I'm not sure this will always be the case. Should we be explicit here?

melisaguevara: The idea with this approach is that we avoid adding a new chain for each custom gas token. I think we can go with this approach and revisit when there's a case that doesn't fit with this logic. wdyt?

pxrl: @james-a-morris My thinking is that it's worth proceeding with until we find the exception to the rule, and then re-evaluate in that case. The cost of adapting it later is quite low, and then we'd have the advantage of knowing what the alternatives scenario(s) is/are.

PR #707, opened 2024-10-30
test(svm): add cctp messages replay test

md0x:
Changes proposed in this PR:
- Add test to check that ccpt messages can't be replayed


Reinis-FRP (approved):
nice!

PR #709, opened 2024-10-30
fix(svm): instruction to create multiple token accounts

Reinis-FRP:
Adds method to create multiple associated token accounts. This instruction can be prepended to other instructions that expect ATAs being initialized.

### DIFF
--- a/programs/svm-spoke/src/instructions/create_token_accounts.rs
+++ b/programs/svm-spoke/src/instructions/create_token_accounts.rs
@@ -1,0 +1,46 @@
+use anchor_lang::prelude::*;
+use anchor_spl::{ associated_token::{ self, AssociatedToken }, token_interface::{ Mint, TokenInterface } };
+
+use crate::error::SvmError;
+
+#[derive(Accounts)]
+pub struct CreateTokenAccounts<'info> {
+    #[account(mut)]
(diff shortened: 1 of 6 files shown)


programs/svm-spoke/src/instructions/create_token_accounts.rs
```diff
@@ -0,0 +1,46 @@
...
+    pub mint: InterfaceAccount<'info, Mint>,
+
+    pub token_program: Interface<'info, TokenInterface>,
+
+    pub associated_token_program: Program<'info, AssociatedToken>,
+
+    pub system_program: Program<'info, System>,
+}
+
+pub fn create_token_accounts<'info>(ctx: Context<'_, '_, '_, 'info, CreateTokenAccounts<'info>>) -> Result<()> {
```
chrismaree, line 21: why in some cases do we need this context syntax like this `'_, '_, '_,`? we do the same thing in `handle_receive_message`. I'm not really clear on this.

Reinis-FRP: Context has 4 lifetime parameters. and since we need `'info` when referencing remaining accounts in CPI we need to pass `_` placeholders for all others

PR #710, opened 2024-10-31
feat: improve how depositor ATA interactions work

chrismaree:


programs/svm-spoke/src/instructions/deposit.rs
```diff
@@ -37,17 +35,18 @@ pub struct DepositV3<'info> {
...
     )]
     pub route: Account<'info, Route>,
 
     #[account(
         mut,
-        token::mint = mint,
-        token::authority = signer,
+        associated_token::mint = mint,
+        associated_token::authority = depositor,
         token::token_program = token_program
```
Reinis-FRP, line 47: change also token_program constraint to use associated_token

programs/svm-spoke/src/instructions/deposit.rs
```diff
@@ -17,15 +17,13 @@ use crate::{
...
-    fill_deadline: u32,
-    exclusivity_deadline: u32,
-    message: Vec<u8>
+    destination_chain_id: u64,
 )]
 pub struct DepositV3<'info> {
-    #[account(mut)]
+    #[account(
+        mut,
+        constraint = signer.key() == depositor @ SvmError::DepositorMustBeSigner
```
Reinis-FRP, line 25: this was explicit check in EVM implementation. why move it here in constraints?

programs/svm-spoke/src/instructions/deposit.rs
```diff
@@ -37,17 +35,18 @@ pub struct DepositV3<'info> {
 
     #[account(
         seeds = [b"route", input_token.as_ref(), state.key().as_ref(), destination_chain_id.to_le_bytes().as_ref()],
-        bump
+        bump,
+        constraint = route.enabled @ CommonError::DisabledRoute
```
Reinis-FRP, line 39: this was explicit check in EVM implementation. why move it here in constraints?

programs/svm-spoke/src/instructions/deposit.rs
Reinis-FRP, line 25: We don't have this restriction in EVM where sender can provide any depositor address

chrismaree: you're right, we dont really care about this as long as the authority is the depositor. previously we checked the authority is the signer, but now we check it is the depositor, which is functionally different. the goal I was trying to achieve is that deposit will work for a multi-sig & correct ATA support, where the depositor is the owner of the ATA. I think how it's structured, without this constraint, will support multisigs where the authority _is not the signer_.

programs/svm-spoke/src/instructions/deposit.rs
chrismaree: I moved it to match how we apply similar constraints on `fill.rs` https://github.com/across-protocol/contracts/blob/5c1dc6ed4ecb0bf87430730b91e91a26f11efad8/programs/svm-spoke/src/instructions/fill.rs#L27

Reinis-FRP: For fills EVM implementation has `unpausedFills` modifier, so we moved it to constraints so that implementation logic is as close as possible in SVM

programs/svm-spoke/src/error.rs
```diff
@@ -71,6 +71,8 @@ pub enum SvmError {
     InvalidClaimInitializer,
     #[msg("Seed must be 0 in production!")]
     InvalidProductionSeed,
+    #[msg("Depositor Must be signer!")]
+    DepositorMustBeSigner,
```
Reinis-FRP, line 75: this is now redundant

programs/svm-spoke/src/instructions/deposit.rs
Reinis-FRP: yes, we don't really care that depositor is signer. and in case of multisig the owner will be the token program. we don't need to replicate any authorsation checks here as they will be performed in CPI transfer

programs/svm-spoke/src/instructions/deposit.rs
chrismaree: ye, I recall this. that still feels funny to me though. Would we not rather be consistent between rust implementation? the EVM logic feels inconsistent between the modifier use vs inline checks anyways.

chrismaree: I'm happy to revert it though and safe it for later or keep it like this

Reinis-FRP (approved):

PR #712, opened 2024-10-31
feat: depositNow

chrismaree:
Adds Deposit Now, fixes 2 todos and addresses one small bug found.


programs/svm-spoke/src/instructions/deposit.rs
```diff
@@ -79,7 +74,7 @@ pub fn deposit_v3(
     exclusive_relayer: Pubkey,
     quote_timestamp: u32,
     fill_deadline: u32,
-    exclusivity_deadline: u32,
+    exclusivity_period: u32,
```
chrismaree, line 77: we had miss-named this prop! the implementation was correct but the variable name was wrong. see https://github.com/across-protocol/contracts/blob/31693cd43334a2a61916deaf4d0dac2e129d47d2/contracts/SpokePool.sol#L489

programs/svm-spoke/src/instructions/deposit.rs
```diff
@@ -90,10 +85,7 @@ pub fn deposit_v3(
 
     let current_time = get_current_time(state)?;
 
-    // TODO: if the deposit quote timestamp is bad it is possible to make this error with a subtraction
-    // overflow (from devnet testing). add a test to re-create this and fix it such that the error is thrown,
-    // not caught via overflow.
-    if current_time - quote_timestamp > state.deposit_quote_time_buffer {
+    if current_time.checked_sub(quote_timestamp).unwrap_or(u32::MAX) > state.deposit_quote_time_buffer {
```
chrismaree, line 88: if `checked_sub` underflows then it defaults to the `unwrap_or` value, of `u32::MAX`, forcing the if to return true, and throwing the error. This fixes the TODO above that I saw on testnet (and validated in unit tests).

programs/svm-spoke/src/instructions/deposit.rs
```diff
@@ -17,12 +17,7 @@ use crate::{
     output_token: Pubkey,
     input_amount: u64,
     output_amount: u64,
-    destination_chain_id: u64, // TODO: we can remove some of these instructions props
```
chrismaree, line 20: we dont need the extra fields here in the instructions. also, be removing them, we could re-use this `DepositV3` context for the `depositNow` method.

test/svm/SvmSpoke.common.ts
```diff
@@ -22,9 +22,9 @@ const exclusiveRelayer = Keypair.generate().publicKey;
 const outputToken = new PublicKey("1111111111113EsMD5n1VA94D2fALdb1SAKLam8j"); // TODO: this is lazy. this is cast USDC from Eth mainnet.
 const inputAmount = new BN(500000);
 const outputAmount = inputAmount;
-const quoteTimestamp = new BN(Math.floor(Date.now() / 1000) - 10); // 10 seconds ago.
+const quoteTimestamp = new BN(Math.floor(Date.now() / 1000) - 50); // 10 seconds ago.
 const fillDeadline = new BN(Math.floor(Date.now() / 1000) + 600); // 600 seconds from now.
-const exclusivityDeadline = new BN(Math.floor(Date.now() / 1000) + 300); // 300 seconds from now.
+const exclusivityPeriod = new BN(300); // 300 seconds.
```
chrismaree, line 27: this update applies to match the EVM changes and the correct input prop and how it's meant to be used.

programs/svm-spoke/src/instructions/deposit.rs
Reinis-FRP: note that EVM implementation has the same underflow issue, so we might fix it there too

programs/svm-spoke/src/instructions/deposit.rs
```diff
@@ -112,7 +109,65 @@ pub fn deposit_v3(
...
+    exclusive_relayer: Pubkey,
+    fill_deadline_offset: u32,
+    exclusivity_period: u32,
+    message: Vec<u8>,
+) -> Result<()> {
+    let state = &mut ctx.accounts.state;
+
+    if !ctx.accounts.route.enabled {
+        return err!(CommonError::DisabledRoute);
+    }
```
Reinis-FRP, line 140: This is redundant as this uses DepositV3 context that already checked the route.

Reinis-FRP: and we better not reimplement the deposit_v3 logic here, but call it internally

test/svm/SvmSpoke.Deposit.ts
```diff
@@ -15,7 +15,7 @@ const { provider, connection, program, owner, seedBalance, initializeState, depo
 const { createRoutePda, getVaultAta, assertSE, assert, getCurrentTime, depositQuoteTimeBuffer, fillDeadlineBuffer } =
   common;
 
-describe("svm_spoke.deposit", () => {
+describe.only("svm_spoke.deposit", () => {
```
Reinis-FRP, line 18: remove only

programs/svm-spoke/src/instructions/deposit.rs
chrismaree: Originally before removing the extra instructions, I was not able to re-call deposit_v3. I've now tried again and with the removed extra instructions (in particular, the ones missing when calling deposit_v3_now) I was able to re-use the original deposit_v3 method. take a look again.

programs/svm-spoke/src/instructions/deposit.rs
chrismaree: on master EVM implementation it's still called [exclusivityPeriod](https://github.com/across-protocol/contracts/blob/31693cd43334a2a61916deaf4d0dac2e129d47d2/contracts/SpokePool.sol#L489)

so I think as it is now we are consistent?

programs/svm-spoke/src/instructions/deposit.rs
chrismaree: in the evm implementation if it fails at that point does it revert with an underflow error or does it throw the associated revert message that we are expecting?

programs/svm-spoke/src/instructions/deposit.rs
```diff
@@ -134,45 +134,22 @@ pub fn deposit_v3_now(
     message: Vec<u8>,
 ) -> Result<()> {
     let state = &mut ctx.accounts.state;
```
Reinis-FRP, line 136: state does not need to be mut here anymore as we only use it for getting current time.

programs/svm-spoke/src/instructions/deposit.rs
Reinis-FRP: as per [this comment](https://github.com/across-protocol/contracts/blob/31693cd43334a2a61916deaf4d0dac2e129d47d2/contracts/SpokePool.sol#L500), it would underflow in EVM

chrismaree: right but my question was would it throw the error `InvalidQuoteTimestamp` or `underflow`. it seems it will throw with `underflow (see [here](https://github.com/across-protocol/contracts/blob/chrismaree/deposit-now/test/evm/hardhat/SpokePool.Deposit.ts#L316)).

my original point is that this is actually a bit confusing for a caller. there are situations where multiple invalid input values return the same error. equally, there are situations where some invalid input values, depending on the input, return inconsistent errors.

I agree with your original reply that we should fix this consistently, changing the evm implementation such that all invalid errors return the same error.

Reinis-FRP (approved):
cool!

programs/svm-spoke/src/instructions/deposit.rs
chrismaree: `pub fn get_current_time(_state: &State)` expects mutable state for this function. we might want to remove this requirement for this as it does not need to mutate it.

PR #714, opened 2024-11-01
improve(OP): More generic OP chain adapter deployment

pxrl:
This change permits an incremental step towards a totally generic chain adapter deployment for OP stack chains. It does this be restructuring the address mappings for OP stack chains, such that the L1 contract addresses can be resolved solely with the combination of HubPool and SpokPool chain IDs. It also updates the new OP_Adapter deployment script to support a runtime-defined SpokePool chain ID. This is clunkily supplied via the environment because hardhat deploy doesn't seem to permit arbitrary runtime arguments being supplied, but that aside, this should ease subsequent deployments.

### DIFF
--- a/deploy/002_deploy_optimism_adapter.ts
+++ b/deploy/002_deploy_optimism_adapter.ts
@@ -1,6 +1,7 @@
-import { L1_ADDRESS_MAP, USDC, WETH } from "./consts";
 import { DeployFunction } from "hardhat-deploy/types";
 import { HardhatRuntimeEnvironment } from "hardhat/types";
+import { CHAIN_IDs } from "../utils";
+import { L1_ADDRESS_MAP, OP_STACK_ADDRESS_MAP, USDC, WETH } from "./consts";
 
 const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
   const { deployer } = await hre.getNamedAccounts();
@@ -8,8 +9,8 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
 
   const args = [
     WETH[chainId],
-    L1_ADDRESS_MAP[chainId].optimismCrossDomainMessenger,
-    L1_ADDRESS_MAP[chainId].optimismStandardBridge,
+    OP_STACK_ADDRESS_MAP[chainId][CHAIN_IDs.OPTIMISM].L1CrossDomainMessenger,
+    OP_STACK_ADDRESS_MAP[chainId][CHAIN_IDs.OPTIMISM].L1StandardBridge,
     USDC[chainId],
     L1_ADDRESS_MAP[chainId].cctpTokenMessenger,
   ];
--- a/deploy/002_deploy_optimism_adapter.ts
+++ b/deploy/002_deploy_optimism_adapter.ts
@@ -3,14 +3,17 @@ import { HardhatRuntimeEnvironment } from "hardhat/types";
 import { CHAIN_IDs } from "../utils";
 import { L1_ADDRESS_MAP, OP_STACK_ADDRESS_MAP, USDC, WETH } from "./consts";
 
+const SPOKE_CHAIN_ID = CHAIN_IDs.OPTIMISM;
+
 const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
   const { deployer } = await hre.getNamedAccounts();
   const chainId = parseInt(await hre.getChainId());
+  const opStack = OP_STACK_ADDRESS_MAP[chainId][SPOKE_CHAIN_ID];
 
   const args = [
     WETH[chainId],
-    OP_STACK_ADDRESS_MAP[chainId][CHAIN_IDs.OPTIMISM].L1CrossDomainMessenger,
-    OP_STACK_ADDRESS_MAP[chainId][CHAIN_IDs.OPTIMISM].L1StandardBridge,
+    opStack.L1CrossDomainMessenger,
+    opStack.L1StandardBridge,
     USDC[chainId],
     L1_ADDRESS_MAP[chainId].cctpTokenMessenger,
   ];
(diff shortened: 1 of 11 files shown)


pxrl:
nb. This only touches constants and deployment scripts, so there should be no audit sensitivity.

deploy/consts.ts
```diff
@@ -69,24 +45,82 @@ export const L1_ADDRESS_MAP: { [key: number]: { [contractName: string]: string }
...
 
-    // https://docs.lisk.com/contracts
-    liskCrossDomainMessenger: "0x857824E6234f7733ecA4e9A76804fd1afa1A3A2C",
-    liskStandardBridge: "0x1Fb30e446eA791cd1f011675E5F3f5311b70faF5",
-
-    // https://docs.blast.io/building/contracts
-    l1BlastBridge: "0xc644cc19d2A9388b71dd1dEde07cFFC73237Dca8",
-    blastCrossDomainMessenger: "0x9338F298F29D3918D5D1Feb209aeB9915CC96333",
-    blastStandardBridge: "0xDeDa8D3CCf044fE2A16217846B6e1f1cfD8e122f",
+export const OP_STACK_ADDRESS_MAP: {
```
pxrl, line 58: Note also we use a lot of these contract addresses in the relayer repository as well, and there are duplicate definitions for them. I think a subsequent step could be to relocate these defs over to the constants repo, so we can define them centrally and inherit in the contracts and relayer repositories. That eases new OP stack chain introductions and helps to reduce risk.

deploy/consts.ts
```diff
@@ -24,11 +20,7 @@ export const L1_ADDRESS_MAP: { [key: number]: { [contractName: string]: string }
     polygonERC20Predicate: "0x40ec5B33f54e0E8A33A975908C5BA1c14e5BbbDf",
     polygonRegistry: "0x33a02E6cC863D393d6Bf231B697b82F6e499cA71",
     polygonDepositManager: "0x401F6c983eA34274ec46f84D70b31C151321188b",
-    baseCrossDomainMessenger: "0x866E82a600A1414e583f7F13623F1aC5d58b0Afa",
-    baseStandardBridge: "0x3154Cf16ccdb4C6d922629664174b904d80F2C35",
-    l1BlastBridge: "0x3a05E5d33d7Ab3864D53aaEc93c8301C1Fa49115",
-    blastCrossDomainMessenger: "0x5D4472f31Bd9385709ec61305AFc749F0fA8e9d0",
-    blastStandardBridge: "0x697402166Fbf2F22E970df8a6486Ef171dbfc524",
+    l1BlastBridge: "0xc644cc19d2A9388b71dd1dEde07cFFC73237Dca8",
```
bmzig, line 23: I think this was meant to be deleted?

deploy/002_deploy_optimism_adapter.ts
```diff
@@ -1,15 +1,16 @@
...
 
 const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
   const { deployer } = await hre.getNamedAccounts();
   const chainId = parseInt(await hre.getChainId());
 
   const args = [
     WETH[chainId],
-    L1_ADDRESS_MAP[chainId].optimismCrossDomainMessenger,
-    L1_ADDRESS_MAP[chainId].optimismStandardBridge,
+    OP_STACK_ADDRESS_MAP[chainId][CHAIN_IDs.OPTIMISM].L1CrossDomainMessenger,
```
bmzig, line 12: Since `CHAIN_IDs.OPTIMISM` is hardcoded, these scripts will only ever work for testnet. Do you want to do something like `const spokeChainId = chainId === CHAIN_IDs.MAINNET ? CHAIN_IDs.OPTIMISM : CHAIN_IDs.OPTIMISM_SEPOLIA`? (Although this might not work for all chains). I'm mentioning this since you defined the testnet addresses in `OP_STACK_ADDRESS_MAP`, but they won't be used without changing the script. Maybe that's not that big of a deal?

bmzig (approved):
I checked also and the bridge addresses check out.

deploy/042_deploy_lisk_adapter.ts
pxrl, line 1: We can actually delete this adapter as well as World Chain because they are both now served by the OP Adapter contract. As other chains migrate to upgradable USDC we'll be able to delete those too.

deploy/002_deploy_optimism_adapter.ts
nicholaspai, line 12: To reduce chance of error, can we do something like `const addressMap = OP_STACK_ADDRESS_MAP[chainId][CHAIN_IDs.OPTIMISM]` above this line?  I worry we accidentally type in the wrong chain ID here and use the wrong address, which could be hard to detect

Same ask for all the deploy files, of course

pxrl: Yep; that's a good call.

In terms of catching errors, I think it'd be standard practice to test any deployed adapter with the chain adapter test task that was added in this commit: 8cc0bb59a4d70c0a69185f981a61964cd50d15c5.

With this we can verify that that tokens are relayed correctly. Bennett separately added a task to test messaging.

deploy/consts.ts
pxrl: Good catch - it was actually incorrect. Fixed here: https://github.com/across-protocol/contracts/pull/714/commits/4e81f7a0b72c310c80e30242ec5c7ca22acfd529

pxrl:
> Looks good. I'm guessing we are just keeping the `deploy_[op network]_adapter` scripts for the sake of bookkeeping at this point?

Kinda, yeah. I do think we could additionally consolidate:
- Optimism & Base (OP stack, native USDC via CCTP)
- Zora & Mode (OP stack, bridged USDC)

Blast is a total snowflake of course, and Redstone now seems to have an upgradable USDC deployment [here](https://explorer.redstone.xyz/token/0xD5d59fC063e7548b6015A36fEb10B875924A19be?tab=contract), so in theory we could remove the Redstone Adapter and in favour of the OP Adapter instead (even irrespective of whether we support USDC there).

However...I probably don't expect many (or any) _new_ chain deployments to fit the Optimism/Base pattern or the Zora/Mode pattern, so I think this optimisation for the OP Stack + upgradable USDC combination gives us the most benefit for the least effort.

PR #716, opened 2024-11-01
feat: superseed deployment

bmzig:
This deploys a `Superseed_SpokePool`, which is a clone of `WorldChain_SpokePool` up to the `USDC_BRIDGE` address.


deploy/056_deploy_superseed_adapter.ts
pxrl, line 1: We might be able to drop this one in anticipation of https://github.com/across-protocol/contracts/pull/714 being merged.

PR #718, opened 2024-11-04
fix(svm): use free function in handling cctp

Reinis-FRP:
For the sake of consistency use free function in internal handle_receive_message call. This also simplifies the external function.

Fixes: https://linear.app/uma/issue/ACX-3102/librs-refactor-handle-receive-message-to-be-consistent-with-using-free

### DIFF
--- a/programs/svm-spoke/src/instructions/handle_receive_message.rs
+++ b/programs/svm-spoke/src/instructions/handle_receive_message.rs
@@ -47,13 +47,15 @@ pub struct HandleReceiveMessageParams {
     pub authority_bump: u8,
 }
 
-// TODO: consider refactoring this to be consistent with using free functions.
-impl<'info> HandleReceiveMessage<'info> {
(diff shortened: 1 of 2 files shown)


programs/svm-spoke/src/instructions/handle_receive_message.rs
```diff
@@ -47,13 +47,15 @@ pub struct HandleReceiveMessageParams {
...
-impl<'info> HandleReceiveMessage<'info> {
-    pub fn handle_receive_message(&self, params: &HandleReceiveMessageParams) -> Result<Vec<u8>> {
-        // Return instruction data for the self invoked CPI based on the received message body.
-        translate_message(&params.message_body)
-    }
+pub fn handle_receive_message<'info>(
+    ctx: Context<'_, '_, '_, 'info, HandleReceiveMessage<'info>>,
+    params: HandleReceiveMessageParams,
+) -> Result<()> {
+    let self_ix_data = translate_message(&params.message_body)?;
```
chrismaree, line 54: this order of operators is easier to understand

chrismaree (approved):
nice.

PR #720, opened 2024-11-04
proposal: ensure that EVM errors are always consistant on underflows

chrismaree:
This PR is a proposal to change how quote timestamp values are validated. Previously, depending on the value for `quoteTimestamp` this require would either revert with `InvalidQuoteTimestamp` OR `underflow`. this makes interfacing with this method more difficult. 

Now, it will always return `InvalidQuoteTimestamp`. The downside to this change is it increased the deposit gas cost from `100800` to `100870`. This increase might be unavoidable if we want this.

So the question is: is it work the extra gas for the better errors?


pxrl (approved):
This has confused some integrators on occasion, +1 from me.

contracts/SpokePool.sol
```diff
@@ -502,7 +502,8 @@ abstract contract SpokePool is
 
         // slither-disable-next-line timestamp
         uint256 currentTime = getCurrentTime();
-        if (currentTime - quoteTimestamp > depositQuoteTimeBuffer) revert InvalidQuoteTimestamp();
```
Reinis-FRP, line 505: need to update comment above as there should be no overflow anymore

Reinis-FRP (approved):

pxrl (approved):

PR #723, opened 2024-11-05
proposal: refine how we pull in relay_hash into account rather than prop

chrismaree:
This PR refines how we pull in relay_hash into functions that need it. rather than it being an unused prop, only used within the instructions, it is passed in as an unchecked account.


programs/svm-spoke/src/lib.rs
```diff
@@ -165,7 +165,6 @@ pub mod svm_spoke {
     // Relayer methods.
     pub fn fill_v3_relay(
         ctx: Context<FillV3Relay>,
-        _relay_hash: [u8; 32],
```
chrismaree, line 168: able to remove it from the props, and therefore the instructions.

programs/svm-spoke/src/instructions/slow_fill.rs
```diff
@@ -131,12 +134,15 @@ pub struct ExecuteV3SlowRelayLeaf<'info> {
...
+    pub relay_hash: UncheckedAccount<'info>,
+
     #[account(
         mut,
-        seeds = [b"fills", relay_hash.as_ref()],
+        seeds = [b"fills", relay_hash.key().as_ref()],
         bump,
         // Make sure caller provided relay_hash used in PDA seeds is valid.
-        constraint = is_relay_hash_valid(&relay_hash, &slow_fill_leaf.relay_data, &state) @ SvmError::InvalidRelayHash
+        constraint = is_relay_hash_valid(&relay_hash.key().as_ref().try_into().unwrap(), &slow_fill_leaf.relay_data, &state) @ SvmError::InvalidRelayHash
```
md0x, line 145: What happens if we pass an account that fails the `try_into()` conversion, i.e. cannot be unwrapped? Does it throw `SvmError::InvalidRelayHash`, or should we handle it explicitly with something like `.as_ref().try_into().map_err(|_| ...`

chrismaree: good question. will test it and report back.

chrismaree:
```
try {
      const invalidRelayHash = new PublicKey("InvalidPublicKeyString");
      await program.methods
        .fillV3Relay(relayData, new BN(1), relayer.publicKey)
        .accounts({
          ...accounts,
          relayHash: invalidRelayHash,
        })
        .signers([relayer])
        .rpc();
      assert.fail("Should not be able to fill relay with an invalid relayHash");
    } catch (err: any) {
      console.log("Error with non-correct public key relayHash:", err.toString());
    }

    // Test with a relayHash that is an invalid public key
    try {
      const invalidRelayHash = new PublicKey(new Uint8Array(32).fill(255)); // Invalid public key
      await program.methods
        .fillV3Relay(relayData, new BN(1), relayer.publicKey)
        .accounts({
          ...accounts,
          relayHash: invalidRelayHash,
        })
        .signers([relayer])
        .rpc();
      assert.fail("Should not be able to fill relay with an invalid relayHash");
    } catch (err: any) {
      console.log("Error with invalid public key relayHash:", err.toString());
    }
```
returns:
```
Error with non-correct public key relayHash: Error: Non-base58 character
Error with invalid public key relayHash: AnchorError caused by account: fill_status. Error Code: ConstraintSeeds. Error Number: 2006. Error Message: A seeds constraint was violated.
```
in either case we don't actually get the expected error, rather the more cryptic issue on seed constrain violation, which is a bit painful. this makes this a bit worse than the previous implementation as it kinda hides the errors, which is less good.

WDYT?

chrismaree: a bit more info on this...I played with it some more tonight and _we generate the same errors with the previous implementation_. i.e we did not loose error quality.

md0x: Ok, I think this solution is good IMO, I would just add a @dev comment explaining why we are doing this.

md0x (approved):

Reinis-FRP (approved):

chrismaree:
we've decided to not use `UncheckedAccount` and rather stick to using the input props as this is potentially more complex for callers, even if the rust is a bit cleaner.

PR #725, opened 2024-11-06
fix: support native token bridging in the forwarder/withdrawal helper

bmzig:
> The [WithdrawalHelperBase](https://github.com/across-protocol/contracts/blob/5a0c67c984d19a3bb843a4cec9bb081734583dd1/contracts/chain-adapters/l2/WithdrawalHelperBase.sol#L18) and [ForwarderBase](https://github.com/across-protocol/contracts/blob/5a0c67c984d19a3bb843a4cec9bb081734583dd1/contracts/chain-adapters/ForwarderBase.sol#L18) contracts are designed to be deployed on L2s and assist in moving tokens and messages to and from L3 chains. These contracts are inherited by the chain-specific contracts, currently designed for Arbitrum and OVM-based blockchains. However, none of these contracts handle ETH transfers correctly. This is because they do not contain the receive function, which causes any attempt to transfer ETH to these contracts to fail.

> In the case of forwarder contracts, the lack of the receive function means that WETH transfers, which rely on unwrapping before bridging such as [the transfers made through the Optimism_Adapter](https://github.com/across-protocol/contracts/blob/5a0c67c984d19a3bb843a4cec9bb081734583dd1/contracts/chain-adapters/Optimism_Adapter.sol#L101-L105), will fail, leaving the ETH in the bridge until the contract can be upgraded. In the case of withdrawal helper contracts, the lack of the receive function implies that they will not be capable of unwrapping WETH [in an attempt to transfer it to Ethereum](https://github.com/across-protocol/contracts/blob/5a0c67c984d19a3bb843a4cec9bb081734583dd1/contracts/chain-adapters/l2/Ovm_WithdrawalHelper.sol#L114-L124). Additionally, they will not be capable of receiving ETH bridged from L3s. Moreover, while the withdrawal helper contracts contain [token-bridging logic](https://github.com/across-protocol/contracts/blob/5a0c67c984d19a3bb843a4cec9bb081734583dd1/contracts/chain-adapters/l2/WithdrawalHelperBase.sol#L87-L101), they do not support bridging ETH. This means that even if they were capable of receiving ETH and ETH was bridged to them from L3s, it could not be routed to L1.

> Consider adding a receive function to the ForwarderBase and WithdrawalHelperBase contracts to facilitate incoming ETH transfers and the unwrapping of WETH tokens during bridging. As the contracts do not support bridging ETH directly, the receive function should include logic to ensure that incoming ETH is handled correctly and can be sent on to the target chain.

This PR adds `receive` to both `ForwarderBase` and `WithdrawalHelperBase` so that it can receive native token transfers from L1 or L3 bridges. Both the `ForwarderBase` and `WithdrawalHelperBase` now contain a hook named `_wrapNativeToken` which is called whenever a L2-L3 bridge or withdrawal which includes the L2's native token is initiated on the respective contracts. Both the `ForwarderBase` and `WithdrawalHelperBase` now take an additional argument `_wrappedNativeToken` in their constructor, which is used to deposit the native token in the `_wrapNativeToken` hook.

### DIFF
--- a/contracts/chain-adapters/Arbitrum_Forwarder.sol
+++ b/contracts/chain-adapters/Arbitrum_Forwarder.sol
@@ -3,6 +3,7 @@ pragma solidity ^0.8.0;
 
 import { ForwarderBase } from "./ForwarderBase.sol";
 import { CrossDomainAddressUtils } from "../libraries/CrossDomainAddressUtils.sol";
+import { WETH9Interface } from "../external/interfaces/WETH9Interface.sol";
 
 /**
  * @title Arbitrum_Forwarder
@@ -19,10 +20,9 @@ contract Arbitrum_Forwarder is ForwarderBase {
 
     /**
      * @notice Constructs an Arbitrum-specific forwarder contract.
-     * @dev Since this is a proxy contract, we only set immutable variables in the constructor, and leave everything else to be initialized.
-     * This includes variables like the cross domain admin.
+     * @param _wrappedNativeToken Address of the wrapped native token contract on the L2.
      */
-    constructor() ForwarderBase() {}
+    constructor(WETH9Interface _wrappedNativeToken) ForwarderBase(_wrappedNativeToken) {}
 
     /**
      * @notice Initializes the forwarder contract.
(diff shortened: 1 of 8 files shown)


nicholaspai (approved):

pxrl (approved):

pxrl:
@bmzig If you merge in the latest change from master you'll get a fix from Reinis that addresses the SVM test failure.

PR #728, opened 2024-11-07
fix: N01-N14: Address Noncritical Issues

bmzig:
This updates parts of the L3 code which were deemed as non-critical issues.


contracts/chain-adapters/ZkStack_CustomGasToken_Adapter.sol
```diff
@@ -70,7 +70,7 @@ contract ZkStack_CustomGasToken_Adapter is AdapterInterface {
     // Custom gas token funder
     FunderInterface public immutable CUSTOM_GAS_TOKEN_FUNDER;
 
-    event ZkStackMessageRelayed(bytes32 canonicalTxHash);
+    event ZkStackMessageRelayed(bytes32 indexed canonicalTxHash);
```
bmzig, line 73: If we never see ourselves filtering on this event. We should remove the `indexed` attribute.

pxrl (commented):
LGTM - I was able to reconcile all changes to the notes raised by the audit. I have a couple of residual queries to align on cc @mrice32 @nicholaspai

contracts/Arbitrum_SpokePool.sol
```diff
@@ -1,7 +1,4 @@
 // SPDX-License-Identifier: BUSL-1.1
-
-// Arbitrum only supports v0.8.19
-// See https://docs.arbitrum.io/for-devs/concepts/differences-between-arbitrum-ethereum/solidity-support#differences-from-solidity-on-ethereum
 pragma solidity ^0.8.19;
```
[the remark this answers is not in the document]
bmzig: I think we just don't _need_ to use 0.8.19 anymore: https://docs.arbitrum.io/build-decentralized-apps/arbitrum-vs-ethereum/solidity-support, so we can change `hardhat.config.ts` to use the same solc version as all our other contracts for the Arbitrum ones.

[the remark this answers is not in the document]
nicholaspai: Can we confirm this? If so, then yes we can change this. We should also check on these other contracts we cap at 8.19 [here](https://github.com/across-protocol/contracts/blob/d2b2b98dad18667b32cc022303ebc30b3acba5f7/hardhat.config.ts#L68)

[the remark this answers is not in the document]
bmzig: Sounds good. I updated the hardhat config as well here: https://github.com/across-protocol/contracts/pull/728/commits/12c4d532508329d9d7b12b22d33e467f28414532. Arbitrum's support for `PUSH0` was confirmed in the writeup for `N-13` as well. We should also be fine for Aleph Zero since they deployed ArbOS20, and the documentation says it was supported in an earlier version.

nicholaspai:
Can you comment on which lines are related to which issues? would make it faster to review

contracts/Arbitrum_SpokePool.sol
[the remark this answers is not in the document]
nicholaspai: So should we change this line to be ^0.8.0?

contracts/chain-adapters/l2/Ovm_WithdrawalHelper.sol
```diff
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: BUSL-1.1
 
-pragma solidity ^0.8.0;
+pragma solidity ^0.8.19;
```
nicholaspai, line 3: If we're going to change Arbitrum to 0.8.0 let's keep this, right?

contracts/chain-adapters/l2/WithdrawalHelperBase.sol
```diff
@@ -2,7 +2,6 @@
 pragma solidity ^0.8.19;
```
nicholaspai, line 2: Reminder to set this to 0.8.0 if we're going to do the same with arbitrum

bmzig: Good point. https://github.com/across-protocol/contracts/pull/728/commits/b6cae1c347123855208db93f43a03cb0736ba15b

contracts/chain-adapters/l2/Ovm_WithdrawalHelper.sol
bmzig: Yeah, we should. It's nice to keep things consistent. https://github.com/across-protocol/contracts/pull/728/commits/b6cae1c347123855208db93f43a03cb0736ba15b

contracts/erc7683/ERC7683Across.sol
```diff
@@ -59,17 +59,17 @@ library ERC7683Permit2Lib {
             "AcrossOrderData orderData)"
         );
 
-    bytes internal constant CROSS_CHAIN_ORDER_EIP712_TYPE =
-        abi.encodePacked(CROSS_CHAIN_ORDER_TYPE, ACROSS_ORDER_DATA_TYPE);
-    bytes32 internal constant CROSS_CHAIN_ORDER_TYPE_HASH = keccak256(CROSS_CHAIN_ORDER_EIP712_TYPE);
+    bytes internal constant GASLESS_CROSS_CHAIN_ORDER_EIP712_TYPE =
```
bmzig, line 62: N-10

contracts/erc7683/ERC7683Permit2Lib.sol
bmzig, line 1: N-11

contracts/erc7683/ERC7683Permit2Lib.sol
```diff
@@ -59,17 +59,17 @@ library ERC7683Permit2Lib {
             "AcrossOrderData orderData)"
         );
 
-    bytes internal constant CROSS_CHAIN_ORDER_EIP712_TYPE =
-        abi.encodePacked(CROSS_CHAIN_ORDER_TYPE, ACROSS_ORDER_DATA_TYPE);
-    bytes32 internal constant CROSS_CHAIN_ORDER_TYPE_HASH = keccak256(CROSS_CHAIN_ORDER_EIP712_TYPE);
+    bytes internal constant GASLESS_CROSS_CHAIN_ORDER_EIP712_TYPE =
```
bmzig, line 62: N-10

bmzig:
> Can you comment on which lines are related to which issues? would make it faster to review

Paul made comments on most of them, and I covered the gaps. There should be comments for 1-14

contracts/chain-adapters/ForwarderBase.sol
```diff
@@ -98,8 +96,15 @@ abstract contract ForwarderBase is UUPSUpgradeable, ForwarderInterface {
...
-        chainAdapters[_destinationChainId] = _l2Adapter;
-        emit ChainAdaptersUpdated(_destinationChainId, _l2Adapter);
+        _updateAdapter(_destinationChainId, _l2Adapter);
+    }
+
+    /**
+     * @notice Removes this contract's set adapter for the specified chain ID.
+     * @param _destinationChainId The chain ID of the target network.
+     */
+    function removeAdapter(uint256 _destinationChainId) external onlyAdmin {
```
nicholaspai, line 106: Do you think we should allow `updateAdapter` to be called with `address(0)` as the `_l2Adapter` to allow the admin to "remove" an adapter? I'm pretty indifferent, the tradeoff would be a smaller contract interface versus adding this intuitive function and preventing resetting adapter to address(0) accidentally

contracts/chain-adapters/ZkStack_Adapter.sol
```diff
@@ -21,8 +21,11 @@ import { BridgeHubInterface } from "../interfaces/ZkStackBridgeHub.sol";
 contract ZkStack_Adapter is AdapterInterface {
     using SafeERC20 for IERC20;
 
+    // The ZkSync bridgehub contract treats address(1) to represent ETH.
+    address private constant ETH_TOKEN_ADDRESS = address(1);
```
nicholaspai, line 25: Is this token an ERC20? that's so weird

contracts/erc7683/ERC7683OrderDepositor.sol
```diff
@@ -352,7 +352,7 @@ abstract contract ERC7683OrderDepositor is IOriginSettler {
         address exclusiveRelayer,
         uint32 quoteTimestamp,
         uint32 fillDeadline,
-        uint32 exclusivityDeadline,
+        uint32 exclusivityPeriod,
```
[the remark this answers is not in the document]
nicholaspai: > I think exclusivityDeadline is objectively a better name than exclusivityPeriod

Are you saying this because we'd only specify exclusivity as a timestamp when initiating 7683 orders? Even for both gasless and gasful deposits?

nicholaspai (commented):
https://github.com/across-protocol/contracts/pull/728/files#r1844570282
https://github.com/across-protocol/contracts/pull/728/files#r1846744574
https://github.com/across-protocol/contracts/pull/728/files#r1846747834

A few outstanding comments then we'll be good, good work!

contracts/Arbitrum_SpokePool.sol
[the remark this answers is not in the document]
bmzig: Yeah. Done here https://github.com/across-protocol/contracts/pull/728/commits/8e4376964270bb50edfafa3d41becf02e9e81823

contracts/chain-adapters/ForwarderBase.sol
bmzig: I originally separated the two since I didn't want to run the risk of inadvertently setting the adapter to 0, but on second thought, we really won't run into this issue, since I'd think we would catch the error before executing the update transaction. I removed it https://github.com/across-protocol/contracts/pull/728/commits/8e4376964270bb50edfafa3d41becf02e9e81823

contracts/chain-adapters/ZkStack_Adapter.sol
bmzig: It looks like they just use it as a special case for ETH https://github.com/matter-labs/era-contracts/blob/aafee035db892689df3f7afe4b89fd6467a39313/l1-contracts/contracts/bridge/L1SharedBridge.sol#L326

contracts/chain-adapters/ForwarderBase.sol
nicholaspai: yeah I don't think that's a risk since its an onlyAdmin function and we can assume the admin wouldn't do this

PR #737, opened 2024-11-08
fix: Reset anchor step in publish workflow

melisaguevara:


programs/svm-spoke/Cargo.toml
```diff
@@ -3,7 +3,7 @@ name = "svm-spoke"
 version = "0.1.0"
 description = "Created with Anchor"
 edition = "2021"
-
+publish = false
```
melisaguevara, line 6: cc @chrismaree @Reinis-FRP We are enforcing that the rust files aren't published to crate.io for now

PR #747, opened 2024-11-13
fix(svm): across plus to solana

Reinis-FRP:
This introduces support for Across+ targeting Solana. Since Solana instructions require additional context on account keys and their permissions we require that messages targeting Solana adhere to common format to encapsulate this additional information.

This also adds multicall handler implementation that allows to execute multiple instructions resembling [compiled instruction](https://solana.com/docs/core/transactions#array-of-instructions) format.

Considering the current structure of `FilledV3Relay` event we get limited by inner instruction size bounds when handler message exceeds 472 bytes, which based on tests allows up to 5 distributions to different recipients within the handler. This could be increased approx 1.6x if we were to emit the hashed version of the message before hitting outer transaction size limits. This could then be further increased if we were loading instruction parameters via buffer account for fills, like we do for executing relayer refunds.

At current implementation the Spoke would invoke the handler without signatures, assuming the handler does not need to retain any privileges on behalf of end users (e.g. swapping out to user controlled ATA). If however the handler needs to retain user privileges (e.g. withdrawing from money market) we would need to introduce signed handler invocations where PDA seeds are derived from some combination of depositor address and origin chain.


programs/svm-spoke/src/utils/message_utils.rs
```diff
@@ -0,0 +1,102 @@
...
+                }
+                accounts.push(AccountMeta::new_readonly(message_account_key, false));
+            }
+        }
+    }
+
+    // Transfer value amount from the relayer to the first account in the message accounts.
+    // Note that the depositor is responsible to make sure that after invoking the handler the recipient account will
+    // not hold any balance that is below its rent-exempt threshold, otherwise the fill would fail.
+    if message.value_amount > 0 {
```
Reinis-FRP, line 78: This is mostly useful when invoked handler transaction requires account initialization payer (e.g. when creating ATAs)

programs/svm-spoke/src/utils/message_utils.rs
```diff
@@ -0,0 +1,102 @@
...
+    data.extend_from_slice(&HANDLE_V3_ACROSS_MESSAGE_DISCRIMINATOR);
+    AnchorSerialize::serialize(&message.handler_message, &mut data)?;
+
+    let instruction = Instruction {
+        program_id: message.handler,
+        accounts,
+        data,
+    };
+
+    // TODO: consider if the message handler requires signed invocation.
```
Reinis-FRP, line 98: This mostly depends on whether the handler needs to retain any privileges on behalf of the user. E.g. if the handler performs a swap and forwards swapped-in tokens to user's ATA then its not important. On the other hand, if the handler deposits funds in a money market that does not allow depositing on behalf of end user then the handler would need to retain rights to withdraw the funds where such call must be authenticated by spoke pool PDA that is derived from depositing user address and origin chain.

programs/svm-spoke/src/instructions/slow_fill.rs
```diff
@@ -225,6 +225,10 @@ pub fn execute_v3_slow_relay_leaf(
     // Emit the FilledV3Relay event
     let message_clone = relay_data.message.clone(); // Clone the message before it is moved
 
+    if message_clone.len() > 0 {
+        invoke_handler(ctx.accounts.signer.as_ref(), ctx.remaining_accounts, &message_clone)?;
```
chrismaree, line 229: `execute_v3_slow_relay_leaf` will use more transaction size than `fill_v3_relay`, due to the proof. This might further limit the complexity of the across+ style hooks that can be used depending on whether they were executed fast or slow.

programs/svm-spoke/src/utils/message_utils.rs
chrismaree: right. main issue being how much size we have left in the tx. I think we can circle back to this once we've totally validated the upper bound of the tx sizing here and that it covers enough basic usecases.

chrismaree: for example this tx that we did from the dev wallet: https://solscan.io/tx/5swoRhHcqpxRXRhyyaCRWqTGDcNuQ8avZXptLomyUFPEPKjQceiMfSPdypA9pcdZQu3inGHmhoZmBWdEbZwGRZYC

this is a deposit into a money market protocol from the perspective of the user. this would not work in the current pattern given this restriction. it could be possible to re-structure this tx such that the relayer deposits and then transfers ownership of the position to the user but this seems a fair bit more complex and requires this kind of transfer logic within the target protocol. 

in general though not having authentication might be ok? we need to think about the exact user flow that we want to accommodate for across+.

src/SvmUtils.ts
```diff
@@ -175,3 +196,299 @@ export class LargeAccountsCoder<A extends string = string> extends BorshAccounts
...
+    const staticAccountKeys = [
+      ...writableNonSigners.map(([address]) => new PublicKey(address)),
+      ...readonlyNonSigners.map(([address]) => new PublicKey(address)),
+    ];
+
+    return [header, staticAccountKeys];
+  }
+}
+
+// Extended version of legacy Message to handle compilation of unsigned transactions. Base implementation is here:
```
chrismaree, line 265: ^

scripts/svm/fakeFillWithRandomDistribution.ts
```diff
@@ -0,0 +1,199 @@
...
+      BigInt(distributionAmount.toString()),
+      tokenDecimals
+    );
+    transferInstructions.push(transferInstruction);
+  }
+
+  // Encode handler message for the token distribution.
+  const multicallHandlerCoder = new MulticallHandlerCoder(transferInstructions);
+  const handlerMessage = multicallHandlerCoder.encode();
+  const message = new AcrossPlusMessageCoder({
```
chrismaree, line 104: ^

programs/svm-spoke/src/utils/message_utils.rs
chrismaree: ok after discussion this more we've decided that this is infact _not_ a requirement as the filler on evm is not permissioned (no preservation of msg.sender on evm) so this is fine to to have on svm.

programs/svm-spoke/src/utils/message_utils.rs
```diff
@@ -0,0 +1,102 @@
...
+) -> Result<()> {
+    let message =
+        AcrossPlusMessage::deserialize(&mut &message[..]).map_err(|_| AcrossPlusError::MessageDidNotDeserialize)?;
+
+    // First remaining account is the handler and the rest are accounts to be passed to the message handler.
+    let message_accounts_len = message.accounts.len();
+    if remaining_accounts.len() != message_accounts_len + 1 {
+        return err!(AcrossPlusError::InvalidMessageKeyLength);
+    }
+    if (message.read_only_len as usize) > message_accounts_len {
```
md0x, line 33: Are accounts sorted by read_only or writable?

programs/svm-spoke/src/event.rs
```diff
@@ -79,7 +79,8 @@ pub struct FilledV3Relay {
     pub relayer: Pubkey,
     pub depositor: Pubkey,
     pub recipient: Pubkey,
-    pub message: Vec<u8>,
+    // TODO: update EVM implementation to use message_hash in all fill related events.
```
chrismaree, line 82: yup, we'll need to update this separately.

scripts/svm/fakeFillWithRandomDistribution.ts
```diff
@@ -0,0 +1,216 @@
+// This script implements a fill where relayed tokens are distributed to random recipients via the message handler.
```
chrismaree, line 1: this is very useful to see the intended encoding structure end to end.

test/svm/SvmSpoke.Fill.AcrossPlus.ts
```diff
@@ -0,0 +1,355 @@
...
+
+    const message = new AcrossPlusMessageCoder({
+      handler: handlerProgram.programId,
+      readOnlyLen: multicallHandlerCoder.readOnlyLen,
+      valueAmount: new BN(0),
+      accounts: multicallHandlerCoder.compiledMessage.accountKeys,
+      handlerMessage,
+    });
+
+    const encodedMessage = message.encode();
```
chrismaree, line 165: having the `AcrossPlusMessageCoder` really helps with this. great job.

test/svm/SvmSpoke.Fill.AcrossPlus.ts
```diff
@@ -0,0 +1,355 @@
...
+    const initialRelayData = {
+      depositor: finalRecipient,
+      recipient: handlerSigner, // Handler PDA that can forward tokens as needed within the message call.
+      exclusiveRelayer: relayer.publicKey,
+      inputToken: mint, // This is lazy. it should be an encoded token from a separate domain most likely.
+      outputToken: mint,
+      inputAmount: new BN(relayAmount),
+      outputAmount: new BN(relayAmount),
+      originChainId: new BN(1),
+      depositId: new BN(Math.floor(Math.random() * 1000000)), // force that we always have a new deposit id.
```
chrismaree, line 131: note we are going to have merge conflicts & breaks between your PR and this https://github.com/across-protocol/contracts/pull/759/files due to `depositId` and how it's defined changing some.

test/svm/SvmSpoke.Bundle.ts
```diff
@@ -1092,42 +1091,6 @@ describe("svm_spoke.bundle", () => {
           ])
         : [];
 
-      // Consolidate all above addresses into a single array for the  Address Lookup Table (ALT).
```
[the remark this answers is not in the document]
Reinis-FRP: now we have `maxExtendedAccounts` inside the `sendTransactionWithLookupTable` helper, as that is not instruction specific

chrismaree (approved):

PR #748, opened 2024-11-13
improve: use immutable `SHARED_BRIDGE` whenever possible

bmzig:
In the `ZkStack_*_Adapter`, we assume that the bridge hub's shared bridge is immutable. While this is not true -- the shared bridge may be changed in the future -- it is assumed to be immutable so that L1 calls will be cheaper, as we stand to save ~2 `staticcall`s per message. The hope is in the long run, if these bridges change, the cost of redeploying an adapter will have already been "paid off" by the gas savings of assuming immutability.


nicholaspai (approved):

nicholaspai:
Is this an audit related issue? Do we need to change the PR title?

bmzig:
> Is this an audit related issue? Do we need to change the PR title?

It isn't addressing the issue outlined in `L-03`, but this was a suggestion given to us based on that issue, so I thought marking it as part of `L-03` would be misleading.

PR #751, opened 2024-11-18
feat: handle EIP-2612 and EIP-3009 in `SwapAndBridge.sol`

bmzig:
# Summary
We need to have a way to deposit into the spoke pool **and** perform a `swapAndBridge` using [EIP-2612](https://eips.ethereum.org/EIPS/eip-2612) and [EIP-3009](https://eips.ethereum.org/EIPS/eip-3009). 

This functionality is added with four new functions in the `UniversalSwapAndBridge` contract: `swapAndBridgeWithPermit`, `depositWithPermit`, `swapAndBridgeWithAuthorization`, and `depositWithAuthorization`, where `*WithPermit` indicates usage of an EIP-2612 compliant token, and `*WithAuthorization` indicates usage of an EIP-3009 compliant token.

# Details
EIP-2616 is used solely to give an approval to an address without calling `approve` first. This allows us to still use `transferFrom` when using any EIP-2612 token. 
On the other hand, EIP-3009 will simply transfer tokens to the designated receiver contract and not raise the contract's allowance, so all usage of these types of functions check whether a sufficient token amount was transferred to the contract before proceeding with the rest of the swap/deposit.

### DIFF
--- a/contracts/SwapAndBridge.sol
+++ b/contracts/SwapAndBridge.sol
@@ -166,13 +166,27 @@ abstract contract SwapAndBridgeBase is Lockable, MultiCaller {
             depositData.outputAmount
         );
         // Deposit the swapped tokens into Across and bridge them using remainder of input params.
-        _acrossInputToken.safeIncreaseAllowance(address(SPOKE_POOL), returnAmount);
+        _depositV3(_acrossInputToken, returnAmount, depositData);
+    }
+
+    /**
+     * @notice Approves the spoke pool and calls `depositV3` function with the specified input parameters.
+     * @param _acrossInputToken Token to deposit into the spoke pool.
+     * @param _acrossInputAmount Amount of the input token to deposit into the spoke pool.
+     * @param depositData Specifies the Across deposit params to use.
+     */
+    function _depositV3(
+        IERC20 _acrossInputToken,
+        uint256 _acrossInputAmount,
+        DepositData calldata depositData
+    ) internal {
+        _acrossInputToken.safeIncreaseAllowance(address(SPOKE_POOL), _acrossInputAmount);
         SPOKE_POOL.depositV3(
             depositData.depositor,
             depositData.recipient,
             address(_acrossInputToken), // input token
             depositData.outputToken, // output token
-            returnAmount, // input amount.
+            _acrossInputAmount, // input amount.
             depositData.outputAmount, // output amount
             depositData.destinationChainid,
             depositData.exclusiveRelayer,
@@ -255,6 +269,9 @@ contract SwapAndBridge is SwapAndBridgeBase {
  * bridging the received token via Across atomically. Provides safety checks post-swap and before-deposit.
  */
 contract UniversalSwapAndBridge is SwapAndBridgeBase {
+    using SafeERC20 for IERC20;
+    error ReceiveWithAuthorizationFailed();
+
     /**
      * @notice Construct a new SwapAndBridgeBase contract.
      * @param _spokePool Address of the SpokePool contract that we'll submit deposits to.
@@ -311,6 +328,10 @@ contract UniversalSwapAndBridge is SwapAndBridgeBase {
      * @param minExpectedInputTokenAmount Minimum amount of received depositData.inputToken that we'll submit bridge
      * deposit with.
      * @param depositData Specifies the Across deposit params we'll send after the swap.
+     * @param deadline Deadline before which the permit signature is valid.
+     * @param v v of the permit signature.
+     * @param r r of the permit signature.
+     * @param s s of the permit signature.
      */
     function swapAndBridgeWithPermit(
         IERC20Permit swapToken,
@@ -347,6 +368,12 @@ contract UniversalSwapAndBridge is SwapAndBridgeBase {
      * @param minExpectedInputTokenAmount Minimum amount of received depositData.inputToken that we'll submit bridge
      * deposit with.
      * @param depositData Specifies the Across deposit params we'll send after the swap.
+     * @param validAfter The unix time after which the `receiveWithAuthorization` signature is valid.
+     * @param validBefore The unix time before which the `receiveWithAuthorization` signature is valid.
+     * @param nonce Unique nonce used in the `receiveWithAuthorization` signature.
+     * @param v v of the EIP-3009 signature.
+     * @param r r of the EIP-3009 signature.
+     * @param s s of the EIP-3009 signature.
      */
     function swapAndBridgeWithAuthorization(
         IERC20Auth swapToken,
@@ -383,5 +410,72 @@ contract UniversalSwapAndBridge is SwapAndBridgeBase {
             IERC20(address(swapToken)), // Cast IERC20Auth to IERC20
             acrossInputToken
         );
+    }
+
+    /**
+     * @notice Deposits an EIP-2612 token Across input token into the Spoke Pool contract.
+     * @dev If `acrossInputToken` does not implement `permit` to the specifications of EIP-2612, this function will fail.
+     * @param acrossInputToken EIP-2612 compliant token to deposit.
+     * @param acrossInputAmount Amount of the input token to deposit.
+     * @param depositData Specifies the Across deposit params to send.
+     * @param deadline Deadline before which the permit signature is valid.
+     * @param v v of the permit signature.
+     * @param r r of the permit signature.
+     * @param s s of the permit signature.
+     */
+    function depositWithPermit(
+        IERC20Permit acrossInputToken,
+        uint256 acrossInputAmount,
+        DepositData calldata depositData,
+        uint256 deadline,
+        uint8 v,
+        bytes32 r,
+        bytes32 s
+    ) external nonReentrant {
+        acrossInputToken.permit(msg.sender, address(this), acrossInputAmount, deadline, v, r, s);
+        IERC20 _acrossInputToken = IERC20(address(acrossInputToken)); // Cast IERC20Permit to an IERC20 type.
+
+        _acrossInputToken.safeTransferFrom(msg.sender, address(this), acrossInputAmount);
+        _depositV3(_acrossInputToken, acrossInputAmount, depositData);
+    }
+
+    /**
+     * @notice Deposits an EIP-3009 compliant Across input token into the Spoke Pool contract.
+     * @dev If `acrossInputToken` does not implement `receiveWithAuthorization` to the specifications of EIP-3009, this call will revert.
+     * @param acrossInputToken EIP-3009 compliant token to deposit.
+     * @param acrossInputAmount Amount of the input token to deposit.
+     * @param depositData Specifies the Across deposit params to send.
+     * @param validAfter The unix time after which the `receiveWithAuthorization` signature is valid.
+     * @param validBefore The unix time before which the `receiveWithAuthorization` signature is valid.
+     * @param nonce Unique nonce used in the `receiveWithAuthorization` signature.
+     * @param v v of the EIP-3009 signature.
+     * @param r r of the EIP-3009 signature.
+     * @param s s of the EIP-3009 signature.
+     */
+    function depositWithAuthorization(
+        IERC20Auth acrossInputToken,
+        uint256 acrossInputAmount,
+        DepositData calldata depositData,
+        uint256 validAfter,
+        uint256 validBefore,
+        bytes32 nonce,
+        uint8 v,
+        bytes32 r,
+        bytes32 s
+    ) external nonReentrant {
+        acrossInputToken.receiveWithAuthorization(
+            msg.sender,
+            address(this),
+            acrossInputAmount,
+            validAfter,
+            validBefore,
+            nonce,
+            v,
+            r,
+            s
+        );
+        IERC20 _acrossInputToken = IERC20(address(acrossInputToken)); // Cast the input token to an IERC20.
+        if (_acrossInputToken.balanceOf(address(this)) < acrossInputAmount) revert ReceiveWithAuthorizationFailed();
+        _depositV3(_acrossInputToken, acrossInputAmount, depositData);
     }
 }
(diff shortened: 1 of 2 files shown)


contracts/SwapAndBridge.sol
```diff
@@ -291,4 +300,78 @@ contract UniversalSwapAndBridge is SwapAndBridgeBase {
...
+            depositData,
+            IERC20(address(swapToken)), // Cast IERC20Permit to IERC20
+            acrossInputToken
+        );
+    }
+
+    /**
+     * @notice Swaps an EIP-3009 token on this chain via specified router before submitting Across deposit atomically.
+     * Caller can specify their slippage tolerance for the swap and Across deposit params.
+     // * @dev If swapToken does not implement `permit` to the specifications of EIP-2612, this function will fail.
```
bmzig, line 343: Need to edit this comment

contracts/SwapAndBridge.sol
```diff
@@ -103,8 +104,13 @@ abstract contract SwapAndBridgeBase is Lockable, MultiCaller {
         // user on this contract, a malicious actor could call transferFrom to steal the user's tokens.
         if (!allowedSelectors[bytes4(routerCalldata)]) revert InvalidFunctionSelector();
 
-        // Pull tokens from caller into this contract.
-        _swapToken.safeTransferFrom(msg.sender, address(this), swapTokenAmount);
+        // Pull tokens from caller into this contract. This contract may already have tokens if, for example,
+        // a user sent the contract tokens and subsequently called `swapAndBridge`, so only pull funds if
+        // the contract needs to.
```
nicholaspai, line 109: This is a nice feature but curious why you're adding it to this PR, should it be in a separate PR?

bmzig: I added it here since `receiveWithAuthorization` should send tokens to the contract. So ideally, whenever `swapAndBridgeWithAuthorization` is called, `transferFrom` should never be called, and for the normal and permit swap and bridge, `transferFrom` should be called.

contracts/SwapAndBridge.sol
bmzig: https://github.com/across-protocol/contracts/pull/751/commits/7bdf27a7ab2d78508e8e86ef519b5fdf63b99ff9

contracts/SwapAndBridge.sol
nicholaspai: I see. Part of me thinks we should separate the token transfer logic from the swapping logic. Perhaps we should add a separate internal function called `_pullTokens` or something which has different behavior for `transferFrom` versus `permit` versus `receiveWithAuthorization `. What do you think?

bmzig: Yes, I think this will simplify things. I added a `_pullTokens` function in this commit: https://github.com/across-protocol/contracts/pull/751/commits/5f7ee7ba84f4b759ba9525a7d49dfd3455b434dc. I think the good thing about doing this is now all we really care about is whether the contract has sufficient balance to make the deposit/swap, and the way it gets the tokens is irrelevant.

nicholaspai: exactly

contracts/SwapAndBridge.sol
```diff
@@ -103,8 +105,11 @@ abstract contract SwapAndBridgeBase is Lockable, MultiCaller {
         // user on this contract, a malicious actor could call transferFrom to steal the user's tokens.
         if (!allowedSelectors[bytes4(routerCalldata)]) revert InvalidFunctionSelector();
 
-        // Pull tokens from caller into this contract.
-        _swapToken.safeTransferFrom(msg.sender, address(this), swapTokenAmount);
+        // Pull tokens from caller into this contract. This contract may already have tokens if, for example,
+        // a user sent the contract tokens and subsequently called `swapAndBridge`, so only pull funds if
```
nicholaspai, line 109: I think this comment is a bit out of place now that we've added _pullTokens

contracts/SwapAndBridge.sol
nicholaspai: so actually what I was suggesting is something more along the lines of removing this pulling logic from this function and leaving it the caller of `_swapAndBridge` to pull tokens before entering the function. `swapAndBridgeWithPermit`, `swapAndBridgeWithAuthorization` and `swapAndBridge` each have their own ways of pulling the user's funds. They can perform any checks they want to after pulling, and then they can call `_swapAndBridge`. What do you think of that? Am open to other ideas

contracts/SwapAndBridge.sol
bmzig: Removed https://github.com/across-protocol/contracts/pull/751/commits/7d67ab56d693289fd6d71933b4a5084503139a07

contracts/SwapAndBridge.sol
bmzig: Ah I see. So basically, leave this part to function in case we were to add more ways of depositing/authenticating. Makes sense. I'm just jumping around these different options. Under the impression that this contract will be modified more times in the future, I think it probably would be better this way. https://github.com/across-protocol/contracts/pull/751/commits/7d67ab56d693289fd6d71933b4a5084503139a07

contracts/SwapAndBridge.sol
```diff
@@ -291,4 +310,178 @@ contract UniversalSwapAndBridge is SwapAndBridgeBase {
...
+        uint256 validAfter,
+        uint256 validBefore,
+        bytes32 nonce,
+        uint8 v,
+        bytes32 r,
+        bytes32 s
+    ) external nonReentrant {
+        // While any contract can vacuously implement `transferWithAuthorization` (or just have a fallback),
+        // If tokens were not sent to this contract, the call to `transferFrom` in _swapAndBridge will revert.
+        swapToken.receiveWithAuthorization(
```
nicholaspai, line 392: Why can't you have `swapAndBridgeWithAuthorization`, `swapAndBridgeWithPermit`, and `swapAndBridge` all call:

```solidity
_pullTokens()
_swapAndBridge()
```
Where each of them defines `_pullTokens` differently..?

nicholaspai (commented):
I don't think my message https://github.com/across-protocol/contracts/pull/751#discussion_r1847357534 came across clearly

contracts/SwapAndBridge.sol
```diff
@@ -291,4 +310,178 @@ contract UniversalSwapAndBridge is SwapAndBridgeBase {
...
+            validBefore,
+            nonce,
+            v,
+            r,
+            s
+        );
+        IERC20 _swapToken = IERC20(address(swapToken)); // Cast IERC20Auth to IERC20
+
+        uint256 tokenBalance = _swapToken.balanceOf(address(this));
+        if (tokenBalance < swapTokenAmount)
```
nicholaspai, line 406: Do you think we need this check? Won't `_swapAndBridge` fail if we have insufficient balance? Or do you think we want to just catch it earlier?

I guess i'm thinking, what's the difference between checking the balance here after calling `receiveWithAuthorization()` but not checking the balance after calling `safeTransferFrom` in the other two methods? Is it because in the latter we are more familiar with that function and knowing it **should just work**, whereas `receiveWithAuthorization` is a newer function that we're not used to?

bmzig: Yeah... That's a good point. I added this since I have the expectation that `receiveWithAuthorization` is just an uncommon function in most ERC20s, but whether a token implements this function can probably just be checked off-chain, so the check is just unnecessary.

contracts/SwapAndBridge.sol
```diff
@@ -291,4 +327,171 @@ contract UniversalSwapAndBridge is SwapAndBridgeBase {
...
+        uint256 minExpectedInputTokenAmount,
+        DepositData calldata depositData,
+        uint256 validAfter,
+        uint256 validBefore,
+        bytes32 nonce,
+        uint8 v,
+        bytes32 r,
+        bytes32 s
+    ) external nonReentrant {
+        // While any contract can vacuously implement `transferWithAuthorization` (or just have a fallback),
```
nicholaspai, line 407: is this comment line still relevant?

nicholaspai (approved):
Just one more Q on a comment!

contracts/SwapAndBridge.sol
bmzig: It's still technically true. Maybe it's self explanatory, though. It's just to highlight that calling this function with a token that doesn't implement EIP-3009 will (probably) revert.

contracts/SwapAndBridge.sol
```diff
@@ -291,4 +327,172 @@ contract UniversalSwapAndBridge is SwapAndBridgeBase {
...
+        bytes calldata routerCalldata,
+        uint256 swapTokenAmount,
+        uint256 minExpectedInputTokenAmount,
+        DepositData calldata depositData,
+        uint256 deadline,
+        uint8 v,
+        bytes32 r,
+        bytes32 s
+    ) external nonReentrant {
+        swapToken.permit(msg.sender, address(this), swapTokenAmount, deadline, v, r, s);
```
mrice32, line 360: What happens if the permit is frontrun?

bmzig: I think this call to permit will revert... So we should probably have a check beforehand which only calls permit if `allowance[msg.sender][address(this)]` is not enough to cover. Does that sound right?

contracts/SwapAndBridge.sol
```diff
@@ -291,4 +328,174 @@ contract UniversalSwapAndBridge is SwapAndBridgeBase {
...
+        uint256 swapTokenAmount,
+        uint256 minExpectedInputTokenAmount,
+        DepositData calldata depositData,
+        uint256 deadline,
+        uint8 v,
+        bytes32 r,
+        bytes32 s
+    ) external nonReentrant {
+        IERC20 _swapToken = IERC20(address(swapToken)); // Cast IERC20Permit to IERC20.
+        if (_swapToken.allowance(msg.sender, address(this)) < swapTokenAmount)
```
nicholaspai, line 362: Why should we be checking for front-running with permit but not with transferFrom? I'd argue we'd want the same allowance() check in both swapAndBridge AND swapAndBridgeWithPermit.

But I favor having no checks at all and just letting the subsequent safeTransferFrom revert

bmzig: I thought the front-running was with respect to somebody calling `permit` on the token before this transaction is executed, so the approval would be there, but the call to permit would revert since the signature input in this function would have already been used.

nicholaspai: is this something that would happen in the normal course of transactions?

bmzig: It's only possible if people want to grief users. Updated here: https://github.com/across-protocol/contracts/pull/751/commits/ab96160a7580ed6a0bfb2aabe08b9b14ef16ded5

contracts/SwapAndBridge.sol
```diff
@@ -291,4 +328,176 @@ contract UniversalSwapAndBridge is SwapAndBridgeBase {
...
+        uint256 swapTokenAmount,
+        uint256 minExpectedInputTokenAmount,
+        DepositData calldata depositData,
+        uint256 deadline,
+        uint8 v,
+        bytes32 r,
+        bytes32 s
+    ) external nonReentrant {
+        IERC20 _swapToken = IERC20(address(swapToken)); // Cast IERC20Permit to IERC20.
+        // For permit transactions, we wrap the call in a try/catch block so that the transaction will continue even if the call to
```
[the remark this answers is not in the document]
bmzig: https://github.com/across-protocol/contracts/pull/751/commits/7bc3b60d0b1b2bf475c41ddd5605597643687204

contracts/SwapAndBridge.sol
```diff
@@ -79,11 +86,13 @@ abstract contract SwapAndBridgeBase is Lockable, MultiCaller {
      */
     constructor(
         V3SpokePoolInterface _spokePool,
+        WETH9Interface _wrappedNativeToken,
```
pxrl, line 89: @bmzig This is also able to be queried directly from the SpokePool contract. Thoughts on doing that within the constructor instead? It'd remove a constructor argument and would also eliminate the possibility for an (admittedly unlikely) misalignment on the address.

bmzig: I thought of doing that but `wrappedNativeTokens` isn't exposed by the V3SpokePoolInterface/SpokePoolInterface, so we would need to modify the interface for this to work (or just create a new interface). I figured we probably shouldn't do that.

contracts/SwapAndBridge.sol
```diff
@@ -291,4 +328,178 @@ contract UniversalSwapAndBridge is SwapAndBridgeBase {
             acrossInputToken
         );
     }
+
+    /**
+     * @notice Swaps an EIP-2612 token on this chain via specified router before submitting Across deposit atomically.
+     * Caller can specify their slippage tolerance for the swap and Across deposit params.
+     * @dev If swapToken does not implement `permit` to the specifications of EIP-2612, this function will fail.
```
pxrl, line 335: Is this comment still correct in light of the try/catch wrapper?

bmzig: It is. This comment is a bit more general, and just saying that you can't expect permit to work for a token which doesn't implement permit.

contracts/SwapAndBridge.sol
```diff
@@ -291,4 +328,178 @@ contract UniversalSwapAndBridge is SwapAndBridgeBase {
...
+            minExpectedInputTokenAmount,
+            depositData,
+            _swapToken,
+            acrossInputToken
+        );
+    }
+
+    /**
+     * @notice Deposits an EIP-2612 token Across input token into the Spoke Pool contract.
+     * @dev If `acrossInputToken` does not implement `permit` to the specifications of EIP-2612, this function will fail.
```
pxrl, line 439: Same q. RE the try/catch wrapper - is this comment still correct?

bmzig: You can think of this as: if a user uses a token which doesn't implement permit, then all else equal, this function will fail on `safeTransferFrom`, since `permit` does nothing to increase the approval. This is different from a token which does implement permit and has had their signature executed before the call to this function, since at least in that case, the allowance is expected to have changed.

PR #752, opened 2024-11-19
fix: delegated transfers in deposits

Reinis-FRP:
Clients might have enabled [CPI-guards](https://spl.solana.com/token-2022/extensions#cpi-guard) that would prohibit transfers within CPI (e.g. deposits, fills). To support this, we need to transfer all user tokens with spoke pool authority, assuming the client has delegated exact deposit/fill amount.

Fixes: https://linear.app/uma/issue/ACX-3351/use-delegated-transfers-in-svm-spoke

### DIFF
--- a/programs/svm-spoke/src/utils/transfer_utils.rs
+++ b/programs/svm-spoke/src/utils/transfer_utils.rs
@@ -1,0 +1,29 @@
+use anchor_lang::prelude::*;
+use anchor_spl::token_interface::{transfer_checked, Mint, TokenAccount, TokenInterface, TransferChecked};
+
+use crate::State;
+
+pub fn transfer_from<'info>(
+    from: &InterfaceAccount<'info, TokenAccount>,
+    to: &InterfaceAccount<'info, TokenAccount>,
+    amount: u64,
+    state: &Account<'info, State>,
+    state_bump: u8,
+    mint: &InterfaceAccount<'info, Mint>,
+    token_program: &Interface<'info, TokenInterface>,
+) -> Result<()> {
+    let transfer_accounts = TransferChecked {
+        from: from.to_account_info(),
+        mint: mint.to_account_info(),
+        to: to.to_account_info(),
+        authority: state.to_account_info(),
+    };
+
+    let state_seed_bytes = state.seed.to_le_bytes();
+    let seeds = &[b"state", state_seed_bytes.as_ref(), &[state_bump]];
+    let signer_seeds = &[&seeds[..]];
+
+    let cpi_context = CpiContext::new_with_signer(token_program.to_account_info(), transfer_accounts, signer_seeds);
+
+    transfer_checked(cpi_context, amount, mint.decimals)
+}
--- a/scripts/svm/simpleDeposit.ts
+++ b/scripts/svm/simpleDeposit.ts
@@ -2,8 +2,14 @@
 
 import * as anchor from "@coral-xyz/anchor";
 import { BN, Program, AnchorProvider } from "@coral-xyz/anchor";
-import { PublicKey } from "@solana/web3.js";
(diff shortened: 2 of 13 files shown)


test/svm/SvmSpoke.Deposit.ts
```diff
@@ -61,17 +92,40 @@ describe("svm_spoke.deposit", () => {
...
 
-  before("Creates token mint and associated token accounts", async () => {
-    await setupInputToken();
-  });
+  const approvedDepositV3 = async (
+    depositDataValues: DepositDataValues,
+    calledDepositAccounts: DepositAccounts = depositAccounts
+  ) => {
+    // Delegate state PDA to pull depositor tokens.
+    const approveIx = await createApproveCheckedInstruction(
```
md0x, line 105: Do we have a test where we catch the error when trying to do this without approving? It might be interesting to document that.

md0x (approved):
Looks good!

programs/svm-spoke/src/utils/transfer_utils.rs
```diff
@@ -0,0 +1,29 @@
...
+    state: &Account<'info, State>,
+    state_bump: u8,
+    mint: &InterfaceAccount<'info, Mint>,
+    token_program: &Interface<'info, TokenInterface>,
+) -> Result<()> {
+    let transfer_accounts = TransferChecked {
+        from: from.to_account_info(),
+        mint: mint.to_account_info(),
+        to: to.to_account_info(),
+        authority: state.to_account_info(),
```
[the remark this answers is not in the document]
Reinis-FRP: yes, as the caller must have approved first

test/svm/SvmSpoke.Deposit.ts
Reinis-FRP: sure, I can add that

programs/svm-spoke/src/utils/transfer_utils.rs
```diff
@@ -0,0 +1,29 @@
+use anchor_lang::prelude::*;
+use anchor_spl::token_interface::{transfer_checked, Mint, TokenAccount, TokenInterface, TransferChecked};
+
+use crate::State;
+
+pub fn transfer_from<'info>(
```
chrismaree, line 6: this is great. it's similar to how we were doing transfers previously, but now wrapped in the `TransferChecked` block.

scripts/svm/simpleDeposit.ts
```diff
@@ -83,9 +89,25 @@ async function depositV3(): Promise<void> {
     { property: "vault", value: vault.toString() },
   ]);
 
-  const tx = await (
+  const userTokenAccount = getAssociatedTokenAddressSync(inputToken, signer.publicKey);
+
+  const tokenDecimals = (await getMint(provider.connection, inputToken, undefined, TOKEN_PROGRAM_ID)).decimals;
+
+  // Delegate state PDA to pull depositor tokens.
+  const approveIx = await createApproveCheckedInstruction(
```
chrismaree, line 97: cool. this is not too bad all things considered. did you test this on testnet with both token2022 and legacy token types?

chrismaree (approved):
This is great, given it's backwards/forwards compatible between token2022 and previous versions.

scripts/svm/simpleDeposit.ts
Reinis-FRP: yes, here is old token [deposit](https://explorer.solana.com/tx/3x95k3WxgaqPWTQfUGaLzQr1udCat54oYtztUwJ31DqdG8TAyDq83dmXwNnqF1dzw7KtiPRPSDBkmdmsE2qohZHF?cluster=devnet)
and here is [2022 token deposit](https://explorer.solana.com/tx/58hDuHgDq735iv2KJoAj9Y5KaYUWzW1iaKZKZGTVin7eXWQNneu6WCEYDxozvSMRwHJ3yR9HTbkk1PZLNK7qau98?cluster=devnet) with enabled CPI guards

PR #753, opened 2024-11-19
proposal: refine how we define PDA paths for some account types

chrismaree:
this PR refines how we define PDAs in some places to make them more consistently dependent on the seed, rather than state


programs/svm-spoke/src/event.rs
```diff
@@ -121,7 +121,6 @@ pub struct ClaimedRelayerRefund {
     pub refund_address: Pubkey,
 }
 
-// TODO: update the name of this to EmergencyDeletedRootBundle and in EVM.
```
Reinis-FRP, line 124: we have not fixed this in EVM

programs/svm-spoke/src/instructions/handle_receive_message.rs
```diff
@@ -11,9 +11,6 @@ use crate::{
     State,
 };
 
-//TODO: we have inconsistent imports in this file, in some places referencing from source crates (SvmSpoke::id)
```
Reinis-FRP, line 14: we have not addressed this

programs/svm-spoke/src/state/state.rs
```diff
@@ -12,7 +12,7 @@ pub struct State {
     pub current_time: u32,              // Only used in testable mode, else set to 0 on mainnet.
     pub remote_domain: u32,             // CCTP domain for Mainnet Ethereum.
     pub cross_domain_admin: Pubkey,     // HubPool on Mainnet Ethereum.
-    pub root_bundle_id: u32,            // TODO rename to next_root_bundle_id
+    pub root_bundle_id: u32,            // Tracks the current root bundle id.
```
Reinis-FRP, line 15: but it is not the current root bundle id. It is next one, thus we had comment to rename this prop

Reinis-FRP (approved):
Using state.seed seems more consistent. though there are some comments we delete that are not yet addressed.

programs/svm-spoke/src/state/state.rs
chrismaree: I chose to just update the comment accordingly.

programs/svm-spoke/src/instructions/handle_receive_message.rs
chrismaree: right but the thing it was referencing for some reason could not be refactored like other imports.

programs/svm-spoke/src/event.rs
chrismaree: done here: https://github.com/across-protocol/contracts/pull/760

PR #754, opened 2024-11-20
fix: prevent any situation where SVM leaves could ever be used in EVM and viceversa

chrismaree:
contrived situations can be created where EVM and SVM leaves can be re-used between domains given how encoding works on these chains and the difference with associated types. this pr enforces that under all situations this is impossible.

### DIFF
--- a/programs/svm-spoke/src/instructions/bundle.rs
+++ b/programs/svm-spoke/src/instructions/bundle.rs
@@ -81,6 +81,11 @@ impl RelayerRefundLeaf {
     pub fn to_bytes(&self) -> Vec<u8> {
         let mut bytes = Vec::new();
 
+        // This requires the first 64 bytes to be 0 within the encoded leaf data. This protects any kind of EVM leaf
+        // from ever being used on SVM. Note that the chain_id field in theory should protect this but this 64 blank
+        // slot protects it under all cases.
+        bytes.extend_from_slice(&[0u8; 64]);
+
         bytes.extend_from_slice(&self.amount_to_return.to_le_bytes());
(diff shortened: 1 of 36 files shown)


programs/svm-spoke/src/instructions/bundle.rs
```diff
@@ -81,6 +81,11 @@ impl RelayerRefundLeaf {
     pub fn to_bytes(&self) -> Vec<u8> {
         let mut bytes = Vec::new();
 
+        // This requires the first 64 bytes to be 0 within the encoded leaf data. This protects any kind of EVM leaf
```
Reinis-FRP, line 84: Maybe we can expand a bit here where 64 bytes come from, like 2 uint256 fields for EVM amountToReturn and chainId

programs/svm-spoke/src/instructions/bundle.rs
```diff
@@ -81,6 +81,11 @@ impl RelayerRefundLeaf {
...
+        // from ever being used on SVM. Note that the chain_id field in theory should protect this but this 64 blank
+        // slot protects it under all cases.
+        bytes.extend_from_slice(&[0u8; 64]);
+
         bytes.extend_from_slice(&self.amount_to_return.to_le_bytes());
```
Reinis-FRP, line 89: Not directly related to this PR, but only now I realize we might been serializing this wrong below. Particularly, nowhere do we encode the lengths of the vectors, we just concatenate them. I think instead of implementing our custom serialization we should have used Anchor's `serialize` that uses Borch encoding where before each vector there is also a 4 byte length inserted.

Reinis-FRP (approved):
we can also have Borch serialization fix in a separate PR

PR #758, opened 2024-11-21
feat: consolidate spoke pool utility functions into a periphery contract

bmzig:
This consolidates our `SpokePoolVerifier` and `SwapAndBridge*` contracts into a single `SpokePoolV3Periphery` contract, which can be used to construct more expressive deposit types, like gasless deposits by using extra, ephemeral input data.

### DIFF
--- a/contracts/SpokePoolV3Periphery.sol
+++ b/contracts/SpokePoolV3Periphery.sol
@@ -17,7 +17,7 @@ import { WETH9Interface } from "./external/interfaces/WETH9Interface.sol";
  * @dev Variables which may be immutable are not marked as immutable, nor defined in the constructor, so that this contract may be deployed deterministically.
  * @custom:security-contact bugs@across.to
  */
-abstract contract SpokePoolV3Periphery is Lockable, MultiCaller {
+contract SpokePoolV3Periphery is Lockable, MultiCaller {
     using SafeERC20 for IERC20;
     using Address for address;
 
--- a/contracts/SpokePoolV3Periphery.sol
+++ b/contracts/SpokePoolV3Periphery.sol
@@ -1,23 +1,25 @@
 //SPDX-License-Identifier: Unlicense
 pragma solidity ^0.8.0;
 
-import "./interfaces/V3SpokePoolInterface.sol";
-import "./external/interfaces/IERC20Auth.sol";
-import "./external/interfaces/WETH9Interface.sol";
-import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
-import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
-import "@openzeppelin/contracts/token/ERC20/extensions/IERC20Permit.sol";
-import "./Lockable.sol";
-import "@uma/core/contracts/common/implementation/MultiCaller.sol";
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import { IERC20Permit } from "@openzeppelin/contracts/token/ERC20/extensions/IERC20Permit.sol";
+import { Address } from "@openzeppelin/contracts/utils/Address.sol";
+import { MultiCaller } from "@uma/core/contracts/common/implementation/MultiCaller.sol";
+import { Lockable } from "./Lockable.sol";
+import { V3SpokePoolInterface } from "./interfaces/V3SpokePoolInterface.sol";
+import { IERC20Auth } from "./external/interfaces/IERC20Auth.sol";
+import { WETH9Interface } from "./external/interfaces/WETH9Interface.sol";
 
 /**
- * @title SwapAndBridgeBase
- * @notice Base contract for both variants of SwapAndBridge.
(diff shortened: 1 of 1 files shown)


contracts/SpokePoolV3Periphery.sol
```diff
@@ -77,216 +82,99 @@ abstract contract SwapAndBridgeBase is Lockable, MultiCaller {
...
-        spokePool = _spokePool;
-        exchange = _exchange;
-        wrappedNativeToken = _wrappedNativeToken;
+    constructor(bytes4[] memory _allowedSelectors) {
         for (uint256 i = 0; i < _allowedSelectors.length; i++) {
             allowedSelectors[_allowedSelectors[i]] = true;
         }
     }
 
-    // This contract supports two variants of swap and bridge, one that allows one token and another that allows the caller to pass them in.
```
bmzig, line 101: The code to these functions were not changed, (although since it is a single contract, their visibility was switched to `private` for gas savings), but they were moved to the bottom of the contract to follow the Solidity style guide.

contracts/SpokePoolV3Periphery.sol
```diff
@@ -77,216 +82,99 @@ abstract contract SwapAndBridgeBase is Lockable, MultiCaller {
...
-     * @param _acrossInputToken Address of the token that will be bridged via Across as the inputToken.
+     * @dev These values are initialized in a function and not in the constructor so that the creation code of this contract
+     * is the same across networks with different addresses for the wrapped native token, the exchange this contract uses to
+     * swap and bridge, and this network's corresponding spoke pool contract. This is to allow this contract to be deterministically
+     * deployed with CREATE2.
+     * @dev This function can be front-run by anybody, so it is critical to check that the `spokePool`, `wrappedNativeToken`, and `exchange`
+     * values used in the single call to this function were passed in correctly before enabling the usage of this contract.
      */
-    constructor(
+    function initialize(
```
bmzig, line 112: This is new code

contracts/SpokePoolV3Periphery.sol
```diff
@@ -77,216 +82,99 @@ abstract contract SwapAndBridgeBase is Lockable, MultiCaller {
...
-        DepositData calldata depositData
-    ) external nonReentrant {
-        _swapAndBridge(
-            routerCalldata,
-            swapTokenAmount,
-            minExpectedInputTokenAmount,
-            depositData,
-            SWAP_TOKEN,
-            ACROSS_INPUT_TOKEN
+    function deposit(
```
bmzig, line 146: This is taken from the spoke pool verifier. The added changes were to remove the `spokePool` argument, since it should be defined in this contract on initialization, and to add a `nonReentrant` modifier to the function.

contracts/SpokePoolV3Periphery.sol
```diff
@@ -77,216 +82,99 @@ abstract contract SwapAndBridgeBase is Lockable, MultiCaller {
...
+        address recipient,
+        address inputToken,
+        uint256 inputAmount,
+        uint256 outputAmount,
+        uint256 destinationChainId,
+        address exclusiveRelayer,
+        uint32 quoteTimestamp,
+        uint32 fillDeadline,
+        uint32 exclusivityDeadline,
```
nicholaspai, line 155: Can we rename this to exclusivityParameter now?

contracts/SpokePoolV3Periphery.sol
```diff
@@ -502,4 +390,103 @@ contract UniversalSwapAndBridge is SwapAndBridgeBase {
...
         _depositV3(_acrossInputToken, acrossInputAmount, depositData);
     }
+
+    /**
+     * @notice Approves the spoke pool and calls `depositV3` function with the specified input parameters.
+     * @param _acrossInputToken Token to deposit into the spoke pool.
+     * @param _acrossInputAmount Amount of the input token to deposit into the spoke pool.
+     * @param depositData Specifies the Across deposit params to use.
+     */
+    function _depositV3(
```
nicholaspai, line 400: These functions were all copied right?

bmzig: Yes exactly.

contracts/SpokePoolV3Periphery.sol
bmzig: Good call. All instances and comments now refer to it as the exclusivityParameter: https://github.com/across-protocol/contracts/pull/758/commits/daff3b73ac3394924bb7a973c6d7997409cef297

contracts/SpokePoolV3Periphery.sol
```diff
@@ -48,15 +53,16 @@ abstract contract SwapAndBridgeBase is Lockable, MultiCaller {
         address recipient;
         // The destination chain identifier.
         uint256 destinationChainid;
-        // The account that can exclusively fill the deposit before the exclusivity deadline.
+        // The account that can exclusively fill the deposit before the  deadline.
```
bmzig, line 56:
```suggestion
        // The account that can exclusively fill the deposit before the exclusivity parameter.
```

pxrl (commented):
@bmzig I'm OK with this new functionality but I think that updating `SwapAndBridge` will potentially cause problems because this contract should be deployed _beside_ an existing `SwapAndBridge` contract. wdyt about moving it all to a new file, rather than updating the existing one? We'd later deprecate `SwapAndBridge` once we have the new implementation fully audited.

PR #759, opened 2024-11-22
feat: add determinstic deposit hash & new exclusivity logic

chrismaree:
This PR adds the logic defined in this PR: https://github.com/across-protocol/contracts/pull/756 to the solana implementation.

there is one gotcha and opinionated change that was needed to make this work: deposit_id is now a `[u8;32]` to accommodate this change. practically, this is not really a big deal for relayers or data worker but should be noted.


programs/svm-spoke/src/instructions/deposit.rs
```diff
@@ -122,6 +142,41 @@ pub fn deposit_v3(
...
+        ctx,
+        depositor,
+        recipient,
+        input_token,
+        output_token,
+        input_amount,
+        output_amount,
+        destination_chain_id,
+        exclusive_relayer,
+        [0u8; 32], // deposit_id of informs internal function to use state.number_of_deposits as id.
```
md0x, line 170: Nit: store this in constants as it's reused multiple times

programs/svm-spoke/src/instructions/deposit.rs
```diff
@@ -156,3 +211,53 @@ pub fn deposit_v3_now(
...
+        message,
+    )?;
+
+    Ok(())
+}
+
+// Define a dummy context struct so we can export this as a view function in lib.
+#[derive(Accounts)]
+pub struct Null {}
+pub fn get_unsafe_deposit_id(msg_sender: Pubkey, depositor: Pubkey, deposit_nonce: u64) -> [u8; 32] {
```
md0x, line 256: Do you think it makes sense to expose this? Are we planning to use it off-chain, or should we implement a similar function in TS for efficiency?

programs/svm-spoke/src/instructions/deposit.rs
```diff
@@ -101,18 +114,25 @@ pub fn deposit_v3(
         &ctx.accounts.token_program,
     )?;
 
-    state.number_of_deposits += 1; // Increment number of deposits
+    let mut applied_deposit_id = deposit_id;
+    // If the passed in deposit_id is all zeros, then we use the state's number of deposits as deposit_id.
+    if deposit_id.iter().all(|&x| x == 0) {
```
Reinis-FRP, line 119:
```suggestion
    if deposit_id == [0u8; 32] {
```

programs/svm-spoke/src/instructions/deposit.rs
```diff
@@ -156,3 +211,53 @@ pub fn deposit_v3_now(
...
+// Define a dummy context struct so we can export this as a view function in lib.
+#[derive(Accounts)]
+pub struct Null {}
+pub fn get_unsafe_deposit_id(msg_sender: Pubkey, depositor: Pubkey, deposit_nonce: u64) -> [u8; 32] {
+    let mut data = Vec::new();
+    data.extend_from_slice(&msg_sender.to_bytes());
+    data.extend_from_slice(&depositor.to_bytes());
+    data.extend_from_slice(&deposit_nonce.to_le_bytes());
+
+    keccak::hash(&data).0
```
Reinis-FRP, line 262:
```suggestion
    keccak::hash(&data).to_bytes()
```
Better not access internal elements

programs/svm-spoke/src/instructions/deposit.rs
```diff
@@ -156,3 +211,53 @@ pub fn deposit_v3_now(
...
+}
+
+// Define a dummy context struct so we can export this as a view function in lib.
+#[derive(Accounts)]
+pub struct Null {}
+pub fn get_unsafe_deposit_id(msg_sender: Pubkey, depositor: Pubkey, deposit_nonce: u64) -> [u8; 32] {
+    let mut data = Vec::new();
+    data.extend_from_slice(&msg_sender.to_bytes());
+    data.extend_from_slice(&depositor.to_bytes());
+    data.extend_from_slice(&deposit_nonce.to_le_bytes());
```
Reinis-FRP, line 260: we might better reuse anchor serializer (should have the same result):
```
AnchorSerialize::serialize(&(msg_sender, depositor, deposit_nonce), &mut data)
```
this though requires converting this function to return Result

Reinis-FRP: note, I ended up using AnchorSerialize in [this PR](https://github.com/across-protocol/contracts/pull/765)

programs/svm-spoke/src/instructions/deposit.rs
chrismaree: yes, I think having it exposed is useful. it's done in solidity and it's good to not force users to have to do this in js. for example say you wanted to do the deposit with etherscan for example. i don't think there is any downside in having it public. do you agree?

programs/svm-spoke/src/instructions/deposit.rs
```diff
@@ -101,18 +114,25 @@ pub fn deposit_v3(
...
 
-    state.number_of_deposits += 1; // Increment number of deposits
+    let mut applied_deposit_id = deposit_id;
+    // If the passed in deposit_id is all zeros, then we use the state's number of deposits as deposit_id.
+    if deposit_id.iter().all(|&x| x == 0) {
+        state.number_of_deposits += 1;
+        let mut deposit_id_bytes = [0u8; 32];
+        deposit_id_bytes[..4].copy_from_slice(&state.number_of_deposits.to_le_bytes());
+        applied_deposit_id = deposit_id_bytes;
+    }
```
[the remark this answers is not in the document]
chrismaree: good call.

md0x (approved):
Looks good!

programs/svm-spoke/src/instructions/deposit.rs
```diff
@@ -255,9 +253,8 @@ pub fn unsafe_deposit_v3(
 pub struct Null {}
 pub fn get_unsafe_deposit_id(msg_sender: Pubkey, depositor: Pubkey, deposit_nonce: u64) -> [u8; 32] {
     let mut data = Vec::new();
-    data.extend_from_slice(&msg_sender.to_bytes());
-    data.extend_from_slice(&depositor.to_bytes());
-    data.extend_from_slice(&deposit_nonce.to_le_bytes());
+    // Use AnchorSerialize to serialize the tuple of values
+    (msg_sender, depositor, deposit_nonce).serialize(&mut data).unwrap();
```
Reinis-FRP, line 257: note that this is not AnchorSerialize wrapper (like mentioned in comment above), but using borch directly. Though, I guess functionally its the same in this case.

chrismaree: done.

PR #760, opened 2024-11-22
fix: make event case consistant between evm & svm

chrismaree:
The `EmergencyDeleteRootBundle` should be `EmergencyDeletedRootBundle` to be consistent with everything being past tense within evm & SVM events such as `ClaimedRelayerRefund`, `ExecutedRelayerRefundRoot` etc.

PR only updates solidity, rust updated here https://github.com/across-protocol/contracts/pull/774

### DIFF
--- a/programs/svm-spoke/src/event.rs
+++ b/programs/svm-spoke/src/event.rs
@@ -121,8 +121,9 @@ pub struct ClaimedRelayerRefund {
     pub refund_address: Pubkey,
 }
 
+// TODO: update the name of this to EmergencyDeletedRootBundle and in EVM.
 #[event]
-pub struct EmergencyDeletedRootBundle {
+pub struct EmergencyDeleteRootBundle {
     pub root_bundle_id: u32,
 }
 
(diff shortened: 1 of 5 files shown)


test/svm/SvmSpoke.HandleReceiveMessage.ts
```diff
@@ -520,35 +520,35 @@ describe("svm_spoke.handle_receive_message", () => {
       messageBody,
     });
 
-    // Remaining accounts specific to EmergencyDeleteRootBundle.
+    // Remaining accounts specific to EmergencyDeletedRootBundle.
```
Reinis-FRP, line 523: I don't think this and below should be renamed as it refers to accounts, not an event

programs/svm-spoke/src/event.rs
```diff
@@ -121,9 +121,8 @@ pub struct ClaimedRelayerRefund {
     pub refund_address: Pubkey,
 }
 
-// TODO: update the name of this to EmergencyDeletedRootBundle and in EVM.
```
Reinis-FRP, line 124: we are merging all other SVM PRs to master, but maybe thats ok if we are merging `svm-dev` branch soon

md0x (approved):

PR #765, opened 2024-11-22
fix(svm): use borch serializer

Reinis-FRP:
We had custom serialization implementation when hashing relayer refund and slow fill leaves. This just concatenated array elements without encoding array length. This PR fixes by using Anchor's borch serializer that encodes this more consistently.

### DIFF
--- a/programs/svm-spoke/src/instructions/bundle.rs
+++ b/programs/svm-spoke/src/instructions/bundle.rs
@@ -78,7 +78,7 @@ pub struct RelayerRefundLeaf {
 }
 
 impl RelayerRefundLeaf {
-    pub fn to_bytes(&self) -> Vec<u8> {
+    pub fn to_bytes(&self) -> Result<Vec<u8>> {
         let mut bytes = Vec::new();
 
         // This requires the first 64 bytes to be 0 within the encoded leaf data. This protects any kind of EVM leaf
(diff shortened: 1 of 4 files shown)


programs/svm-spoke/src/instructions/bundle.rs
```diff
@@ -88,23 +88,14 @@ impl RelayerRefundLeaf {
...
-        bytes.extend_from_slice(&self.chain_id.to_le_bytes());
-        for amount in &self.refund_amounts {
-            bytes.extend_from_slice(&amount.to_le_bytes());
-        }
-        bytes.extend_from_slice(&self.leaf_id.to_le_bytes());
-        bytes.extend_from_slice(self.mint_public_key.as_ref());
-        for address in &self.refund_addresses {
-            bytes.extend_from_slice(address.as_ref());
-        }
+        AnchorSerialize::serialize(&self, &mut bytes)?;
```
md0x, line 91: Wow! Much better like this 👍

test/svm/SvmSpoke.SlowFill.ts
```diff
@@ -381,7 +381,9 @@ describe("svm_spoke.slow_fill", () => {
       )
       .accounts(executeSlowRelayLeafAccounts)
       .remainingAccounts(fillRemainingAccounts)
-      .rpc();
+      .instruction();
+    const computeBudgetInstruction = ComputeBudgetProgram.setComputeUnitLimit({ units: 1_400_000 });
```
md0x, line 385: Has the borsh serialisation increased the CU?

test/svm/utils.ts
```diff
@@ -164,14 +164,17 @@ export function calculateRelayerRefundLeafHashUint8Array(relayData: RelayerRefun
 
   const refundAddressesBuffer = Buffer.concat(relayData.refundAddresses.map((address) => address.toBuffer()));
 
+  // TODO: We better consider reusing Borch serializer in production.
```
md0x, line 167: +1 we should probably test this and make sure they produce the same byte layout

test/svm/SvmSpoke.SlowFill.ts
Reinis-FRP: Its hard to tell precisely as our tests are not deterministic and have much more variance due to differing bumps is PDA derivation, but over 10 runs I observed on average +2% in CU, which I guess is acceptable here.

chrismaree (approved):
this is great!

PR #766, opened 2024-11-22
fix: stack too deep

md0x:
Changes proposed in this PR:
- Fix stack to deep error in _depositV3 by combining the function arguments into a struct `DepositV3Params`
- Fix tests

### DIFF
--- a/contracts/SpokePool.sol
+++ b/contracts/SpokePool.sol
@@ -514,21 +514,22 @@ abstract contract SpokePool is
         uint32 exclusivityParameter,
         bytes calldata message
     ) public payable override nonReentrant unpausedDeposits {
-        _depositV3(
-            depositor,
-            recipient,
-            inputToken.toAddress(), // Input token will always be an address when deposits originate from EVM.
-            outputToken,
-            inputAmount,
-            outputAmount,
-            destinationChainId,
-            exclusiveRelayer,
-            numberOfDeposits++, // Increment count of deposits so that deposit ID for this spoke pool is unique.
-            quoteTimestamp,
-            fillDeadline,
-            exclusivityParameter,
-            message
-        );
+        DepositV3Params memory params = DepositV3Params({
+            depositor: depositor,
+            recipient: recipient,
+            inputToken: inputToken,
+            outputToken: outputToken,
+            inputAmount: inputAmount,
+            outputAmount: outputAmount,
+            destinationChainId: destinationChainId,
+            exclusiveRelayer: exclusiveRelayer,
+            depositId: numberOfDeposits++, // Increment count of deposits so that deposit ID for this spoke pool is unique.
+            quoteTimestamp: quoteTimestamp,
+            fillDeadline: fillDeadline,
+            exclusivityParameter: exclusivityParameter,
+            message: message
+        });
+        _depositV3(params);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


contracts/SpokePool.sol
```diff
@@ -1270,7 +1257,8 @@ abstract contract SpokePool is
 
         // slither-disable-next-line timestamp
         uint256 currentTime = getCurrentTime();
-        if (currentTime - quoteTimestamp > depositQuoteTimeBuffer) revert InvalidQuoteTimestamp();
```
md0x, line 1273: A future quote timestamp was causing underflow errors. I added this additional check to handle the issue. Should we introduce a separate error, `InvalidFutureQuoteTimestamp()`, or is it acceptable to combine both cases into `InvalidQuoteTimestamp()`?

test/evm/hardhat/SpokePool.Deposit.ts
```diff
@@ -407,40 +390,44 @@ describe("SpokePool Depositor Logic", async function () {
       depositArgs = getDepositArgsFromRelayData(relayData);
     });
     it("placeholder: gas test", async function () {
-      await spokePool.connect(depositor)[depositV3Bytes](...depositArgs);
+      await spokePool.connect(depositor)[SpokePoolFuncs.depositV3Bytes](...depositArgs);
```
md0x, line 393: Hardhat does not automatically pick overloaded functions, so we need to explicitly specify the version.

contracts/SpokePool.sol
```diff
@@ -1242,24 +1243,10 @@ abstract contract SpokePool is
      *         INTERNAL FUNCTIONS         *
      **************************************/
 
-    function _depositV3(
-        bytes32 depositor,
-        bytes32 recipient,
-        address inputToken,
```
md0x, line 1248: Note that I’ve updated inputToken to be of type `bytes32`. It seems this was reverted to an address due to branch merges overriding the previous change to `bytes32`.

chrismaree (approved):
I like this solution a lot. I think it's better than https://github.com/across-protocol/contracts/pull/767

contracts/interfaces/V3SpokePoolInterface.sol
```diff
@@ -91,6 +91,22 @@ interface V3SpokePoolInterface {
         uint256 updatedOutputAmount;
         FillType fillType;
     }
+    // Represents the parameters required for a V3 deposit operation in the SpokePool.
+    struct DepositV3Params {
+        bytes32 depositor;
+        bytes32 recipient;
+        bytes32 inputToken;
```
md0x, line 98: See [this comment](https://github.com/across-protocol/contracts/pull/766/files#r1855152491)

contracts/SpokePool.sol
nicholaspai: I think a future quote time causing under flow is intended. We don't want people using future quotes

nicholaspai (approved):

nicholaspai (changes requested):
I also like this more than #767 but one comment here about changing logic: https://github.com/across-protocol/contracts/pull/766#discussion_r1855196658

contracts/SpokePool.sol
mrice32: Yeah, that's right. Quote timestamps should not be in the future.

contracts/SpokePool.sol
mrice32: I thought we wanted internal methods to take address types, not bytes32? Or am I misremembering?

contracts/SpokePool.sol
chrismaree: I'm not sure why this has not come in (it might be due to merge issues) but this was erroring due to this PR not being in your branch:

https://github.com/across-protocol/contracts/pull/720/files

the solution done there was:

```
if (quoteTimestamp > currentTime || currentTime - quoteTimestamp > depositQuoteTimeBuffer)
            revert InvalidQuoteTimestamp();
```

the argument in that PR was that `underflow` OR `InvalidQuoteTimestamp` is bad error handling. After speaking with @pxrl it turns out this is often a problem for integrators and having a consistent error handling flow here is important.

chrismaree:
I'm going to go ahead and merge this to unblock other changes on `svm-dev`

PR #768, opened 2024-11-24
fix(svm): optional instruction params

Reinis-FRP:
This allows the caller to choose whether to pass instruction parameters directly or load them from buffer account. As a result, we can increase inner message size in Across+. The implementation optimizes for path where parameters are passed directly as that is also simpler for caller to execute.

Currently this change is applied to fill and slow fill methods, but we might want to apply it also on relayer refund leaf execution.


programs/svm-spoke/src/instructions/slow_fill.rs
```diff
@@ -252,3 +311,26 @@ pub fn execute_v3_slow_relay_leaf<'info>(
...
+        (Some(slow_fill_leaf), Some(proof)) => ExecuteV3SlowRelayLeafParams {
+            slow_fill_leaf,
+            root_bundle_id: 0, // This is not used in the caller.
+            proof,
+        },
+        _ => account
+            .as_ref()
+            .map(|account| ExecuteV3SlowRelayLeafParams {
+                slow_fill_leaf: account.slow_fill_leaf.clone(),
+                root_bundle_id: account.root_bundle_id,
```
md0x, line 331: Is this `root_bundle_id` used by the caller? do we need it? Why in the fallback we return something and if in the other case we return 0?

md0x (approved):
Nice usage of the optional arguments. Looks good to me.

Just added a question

programs/svm-spoke/src/instructions/fill.rs
```diff
@@ -11,17 +11,21 @@ use crate::{
...
     get_current_time,
-    state::{FillStatus, FillStatusAccount, State},
+    state::{FillStatus, FillStatusAccount, FillV3RelayParams, State},
     utils::{hash_non_empty_message, invoke_handler, transfer_from},
 };
 
 #[event_cpi]
 #[derive(Accounts)]
-#[instruction(relay_hash: [u8; 32], relay_data: V3RelayData)]
+#[instruction(relay_hash: [u8; 32], relay_data: Option<V3RelayData>)]
```
chrismaree, line 20: this is cool! had not seen this syntax before.

scripts/svm/fakeFillWithRandomDistribution.ts
```diff
@@ -46,7 +48,8 @@ const argv = yargs(hideBin(process.argv))
   .option("exclusivityDeadline", { type: "number", demandOption: false, describe: "Exclusivity deadline" })
   .option("repaymentChain", { type: "number", demandOption: false, description: "Repayment chain ID" })
   .option("repaymentAddress", { type: "string", demandOption: false, description: "Repayment address" })
-  .option("distributionCount", { type: "number", demandOption: false, describe: "Distribution count" }).argv;
+  .option("distributionCount", { type: "number", demandOption: false, describe: "Distribution count" })
+  .option("bufferParams", { type: "boolean", demandOption: false, describe: "Use buffer account for params" }).argv;
```
chrismaree, line 52: 🔥

test/svm/SvmSpoke.Fill.ts
```diff
@@ -67,6 +68,7 @@ describe("svm_spoke.fill", () => {
     accounts = {
       state,
       signer: relayer.publicKey,
+      instructionParams: program.programId,
```
chrismaree, line 71: this is not too impactful, remember to update the comment in `lib.rs` to bring this to parity with https://github.com/across-protocol/contracts/pull/769 once that PR is in master

PR #772, opened 2024-11-25
fix: emit hashed message in evm fill events

Reinis-FRP:
When adding Across+ support for Solana we had to change fill related events to emit message hash (or zero bytes32 for empty message) due to Solana limitations to emit CPI events. This PR brings the exact same logic back to EVM, so that both implementations are consistent.

This would require amendments in UMIP and bot logic when validating fills to first apply the same message hashing procedure from valid deposits.


contracts/SpokePool.sol
```diff
@@ -1683,6 +1683,12 @@ abstract contract SpokePool is
         return exclusivityDeadline >= currentTime;
     }
 
+    // Helper for emitting message hash. For easier easier human readability we return bytes32(0) for empty message.
+    function _hashNonEmptyMessage(bytes memory message) internal pure returns (bytes32) {
```
chrismaree, line 1687: I changed this a little bit to make it shorter but otherwise looks perfect.

chrismaree (approved):
great. nice and simple.

mrice32 (approved):

md0x (approved):

PR #776, opened 2024-11-27
feat: update scripts

chrismaree:
After playing with scripts last week, a number of the scripts were broken due to recent changes pre-audit. 

This PR fixes this.


Anchor.toml
```diff
@@ -6,19 +6,19 @@ skip-lint = false
...
 multicall_handler = "6kqWTz3A3ZYMV2FMU24ke8rHzT82SaBz7GkBKTd7Z9BH"
-svm_spoke = "DnLjPzpMCW2CF99URhGF3jDYnVRcCJMjUWsbPb4xVoBn"
+svm_spoke = "YVMQN27RnCNt23NRxzJPumXRd8iovEfKtzkqyMc5vDt"
 
 [registry]
 url = "https://api.apr.dev"
 
 [provider]
-cluster = "localnet"
-wallet = "test/svm/keys/localnet-wallet.json"
```
Reinis-FRP, line 21: this would break forking tests where we have overrided CCTP state with this key as permissioned actor

chrismaree: yes, was not planning on committing when the PR is ready

scripts/svm/queryFills.ts
```diff
@@ -58,18 +59,22 @@ async function queryFills(): Promise<void> {
...
         { Property: "depositId", Value: event.data.depositId.toString() },
+        { Property: "depositIdNum", Value: u8Array32ToInt(event.data.depositId).toString() },
         { Property: "fillDeadline", Value: event.data.fillDeadline.toString() },
         { Property: "exclusivityDeadline", Value: event.data.exclusivityDeadline.toString() },
         { Property: "exclusiveRelayer", Value: new PublicKey(event.data.exclusiveRelayer).toString() },
         { Property: "relayer", Value: new PublicKey(event.data.relayer).toString() },
         { Property: "depositor", Value: new PublicKey(event.data.depositor).toString() },
         { Property: "recipient", Value: new PublicKey(event.data.recipient).toString() },
-        { Property: "message", Value: event.data.message.toString() },
+        { Property: "messageHash", Value: new PublicKey(event.data.messageHash).toString() },
```
Reinis-FRP, line 69: why format as Pubkey? Its more consistent with emitted EVM event to log 32 byte long hex string

scripts/svm/simpleFill.ts
```diff
@@ -49,7 +50,7 @@ async function fillV3Relay(): Promise<void> {
   const inputAmount = new BN(resolvedArgv.inputAmount);
   const outputAmount = new BN(resolvedArgv.outputAmount);
   const originChainId = new BN(resolvedArgv.originChainId);
-  const depositId = resolvedArgv.depositId;
+  const depositId = intToU8Array32(resolvedArgv.depositId);
```
Reinis-FRP, line 53: this might not work with deterministic ids because of JS number size limitations. maybe better let `resolvedArgv.depositId` be string and parse it as BigNum->U8Array32 if it starts with 0x?

chrismaree: ye, this is a good point. I've updated accordingly and updated `intToU8Array32` to support BN.

scripts/svm/simpleFill.ts
```diff
@@ -50,7 +50,7 @@ async function fillV3Relay(): Promise<void> {
   const inputAmount = new BN(resolvedArgv.inputAmount);
   const outputAmount = new BN(resolvedArgv.outputAmount);
   const originChainId = new BN(resolvedArgv.originChainId);
-  const depositId = intToU8Array32(resolvedArgv.depositId);
+  const depositId = intToU8Array32(new BN(resolvedArgv.depositId));
```
Reinis-FRP, line 53: I don't think BN constructor automatically detects and parses hex strings

scripts/svm/addressToPublicKey.ts
```diff
@@ -0,0 +1,18 @@
+import { evmAddressToPublicKey } from "../../src/SvmUtils";
+import yargs from "yargs";
+import { hideBin } from "yargs/helpers";
+const argv = yargs(hideBin(process.argv)).option("address", {
+  type: "string",
+  demandOption: true,
+  describe: "Ethereum address to convert",
+}).argv;
+
+async function logPublicKey(): Promise<void> {
```
md0x, line 10: Do we want to add this to Anchor scripts?

md0x (approved):
Looks good after addressing the BN issue that Reinis raised. Also, some scripts are not included in the Anchor.yml scripts section. Were they intended to be run differently than with `anchor run`?

scripts/svm/addressToPublicKey.ts
chrismaree: can run it with node as well but I'll add it, sure.

scripts/svm/simpleFill.ts
chrismaree: it's assumed its just a large number in a string so should be fine.

PR #777, opened 2024-11-27
feat(SpokePoolPeriphery): Support multiple exchanges 

nicholaspai:
Currently we can only initialize the periphery contract with a single exchange to swap with. This PR allows us to initialize it with multiple exchanges to swap with. Like before, these initial set of exchanges and function selectors cannot be changed post-initialization, which gives the user assurances.

~This also adds a method `whitelistExchanges` which can update the whitelisted exchange list. This means that the contract is now `Ownable` and both `initialize` and `whitelistExchanges` are marked `onlyOwner`. This means that users need to trust the owner of the contract now, and the worst thing that the owner can do is whitelist a malicious exchange. The caller of the `swapX` functions would still need to knowingly sign off on `routerCalldata` that would target this malicious exchange, so the user still retains control~

~I don't think we're going to make the contract `Ownable` and let the owner update the whitelisted exchange list post-deployment, but if we want to we can revert [this commit](https://github.com/across-protocol/contracts/pull/777/commits/dda84998fe9c9501da08b03b41016c47f5e4c3d4). The reasoning is to give better assurances to the user about the contract's expected behavior.~

This PR removes the whitelist of exchanges that can be called by the SpokePoolPeriphery, which was useful for protecting the caller from `approve`-ing the SpokePoolPeriphery contract and leaving an approval that could be stolen by another user who could direct the SpokePoolPeriphery to `transferFrom` the user who left their approval hanging. The reasoning for removing the whitelist is so that multiple exchanges can be used by the caller. In order to keep protecting the user from this "approval abuse", we add a new "Proxy" contract that should be used by the caller to call functions like `SpokePoolPeriphery.swapAndBridge` that require approvals. Now the user only approves the proxy contract which can only be used to call `SpokePoolPeriphery.swapAndBridge` rather than execute any arbitrary calldata.


bmzig (approved):

contracts/SpokeV3PoolPeriphery.sol
```diff
@@ -87,40 +90,47 @@ contract SpokePoolV3Periphery is Lockable, MultiCaller {
...
+        WhitelistedExchanges[] calldata exchanges
     ) external {
         if (initialized) revert ContractInitialized();
         initialized = true;
 
+        if (!address(_spokePool).isContract()) revert InvalidSpokePool();
         spokePool = _spokePool;
         wrappedNativeToken = _wrappedNativeToken;
-        exchange = _exchange;
+        for (uint256 i = 0; i < exchanges.length; i++) {
```
bmzig, line 126: Very nitty, but I think we can optimize gas here if we do something like:
```
WhitelistedExchanges calldata _exchange;
uint256 nExchanges = exchanges.length;
for (uint256 i = 0; i < nExchanges; i++) {
  _exchange = exchanges[i];
  ...
}
```
Although, this _is_ just an init function, so probably not necessary.

nicholaspai: fair points! https://github.com/across-protocol/contracts/pull/777/commits/c1b6d4f39b626f1b1371a7b24854156cbb33be83

nicholaspai:
I just added some forge unit tests for:
- `swapAndBridge`
- `initialize`
- `deposit`

contracts/SpokeV3PoolPeriphery.sol
```diff
@@ -13,26 +13,29 @@ import { WETH9Interface } from "./external/interfaces/WETH9Interface.sol";
...
 
     // Across SpokePool we'll submit deposits to with acrossInputToken as the input token.
     V3SpokePoolInterface public spokePool;
 
-    // Exchange address or router where the swapping will happen.
-    address public exchange;
-
     // Wrapped native token contract address.
-    WETH9Interface internal wrappedNativeToken;
+    WETH9Interface public wrappedNativeToken;
```
nicholaspai, line 38: any harm in making this public? seems potentially useful for deployer to verify

contracts/SpokeV3PoolPeriphery.sol
```diff
@@ -52,7 +55,7 @@ contract SpokePoolV3Periphery is Lockable, MultiCaller {
         // chain and if this is a contract then they will receive an ERC20.
         address recipient;
         // The destination chain identifier.
-        uint256 destinationChainid;
+        uint256 destinationChainId;
```
nicholaspai, line 58: typo that doesn't affect anything but better to remain consistent

contracts/SpokeV3PoolPeriphery.sol
```diff
@@ -87,40 +90,49 @@ contract SpokePoolV3Periphery is Lockable, MultiCaller {
...
+     * @dev This function can be front-run by anybody, so it is critical to check that the values used in the
+     * single call to this function were passed in correctly before enabling the usage of this contract.
      */
     function initialize(
         V3SpokePoolInterface _spokePool,
         WETH9Interface _wrappedNativeToken,
-        address _exchange
-    ) external {
+        WhitelistedExchanges[] calldata exchanges
+    ) external nonReentrant {
```
nicholaspai, line 119: i think making this nonReentrant makes sense even though there are no external calls--just in case we add an external call in the ftuure

contracts/SpokeV3PoolPeriphery.sol
bmzig: No harm. I made it internal only to save gas, but I agree it's annoying to have to check the `initialize` transaction to verify a correct value over just making it public.

contracts/SpokeV3PoolPeriphery.sol
bmzig: Do you also think it' worthwhile to declare the exchanges as `calldata` over `memory`. Right now, it doesn't look like it needs to be memory, and I think that might save gas if we want to be super particular.

contracts/SpokeV3PoolPeriphery.sol
nicholaspai: how does it save gas?

bmzig: It's very small, but public variables cost more than private/internal variables on deployment.

nicholaspai: ah ok i thought you were referencing gas costs on post deployment function calls, got it. we can live with that

contracts/SpokePoolV3Periphery.sol
```diff
@@ -11,28 +11,71 @@ import { V3SpokePoolInterface } from "./interfaces/V3SpokePoolInterface.sol";
...
+    function callSpokePoolPeriphery(
+        address inputToken,
+        uint256 inputAmount,
+        bytes memory peripheryFunctionCalldata
+    ) external payable nonReentrant {
+        uint256 balanceBefore = IERC20(inputToken).balanceOf(address(this));
+        IERC20(inputToken).safeTransferFrom(msg.sender, address(this), inputAmount);
+        IERC20(inputToken).forceApprove(SPOKE_POOL_PERIPHERY, inputAmount);
+        // solhint-disable-next-line avoid-low-level-calls
+        (bool success, bytes memory result) = SPOKE_POOL_PERIPHERY.call{ value: msg.value }(peripheryFunctionCalldata);
```
bmzig, line 56: I think we only want to use this proxy contract if we are calling `swapAndBridge` on the periphery, so should we just make this function a wrapper for `swapAndBridge` specifically?

nicholaspai: are you saying that because the other flows wouldn't require an `approve`?

If so, do you think the user of this Periphery contract would prefer to just go through a single contract address (i.e.g the proxy)? Or do you think the moderate gas cost increase is worth avoiding and switching on the address to call?

nicholaspai: https://github.com/across-protocol/contracts/pull/777/commits/95116666524b72d84901c6e0e083d306ecf63700

bmzig: Ah, I do see the value here. So for the permit/2/transferWithAuth scenarios, the `transferFrom` would be transferring 0 tokens to the contract, and the benefit would be for integrators (and our FE) to not need to juggle two addresses for a special case of swap and bridging, right? @dohaki, would you prefer just one address here? I personally think the extra gas is not awful, and maybe you could then just conditionally approve the peripery conract /transfer from the user if the amount to transfer is not zero.

contracts/SpokePoolV3Periphery.sol
```diff
@@ -11,28 +11,85 @@ import { V3SpokePoolInterface } from "./interfaces/V3SpokePoolInterface.sol";
...
+     * @param acrossInputToken Address of the token that will be bridged via Across as the inputToken.
+     * @param exchange Address of the exchange contract to call.
+     * @param routerCalldata ABI encoded function data to call on router. Should form a swap of swapToken for
+     * enough of acrossInputToken, otherwise this function will revert.
+     * @param swapTokenAmount Amount of swapToken to swap for a minimum amount of depositData.inputToken.
+     * @param minExpectedInputTokenAmount Minimum amount of received depositData.inputToken that we'll submit bridge
+     * deposit with.
+     * @param depositData Specifies the Across deposit params we'll send after the swap.
+     */
+    function swapAndBridge(
```
nicholaspai, line 54: Question for others:
- Do we need to force other calls on SpokePoolPeriphery to be sent through this proxy because they can be similarly "approval abused"? For example, can `permit`, `receiveWithAuthorization`, and `permit2` be similarly abused? I'm guessing no because these systems have nonces that prevent re-broadcasting signed permits/auths but I'm not sure

contracts/SpokePoolV3Periphery.sol
nicholaspai: the downside of generalizing this function is we would probably want to add some protection to the user that they don't leave any tokens in this contract:

```solidity
uint balanceBefore = x;
spokePoolPeriphery.call();
require(balanceAfter == balanceBefore, "user left tokens on proxy");
```

which would add gas.

contracts/SpokePoolV3Periphery.sol
nicholaspai: @mrice32 @bmzig @pxrl

bmzig: `receiveWithAuthorization` should be OK since it enforces that the `msg.sender` to that call is the receiver.

I do think you are right in that we may need to do something clever for permit/permit2. At least for permit, now if the signature is front-run, the actual call to the periphery may also be front-run, and the user will lose funds. So we may just need to rethink how the proxy works or if we should do a proxy. I think something similar would happen with permit2.

nicholaspai: why would user lose funds? Doesn't the permit include who the `recipient` is and it is signed by the `sender`?

bmzig: I was just thinking that if you front-run the permit signature, there will always be a race between calling `swapAndBridge` honestly and calling `swapAndBridge` with the token as the exchange and `transferFrom` as the routerCalldata. I'm still not certain on permit2, but I'd imagine if we can front-run the permit2 signature, then it's basically the same thing.

contracts/SpokePoolV3Periphery.sol
```diff
@@ -311,18 +191,35 @@ contract SpokePoolV3Periphery is Lockable, MultiCaller {
...
+    /**
+     * @notice Deposits an ERC20 token into the Spoke Pool contract.
+     * @dev User should probably just call SpokePool.depositV3 directly to save marginal gas costs,
+     * but this is added here for convenience in case caller only wants to interface Across through a single
+     * address.
+     * @param acrossInputToken ERC20 token to deposit.
+     * @param acrossInputAmount Amount of the input token to deposit.
+     * @param depositData Specifies the Across deposit params to send.
+     */
+    function depositERC20(
```
bmzig, line 214: Do we need this? Can we just call the spoke pool directly and save the extra gas?

contracts/SpokePoolV3Periphery.sol
```diff
@@ -388,8 +284,271 @@ contract SpokePoolV3Periphery is Lockable, MultiCaller {
...
+     * @param acrossInputToken Address of the token that will be bridged via Across as the inputToken.
+     * @param exchange Address of the exchange contract to call.
+     * @param routerCalldata ABI encoded function data to call on router. Should form a swap of swapToken for
+     * enough of acrossInputToken, otherwise this function will revert.
+     * @param swapTokenAmount Amount of swapToken to swap for a minimum amount of depositData.inputToken.
+     * @param minExpectedInputTokenAmount Minimum amount of received depositData.inputToken that we'll submit bridge
+     * deposit with.
+     * @param depositData Specifies the Across deposit params we'll send after the swap.
+     */
+    function swapAndBridge(
```
bmzig, line 507: Should this also be `onlyProxy`?

bmzig: I see this is the payable entrypoint.

contracts/SpokePoolV3Periphery.sol
nicholaspai: removed

mrice32 (commented):
Just one comment

contracts/SpokePoolV3Periphery.sol
```diff
@@ -12,169 +12,52 @@ import { IERC20Auth } from "./external/interfaces/IERC20Auth.sol";
...
-     * values used in the single call to this function were passed in correctly before enabling the usage of this contract.
+     * @notice Initialize the SpokePoolPeripheryProxy contract.
+     * @param _spokePoolPeriphery Address of the SpokePoolPeriphery contract that this proxy will call.
      */
-    function initialize(
-        V3SpokePoolInterface _spokePool,
-        WETH9Interface _wrappedNativeToken,
-        address _exchange
-    ) external {
+    function initialize(SpokePoolV3Periphery _spokePoolPeriphery) external nonReentrant {
```
mrice32, line 56: Why an initialize function? Do we expect this to be an upgradable contract?

contracts/SpokePoolV3Periphery.sol
```diff
@@ -388,8 +271,275 @@ contract SpokePoolV3Periphery is Lockable, MultiCaller {
...
+        );
+    }
+
+    /**
+     * @notice Calls depositV3 on spokepool.
+     * @param acrossInputToken inputToken to deposit.
+     * @param acrossInputAmount inputAmount to deposit.
+     * @param depositData DepositData required to specify Across deposit.
+     */
+    function depositERC20(
```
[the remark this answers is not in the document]
nicholaspai: I removed this function but I agree with you

contracts/SpokePoolV3Periphery.sol
nicholaspai: We want to deploy this contract at the same create2 address everywhere so I wanted to avoid passing any constructor params that would modify the bytecode. Do you think that's not needed?

contracts/SpokePoolV3Periphery.sol
nicholaspai: i guess as long as you put the routerData in the permit2 witness data that the user signs, there is no ability to front-run the permit2 submission?

contracts/SpokePoolV3Periphery.sol
mrice32: Oh, I see.

I do think that makes sense! To avoid frontrunning, should we include a deployer contract that has no input params (and thus would have a consistent address) and has an function that creates the contract and initializes it atomically? This might also be useful since there's two contracts that are linked that we want to deploy together, so this deployer can ensure that both get deployed and setup correctly.

nicholaspai: Good call: 6db7d87b23924385680d9326ae52a96db82995d8

contracts/SpokePoolV3Periphery.sol
bmzig, line 1: This would greatly increase the diff, but what do you think about replacing all the duplicate natspec comments on the functions here which are defined in `SpokePoolV3PeripheryInterface` with a single `// @inheritdoc ISpokePoolV3PeripheryInterface`? This way we don't need to maintain two identical pieces of documentation.

nicholaspai: Yeah let's do that

contracts/SpokePoolV3Periphery.sol
```diff
@@ -10,65 +10,132 @@ import { Lockable } from "./Lockable.sol";
...
+    using SafeERC20 for IERC20;
+    using Address for address;
+
+    // Flag set for one time initialization.
+    bool private initialized;
+
+    // The SpokePoolPeriphery should be deterministically deployed at the same address across all networks,
+    // so this contract should also be able to be deterministically deployed at the same address across all networks
+    // since the periphery address is the only constructor argument.
+    SpokePoolV3Periphery public SPOKE_POOL_PERIPHERY;
```
bmzig, line 39: nit: The comment is outdated/ `SPOKE_POOL_PERIPHERY` should not be in all caps anymore. Although, I'd think we could get away with defining this in the constructor.

bmzig (dismissed):
This is a really good improvement. Another thing I was wondering was if we could not pass in a struct to `_swapAndDeposit` but IIRC that causes a stack too deep error.

contracts/SpokePoolV3Periphery.sol
nicholaspai: f4250d0c3483f1a9bad6cc733436b872d52b96a0

nicholaspai:
> _swapAndDeposit

Does this function exist anymore?

contracts/SpokePoolV3Periphery.sol
```diff
@@ -155,7 +230,7 @@ contract SpokePoolV3Periphery is Lockable, MultiCaller {
         uint32 fillDeadline,
         uint32 exclusivityParameter,
         bytes memory message
-    ) external payable nonReentrant {
```
dohaki, line 158: Not directly related to this specific diff, but is there a reason why you can not specify an explicit `outputToken` here?

contracts/SpokePoolV3Periphery.sol
```diff
@@ -7,68 +7,137 @@ import { IERC20Permit } from "@openzeppelin/contracts/token/ERC20/extensions/IER
...
+     * @notice Swaps tokens on this chain via specified router before submitting Across deposit atomically.
+     * Caller can specify their slippage tolerance for the swap and Across deposit params.
+     * @dev If swapToken or acrossInputToken are the native token for this chain then this function might fail.
+     * the assumption is that this function will handle only ERC20 tokens.
+     * @param swapAndDepositData Specifies the params we need to perform a swap on a generic exchange.
+     */
+    function swapAndBridge(SpokePoolV3PeripheryInterface.SwapAndDepositData calldata swapAndDepositData)
+        external
+        override
+        nonReentrant
```
dohaki, line 76: In `SpokePoolV3Periphery` this is `payable`. Should this one be as well?

contracts/SpokePoolV3Periphery.sol
bmzig: This is a copy/paste of the spoke pool verifier deposit function, so we just assume that this is only being used with the native token,

contracts/SpokePoolV3Periphery.sol
bmzig: https://github.com/across-protocol/contracts/pull/777/commits/c1f618173eace1dd03ea2a80124c07264e473840

contracts/SpokePoolV3Periphery.sol
bmzig: This is not payable since the proxy is only used for ERC20 approval based flows. If we swap and bridge with ETH, it is more gas-efficient to call the periphery directly, so we force that the user does that.

nicholaspai:
I think we need to re-review this PR given the possibility of merge conflicts since it was last "finalized"

PR #778, opened 2024-11-27
fix: hash message hash in relay hash

Reinis-FRP:
After having implemented Across+ that replaces message with its hash in the emitted fill events, it becomes overly complex to discover `fill_status` PDA addresses as its derivation requires knowledge of full message. This PR overcomes this issue by changing how relay hash is calculated, so that message is replaced with its hash before calculating the hash of the whole relay data struct. This also drops relay hash and relay data parameters in the `close_fill_pda` by storing fill deadline in the `fill_status` account.

Same EVM changes are done in #779

### DIFF
--- a/programs/svm-spoke/src/instructions/fill.rs
+++ b/programs/svm-spoke/src/instructions/fill.rs
@@ -201,6 +201,7 @@ pub struct CloseFillPda<'info> {
     #[account(seeds = [b"state", state.seed.to_le_bytes().as_ref()], bump)]
     pub state: Account<'info, State>,
 
+    // No need to check seed derivation as this method only evaluates fill deadline that is recorded in this account.
     #[account(mut, close = signer)]
     pub fill_status: Account<'info, FillStatusAccount>,
 }
--- a/programs/svm-spoke/src/instructions/fill.rs
+++ b/programs/svm-spoke/src/instructions/fill.rs
@@ -133,9 +133,10 @@ pub fn fill_v3_relay<'info>(
         )?;
     }
 
-    // Update the fill status to Filled and set the relayer
+    // Update the fill status to Filled, set the relayer and fill deadline
     fill_status_account.status = FillStatus::Filled;
     fill_status_account.relayer = *ctx.accounts.signer.key;
+    fill_status_account.fill_deadline = relay_data.fill_deadline;
 
     if !relay_data.message.is_empty() {
         invoke_handler(ctx.accounts.signer.as_ref(), ctx.remaining_accounts, &relay_data.message)?;
@@ -193,7 +194,6 @@ fn unwrap_fill_v3_relay_params(
 }
 
 #[derive(Accounts)]
-#[instruction(relay_hash: [u8; 32], relay_data: V3RelayData)]
 pub struct CloseFillPda<'info> {
     #[account(mut, address = fill_status.relayer @ SvmError::NotRelayer)]
     pub signer: Signer<'info>,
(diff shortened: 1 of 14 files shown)


programs/svm-spoke/src/instructions/fill.rs
```diff
@@ -193,30 +194,23 @@ fn unwrap_fill_v3_relay_params(
...
     pub state: Account<'info, State>,
 
-    #[account(
-        mut,
-        seeds = [b"fills", relay_hash.as_ref()],
-        bump,
-        close = signer,
-        constraint = is_relay_hash_valid(&relay_hash, &relay_data, &state) @ SvmError::InvalidRelayHash
-    )]
+    #[account(mut, close = signer)]
```
chrismaree, line 204: it might be a useful to say why this is still safe.

programs/svm-spoke/src/instructions/slow_fill.rs
```diff
@@ -254,7 +255,8 @@ pub fn execute_v3_slow_relay_leaf<'info>(
         CpiContext::new_with_signer(ctx.accounts.token_program.to_account_info(), transfer_accounts, signer_seeds);
     transfer_checked(cpi_context, slow_fill_leaf.updated_output_amount, ctx.accounts.mint.decimals)?;
 
-    // Update the fill status to Filled. Note we don't set the relayer here as it is set when the slow fill was requested.
+    // Update the fill status to Filled. Note we don't set the relayer and fill deadline here as it is set when the slow
```
chrismaree, line 258: I know I'm irritating with this request but can we keep this on one line? might require to re-word it but its cleaner to have one line comments.

programs/svm-spoke/src/utils/merkle_proof_utils.rs
```diff
@@ -1,11 +1,16 @@
 use anchor_lang::{prelude::*, solana_program::keccak};
 
-use crate::{common::V3RelayData, error::CommonError};
+use crate::{common::V3RelayData, error::CommonError, utils::hash_non_empty_message};
 
 pub fn get_v3_relay_hash(relay_data: &V3RelayData, chain_id: u64) -> [u8; 32] {
     let mut input = relay_data.try_to_vec().unwrap();
+
+    // Trim off the serialized message (4 bytes length + message bytes) as it is replaced with the hash.
```
chrismaree, line 8: why do we need to do this?

test/svm/SvmSpoke.Fill.ts
```diff
@@ -302,7 +307,7 @@ describe("svm_spoke.fill", () => {
     await setCurrentTime(program, state, relayer, new BN(relayData.fillDeadline + 1));
 
     // Close the fill PDA
-    await program.methods.closeFillPda(relayHash, relayData).accounts(closeFillPdaAccounts).signers([relayer]).rpc();
+    await program.methods.closeFillPda().accounts(closeFillPdaAccounts).signers([relayer]).rpc();
 
     // Verify the fill PDA is closed
```
chrismaree, line 312: could you add a test to show that closing the fill PDA works if the fill was done with across plus alternative structure?

programs/svm-spoke/src/utils/merkle_proof_utils.rs
Reinis-FRP: Alternative is to serialize all struct fields manually, except for message and then serialize the message hash instead. But that is error prone and also turned out to be more CU intensive than this implementation where we serialize the whole struct with original message, trim off the message and replace it with message hash. I can add bit more comments on this, but that would be more than 1 line.

test/svm/SvmSpoke.Fill.ts
Reinis-FRP: sure, will add that

programs/svm-spoke/src/instructions/fill.rs
md0x: I know this is not in the scope of this changes but curious if we could do:
`close = fill_status.relayer` and remove the `address = fill_status.relayer` constraint in the signer

Reinis-FRP: anchor `close` method requires `sol_destination` to be `AccountInfo` as it is moving account lamports upon close

src/SvmUtils.ts
```diff
@@ -55,28 +55,33 @@ export async function readEvents<IDL extends Idl = Idl>(
...
     );
   }
 
-  let events: any[] = [];
+  const events = [];
 
-  // TODO: Add support for version 0 transactions.
-  if (!txResult || txResult.transaction.message.version !== "legacy") return events;
+  // Resolve any potential addresses that were passed from address lookup tables.
+  const messageAccountKeys = txResult.transaction.message.getAccountKeys({
```
md0x, line 71: This is still capturing the accounts not passed in ALT right?

Reinis-FRP: yes, it covers both static accounts and also the ones passed in ALT (see accountKeysFromLookups prop below)

md0x (approved):

chrismaree (approved):
nice. this is simple and matches with the implementation now on EVM.

scripts/svm/closeRelayerPdas.ts
```diff
@@ -6,8 +6,7 @@ import { PublicKey, SystemProgram } from "@solana/web3.js";
 import { SvmSpoke } from "../../target/types/svm_spoke";
```
chrismaree, line 6: thanks for updating this script here as well!

PR #780, opened 2024-11-28
feat: remote hub pool set deposit route script

md0x:
Changes proposed in this PR:
- Script to enable deposit route from HubPool


scripts/svm/constants.ts
```diff
@@ -0,0 +1,6 @@
+export const CIRCLE_IRIS_API_URL_DEVNET = "https://iris-api-sandbox.circle.com";
+export const CIRCLE_IRIS_API_URL_MAINNET = "https://iris-api.circle.com";
+export const HUB_POOL_ADDRESS_MAINNET = "0x75Baa2565C710644679F56E21ce3FF54Ef2b6192";
+export const HUB_POOL_ADDRESS_SEPOLIA = "0x75Baa2565C710644679F56E21ce3FF54Ef2b6192";
+export const SOLANA_USDC_MAINNET = "EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v";
+export const SOLANA_USDC_DEVNET = "4zMMC9srt5Ri5X14GAgXhaHii3GnPAEERYPJgZJDncDU";
```
Reinis-FRP, line 6: should we hardcode them? we might have new deployments, so maybe better to set these in env?

md0x: I've moved the `HUB_POOL_ADDRESS` to an env var but kept `SOLANA_USDC_DEVNET` and `SOLANA_USDC_MAINNET` as constants as these are managed by CCTP

scripts/svm/remoteHubPoolSetDepositRoute.ts
```diff
@@ -0,0 +1,298 @@
...
+} from "./utils/constants";
+import { fromBase58ToBytes32, fromBytes32ToAddress } from "./utils/helpers";
+
+// Set up Solana provider.
+const provider = AnchorProvider.env();
+anchor.setProvider(provider);
+
+// Parse arguments
+const argv = yargs(hideBin(process.argv))
+  .option("seed", { type: "string", demandOption: true, describe: "Seed for the state account PDA" })
```
Reinis-FRP, line 32: this might be redundant now or we still expect to deploy test builds that allow non-zero seeds?

scripts/svm/remoteHubPoolSetDepositRoute.ts
```diff
@@ -0,0 +1,298 @@
...
+  .option("originChainId", { type: "string", demandOption: true, describe: "Origin chain ID" })
+  .option("destinationChainId", { type: "string", demandOption: true, describe: "Destination chain ID" })
+  .option("depositsEnabled", { type: "boolean", demandOption: true, describe: "Deposits enabled" })
+  .option("resumeRemoteTx", { type: "string", demandOption: false, describe: "Resume receiving remote tx" })
+  .check((argv) => {
+    return true;
```
Reinis-FRP, line 38: this check is reduntant.

Reinis-FRP (approved):

PR #781, opened 2024-11-29
feat: add pause deposits script

md0x:
Changes proposed in this PR:
- Add pause deposits script

### DIFF
--- a/scripts/svm/remoteHubPoolPauseDeposits.ts
+++ b/scripts/svm/remoteHubPoolPauseDeposits.ts
@@ -1,0 +1,211 @@
+// This script bridges remote call to pause deposits on Solana Spoke Pool. Required environment:
+// - ETHERS_PROVIDER_URL: Ethereum RPC provider URL.
+// - ETHERS_MNEMONIC: Mnemonic of the wallet that will sign the sending transaction on Ethereum
+// - HUB_POOL_ADDRESS: Hub Pool address
+
+import * as anchor from "@coral-xyz/anchor";
+import { AnchorProvider, BN, Program, web3 } from "@coral-xyz/anchor";
+import { AccountMeta, PublicKey } from "@solana/web3.js";
+import "dotenv/config";
+import { ethers } from "ethers";
+import yargs from "yargs";
+import { hideBin } from "yargs/helpers";
+import { MessageTransmitter } from "../../target/types/message_transmitter";
+import { SvmSpoke } from "../../target/types/svm_spoke";
+import { decodeMessageHeader, getMessages } from "../../test/svm/cctpHelpers";
+import { HubPool__factory } from "../../typechain";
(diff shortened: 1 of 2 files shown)


scripts/svm/remoteHubPoolPauseDeposits.ts
```diff
@@ -0,0 +1,211 @@
+// This script bridges remote call to pause deposits on Solana Spoke Pool. Required environment:
+// - ETHERS_PROVIDER_URL: Ethereum RPC provider URL.
```
chrismaree, line 2: NIT: we should think about cleaning these up to be more consistent with other scripts. this is a unique variable that is not named like this in other places. might be ok, for now but as we make these more mature it should standardize.

scripts/svm/remoteHubPoolPauseDeposits.ts
```diff
@@ -0,0 +1,211 @@
...
+  const seed = new BN(0);
+  const resumeRemoteTx = resolvedArgv.resumeRemoteTx;
+  const pause = resolvedArgv.pause;
+
+  // Set up Ethereum provider.
+  if (!process.env.ETHERS_PROVIDER_URL) {
+    throw new Error("Environment variable ETHERS_PROVIDER_URL is not set");
+  }
+  const ethersProvider = new ethers.providers.JsonRpcProvider(process.env.ETHERS_PROVIDER_URL);
+  if (!process.env.ETHERS_MNEMONIC) {
```
chrismaree, line 42: https://github.com/across-protocol/contracts/blob/chrismaree/add-missing-events/scripts/deployZkSyncSpokePoolProxy.ts#L16

this structure might be better

chrismaree, line 42: https://github.com/across-protocol/contracts/blob/chrismaree/add-missing-events/scripts/deployZkSyncSpokePoolProxy.ts#L16

this structure might be better

scripts/svm/remoteHubPoolPauseDeposits.ts
```diff
@@ -0,0 +1,211 @@
...
+async function remoteHubPoolPauseDeposit(): Promise<void> {
+  const resolvedArgv = await argv;
+
+  const chainId = resolvedArgv.chainId;
+  const seed = new BN(0);
+  const resumeRemoteTx = resolvedArgv.resumeRemoteTx;
+  const pause = resolvedArgv.pause;
+
+  // Set up Ethereum provider.
+  if (!process.env.ETHERS_PROVIDER_URL) {
```
chrismaree, line 38: also this is diffrent to how the rest of the repo does this. either you should should use `CUSTOM_NODE_URL` or use `getNodeUrl` like done in other places

Reinis-FRP (approved):
confirm that the script is working as expected for me

PR #782, opened 2024-11-29
feat: add permit2 entrypoints to the periphery

bmzig:
This adds two more functions to the spoke pool periphery contract: `depositWithPermit2` and `swapAndBridgeWithPermit2`. Both use the signature transfer logic in permit2. 

There is an assumption that the amount to spend on the permit2 signature is equal to the amount to swap (for a swap and bridge tx) or the amount to deposit (for a deposit tx). I believe that this assumption is correct, since a user should not give the periphery an allowance, so it should only ever `transferFrom` the amount specified in the permit details anyways.


nicholaspai (commented):
This looks good. Can you add a forge unit test for swapAndBridgeWithPermit2?

contracts/SpokePoolV3Periphery.sol
```diff
@@ -163,12 +167,17 @@ contract SpokePoolV3Periphery is Lockable, MultiCaller {
...
-    function initialize(V3SpokePoolInterface _spokePool, WETH9Interface _wrappedNativeToken) external nonReentrant {
+    function initialize(
+        V3SpokePoolInterface _spokePool,
+        WETH9Interface _wrappedNativeToken,
+        IPermit2 _permit2
+    ) external nonReentrant {
         if (initialized) revert ContractInitialized();
         initialized = true;
 
         if (!address(_spokePool).isContract()) revert InvalidSpokePool();
```
nicholaspai, line 178: Maybe we should add a check that permit2 is initialized to a real contract similar to the spoke pool address check?

contracts/external/interfaces/IPermit2.sol
```diff
@@ -18,6 +18,13 @@ interface IPermit2 {
         uint256 requestedAmount;
     }
 
+    function permitTransferFrom(
```
nicholaspai, line 21: Do we not want to use the permitWitnessTransferFrom function? I'm not sure myself, just asking the question. Maybe @dohaki would know better

test/evm/foundry/local/SpokePoolPeriphery.t.sol
```diff
@@ -54,6 +56,7 @@ contract SpokePoolPeripheryTest is Test {
         depositor = vm.addr(1);
         owner = vm.addr(2);
         recipient = vm.addr(3);
+        permit2 = IPermit2(vm.addr(4)); // Permit2 is usused in this test.
```
nicholaspai, line 59:
```suggestion
        permit2 = IPermit2(vm.addr(4));
```

contracts/SpokePoolV3Periphery.sol
```diff
@@ -551,6 +644,18 @@ contract SpokePoolV3Periphery is Lockable, MultiCaller {
         _depositV3(acrossInputToken, acrossInputAmount, depositData);
     }
 
+    /**
+     * @notice Verifies that the signer is the owner of the signing contract.
+     * @dev The _hash and _signature fields are intentionally ignored since this contract will accept
+     * any signature which originated from permit2 after the call to the exchange.
+     * @dev This is safe since this contract should never hold funds nor approvals, other than when it is depositing or swapping.
+     */
+    function isValidSignature(bytes32, bytes calldata) external view returns (bytes4 magicBytes) {
```
bmzig, line 653: Can we actually just always return true? My reasoning is that this contract can already perform arbitrary calls -- it should _never_ have tokens or approvals -- and presumably the only thing this is used for is to transfer funds from this contract to an exchange while in the middle of a swap transaction.

contracts/external/interfaces/IPermit2.sol
bmzig: You're right. We do want it https://github.com/across-protocol/contracts/pull/782/commits/4f3b2d6b094548efeeb11dacab40acd11f67bccd

contracts/SpokePoolV3Periphery.sol
```diff
@@ -581,30 +686,50 @@ contract SpokePoolV3Periphery is Lockable, MultiCaller {
...
         uint256 srcBalanceBefore = _swapToken.balanceOf(address(this));
         uint256 dstBalanceBefore = _acrossInputToken.balanceOf(address(this));
 
-        _swapToken.forceApprove(exchange, swapTokenAmount);
+        if (_transferType == TransferType.Approval) _swapToken.forceApprove(_exchange, _swapTokenAmount);
+        else if (_transferType == TransferType.Transfer) _swapToken.transfer(_exchange, _swapTokenAmount);
+        else {
+            permit2.permit(
+                address(this), // owner
+                IPermit2.PermitSingle({
```
bmzig, line 707: Seems very strange, but we should also just be able to make these all just dummy values, since we never read anything from permit2, which is the contract that uses these values to process the approval. Sounds odd though.

contracts/SpokePoolV3Periphery.sol
```diff
@@ -53,153 +84,153 @@ contract SpokePoolPeripheryProxy is Lockable, MultiCaller {
...
+                depositData.exclusiveRelayer,
+                depositData.quoteTimestamp,
+                depositData.fillDeadline,
+                depositData.exclusivityParameter,
+                depositData.message
+            )
+        );
+        permit2.permitWitnessTransferFrom(
+            permit,
+            transferDetails,
```
nicholaspai, line 188: Shouldn't we be enforcing that the transferDetails.to == address(this)? We only want to pull funds into this contract, right?

contracts/SpokePoolV3Periphery.sol
```diff
@@ -249,6 +277,48 @@ contract SpokePoolPeripheryProxy is Lockable, MultiCaller {
...
```
nicholaspai, line 313: same question as https://github.com/across-protocol/contracts/pull/782/files#r1868191708 about `transferDetails`

contracts/SpokePoolV3Periphery.sol
```diff
@@ -581,30 +686,50 @@ contract SpokePoolV3Periphery is Lockable, MultiCaller {
...
+                address(this), // owner
+                IPermit2.PermitSingle({
+                    details: IPermit2.PermitDetails({
+                        token: address(_swapToken),
+                        amount: uint160(_swapTokenAmount),
+                        expiration: uint48(block.timestamp),
+                        nonce: nonce++
+                    }),
+                    spender: _exchange,
+                    sigDeadline: block.timestamp
```
nicholaspai, line 715: so this sigDeadline enforces that this permit is atomically used right?

contracts/SpokePoolV3Periphery.sol
nicholaspai: i think you're correct in checking `expectingPermit2Callback`, but want to see what others think

nicholaspai (commented):
This looks right to me, just have a couple q's

contracts/SpokePoolV3Periphery.sol
bmzig: Correct: https://github.com/Uniswap/permit2/blob/cc56ad0f3439c502c246fc5cfcc3db92bb8b7219/src/AllowanceTransfer.sol#L34

contracts/SpokePoolV3Periphery.sol
bmzig: We can, but if somebody did this, I think it means that they just passed in some random permit signature. The function would revert eventually, since the periphery would have no funds to swap or deposit.

contracts/SpokePoolV3Periphery.sol
```diff
@@ -53,153 +84,153 @@ contract SpokePoolPeripheryProxy is Lockable, MultiCaller {
...
+        IERC20 acrossInputToken,
+        address depositor,
+        SpokePoolV3Periphery.SwapData calldata swapData,
+        SpokePoolV3Periphery.DepositData calldata depositData,
+        IPermit2.PermitTransferFrom calldata permit,
+        IPermit2.SignatureTransferDetails calldata transferDetails,
+        bytes calldata signature
+    ) external nonReentrant {
+        bytes32 witness = keccak256(
+            abi.encode(
```
bmzig, line 172: This will be updated to a new ACROSS_SWAP_AND_DEPOSIT_DATA_TYPE in a follow-up PR (along with the other swapAndBridge functions) so that the user can't get destroyed if a malicious person modified the swapData to accept a low output amount (and thus a low inputAmount for the across input token).

contracts/SpokePoolV3Periphery.sol
```diff
@@ -53,153 +84,153 @@ contract SpokePoolPeripheryProxy is Lockable, MultiCaller {
...
+    function swapAndBridgeWithPermit2(
+        IERC20 acrossInputToken,
+        address depositor,
+        SpokePoolV3Periphery.SwapData calldata swapData,
+        SpokePoolV3Periphery.DepositData calldata depositData,
+        IPermit2.PermitTransferFrom calldata permit,
+        IPermit2.SignatureTransferDetails calldata transferDetails,
+        bytes calldata signature
+    ) external nonReentrant {
+        bytes32 witness = keccak256(
```
mrice32, line 171: Permit2 cannot be approval abused, so there should be no need for an extra hop as long as we include _all_ relevant swap and deposit data in the permit2 signed witness data.

Should we consider only using the extra hop for approvals, permit, and transferWithAuthorization?

nicholaspai: Bennett and I discussed this as well. We discussed moving swapAndBridgeWithPermit2 to the Periphery in a separate PR but since this PR only adds Permit2 functionality we should be good to put all that permit2 logic in the Periphery rather than the Proxy. @bmzig thoughts?

bmzig: Yeah I think this works. I agree that we will need to refactor this in general once the correct functionality to the non-approval based flows is finished.

nicholaspai: I've moved the permit and auth functions back to the periphery, which reduced the code diff a lot 

https://github.com/across-protocol/contracts/pull/777/files

contracts/SpokePoolV3Periphery.sol
```diff
@@ -350,93 +376,113 @@ contract SpokePoolV3Periphery is Lockable, MultiCaller {
...
+     * @param transferDetails The spender's requested transfer details for the permitted token.
+     * @param signature The permit2 signature to verify against the deposit data.
+     */
+    function swapAndBridgeWithPermit2(
+        IERC20 acrossInputToken,
+        address depositor,
+        SwapData calldata swapData,
+        DepositData calldata depositData,
+        IPermit2.PermitTransferFrom calldata permit,
+        IPermit2.SignatureTransferDetails calldata transferDetails,
```
nicholaspai, line 424: Can you not infer the transferDetails from the depositData or swapData?

contracts/SpokePoolV3Periphery.sol
nicholaspai: i'm not sure I understand this comment, shouldn't you include the types here if we need to get this to get permit2 to work?

bmzig: You're right. This will be added here. I don't think this would revert per se, but right now it is vulnerable to griefing, since the swapData is not part of the permit witness.

contracts/SpokePoolV3Periphery.sol
```diff
@@ -133,15 +104,56 @@ contract SpokePoolV3Periphery is Lockable, MultiCaller {
...
+    bool private expectingPermit2Callback;
+
+    // EIP 1271 magic bytes indicating a valid signature.
+    bytes4 private constant EIP1271_VALID_SIGNATURE = 0x1626ba7e;
+
+    // EIP 1271 bytes indicating an invalid signature.
+    bytes4 private constant EIP1271_INVALID_SIGNATURE = 0xffffffff;
+
+    // Permit2 constants:
+    bytes private constant ACROSS_DEPOSIT_DATA_TYPE =
```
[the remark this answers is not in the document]
bmzig: Yes they are. Good call. I added it here https://github.com/across-protocol/contracts/pull/782/commits/e5473c57e07ab0a517ab11e93b819265a4e7ec5e.

The commit here was pretty large partially due to adding the hashing functions needed for EIP712.

contracts/SpokePoolV3Periphery.sol
bmzig: Added here. I needed to refactor the structs since there are some overlaps between the output tokens in swaps and the input tokens in deposits. Also, since we need to include all the swap and deposit data in the permit2 witness, I had to move some arguments which were formerly not in the `SwapData/DepositData` struct into those structs so that they would be included in the witness (things like `inputToken/inputAmount`). This was done here https://github.com/across-protocol/contracts/pull/782/commits/e5473c57e07ab0a517ab11e93b819265a4e7ec5e

contracts/SpokePoolV3Periphery.sol
```diff
@@ -289,323 +377,471 @@ contract SpokePoolV3Periphery is Lockable, MultiCaller {
...
+            swapAndDepositData.depositData.message
+        );
     }
 
+    /**
+     * @notice Creates the EIP712 compliant hashed data corresponding to the BaseDepositData struct.
+     * @param baseDepositData Input struct whose values are hashed.
+     * @dev BaseDepositData is only used as a nested struct for both DepositData and SwapAndDepositData.
+     */
+    function _hashBaseDepositData(BaseDepositData calldata baseDepositData) private pure returns (bytes32) {
```
bmzig, line 769: Thoughts on moving these hashing functions and `DepositData/BaseDepositData/SwapAndDepositData` and all EIP712 stuff to a library? Something like PeripherySigningLib?

nicholaspai: yes please!

nicholaspai: This contract is getting too big

bmzig: Library added here: https://github.com/across-protocol/contracts/pull/782/commits/e1162402065398e41221f13ffb2a25a39f2d03ef

contracts/SpokePoolV3Periphery.sol
bmzig: Good call. I checked with forge and it seems like we also save ~800 gas if we do this. https://github.com/across-protocol/contracts/pull/782/commits/e1162402065398e41221f13ffb2a25a39f2d03ef

contracts/SpokePoolV3Periphery.sol
nicholaspai: I wouldn't have moved the Swap/Deposit data into the "Signing" lib, i would have kept those in the Periphery where they are used. Just a nit

bmzig: I'll change it.

contracts/SpokePoolV3Periphery.sol
```diff
@@ -289,323 +251,405 @@ contract SpokePoolV3Periphery is Lockable, MultiCaller {
...
+        uint256 _swapTokenAmount = swapAndDepositData.swapTokenAmount;
+
         // Swap and run safety checks.
         uint256 srcBalanceBefore = _swapToken.balanceOf(address(this));
         uint256 dstBalanceBefore = _acrossInputToken.balanceOf(address(this));
 
-        _swapToken.forceApprove(exchange, swapTokenAmount);
+        // The exchange will either receive funds from this contract via a direct transfer, an approval to spend funds on this contract, or via an
+        // EIP1271 permit2 signature.
+        if (_transferType == PeripherySigningLib.TransferType.Approval)
```
nicholaspai, line 586: Can you add unit tests for these three `TransferType` flows? I also think we should move `TransferType` into this contract as its not really a "signing" related struct

nicholaspai (approved):
let's just get unit tests in especially for any behavior in _swapAndBridge. I'm more ok with not having unit tests yet for the external permit/auth/permit2 logic yet but I want to make sure the swapAndBridge() and deposit() functions work well

contracts/SpokePoolV3Periphery.sol
```diff
@@ -587,6 +587,8 @@ contract SpokePoolV3Periphery is SpokePoolV3PeripheryInterface, Lockable, MultiC
         if (_transferType == TransferType.Approval) _swapToken.forceApprove(_exchange, _swapTokenAmount);
         else if (_transferType == TransferType.Transfer) _swapToken.transfer(_exchange, _swapTokenAmount);
         else {
+            _swapToken.forceApprove(address(permit2), _swapTokenAmount);
```
nicholaspai, line 590: nice that the test caught this

contracts/SpokePoolV3Periphery.sol
```diff
@@ -587,6 +587,8 @@ contract SpokePoolV3Periphery is SpokePoolV3PeripheryInterface, Lockable, MultiC
...
+            expectingPermit2Callback = true;
```
nicholaspai, line 591: And this, unit tests doing their job!

contracts/SpokePoolV3Periphery.sol
nicholaspai, line 590: This third TransferType shouldn't be called `EIP1271Signature` i don't think, I think its more appropriately named something to do with Permit2. Technically, we could use another method that requires checking the Periphery's EIP1271Signature that doesn't involve Permit2, right?

bmzig: Yeah that's correct. We could use it for other things. I renamed it to `Permit2Approval` instead here https://github.com/across-protocol/contracts/pull/782/commits/f77331676ed7058639c70429ba370b7db9359112

PR #784, opened 2024-12-02
feat: add missing TokensBridged event when amount_to_return > 0

chrismaree:


md0x (approved):
Looks perfect.
This matches the `TokensBridged` EVM version, and I agree that this couldn’t be emitted in [bridge_tokens_to_hub_pool](https://github.com/across-protocol/contracts/blob/3dcc32518a6a760f93a42c6b8e91fbc981b4791a/programs/svm-spoke/src/instructions/token_bridge.rs#L79) as we would miss some of the arguments.

PR #786, opened 2024-12-03
feat: rebalance to hub pool script

md0x:
#### Changes Proposed in this PR:
This PR introduces scripts to manage the full lifecycle of a rootBundle workflow for rebalancing USDC between the SVM Spoke Pool and the Hub Pool on mainnet. The process includes proposing a rootBundle, executing it after the liveness period, bridging via CCTP, and handling liabilities. The workflow is divided into three scripts:
- `proposeRebalanceToHubPool`: Proposes the rebalance from the Spoke Pool to the Hub Pool in the Hub Pool.
- `executeRebalanceToHubPool`: Executes the approved rebalance in the Hub Pool, bridges it via CCTP to the Spoke Pool, and processes the relayer refund leaf to create the liability.
- `bridgeLiabilityToHubPool`: Bridges any pending liability from the SVM Spoke Pool back to the Hub Pool via CCTP.


scripts/svm/utils/constants.ts
```diff
@@ -1,4 +1,12 @@
+import BN from "bn.js";
+
 export const CIRCLE_IRIS_API_URL_DEVNET = "https://iris-api-sandbox.circle.com";
 export const CIRCLE_IRIS_API_URL_MAINNET = "https://iris-api.circle.com";
 export const SOLANA_USDC_MAINNET = "EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v";
 export const SOLANA_USDC_DEVNET = "4zMMC9srt5Ri5X14GAgXhaHii3GnPAEERYPJgZJDncDU";
+export const SEPOLIA_CCTP_MESSAGE_TRANSMITTER_ADDRESS = "0x7865fAfC2db2093669d92c0F33AeEF291086BEFD";
+export const MAINNET_CCTP_MESSAGE_TRANSMITTER_ADDRESS = "0x0a992d191deec32afe36203ad87d7d289a738f81";
+export const SEPOLIA_USDC_ADDRESS = "0x1c7D4B196Cb0C7B01d743Fbc6116a902379C7238";
+export const MAINNET_USDC_ADDRESS = "0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48";
```
Reinis-FRP, line 10: better reuse `TOKEN_SYMBOLS_MAP` as exported in `utils/constants.ts`

scripts/svm/proposeRebalanceToHubPool.ts
```diff
@@ -0,0 +1,136 @@
...
+    { Property: "netSendAmount (formatted)", Value: formatUsdc(netSendAmount) },
+    { Property: "poolRebalanceRoot", Value: poolRebalanceTree.getHexRoot() },
+    { Property: "relayerRefundRoot", Value: merkleTree.getHexRoot() },
+  ]);
+
+  console.log("Submitting proposal...");
+  const tx = await hubPool.connect(ethersSigner).proposeRootBundle(
+    [0], // bundleEvaluationBlockNumbers, not checked in this script.
+    1, // poolRebalanceLeafCount, only one leaf in this script.
+    poolRebalanceTree.getHexRoot(), // poolRebalanceRoot.
```
Reinis-FRP, line 126: why is pool rebalance needed, cannot it be zero hash?

md0x: I haven’t been able to do it with a zero hash in the poolRebalanceTree. I think we need to pass at least a tree with one leaf; the leaf can be constructed with zero hashes, except for the chain ID. Otherwise, we cannot get past this line:
https://github.com/across-protocol/contracts/blob/16060638d33df763436ec3cbc1336f11aba6be63/contracts/HubPool.sol#L635-L651

Additionally, the chain ID needs to be valid; otherwise, this line fails:
https://github.com/across-protocol/contracts/blob/16060638d33df763436ec3cbc1336f11aba6be63/contracts/HubPool.sol#L655

scripts/svm/utils/constants.ts
md0x: Do you mean doing this?
https://github.com/across-protocol/contracts/pull/786/commits/c17299c65184e069af3dfda37afa39e61e9c1e78

Reinis-FRP: kindof, but better avoid putting the var in constants and not pass chainId. E.g. can pull required token for required chain dynamically, like [here](https://github.com/across-protocol/contracts/blob/0785cf2519e52d40629c6d3337f4a7c3441f3c7d/scripts/svm/proposeRebalanceToSpokePool.ts#L51)

scripts/svm/proposeRebalanceToHubPool.ts
Reinis-FRP: oh, yes, you're totally right on this!

scripts/svm/utils/constants.ts
md0x: You are absolutely right. I fixed it here: https://github.com/across-protocol/contracts/pull/786/commits/10a60f2b7d5f9bbd443fa8342df6f24b47b43e2b

scripts/svm/bridgeLiabilityToHubPool.ts
```diff
@@ -0,0 +1,305 @@
...
+  console.log(`SVM Spoke Pool Final Vault balance: ${formatUsdc(BigNumber.from(finalVaultBalance))} USDC.`);
+  console.log(
+    `Sent ${formatUsdc(BigNumber.from(initialVaultBalance).sub(BigNumber.from(finalVaultBalance)))} USDC through CCTP.`
+  );
+
+  return tx;
+}
+
+// Run the bridgeLiabilityToHubPool function
+bridgeLiabilityToHubPool();
```
chrismaree, line 305: this script is SUPER useful. This will basically be what the finalizer needs to do in production.

scripts/svm/bridgeLiabilityToHubPool.ts
```diff
@@ -0,0 +1,305 @@
...
+  console.table([
+    { Property: "isTestnet", Value: process.env.TESTNET === "true" },
+    { Property: "hubPoolAddress", Value: hubPoolAddress },
+    { Property: "svmSpokeProgramProgramId", Value: svmSpokeProgram.programId.toString() },
+    { Property: "providerPublicKey", Value: provider.wallet.publicKey.toString() },
+    { Property: "usdcBalanceBefore", Value: usdcBalanceBefore.toString() },
+  ]);
+
+  const [transferLiability] = PublicKey.findProgramAddressSync(
+    [Buffer.from("transfer_liability"), new PublicKey(svmUsdc).toBuffer()],
```
chrismaree, line 164: only comment I could make on this script (but I think it's fine as we do similar things all over the place) is this assumes we are working with USDC.

scripts/svm/executeRebalanceToHubPool.ts
```diff
@@ -0,0 +1,482 @@
+/**
+ * Script: Execute USDC Rebalance to Hub Pool
+ *
+ * This script executes a previously proposed root bundle on the Hub Pool
+ * to rebalance USDC from the Solana Spoke Pool to the Ethereum Hub Pool.
+ * It handles CCTP attestations, relayer refund leaf execution, and prepares
+ * pending liabilities for bridging back to the Hub Pool.
```
chrismaree, line 7: just an irritating nit that like with the other scripts we have 120 chars to work with so we don't need to wrap shorter than 120.

scripts/svm/executeRebalanceToHubPool.ts
```diff
@@ -0,0 +1,482 @@
...
+ *
+ * Example Usage:
+ * TESTNET=true \
+ * NODE_URL_11155111=$NODE_URL_11155111 \
+ * MNEMONIC=$MNEMONIC \
+ * HUB_POOL_ADDRESS=$HUB_POOL_ADDRESS \
+ * anchor run executeRebalanceToHubPool \
+ * --provider.cluster "devnet" \
+ * --provider.wallet $SOLANA_PKEY_PATH \
+ * -- --netSendAmount 7
```
chrismaree, line 28: so this is provided to re-construct the merkle leaf & proof? it also assumes that there are no relayer repayments within the leaf, right?

in practice, I don't think this is how this flow will work, right? it's assumed that the data worker will post the leafs & proofs to a public location so these can be retrieved. this particular script flow is only useful in this particular situation where you don't have any relayer repayments & we only do a net repayment.

It would be useful to be able to pass in the leafs as en env/json or something like this? this would let you take them from the script that does the proposing, save them, and then not need to re-generate them. it would also better track with what things will look like when we interface with the data worker later (executor should not need to generate leafs, just get given them)

chrismaree: (prehaps this is still useful if we are dealing with JUST creating leafs to pull/push tokens to/from SVM via the netSendAmount

md0x: Yes, exactly. The `netSendAmount` is merely to avoid having to pass leaves, roots, proofs, etc., between this script and `proposeRebalanceToHubPool.ts`. It also assumes that the `poolRebalanceTree` has a single leaf with empty data and the solanaChainId. Thus, both scripts are dependent on each other. I believe they do not showcase a real production scenario; they are more like helpers to test the rebalance from the SVM Spoke to the HubPool.

On the other hand, I separated `bridgeLiabilityToHubPool.ts` from them because this one could represent a real use case, as you pointed out. It’s essentially what the finalizer needs to do.

scripts/svm/bridgeLiabilityToHubPool.ts
md0x: Yeah, in many places we have dependencies on USDC, including in our smart contracts (solana adapter) and svm programs. Do you think it’s worth generalizing?

scripts/svm/executeRebalanceToHubPool.ts
md0x: Fixed!

Reinis-FRP (approved):

PR #789, opened 2024-12-04
feat: improve event utils

chrismaree:
simple PR to improve the events utils used on SVM. measurably improves performance and enables us to handle more than 1000 event limt.


Reinis-FRP (approved):
nice!

src/SvmUtils.ts
```diff
@@ -108,19 +109,53 @@ export function getEvent(events: any[], program: PublicKey, eventName: string) {
...
+    }
 
-  for (const signature of pastSignatures) {
-    events.push(...(await readEvents(connection, signature.signature, [program], finality)));
+    if (options.limit && signatures.length < options.limit) break; // Exit early if the number of signatures < limit
   }
-  return events;
+
+  // Fetch events for all signatures in parallel
+  const eventsWithSlots = await Promise.all(
```
md0x, line 144: This could be resource intensive if there are many signatures. Should we consider batching the Promise.all calls with a configurable batch size to improve efficiency?

md0x (approved):
Just a performance observation but overall looks good!

src/SvmUtils.ts
chrismaree: same net efficiency, right? it's the same number of requests. the bigger issue could be the RPC provider blocking this or having an issue with the number of requests but for now we can leave it like this and improve when we get there.

md0x: Yeah, I meant splitting in chunks and run them sequentially. Similarly to how we do it here:
https://github.com/UMAprotocol/protocol/blob/e526e5a1caf68ae7be706f6ef9c996660429282a/packages/common/src/EventUtils.ts#L36-L38

But it's ok for now I think!

PR #790, opened 2024-12-04
feat: sponsored swap and deposits

bmzig:
This PR adds an extra signatureOwner and signature argument to the functions which use `permit` and `receiveWithAuthorization` so that we no longer have a dependency for the depositor to call these functions. This should mean that as long as the depositor signed off on the input deposit/swapAndBridge arguments for the `DepositData` and `SwapAndDepositData` structs respectively, then anybody can call these functions and execute the swap.

Most of the diff for this PR are tests.


test/evm/foundry/local/SpokePoolPeriphery.t.sol
```diff
@@ -413,29 +854,146 @@ contract SpokePoolPeripheryTest is Test {
...
+        spokePoolPeriphery.swapAndBridgeWithPermit2(
+            depositor, // signatureOwner
+            swapAndDepositData,
+            permit,
+            signature
+        );
+    }
 
-        vm.stopPrank();
+    function testPermit2SwapAndBridgeValidWitness(address rando) public {
```
bmzig, line 928:
```suggestion
    function testPermit2SwapAndBridgeInvalidWitness(address rando) public {
```

contracts/SpokePoolV3Periphery.sol
```diff
@@ -491,6 +536,14 @@ contract SpokePoolV3Periphery is SpokePoolV3PeripheryInterface, Lockable, MultiC
             s
         );
 
+        // Verify that the signatureOwner signed the input depositData.
+        if (
```
nicholaspai, line 540: Is this code block the same for all gasless flows? Can we refactor into internal func?

nicholaspai (approved):
Looks good! just one suggestion

contracts/SpokePoolV3Periphery.sol
bmzig: It is the same. Factored out here: https://github.com/across-protocol/contracts/pull/790/commits/dfa6f89bcbbfc3fea759b06ed37cd600b0a64668

PR #795, opened 2024-12-07
fix: Add `--json` flag to checkStorageLayout's `forge inspect` command

nicholaspai:
The `forge inspect` behavior changed in one of the `nightly` builds recently, and this broke CI because our CI pulls from the latest `nightly` version each time it runs. [The default behavior prints a table now](https://github.com/foundry-rs/foundry/blob/00efa0d5965269149f374ba142fb1c3c7edd6c94/crates/forge/bin/cmd/inspect.rs#L184), which honestly is pretty nice to look at and maybe we should use it. However, the table is a bit harder to modify using simple `jq` bash commands like we do today.

However, the simplest fix for now is to add the `--json` flag to `inspect` command. 

I've also added other spoke pools to the script.

### DIFF
--- a/scripts/checkStorageLayout.sh
+++ b/scripts/checkStorageLayout.sh
@@ -1,10 +1,10 @@
 #!/bin/bash
 
-CONTRACTS=("Arbitrum_SpokePool")
+CONTRACTS=("Arbitrum_SpokePool" "Optimism_SpokePool" "Polygon_SpokePool" "Linea_SpokePool" "ZkSync_SpokePool" "Ethereum_SpokePool" "Base_SpokePool" "Mode_SpokePool" "Blast_SpokePool" "AlephZero_SpokePool" "Redstone_SpokePool" "Scroll_SpokePool" "WorldChain_SpokePool" "Zora_SpokePool" "PolygonZkEVM_SpokePool")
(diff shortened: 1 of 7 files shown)


scripts/checkStorageLayout.sh
```diff
@@ -1,10 +1,10 @@
 #!/bin/bash
 
-CONTRACTS=("Arbitrum_SpokePool" "Optimism_SpokePool" "Polygon_SpokePool" "Linea_SpokePool" "ZkSync_SpokePool" "Ethereum_SpokePool" "Base_SpokePool" "Mode_SpokePool" "Blast_SpokePool")
+CONTRACTS=("Arbitrum_SpokePool" "Optimism_SpokePool" "Polygon_SpokePool" "Linea_SpokePool" "ZkSync_SpokePool" "Ethereum_SpokePool" "Base_SpokePool" "Mode_SpokePool" "Blast_SpokePool" "AlephZero_SpokePool" "Redstone_SpokePool" "Scroll_SpokePool" "WorldChain_SpokePool" "Zora_SpokePool" "PolygonZkEVM_SpokePool")
```
chrismaree, line 3: does this mean we were not checking things properly before? I guess so

chrismaree (approved):

PR #799, opened 2024-12-13
improve: Support testing chain adapters w/o token defs

pxrl:


pxrl:
nb. I needed this for testing USDC.e on Aleph Zero, which is not currently mapped via the HubPool.

bmzig (approved):

nicholaspai (approved):

PR #803, opened 2024-12-17
fix: standardize evm env in scripts

Reinis-FRP:
Standardizes environment variables for EVM chain provider URL and signer mnemonics used in scripts.

Fixes: https://linear.app/uma/issue/ACX-3482/replace-ethers-mnemonic-ethers-provider-url-in-scripts-with-a


chrismaree (approved):
great, thanks for fixing this!

md0x (approved):

PR #804, opened 2024-12-17
feat(svm): query events script

md0x:
Changes proposed in this PR:
- Add script to read and print svm events in a human readable format.

### DIFF
--- a/scripts/svm/queryEvents.ts
+++ b/scripts/svm/queryEvents.ts
@@ -36,7 +36,6 @@ const argvPromise = yargs(hideBin(process.argv)).option("eventName", {
   ],
 }).argv;
 
-// Parse argument
 async function queryEvents(): Promise<void> {
   const argv = await argvPromise;
   const eventName = argv.eventName || "any";
--- a/scripts/svm/queryEvents.ts
+++ b/scripts/svm/queryEvents.ts
@@ -1,11 +1,10 @@
 // This script queries the events of the spoke pool and prints them in a human readable format.
 import * as anchor from "@coral-xyz/anchor";
-import { AnchorProvider, BN, Program } from "@coral-xyz/anchor";
-import { publicKeyToEvmAddress, readProgramEvents } from "../../src/svm";
-import { SvmSpoke } from "../../target/types/svm_spoke";
-import { BigNumber } from "ethers";
+import { AnchorProvider, Program } from "@coral-xyz/anchor";
 import yargs from "yargs";
 import { hideBin } from "yargs/helpers";
+import { readProgramEvents, stringifyCpiEvent } from "../../src/svm";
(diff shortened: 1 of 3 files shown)


chrismaree (approved):

src/svm/solanaProgramUtils.ts
```diff
@@ -189,3 +191,69 @@ export async function subscribeToCpiEventsForProgram(
...
+  const depositId = deserialize(depositIdSchema, DepositId, Buffer.from(value));
+  return new BN(depositId.value).toString();
+}
+
+/**
+ * Stringifies a CPI event.
+ */
+export function stringifyCpiEvent(obj: any): any {
+  if (obj?.constructor?.toString()?.includes("PublicKey")) {
+    if (obj.toString().includes("111111111111")) {
```
Reinis-FRP, line 235: unlikely to have zeroes anywhere else than in the beginning, though, still better to restrict that only the first 12 bytes matching 0 are translated as EVM addresses

src/svm/solanaProgramUtils.ts
```diff
@@ -189,3 +191,69 @@ export async function subscribeToCpiEventsForProgram(
...
+      // First 8 bytes are 0 for EVM addresses.
```
Reinis-FRP, line 236:
```suggestion
      // First 12 bytes are 0 for EVM addresses.
```

src/svm/solanaProgramUtils.ts
md0x: 100%

scripts/svm/queryEvents.ts
```diff
@@ -0,0 +1,50 @@
...
+const idl = require("../../target/idl/svm_spoke.json");
+const program = new Program<SvmSpoke>(idl, provider);
+const programId = program.programId;
+console.log("SVM-Spoke Program ID:", programId.toString());
+
+const argvPromise = yargs(hideBin(process.argv)).option("eventName", {
+  type: "string",
+  demandOption: false,
+  describe: "Name of the event to query",
+  choices: [
```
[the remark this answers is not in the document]
md0x: The [EventCoder](https://github.com/across-protocol/contracts/blob/78e35850fc1afa1ffa83d26f116bf3678600313c/src/svm/solanaProgramUtils.ts#L84) returns them with the first letter in lowercase. I guess we could capitalize the events ourselves in the [processEventFromTx](https://github.com/across-protocol/contracts/blob/78e35850fc1afa1ffa83d26f116bf3678600313c/src/svm/solanaProgramUtils.ts#L55) function as a custom treatment, but we aren’t doing that. That’s why the event names should match those returned by the EventCoder.

Reinis-FRP (approved):
nice!

PR #805, opened 2024-12-18
feat: update function and event naming for backwards compatibility

chrismaree:
This PR makes a single functional change: it adds back the previous `fillV3Relay` interface that existed prior to this audit. This method forwards to `fillRelay`

The rest of the PR is entirely renaming to make it an easier upgrade for integrators:
1. The old `deposit` function has been removed and replaced with the colliding function name `depositDeprecated_5947912356`. Note: this means that anyone who fails to update their code will not be broken. However, it means that we are able to reuse the deposit name without function overloading.
2. The `bytes32` overload of `depositV3` has been renamed to just `deposit`.
3. Because unsafeDepositV3 has never been deployed to production, it was renamed to `unsafeDeposit` and only the `bytes32` version was kept.
4. The `bytes32` overload of `depositV3Now` was renamed to `depositNow`.
5. The `bytes32` overload of `speedUpV3Deposit` was renamed to `speedUpDeposit`.
6. The `bytes32` version of `fillV3Relay` was renamed to `fillRelay`.
7. The following transformation was applied to events `V3FundsDeposited`, `RequestedSpeedUpV3Deposit`, `FilledV3Relay`, `RequestedV3SlowFill`:
    1. For the `bytes32` version, the `V3` was dropped from the name.
    2. The legacy event _with_ the `V3` was added back to the interface, so that it persists in the ABI to make migration easier.


nicholaspai (commented):
I was just thinking about this, +1

contracts/interfaces/V3SpokePoolInterface.sol
```diff
@@ -62,6 +62,22 @@ interface V3SpokePoolInterface {
...
+    struct V3RelayDataLegacy {
+        address depositor;
+        address recipient;
+        address exclusiveRelayer;
+        address inputToken;
+        address outputToken;
+        uint256 inputAmount;
+        uint256 outputAmount;
+        uint256 originChainId;
+        uint256 depositId;
```
chrismaree, line 75: ^

contracts/SpokePool.sol
```diff
@@ -633,8 +633,8 @@ abstract contract SpokePool is
         uint32 fillDeadline,
         uint32 exclusivityParameter,
         bytes calldata message
-    ) public payable {
-        unsafeDepositV3(
+    ) public payable override {
```
chrismaree, line 636: this did not have `override`, so I've added it to the interface.

contracts/SpokePool.sol
```diff
@@ -766,8 +766,8 @@ abstract contract SpokePool is
         uint32 fillDeadlineOffset,
         uint32 exclusivityPeriod,
         bytes calldata message
-    ) external payable {
-        depositV3(
+    ) external payable override {
```
chrismaree, line 769: was also missing from interface.

contracts/SpokePool.sol
```diff
@@ -827,7 +827,7 @@ abstract contract SpokePool is
         uint32 fillDeadlineOffset,
         uint32 exclusivityPeriod,
         bytes calldata message
-    ) external payable {
+    ) external payable override {
```
chrismaree, line 830: was also missing from interface.

test/evm/hardhat/constants.ts
```diff
@@ -109,22 +109,3 @@ export const sampleRateModel = {
   R1: toWei(0.07).toString(),
   R2: toWei(0.75).toString(),
 };
-
-export const SpokePoolFuncs = {
```
chrismaree, line 113: no more loverload. yay!

contracts/SpokePoolVerifier.sol
```diff
@@ -42,7 +42,7 @@ contract SpokePoolVerifier {
      * to 0 if exclusiveRelayer is set to 0x0, and vice versa.
      * @param fillDeadline Timestamp after which this deposit can no longer be filled.
      */
-    function deposit(
+    function depositV3Bytes32(
```
chrismaree, line 45: ^

contracts/SpokePool.sol
```diff
@@ -862,7 +827,7 @@ abstract contract SpokePool is
      * account. If depositor is a contract, then should implement EIP1271 to sign as a contract. See
      * _verifyUpdateV3DepositMessage() for more details about how this signature should be constructed.
      */
-    function speedUpV3Deposit(
+    function speedUpDeposit(
```
md0x, line 830: Should we rename for consistency?
`UPDATE_V3_DEPOSIT_DETAILS_HASH` to `UPDATE_DEPOSIT_DETAILS_HASH`
https://github.com/across-protocol/contracts/blob/de61c0a2f06f4981a9352b0ed968acb7ae6337b9/contracts/SpokePool.sol#L140-L141

and
`UPDATE_V3_DEPOSIT_ADDRESS_OVERLOAD_DETAILS_HASH` to `UPDATE_V3_DEPOSIT_DETAILS_HASH`
https://github.com/across-protocol/contracts/blob/de61c0a2f06f4981a9352b0ed968acb7ae6337b9/contracts/SpokePool.sol#L145

contracts/SpokePool.sol
```diff
@@ -862,7 +827,7 @@ abstract contract SpokePool is
      * account. If depositor is a contract, then should implement EIP1271 to sign as a contract. See
      * _verifyUpdateV3DepositMessage() for more details about how this signature should be constructed.
```
md0x, line 828: Should we rename _verifyUpdateV3DepositMessage to _verifyUpdateDepositMessage?

contracts/SpokePool.sol
chrismaree: agreed, done.

foundry.toml
```diff
@@ -23,11 +23,11 @@ remappings = [
     "hardhat/=node_modules/hardhat/",
 ]
 via_ir = true
-optimizer_runs = 1_000_000
-solc_version = "0.8.25"
```
chrismaree, line 27: ^

PR #806, opened 2024-12-18
feat(svm): include solana artifacts in package build output

md0x:
Changes proposed in this PR:
- Include Solana artifacts in the package build output.
- Add utility functions to get connected Anchor program instances with types.


scripts/generate-svm-assets.sh
```diff
@@ -0,0 +1,89 @@
...
+for file in "$SVM_TYPES"/*.ts; do
+  [ "$(basename -- "$file")" = "index.ts" ] && continue
+  filename=$(basename -- "$file")
+  name="${filename%.ts}"
+  camelCaseName=$(echo "$name" | awk -F'_' '{
+    for (i=1; i<=NF; i++) {
+      printf toupper(substr($i,1,1)) tolower(substr($i,2));
+    }
+  }')
+  newName="${camelCaseName}Anchor"
```
md0x, line 85: Prepend "Anchor" to the types to prevent conflicts with evm types

scripts/generate-svm-assets.sh
```diff
@@ -0,0 +1,89 @@
...
+
+for file in "$SVM_IDL"/*.json; do
+  filename=$(basename -- "$file")
+  name="${filename%.json}"
+  camelCaseName=$(echo "$name" | awk -F'_' '{
+    for (i=1; i<=NF; i++) {
+      printf toupper(substr($i,1,1)) tolower(substr($i,2));
+    }
+  }')
+  IMPORTS="${IMPORTS}const ${camelCaseName}Idl = require(\"./${filename}\");\n"
```
md0x, line 45: We use `require()` for the IDLs similarly as before in the scripts to avoid ts issues: https://github.com/across-protocol/contracts/blob/e0cc5a2801d53a72da5bb5db469da8cfdf3468ad/scripts/svm/executeRebalanceToHubPool.ts#L77

index.ts
```diff
@@ -1,3 +1,4 @@
 export * from "./typechain";
 export * from "./src/DeploymentUtils";
 export * from "./utils";
+export * from "./src/svm/assets";
```
md0x, line 4: The program getters, types, and IDLs can be imported directly from `"@across-protocol/contracts"`, but we don’t import the SVM utility functions because a subset of these should be shared in the SDK.

index.ts
```diff
@@ -1,3 +1,4 @@
 export * from "./typechain";
 export * from "./src/DeploymentUtils";
 export * from "./utils";
+export * from "./src/svm";
```
md0x, line 4: The program getters, types, and IDLs can be imported directly from `@across-protocol/contracts`, as well as utility functions (we should move the latter to the SDK next).

src/svm/programConnectors.ts
```diff
@@ -0,0 +1,35 @@
...
+  SvmSpokeAnchor,
+  SvmSpokeIdl,
+  TokenMessengerMinterAnchor,
+  TokenMessengerMinterIdl,
+} from "./assets";
+
+export function getConnectedProgram<P extends Idl>(idl: P, provider: Provider, programId?: string) {
+  if (programId) {
+    // An IDL must have been previously initialized
+    return Program.at<P>(programId, provider);
```
md0x, line 16: Note: In the SVM scripts, we have been connecting to the program ID defined in the SvmSpokeIdl file. Since these IDs are generated automatically, any future changes to the program, if merged into master, could cause the program ID to drift.
I've added the optional programId argument to target specific id's and I think this is safer I general. Wdy?

src/svm/programConnectors.ts
```diff
@@ -0,0 +1,31 @@
...
+  MessageTransmitterAnchor,
+  MessageTransmitterIdl,
+  TokenMessengerMinterAnchor,
+  TokenMessengerMinterIdl,
+  MulticallHandlerAnchor,
+  MulticallHandlerIdl,
+} from "./assets";
+
+export function getConnectedProgram<P extends Idl>(idl: P, provider: Provider) {
+  return new Program<P>(idl, provider);
```
md0x, line 14: Note: In the SVM scripts, we have been connecting to the program ID defined in the IDL file. Since these IDs are generated automatically, any future changes to the program, if merged into master, could cause the program ID to drift.
We should explore alternatives where the caller can pass a specific program ID in a followup PR so we can use this in our bots to prevent this unexpected programIDs changes. Wdyt?

Reinis-FRP: agree, we should better implement swapping out the programId

Reinis-FRP (approved):
nice, this is much better that we had before!

src/DeploymentUtils.ts
```diff
@@ -6,7 +6,11 @@ interface DeploymentExport {
 const deployments: DeploymentExport = deployments_ as any;
 
 // Returns the deployed address of any contract on any network.
-export function getDeployedAddress(contractName: string, networkId: number, throwOnError = true): string | undefined {
+export function getDeployedAddress(
+  contractName: string,
+  networkId: number | string,
```
md0x, line 11: Network id `1810017368444177321` (solana-devnet) overflows js numbers so we need to pass it as a string

src/svm/programConnectors.ts
md0x: Finally implemented the programId override as it enhances the usability of these functions

Reinis-FRP (approved):

package.json
```diff
@@ -26,10 +26,11 @@
     "lint-fix": "yarn prettier --write **/*.js **/*.ts ./programs/**/*.rs ./contracts**/*.sol && cargo +nightly fmt  --all && cargo clippy",
     "clean-fast": "for dir in node_modules cache cache-zk artifacts artifacts-zk dist typechain; do mv \"${dir}\" \"_${dir}\"; rm -rf \"_${dir}\" &; done",
     "clean": "rm -rf node_modules cache cache-zk artifacts artifacts-zk dist typechain",
+    "generate-svm-assets": "sh ./scripts/generate-svm-assets.sh",
```
chrismaree, line 29: we should think about making this file easier to reason about with this new command you've added. the equivalent evm command is `generate-contract-types`. I tried with the other commands to make it clear what they are (solidity/js/rust/evm/svm) but this one is now a bit inconsistent. perhaps if we call it `generate-svm-assets` and then re-name the `generate-contract-types` to `generate-evm-assets`?

also, I think that given this command is not yet within the `prepublish` command we could actually be safe to merge this into master without it joining the released package.

src/svm/programConnectors.ts
chrismaree: ye, there are similar things we'd want to do on the evm side I think wherein the deployments/release process also includes a hash or some deterministic way to derive the code we are revering to on a release.

can fit into this as well.

package.json
md0x: Updated ✅

> also, I think that given this command is not yet within the prepublish command we could actually be safe to
merge this into master without it joining the released package.

I think it is, its part of the build as we need to run `yarn generate-svm-assets` before `yarn build-ts`. See this:
https://github.com/across-protocol/contracts/blob/907b7d7982a90ee456bff87a9cef179f51b23d87/package.json#L33

Let's wait for @mrice32 & @nicholaspai to see what they think the proposed changes.

deployments/deployments.json
```diff
@@ -163,5 +163,10 @@
     "SpokePool": { "address": "0xeF684C38F94F48775959ECf2012D7E864ffb9dd4", "blockNumber": 1139240 },
     "SpokePoolVerifier": { "address": "0xB4A8d45647445EA9FC3E1058096142390683dBC2", "blockNumber": 1152853 },
     "MulticallHandler": { "address": "0x924a9f036260DdD5808007E1AA95f08eD08aA569", "blockNumber": 1145284 }
+  },
+  "1810017368444177321": {
```
md0x, line 167: This is our chain id representation of solana-devnet, see this: https://github.com/across-protocol/contracts/blob/ddbe2fc687c503c7f52d4d0cfb7871ef15816091/src/svm/helpers.ts#L8-L12

deployments/README.md
```diff
@@ -133,7 +133,7 @@ This is because this `deployments.json` file is used by bots in [`@across-protoc
 
 ## Ink mainnet (57073)
 
-| Contract Name | Address                                                                                                                          |
-| ------------- | -------------------------------------------------------------------------------------------------------------------------------- |
-| Ink_SpokePool | [0xeF684C38F94F48775959ECf2012D7E864ffb9dd4](https://explorer.inkonchain.com/address/0xeF684C38F94F48775959ECf2012D7E864ffb9dd4) |
-| MulticallHandler   | [0x924a9f036260DdD5808007E1AA95f08eD08aA569](https://explorer.inkonchain.com/address/0x924a9f036260DdD5808007E1AA95f08eD08aA569) |
+| Contract Name    | Address                                                                                                                          |
```
md0x, line 136: Unrelated to this PR, linter fixes

Reinis-FRP (approved):

PR #808, opened 2024-12-18
feat(ink): configure and deploy Ink contracts

james-a-morris:
All deployments made against origin commit hash `7f9ebb360c9bfcc9960a0226a92098fccc186ce8`. Rebased against master with all deployments made.


contracts/Ink_SpokePool.sol
bmzig, line 1: I think you are going to need to copy specifically the WorldChain_SpokePool contract, since it has a special `_bridgeTokensToHubPool` function you will need to use. It accounts for upgradeable USDC, since for OP stack chains, upgradeable USDC does not use the canonical bridge contract, but instead some other address. (This will also mean you need to specify the upgradeable USDC address when deploying the spoke pool).

james-a-morris: To clarify: we **do not** have the ERC20 token address for USDC nor have we gotten confirmation from the RaaS provider/Ink that this will be upgradeable 😞

pxrl: I think it's probably the right call to `cp WorldChain_SpokePool.sol Ink_SpokePool.sol`, but to proceed with the USDC address set to 0x0. Then we can support USDC by just redeploying with the USDC address set in the constructor.

deploy/057_deploy_ink_spokepool.ts
pxrl, line 1: @james-a-morris Can you try deploying this with `056_deploy_op_adapter.ts` instead? Creating a new deployment script for each new deployment is an antipattern that we're trying to get away from.

The script would need a one-off tweak to revert to 0x0 for the USDC address on Ink; I think that's OK because upgradable USDC will ultimately be supported.

deploy/058_deploy_ink_adapter.ts
pxrl, line 1: @james-a-morris Can you try deploying this with `056_deploy_op_adapter.ts` instead? Creating a new deployment script for each new deployment is an antipattern that we're trying to get away from.

The script would need a one-off tweak to revert to 0x0 for the USDC address on Ink; I think that's OK because upgradable USDC will ultimately be supported.

deploy/consts.ts
```diff
@@ -80,6 +80,10 @@ export const OP_STACK_ADDRESS_MAP: {
       L1CrossDomainMessenger: "0x5D4472f31Bd9385709ec61305AFc749F0fA8e9d0",
       L1StandardBridge: "0x697402166Fbf2F22E970df8a6486Ef171dbfc524",
     },
+    [CHAIN_IDs.INK]: {
+      L1CrossDomainMessenger: "0x69d3cf86b2bf1a9e99875b7e2d9b6a84426c171f",
+      L1StandardBridge: "0x88ff1e5b602916615391f55854588efcbb7663f0",
```
pxrl, line 85: If the OP adapter deployment script is used then we'd want to also define `L1OpUSDCBridgeAdapter` as 0x0 here.

contracts/chain-adapters/Ink_Adapter.sol
pxrl, line 1: Per other comments, I'd really like to stick to deploying the `OP_Adapter.sol` chain adapter for OP stack. As a rule we won't support non-Circle bridged USDC on OP stack chains, so I think it makes more sense to just use the OP adapter so it can easily be redeployed with the correct USDC and bridge addresses defined. This will save us from having to modify contracts on subsequent deployments.

contracts/Ink_SpokePool.sol
james-a-morris: Done: https://github.com/across-protocol/contracts/pull/808/commits/d217938773bd9b52e30d4965703966a0d33c6462

contracts/chain-adapters/Ink_Adapter.sol
pxrl, line 1: We should also be able to proceed without any custom Ink adapter, since `OP_Adapter.sol` is generic and should work for all new OP stack chains. This is part of the new chain automation - not having to copy/paste/modify the chain adapter reduces the risk of something going wrong and streamlines the process.

contracts/Ink_SpokePool.sol
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+ * @notice Ink Spoke pool.
+ * @custom:security-contact bugs@across.to
+ */
+contract Ink_SpokePool is Ovm_SpokePool {
+    using SafeERC20 for IERC20;
+
+    // Address of the custom L2 USDC bridge.
+    address private constant USDC_BRIDGE = 0xbD80b06d3dbD0801132c6689429aC09Ca6D27f82;
```
pxrl, line 17: nb. this is the WorldChain bridge address, so we'd probably want to make it obvious that this is not to be used (might set to 0x0?)

deploy/consts.ts
james-a-morris: https://github.com/across-protocol/contracts/pull/808/commits/8a97edf2a2f96c7527eb8993ec18018443b4015b

contracts/Ink_SpokePool.sol
james-a-morris: https://github.com/across-protocol/contracts/pull/808/commits/2399d8eba61cf4dced94aad2c7c191804ea594fe

deploy/058_deploy_ink_adapter.ts
james-a-morris: Done

contracts/Ink_SpokePool.sol
```diff
@@ -0,0 +1,72 @@
...
+        __OvmSpokePool_init(_initialDepositId, _crossDomainAdmin, _withdrawalRecipient, Lib_PredeployAddresses.OVM_ETH);
+    }
+
+    /**
+     * @notice Ink-specific logic to bridge tokens back to the hub pool contract on L1.
+     * @param amountToReturn Amount of the token to bridge back.
+     * @param l2TokenAddress Address of the l2 Token to bridge back. This token will either be bridged back to the token defined in the mapping `remoteL1Tokens`,
+     * or via the canonical mapping defined in the bridge contract retrieved from `tokenBridges`.
+     * @dev This implementation deviates slightly from `_bridgeTokensToHubPool` in the `Ovm_SpokePool` contract since World Chain has a USDC bridge which uses
+     * a custom interface. This is because the USDC token on World Chain is meant to be upgraded to a native, CCTP supported version in the future.
```
pxrl, line 57:
```suggestion
     * @dev This implementation deviates slightly from `_bridgeTokensToHubPool` in the `Ovm_SpokePool` contract since
     * this chain uses Circle's bridged (upgradable to native) USDC standard, which uses a custom interface.
```

contracts/Ink_SpokePool.sol
```diff
@@ -0,0 +1,72 @@
...
+     */
+    function _bridgeTokensToHubPool(uint256 amountToReturn, address l2TokenAddress) internal virtual override {
+        // Handle custom USDC bridge which doesn't conform to the standard bridge interface. In the future, CCTP may be used to bridge USDC to mainnet, in which
+        // case bridging logic is handled by the Ovm_SpokePool code. In the meantime, if CCTP is not enabled, then use the USDC bridge. Once CCTP is activated on
+        // WorldChain, this block of code will be unused.
```
pxrl, line 62:
```suggestion
        // Handle Circle's bridge for upgradable USDC, which doesn't conform to the standard OP bridge interface. In the future, CCTP may be used to bridge USDC to mainnet, in which
        // case bridging logic is handled by the Ovm_SpokePool code. In the meantime, if CCTP is not enabled, then use the USDC bridge. Once CCTP is activated on this chain, this block of code will be unused.
```

pxrl (approved):
LGTM - we can generalise the SpokePool comments in a follow-up PR.

bmzig (approved):
Should you also add the contract addresses to the README in `deployments/`? Also, will you follow up with deployments of the verifier/multicall handler?

pxrl (approved):

bmzig (approved):
This looks good to me. Are we not able to deploy the spoke pool verifier on this chain? Are we not doing it because the current version is unaudited?

james-a-morris:
> This looks good to me. Are we not able to deploy the spoke pool verifier on this chain? Are we not doing it because the current version is unaudited?

Yep but also because this is a low prio contract to deploy and this will begin to block Alex/myself on the rest of the deployments

nicholaspai (approved):

pxrl:
> > This looks good to me. Are we not able to deploy the spoke pool verifier on this chain? Are we not doing it because the current version is unaudited?
> 
> Yep but also because this is a low prio contract to deploy and this will begin to block Alex/myself on the rest of the deployments

I looked at the deployment process and found it's more convoluted now due to a merge conflict. I was able to make a deployment here: https://explorer.inkonchain.com/address/0xB4A8d45647445EA9FC3E1058096142390683dBC2?tab=contract

Here's a PR to add it: https://github.com/across-protocol/contracts/pull/814

PR #809, opened 2024-12-18
feat: Delete SwapAndBridge and add submission fees to gasless flow

nicholaspai:
SwapAndBridge is to be replaced with SpokePoolV3Periphery

Gasless flows will require user to cover gas cost of whoever submits the transaction, but they can be set to 0 if the user wants to submit themselves.


contracts/libraries/PeripherySigningLib.sol
```diff
@@ -78,6 +83,7 @@ library PeripherySigningLib {
             keccak256(
                 abi.encode(
                     EIP712_BASE_DEPOSIT_DATA_TYPEHASH,
+                    baseDepositData.inputToken,
```
nicholaspai, line 86: I think we were missing this in this typehash

bmzig: Ah we definitely were.

contracts/libraries/PeripherySigningLib.sol
```diff
@@ -35,18 +37,20 @@ library PeripherySigningLib {
         );
 
     // EIP712 Type hashes.
+    bytes32 internal constant EIP712_FEES_TYPEHASH = keccak256(EIP712_FEES_TYPE);
     bytes32 internal constant EIP712_BASE_DEPOSIT_DATA_TYPEHASH = keccak256(EIP712_BASE_DEPOSIT_DATA_TYPE);
     bytes32 internal constant EIP712_DEPOSIT_DATA_TYPEHASH =
-        keccak256(abi.encode(EIP712_DEPOSIT_DATA_TYPE, EIP712_BASE_DEPOSIT_DATA_TYPE));
+        keccak256(abi.encode(EIP712_DEPOSIT_DATA_TYPE, EIP712_FEES_TYPEHASH, EIP712_BASE_DEPOSIT_DATA_TYPE));
```
bmzig, line 43: I think these need to be `EIP712_FEES_TYPE` and not TYPEHASH.

contracts/SpokePoolV3Periphery.sol
```diff
@@ -706,6 +737,16 @@ contract SpokePoolV3Periphery is SpokePoolV3PeripheryInterface, Lockable, MultiC
         );
     }
 
+    function _paySubmissionFees(
+        address swapToken,
```
bmzig, line 741: nit: Do you want to call this `swapToken` since this is also called by the deposit entrypoints? Maybe just `feeToken`?

contracts/libraries/PeripherySigningLib.sol
```diff
@@ -35,18 +37,20 @@ library PeripherySigningLib {
...
 
     // EIP712 Type hashes.
+    bytes32 internal constant EIP712_FEES_TYPEHASH = keccak256(EIP712_FEES_TYPE);
     bytes32 internal constant EIP712_BASE_DEPOSIT_DATA_TYPEHASH = keccak256(EIP712_BASE_DEPOSIT_DATA_TYPE);
     bytes32 internal constant EIP712_DEPOSIT_DATA_TYPEHASH =
-        keccak256(abi.encode(EIP712_DEPOSIT_DATA_TYPE, EIP712_BASE_DEPOSIT_DATA_TYPE));
+        keccak256(abi.encode(EIP712_DEPOSIT_DATA_TYPE, EIP712_FEES_TYPE, EIP712_BASE_DEPOSIT_DATA_TYPE));
     bytes32 internal constant EIP712_SWAP_AND_DEPOSIT_DATA_TYPEHASH =
-        keccak256(abi.encode(EIP712_SWAP_AND_DEPOSIT_DATA_TYPE, EIP712_BASE_DEPOSIT_DATA_TYPE));
+        keccak256(abi.encode(EIP712_SWAP_AND_DEPOSIT_DATA_TYPE, EIP712_FEES_TYPEHASH, EIP712_BASE_DEPOSIT_DATA_TYPE));
```
bmzig, line 45: This also needs to be `TYPE` and not `TYPEHASH`

bmzig (approved):
One comment.

bmzig (approved):
LGTM

PR #810, opened 2024-12-19
feat(svm): Refactor deposit ID to use big-endian encoding in Bytes32 format

md0x:
Changes proposed in this PR:
- Updated deposit ID storage format to use big-endian encoding 
- Refactored byte manipulation logic to ensure consistent big-endian encoding and decoding for 32byte arrays

Motivation:

Previously, "safe" (`deposit_v3`) deposit IDs were stored in little-endian (LE) format, causing friction for relayers and off-chain tools due to format conversions. Challenges included:
1. Relayers had to convert BE (EVM standard) to LE, adding complexity.
2. Special handling was required to extract and parse IDs stored in LE format see [this](https://github.com/across-protocol/contracts/blob/25fc095c2d75da8922a220acfcea8c4dd5c9295f/src/svm/solanaProgramUtils.ts#L213-L228)
3. Diverged from EVM conventions, complicating cross-chain operations.
	
Solution:

This PR stores "safe" (`deposit_v3`) deposit IDs in BE format, storing `number_of_deposits` in the right-most bits. This aligns with EVM, simplifies cross-chain operations and reduces off-chain parsing complexity

Note: The “unsafe” deposit (`unsafe_deposit_v3`) does not require any changes and is unaffected by these challenges, as it uses a generated hash without endianness considerations.

### DIFF
--- a/programs/svm-spoke/src/instructions/deposit.rs
+++ b/programs/svm-spoke/src/instructions/deposit.rs
@@ -117,7 +117,7 @@ pub fn _deposit_v3(
     // If the passed in deposit_id is all zeros, then we use the state's number of deposits as deposit_id.
     if deposit_id == ZERO_DEPOSIT_ID {
         state.number_of_deposits += 1;
-        applied_deposit_id[..4].copy_from_slice(&state.number_of_deposits.to_le_bytes());
+        applied_deposit_id[28..].copy_from_slice(&state.number_of_deposits.to_be_bytes());
     }
 
     emit_cpi!(V3FundsDeposited {
--- a/src/svm/conversionUtils.ts
+++ b/src/svm/conversionUtils.ts
@@ -1,6 +1,6 @@
 import { BN } from "@coral-xyz/anchor";
 import { PublicKey } from "@solana/web3.js";
-import { ethers } from "ethers";
+import { BigNumber, ethers } from "ethers";
 
 /**
  * Converts an integer to a 32-byte Uint8Array.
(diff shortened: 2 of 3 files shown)


Reinis-FRP (approved):
indeed, this should be easier for offchain to work with

programs/svm-spoke/src/instructions/deposit.rs
```diff
@@ -117,7 +117,7 @@ pub fn _deposit_v3(
     // If the passed in deposit_id is all zeros, then we use the state's number of deposits as deposit_id.
     if deposit_id == ZERO_DEPOSIT_ID {
         state.number_of_deposits += 1;
-        applied_deposit_id[..4].copy_from_slice(&state.number_of_deposits.to_le_bytes());
+        applied_deposit_id[28..].copy_from_slice(&state.number_of_deposits.to_be_bytes());
```
chrismaree, line 120: this is great! I love how simple it is to get these structures to match what is expected here.

src/svm/conversionUtils.ts
```diff
@@ -1,33 +1,16 @@
 import { BN } from "@coral-xyz/anchor";
 import { PublicKey } from "@solana/web3.js";
-import { ethers } from "ethers";
+import { BigNumber, ethers } from "ethers";
 
 /**
  * Converts an integer to a 32-byte Uint8Array.
  */
 export function intToU8Array32(num: number | BN): number[] {
```
chrismaree, line 8: it might make sense at some point to have unit tests for these kinds of utils.

chrismaree (approved):

PR #811, opened 2024-12-19
fix: address LUT issues on tests

chrismaree:
addresses an issue we were seeing with unit tests after an anchor update. Now, all tests work.

reason this was needed is as follows:
```
Note that after you create or extend a lookup table, it must "warm up" for one slot before the lookup table or newly added addresses can be used in transactions. You can only access lookup tables and addresses added in slots prior to the current one.
```

Safari 2024-12-19 15 14 03

### DIFF
--- a/src/svm/transactionUtils.ts
+++ b/src/svm/transactionUtils.ts
@@ -39,7 +39,8 @@ export async function sendTransactionWithLookupTable(
 
   // Submit the ALT creation transaction
   await web3.sendAndConfirmTransaction(connection, new web3.Transaction().add(lookupTableInstruction), [sender], {
-    skipPreflight: true, // Avoids recent slot mismatch in simulation.
+    commitment: "confirmed",
+    skipPreflight: true,
   });
 
   // Extend the ALT with all accounts making sure not to exceed the maximum number of accounts per transaction.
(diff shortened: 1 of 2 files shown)


src/svm/transactionUtils.ts
```diff
@@ -51,13 +52,14 @@ export async function sendTransactionWithLookupTable(
...
-    await web3.sendAndConfirmTransaction(connection, new web3.Transaction().add(extendInstruction), [sender], {
-      skipPreflight: true, // Avoids recent slot mismatch in simulation.
-    });
+    await web3.sendAndConfirmTransaction(connection, new web3.Transaction().add(extendInstruction), [sender]);
   }
 
-  // Avoids invalid ALT index as ALT might not be active yet on the following tx.
-  await new Promise((resolve) => setTimeout(resolve, 1000));
+  // Wait for slot to advance. LUTs only active after slot advance.
+  const initialSlot = await connection.getSlot();
```
md0x, line 59: Nice finding!
Suggestion: Could we use something like this [connection.onSlotChange](https://solana-labs.github.io/solana-web3.js/classes/Connection.html#onslotchange) to subscribe to slot changes? Not sure how reliable onSlotChange is though

chrismaree: I tried this and this was better. not sure if onSlotChange was delayed or returned early but it did to result in the tests always working as expected, like this implementation does.

src/svm/transactionUtils.ts
```diff
@@ -51,13 +52,14 @@ export async function sendTransactionWithLookupTable(
       addresses: lookupAddresses.slice(i, i + maxExtendedAccounts),
     });
 
-    await web3.sendAndConfirmTransaction(connection, new web3.Transaction().add(extendInstruction), [sender], {
-      skipPreflight: true, // Avoids recent slot mismatch in simulation.
-    });
+    await web3.sendAndConfirmTransaction(connection, new web3.Transaction().add(extendInstruction), [sender]);
```
Reinis-FRP, line 55: shall we also set this to "confirmed", otherwise might default to "max"

chrismaree: sure & done.

Reinis-FRP (approved):

PR #820, opened 2024-12-23
improve(tests): silence `unsafeAllowDelegatecall` warning in tests

bmzig:
Applies the function introduced by OZ in this PR: https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/228. This is only used on two test fixtures, so we still should see warnings when we deploy.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@across-protocol/contracts",
-  "version": "3.0.19",
+  "version": "3.0.20",
   "author": "UMA Team",
   "license": "AGPL-3.0-only",
   "repository": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@across-protocol/contracts",
-  "version": "3.0.20",
+  "version": "3.0.19",
   "author": "UMA Team",
   "license": "AGPL-3.0-only",
   "repository": {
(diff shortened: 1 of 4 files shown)


james-a-morris (commented):
OOC will this potentially silence other warnings?

bmzig:
> OOC will this potentially silence other warnings?

It would if there were other warnings thrown in those fixtures.

nicholaspai (approved):
fixtures by design are only used in tests so this is nice. Is there  way to target specific warnings?

bmzig:
> fixtures by design are only used in tests so this is nice. Is there way to target specific warnings?

I don't see a way to do this. I was referencing this post: https://forum.openzeppelin.com/t/logging-verbose-during-local-tests-with-upgradeable-contracts/4633, which seemed to have similar issues as ours. I changed this to just instead silence warnings for our contract tests only. The upside is that we should have a clean output for our tests, but the downside is that any warnings emitted from `@openzeppelin/hardhat-upgrades` will be silenced (e.g. the `unsafeAllow` options found [here](https://docs.openzeppelin.com/upgrades-plugins/1.x/api-hardhat-upgrades)) I think this is OK, since we still will need to explicitly supply the `unsafeAllow` type when deploying the contract.

james-a-morris (dismissed):
Curious about Nick's comment. This is what I was probing at. If we can't and this looks good I will approve so that I'm not a blocker

bmzig:
I reverted the package bump since this I don't think this is significant enough to merit a version change. We can just include it whenever we need to upgrade contracts. This should stop the spamming in the relayer repository's unit tests.

As far as silencing specific warnings, I don't think this is possible, but it shouldn't matter, since we know that the warnings will only be emitted if we have an `unsafeAllow` option somewhere in the code. Essentially, we need to already acknowledge and mark that the warning is allowed in the contract for it to be silenced.

james-a-morris (approved):

PR #826, opened 2025-01-02
fix(evm): C01 - Address incorrect use of relayerRefund over msg.sender in claimRelayerRefund function

chrismaree:
OZ identified the following issue in the contracts:

```
The claimRelayerRefund function is meant to give relayers the option to claim outstanding repayments. This can happen in different edge cases, like blacklists not allowing token transfers. In such cases, the relayer can call this function and specify a different refundAddress to claim their funds.

The function first reads the current outstanding refund from the relayerRefund mapping using the l2TokenAddress and msg.sender keys. If this value is greater than 0 then it is transferred to the refundAddress and the appropriate event is emitted.

Before transferring out the tokens, the mapping value is set to 0 for correct accounting. However, the key used to reset the mapping value is refundAddress and not msg.sender. This opens the door for any relayer with a small refund amount to zero out any other relayer's refund by specifying their refundAddress, effectively making the relayers lose their refunds. In addition, since the original mapping value is never reset, a malicious relayer can exploit this by repeatedly calling the function to drain the entire balance of the l2TokenAddress contract.

Consider using the proper msg.sender key instead of refundAddress to correctly set the refund amount to zero.
```

This PR addresses this by moving to using `msg.sender` within this method, as requested.


mrice32 (approved):
LGTM!

Reinis-FRP (approved):

md0x (approved):

PR #833, opened 2025-01-03
improve(hardhat): Autogenerate network config

pxrl:
Network config can be derived from the common network definitions that already exist upstream in the constants repository. This saves time when adding new chains and reduces duplication.

### DIFF
--- a/hardhat.config.ts
+++ b/hardhat.config.ts
@@ -1,8 +1,7 @@
 import * as dotenv from "dotenv";
-
 import { HardhatUserConfig } from "hardhat/config";
 import { getNodeUrl, getMnemonic } from "@uma/common";
-import { CHAIN_IDs } from "./utils/constants";
+import { CHAIN_IDs, PUBLIC_NETWORKS, MAINNET_CHAIN_IDs } from "./utils/constants";
 
 import "@nomicfoundation/hardhat-verify"; // Must be above hardhat-upgrades
 import "@nomiclabs/hardhat-waffle";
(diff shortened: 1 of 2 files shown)


hardhat.config.ts
```diff
@@ -62,6 +61,33 @@ const DEFAULT_CONTRACT_COMPILER_SETTINGS = {
...
 
+const networks = Object.fromEntries(
+  Object.values(CHAIN_IDs)
+    .filter((chainId) => PUBLIC_NETWORKS[chainId] !== undefined)
+    .map((chainId) => {
+      const network = PUBLIC_NETWORKS[chainId];
+      const name = network.name.replace(" ", "").toLowerCase();
+      const hubChainId = Object.values(MAINNET_CHAIN_IDs).includes(chainId) ? CHAIN_IDs.MAINNET : CHAIN_IDs.SEPOLIA;
+      const chainDef = {
+        url: network.publicRPC,
```
pxrl, line 72: nb. depends on this PR: https://github.com/across-protocol/constants/pull/88

hardhat.config.ts
```diff
@@ -62,6 +61,33 @@ const DEFAULT_CONTRACT_COMPILER_SETTINGS = {
...
+        accounts: { mnemonic },
+        saveDeployments: true,
+        chainId: hubChainId,
+        companionNetworks: { l1: hubChainId === CHAIN_IDs.MAINNET ? "mainnet" : "sepolia" },
+      };
+
+      // zkSync is a special snowflake.
+      if ([CHAIN_IDs.ZK_SYNC, CHAIN_IDs.ZK_SYNC_SEPOLIA].includes(chainId)) {
```
pxrl, line 80: Would need to add Lens to this list. zkSync tooling is really annoying :\

hardhat.config.ts
```diff
@@ -104,204 +130,7 @@ const config: HardhatUserConfig = {
...
-      chainId: CHAIN_IDs.ALEPH_ZERO,
-      url: "https://rpc.alephzero.raas.gelato.cloud",
-      saveDeployments: true,
-      accounts: { mnemonic },
-      companionNetworks: { l1: "mainnet" },
-    },
+    ...networks, // autogenerated
   },
   gasReporter: { enabled: process.env.REPORT_GAS !== undefined, currency: "USD" },
   etherscan: {
```
pxrl, line 136: I'll be interested to see if we can autogenerate some of this config as well.

hardhat.config.ts
```diff
@@ -62,6 +61,33 @@ const DEFAULT_CONTRACT_COMPILER_SETTINGS = {
   },
 };
 
+const networks = Object.fromEntries(
+  Object.values(CHAIN_IDs)
+    .filter((chainId) => PUBLIC_NETWORKS[chainId] !== undefined)
+    .map((chainId) => {
+      const network = PUBLIC_NETWORKS[chainId];
+      const name = network.name.replace(" ", "").toLowerCase();
```
pxrl, line 69: nb. might need to add some special handling for the `-sepolia` chains.

pxrl: In https://github.com/across-protocol/contracts/pull/833/commits/61f2d29f02457163c926b10ee0ddbd97a65a99f8 I reverted to using the chainId as the network key, rather than a symbolic name. The problem with symbolic names is that we have no consistent naming scheme for the various networks. Chain IDs are at least universally consistent and don't impose decisions like whether to hyphenate spaces or not.

hardhat.config.ts
pxrl: https://github.com/across-protocol/contracts/pull/833/commits/ec9f8e86f936d2f54fb69b979b65ca04bb69f56d

PR #834, opened 2025-01-06
fix(evm): L02 _destinationSettler Can Return Zero Address

mrice32:
OZ identified the following issue:
```In the _resolveFor function of the ERC7683OrderDepositor contract, the value returned from the internal _destinationSettler function is [not validated](https://github.com/across-protocol/contracts/blob/401e24ccca1b3af919dd521e58acd445297b65b6/contracts/erc7683/ERC7683OrderDepositor.sol#L245) to be non-zero. The _destinationSettler function returns the settler contract for the destination chain. If this value has not been set, the default value will be returned and passed to the [fillInstructions](https://github.com/across-protocol/contracts/blob/401e24ccca1b3af919dd521e58acd445297b65b6/contracts/erc7683/ERC7683OrderDepositor.sol#L245) field of the resolvedOrder parameter.

Consider adding the requirement that the _destinationSettler value used in the FillInstruction struct of the resolved order cannot be the zero address.
```

This PR addresses this using the requested change.


chrismaree (approved):

nicholaspai:
accidentally closed this, replaced here https://github.com/across-protocol/contracts/pull/886

PR #839, opened 2025-01-06
feat(svm): L-06 Unnecessary Use of InitSpace Macro

md0x:
Changes proposed in this PR:
- Fix audit issue:
```
Throughout the codebase, the following structs use the InitSpace macro unnecessarily:

FillV3RelayParams

ExecuteRelayerRefundLeafParams

RequestV3SlowFillParams

ExecuteV3SlowRelayLeafParams

These accounts are initialized using the initialize_instruction_params method, which creates empty accounts with a size specified in the function's instructions. Consequently, the InitSpace macro, which implements the Space trait, becomes redundant as the INIT_SPACE constant generated by this trait is not utilized. The inclusion of the InitSpace macro in these cases adds unnecessary overhead and obscures the intended logic, given that its functionality is not employed.

Consider removing the InitSpace macro from these structs to simplify the codebase and remove superfluous logic. This adjustment would help improve code clarity and align it more closely with actual use patterns.
```

### DIFF
--- a/programs/svm-spoke/src/common/v3_relay_data.rs
+++ b/programs/svm-spoke/src/common/v3_relay_data.rs
@@ -1,6 +1,6 @@
 use anchor_lang::prelude::*;
 
-#[derive(AnchorSerialize, AnchorDeserialize, Clone, InitSpace)]
+#[derive(AnchorSerialize, AnchorDeserialize, Clone)]
 pub struct V3RelayData {
     pub depositor: Pubkey,
     pub recipient: Pubkey,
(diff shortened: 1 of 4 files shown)


Reinis-FRP (commented):
we might be also able to remove `InitSpace` from all nested structs

md0x:
> we might be also able to remove `InitSpace` from all nested structs

@Reinis-FRP, I’ve removed some more InitSpace instances. Let me know if you were referring to those.

chrismaree (approved):

Reinis-FRP (approved):

PR #844, opened 2025-01-07
feat(svm): N-05 Ambiguous Parameter Usage for fill_deadline

md0x:
**Changes proposed in this PR:**
- Fixes the following issue identified by Open Zeppelin:

> The [deposit_v3](https://github.com/across-protocol/contracts/blob/0f2600f83e8a5738d0d62a78b05ff37adda4c1c9/programs/svm-spoke/src/lib.rs#L250) and [deposit_v3_now](https://github.com/across-protocol/contracts/blob/0f2600f83e8a5738d0d62a78b05ff37adda4c1c9/programs/svm-spoke/src/lib.rs#L283) functions use the same parameter name, fill_deadline, with differing semantics. In deposit_v3, it is an absolute timestamp, whereas in deposit_v3_now, it is an offset added to the current time.
> 
> The documentation lacks clarity about this distinction, increasing the likelihood of errors. Users might mistakenly pass an absolute timestamp to deposit_v3_now, leading to a computed deadline (current_time + fill_deadline_offset) that exceeds current_time + state.fill_deadline_buffer. This results in [failed transactions](https://github.com/across-protocol/contracts/blob/0f2600f83e8a5738d0d62a78b05ff37adda4c1c9/programs/svm-spoke/src/instructions/deposit.rs#L90-L92).
> 
> Consider renaming fill_deadline in deposit_v3_now to fill_deadline_offset for clarity. Alternatively, consider improving the documentation to explicitly state the difference in semantics, including examples.

### DIFF
--- a/programs/svm-spoke/src/lib.rs
+++ b/programs/svm-spoke/src/lib.rs
@@ -290,7 +290,7 @@ pub mod svm_spoke {
         output_amount: u64,
         destination_chain_id: u64,
         exclusive_relayer: Pubkey,
-        fill_deadline: u32,
+        fill_deadline_offset: u32,
         exclusivity_parameter: u32,
         message: Vec<u8>,
     ) -> Result<()> {
@@ -304,7 +304,7 @@ pub mod svm_spoke {
             output_amount,
             destination_chain_id,
             exclusive_relayer,
-            fill_deadline,
+            fill_deadline_offset,
             exclusivity_parameter,
             message,
         )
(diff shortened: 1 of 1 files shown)


chrismaree (commented):
you'll need to update the comment block above for `fill_deadline`

md0x:
> you'll need to update the comment block above for `fill_deadline`

Comment updated

chrismaree (approved):

PR #845, opened 2025-01-07
fix(svm): N-03 remove duplicate logic

Reinis-FRP:

> The [claim_relayer_refund](https://github.com/across-protocol/contracts/blob/401e24ccca1b3af919dd521e58acd445297b65b6/programs/svm-spoke/src/instructions/refund_claims.rs#L76) and [claim_relayer_refund_for](https://github.com/across-protocol/contracts/blob/401e24ccca1b3af919dd521e58acd445297b65b6/programs/svm-spoke/src/instructions/refund_claims.rs#L152) functions share almost identical logic, with the primary distinction being the refund address. This duplication introduces maintainability challenges, as future updates may need to be applied to both functions simultaneously, increasing the risk of inconsistencies.
> 
> To streamline the code and reduce redundancy, consider implementing the following refactor:
>
> Abstract the common logic from both functions into a helper function that handles the transfer and event emission. This helper could accept parameters such as the target token account and any additional context-specific data.
Alternatively, consolidate the instructions into a single function that takes an optional refund_address parameter, enabling flexible use cases while maintaining a single code path.
Implementing either of these approaches would simplify future maintenance and reduce the surface area for potential bugs.

This PR addresses the issue by consolidating the instructions into a single function that takes additional `refund_address` account. When it is different from the `signer`, the receiving `token_account` must match ATA for this `refund_address`. Otherwise it can be any custom token account for the given mint and token program as the claim account would be derived from the `signer`.

### DIFF
--- a/programs/svm-spoke/src/constraints.rs
+++ b/programs/svm-spoke/src/constraints.rs
@@ -1,4 +1,8 @@
 use anchor_lang::prelude::*;
+use anchor_spl::{
+    associated_token::get_associated_token_address_with_program_id,
+    token_interface::{Mint, TokenAccount, TokenInterface},
+};
 
 use crate::{
     state::State,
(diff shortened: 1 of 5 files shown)


md0x (dismissed):
Nice! Beautiful refactoring.

md0x (approved):

PR #847, opened 2025-01-08
fix(svm): H-01 optimize claiming relayer refunds

Reinis-FRP:

> When a relayer within a transaction leaf cannot receive direct refunds, such as due to a failed token transfer caused by blacklisting, the protocol uses claim account PDAs to handle the refund. However, this fallback mechanism imposes limitations on other relayers within the same transaction, forcing all of them to claim refunds via a [ClaimAccount](https://github.com/across-protocol/contracts/blob/401e24ccca1b3af919dd521e58acd445297b65b6/programs/svm-spoke/src/state/refund_account.rs#L5) PDA, even if direct refunds are possible for them. This leads to the following issues:
>
> - Risk of Hitting Compute Budget: The [accrue_relayer_refunds](https://github.com/across-protocol/contracts/blob/401e24ccca1b3af919dd521e58acd445297b65b6/programs/svm-spoke/src/instructions/bundle.rs#L219) process involves find_program_address, which can be computationally expensive as it iteratively attempts to validate a valid bump. Since this process can be used for several accounts, there is a high likelihood that the transaction will revert. Consequently, relayers and users may not be able to retrieve their refunds from that leaf.
> - Refund Delays: A relayer might claim their refund from its ClaimAccount PDA right before the execution of a leaf, reducing the vault’s total available funds. This could delay refunds for other users as the protocol [might lack sufficient liquidity at the time of execution](https://github.com/across-protocol/contracts/blob/401e24ccca1b3af919dd521e58acd445297b65b6/programs/svm-spoke/src/instructions/bundle.rs#L142).
> - Increased Execution Overhead: Executors are responsible for ensuring that all relayers have initialized claim accounts and are incurring additional computational and financial overhead, including the cost of covering rent exemption fees.
> - Liquidity-Related Reverts: Relayers can [attempt to claim refunds](https://github.com/across-protocol/contracts/blob/401e24ccca1b3af919dd521e58acd445297b65b6/programs/svm-spoke/src/instructions/refund_claims.rs#L76) when the spoke pool lacks sufficient liquidity. This scenario might result in transaction reversion without providing a meaningful error message, as the hub pool does not track outstanding spoke pool debts.
> - Prone to misuse: Since the function can be invoked by anyone, malicious actors could front-run the data worker, unnecessarily deferring refunds even when direct refund mechanisms are operational. In addition, a malicious relayer might force the use of claim accounts and delay its refund claim until just before the execution of a leaf, deliberately triggering a revert to disrupt system operations.
>
> Consider implementing logic to route refunds only through ClaimAccount PDAs for relayers that are explicitly unable to receive direct transfers. This would optimize operations, prevent errors, and reduce overhead. Moreover, consider enhancing error messaging to clearly indicate insufficient liquidity in the spoke pool during claim attempts. This change would ensure smoother operations and improved clarity for users.

We have opted to apply only a limited fix by optimizing the way how relayer refund claim instructions are composed by moving its parameters to account data.


test/svm/SvmSpoke.Bundle.ts
```diff
@@ -1732,4 +1732,161 @@ describe("svm_spoke.bundle", () => {
...
+          initializer: owner,
+          state,
+          vault,
+          mint,
+          tokenAccount,
+          claimAccount,
+          tokenProgram: TOKEN_PROGRAM_ID,
+          program: program.programId,
+        };
+        // Will not claim for the first relayer as if its token account was blacklisted.
```
md0x, line 1778: Out of curiosity, why do we simulate the first relayer being blacklisted?

md0x (dismissed):
Looks good!

test/svm/SvmSpoke.Bundle.ts
Reinis-FRP: agree, this was rather arbitrary. now simplified the tests by claiming to all accounts

test/svm/SvmSpoke.Bundle.ts
```diff
@@ -1732,4 +1732,220 @@ describe("svm_spoke.bundle", () => {
...
+
+    it("Execute Max multiple refunds with claims in separate phase legacy transactions", async () => {
+      // Larger amount would hit transaction message size limit.
+      const solanaDistributions = 7;
+      await executeMaxRefundClaims({ solanaDistributions, useAddressLookup: false, separatePhases: true });
+    });
+
+    it("Execute Max multiple refunds with claims in separate phase versioned transactions", async () => {
+      // Larger amount would hit maximum instruction trace length limit.
+      const solanaDistributions = 21;
```
md0x, line 1947: Nice improvement!

test/svm/SvmSpoke.Bundle.ts
```diff
@@ -1742,4 +1742,220 @@ describe("svm_spoke.bundle", () => {
...
+          vault,
+          mint,
+          tokenAccount,
+          refundAddress: tokenOwner,
+          claimAccount,
+          tokenProgram: TOKEN_PROGRAM_ID,
+          program: program.programId,
+        };
+        claimInstructions.push(
+          await program.methods.claimRelayerRefundFor().accounts(claimRelayerRefundAccounts).instruction()
```
md0x, line 1796: This one is remaining

chrismaree (approved):

md0x (approved):

PR #850, opened 2025-01-08
feat: lens testnet

bmzig:
Spoke Pool: https://block-explorer.testnet.lens.dev/address/0x6A0a7f39530923911832Dd60667CE5da5449967B#contract
Adapter: https://sepolia.etherscan.io/address/0x8fac6f764ae0b4f632fe2e6c938ed5637e629ff2
Donation Box: https://sepolia.etherscan.io/address/0x74f00724075443Cbbf55129F17CbAB0F77bA0722#code
MulticallHandler: https://block-explorer.testnet.lens.dev/address/0x02D2B95F631E0CF6c203E77f827381B0885F7822#contract


deploy/016_deploy_zkstack_spokepool.ts
```diff
@@ -38,6 +38,7 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
     newAddress = _deployment.address;
     console.log(`New ${contractName} implementation deployed @ ${newAddress}`);
   } else {
+    console.log(constructorArgs);
```
bmzig, line 41:
```suggestion
```

deploy/016_deploy_zkstack_spokepool.ts
bmzig, line 1: I renamed this from "zksync" to "zkstack" since I am assuming that the l2 erc20 bridge works the same as in zksync. The native token is `GRASS`, but it has the [same contract address](https://block-explorer.testnet.lens.dev/address/0x000000000000000000000000000000000000800A#transactions) as we have hardcoded in the [spoke pool](https://github.com/across-protocol/contracts/blob/86f17ee570a8b3005e65e11ea9a2fcec25c716b5/contracts/ZkSync_SpokePool.sol#L61), so I'm assuming that works fine, too. We may need to follow-up with a generalization to the zksync spoke pool if withdrawals do not work.

bmzig: Was able to test that the zksync spoke pool withdrawal logic works as expected with this test [spoke pool](https://block-explorer.testnet.lens.dev/address/0xb60A2Dc4bF1e75dB8B6a1e35f954d69Cf37BCD25#transactions).
[Here](https://block-explorer.testnet.lens.dev/tx/0xdd08ff8d8e6815ccd35af98139fb84e10e5dd8c5cb1430d18ae6cca41453f639) is a call to `_bridgeTokensToHubPool` which withdraws grass to l1, and [here](https://block-explorer.testnet.lens.dev/tx/0xe03ce2e136ebad281ed232e9af2ff8d1eb918cbe90cc8ff1ba18019ce42ec406) is a call which withdraws weth to L1.

nicholaspai (approved):

PR #853, opened 2025-01-10
fix(svm): format docstrings

Reinis-FRP:
Anchor IDL generates docs from lines starting with tripple-slashes (`///`). Some of comments had these unnecessarily causing the IDL docs including information that is not specific to a given instruction or account. This fixes by formatting these lines as regular comments starting with double-slash (`//`).


programs/svm-spoke/src/instructions/token_bridge.rs
```diff
@@ -40,28 +40,28 @@ pub struct BridgeTokensToHubPool<'info> {
     pub token_messenger_minter_sender_authority: UncheckedAccount<'info>,
 
     /// CHECK: MessageTransmitter is checked in CCTP. Seeds must be \["message_transmitter"\] (CCTP Message Transmitter
-    /// program).
+    // program).
```
md0x, line 43: Turning these into double slashes doesn't remove them from docs?

Reinis-FRP: it should remove them from the docs, aren't they?

md0x: Okay, yes, before your change, Anchor was adding program). to the documentation.
Now, it doesn’t add documentation for this account (expected behavior).

Before your fix:
```
        {
          "name": "message_transmitter",
          "docs": [
            "program)."
          ],
          "writable": true
        },
```

After fix:
```
        {
          "name": "message_transmitter",
          "writable": true
        },
```

All good!

md0x (approved):

chrismaree (dismissed):

md0x (approved):

PR #856, opened 2025-01-12
feat(cher): deploy cher contracts

amateima:
All deployments made against origin commit hash 7f9ebb360c9bfcc9960a0226a92098fccc186ce8. 
Rebased against master with all deployments made.

⚠️ `SpokePoolVerifier` contract not deployed yet as it's not clear for me yet what commit hash to use

⚠️ I used a code name for the name of the chain. RPCs and explorer URLs need to be private until public launch


james-a-morris (commented):
I don't think we'll be able to set the xchain contracts with this unless we intend to create a secondary multisig call at a later point with different names

james-a-morris (approved):

deployments/README.md
```diff
@@ -137,3 +138,10 @@ This is because this `deployments.json` file is used by bots in [`@across-protoc
 | ---------------- | -------------------------------------------------------------------------------------------------------------------------------- |
 | Ink_SpokePool    | [0xeF684C38F94F48775959ECf2012D7E864ffb9dd4](https://explorer.inkonchain.com/address/0xeF684C38F94F48775959ECf2012D7E864ffb9dd4) |
 | MulticallHandler | [0x924a9f036260DdD5808007E1AA95f08eD08aA569](https://explorer.inkonchain.com/address/0x924a9f036260DdD5808007E1AA95f08eD08aA569) |
+
+## Cher mainnet (1868)
+
+| Contract Name    | Address                                                                                                               |
+| ---------------- | --------------------------------------------------------------------------------------------------------------------- |
+| Cher_SpokePool   | [0x3baD7AD0728f9917d1Bf08af5782dCbD516cDd96](https://etherscan.io/address/0x3baD7AD0728f9917d1Bf08af5782dCbD516cDd96) |
```
bmzig, line 146: To be clear, we are doing this since we don't want to expose the block explorer?

bmzig (approved):

deployments/README.md
amateima: yes. Will update the urls today/tomorrow

PR #860, opened 2025-01-14
feat(svm): include solana artifacts in package build output (#806)

md0x:
Changes proposed in this PR:
1. Reintroduces the updates from PR #806, which were previously merged
- These changes were reverted due to a build issue where JSON IDL files were not being included in the dist folder.
2. Fixes the build process to ensure JSON IDL files are correctly included in the dist folder.


deployments/deployments.json
```diff
@@ -173,5 +173,10 @@
   "1868": {
     "SpokePool": { "address": "0x3baD7AD0728f9917d1Bf08af5782dCbD516cDd96", "blockNumber": 1709997 },
     "MulticallHandler": { "address": "0x924a9f036260DdD5808007E1AA95f08eD08aA569", "blockNumber": 1711482 }
+  },
+  "1810017368444177321": {
```
Reinis-FRP, line 177: this looks like lowest 8 bytes from `keccak("solana-devnet")`, didn't we intend to use 6 bytes to avoid overflows?

md0x: Good catch.
That was the previous deployment before we shortened the identifier. I will remove it and update it once we deploy the latest version.

Reinis-FRP (approved):

PR #867, opened 2025-01-20
fix: use entire message when calculating relay hash for evm chains

bmzig:
If we have two separate methods to calculate the relay hash, separated by whether the L2's spoke pool is updated or not, then relayers may produce two valid deposits for the same fill, or even worse, a slow fill may be executed for a deposit which has already been filled. To mitigate this, we should consider not changing the way we calculate relay hashes for evm chains.

### DIFF
--- a/contracts/SpokePool.sol
+++ b/contracts/SpokePool.sol
@@ -1011,7 +1011,7 @@ abstract contract SpokePool is
 
         V3RelayExecutionParams memory relayExecution = V3RelayExecutionParams({
             relay: relayData,
-            relayHash: _getV3RelayHash(relayData),
+            relayHash: getV3RelayHash(relayData),
             updatedOutputAmount: relayData.outputAmount,
             updatedRecipient: relayData.recipient,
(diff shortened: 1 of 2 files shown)


contracts/SpokePool.sol
```diff
@@ -1641,24 +1641,7 @@ abstract contract SpokePool is
     }
 
     function _getV3RelayHash(V3RelayData memory relayData) private view returns (bytes32) {
```
pxrl, line 1643: Should we consider making this public? That would permit people to throw a relayData object at this function and get a SpokePool-derived hash back. This could be used by integrators to verify their local hash derivation implementations.

bmzig: I don't see a downside either. Done here: https://github.com/across-protocol/contracts/pull/867/commits/fee01d917109ffa0d8a9120fddea6ff29bb4f115

pxrl (approved):

mrice32 (approved):

PR #870, opened 2025-01-21
feat(SpokePool): Permit historical fillDeadline on deposit

pxrl:
This removes a sharp edge for pre-fill deposits, where the deposit comes after the fill. Permitting a historical fillDeadline gives more flexibility to the relayer around when they submit the deposit on the origin chain.

### DIFF
--- a/contracts/SpokePool.sol
+++ b/contracts/SpokePool.sol
@@ -488,8 +488,8 @@ abstract contract SpokePool is
      *  This must be set to some time between [currentTime - depositQuoteTimeBuffer, currentTime]
      * where currentTime is block.timestamp on this chain or this transaction will revert.
      * @param fillDeadline The deadline for the relayer to fill the deposit. After this destination chain timestamp,
-     * the fill will revert on the destination chain. Must be set between [currentTime, currentTime + fillDeadlineBuffer]
-     * where currentTime is block.timestamp on this chain or this transaction will revert.
+     * the fill will revert on the destination chain. Must be set before currentTime + fillDeadlineBuffer, where
+     * currentTime is block.timestamp on this chain or this transaction will revert.
      * @param exclusivityParameter This value is used to set the exclusivity deadline timestamp in the emitted deposit
      * event. Before this destination chain timestamp, only the exclusiveRelayer (if set to a non-zero address),
      * can fill this deposit. There are three ways to use this parameter:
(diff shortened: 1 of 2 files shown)


contracts/SpokePool.sol
```diff
@@ -1339,12 +1339,7 @@ abstract contract SpokePool is
         // fillDeadline is relative to the destination chain.
         // Don’t allow fillDeadline to be more than several bundles into the future.
         // This limits the maximum required lookback for dataworker and relayer instances.
-        // Also, don't allow fillDeadline to be in the past. This poses a potential UX issue if the destination
-        // chain time keeping and this chain's time keeping are out of sync but is not really a practical hurdle
-        // unless they are significantly out of sync or the depositor is setting very short fill deadlines. This latter
-        // situation won't be a problem for honest users.
-        if (params.fillDeadline < currentTime || params.fillDeadline > currentTime + fillDeadlineBuffer)
-            revert InvalidFillDeadline();
+        if (params.fillDeadline > currentTime + fillDeadlineBuffer) revert InvalidFillDeadline();
```
nicholaspai, line 1342: Is there anywhere in the offchain stack that would break if fill deadline was 0?

nicholaspai (approved):

mrice32 (approved):
LGTM! I second @nicholaspai's question.

contracts/SpokePool.sol
pxrl: The BundleDataClient already handles fillDeadline 0 OK and explicit notehat it's intentional: https://github.com/across-protocol/sdk/blob/db244fbaec15dc049452cd104ca7bec5856ad3da/src/clients/BundleDataClient/BundleDataClient.ts#L815

There are a couple of places in the BundleDataClient where we'll need updates to handle pre-fills, but it's mostly tangential to this SpokePool change. I left comments on what we need to verify in this PR: https://github.com/across-protocol/sdk/pull/834

bmzig:
Just to be clear here, we are _not_ going to allow fillDeadlines of 0. I would vote for this, since it would give more information to the dataworker over a fill deadline of 0, but am not necessarily opposed to it if we think it would really help relayers.

bmzig (approved):

contracts/SpokePool.sol
nicholaspai, line 1342: I just found one case where we assume fill deadline is not 0. Now, I don't think this is an issue but I'm gonna block this PR until we review more: https://github.com/across-protocol/sdk/blob/db244fbaec15dc049452cd104ca7bec5856ad3da/src/clients/BundleDataClient/BundleDataClient.ts#L817

nicholaspai (changes requested):
I think there are some issues where the deposit is sent with a very low fill deadline: https://github.com/across-protocol/contracts/pull/870/files#r1926100789

mrice32:
> I think this means that there is no way someone could get refunded twice for a deposit expiry, right? It's trivial to set a fillDeadline low enough to get refunded in the current bundle, but there is no follow up bundle where that deposit will "newly" expire and a duplicate refund will get sent.

I agree with your logic. (deposit ∈ { this block range } AND expiry ∈ { this block range, past block range }) OR (deposit ∈ { this block range, past block range } AND expiry ∈ { this block range }).

Each of these OR'd conditions requires one of these things (deposit or expiry) to be in the current block range, while the other could be at any past block range. This condenses down to exactly one block range where a refund can happen, which is the _latest_ block range that at least one of them falls in.

PR #898, opened 2025-02-20
feat(svm): fix simple fill script

md0x:
Changes proposed in this PR:
- Fix simple fill script

### DIFF
--- a/scripts/svm/simpleFill.ts
+++ b/scripts/svm/simpleFill.ts
@@ -11,16 +11,10 @@ import {
   getMint,
   getOrCreateAssociatedTokenAccount,
 } from "@solana/spl-token";
-import { AccountMeta, PublicKey, SystemProgram, Transaction, sendAndConfirmTransaction } from "@solana/web3.js";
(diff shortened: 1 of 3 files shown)


scripts/svm/simpleFill.ts
```diff
@@ -145,24 +162,38 @@ async function fillRelay(): Promise<void> {
...
-    })
+  const fillDataValues: FillDataValues = [Array.from(relayHashUint8Array), relayData, chainId, signer.publicKey];
+
+  const handlerProgram = getMulticallHandlerProgram(provider);
+  const multicallHandlerCoder = new MulticallHandlerCoder([]);
+
+  const fillRemainingAccounts: AccountMeta[] = [
+    { pubkey: handlerProgram.programId, isSigner: false, isWritable: false },
+    ...multicallHandlerCoder.compiledKeyMetas,
+  ];
```
Reinis-FRP, line 173: Why do we need remaining accounts in the simple fill where message is empty?

Reinis-FRP (approved):

PR #902, opened 2025-02-25
Add `OFTTransportAdapter` to support cross-chain token transfers of `USDT0` via `OFT` messaging protocol

grasphoper:
- OFT standard: https://docs.layerzero.network/v2/home/token-standards/oft-standard
- currently, `OFTTransportAdapter` is created to be in 1-to-1 relationship with `usdt`, as this PR's most pressing issue is solving `USDT0` bridging

Closes ACX-3761, ACX-3762, ACX-3774


grasphoper:
Please review the code and provide suggestions.
There's an issue with contract sizes that's yet to be fully resolved. I can't really fully resolve it until I get contracts compiling on M4 properly

grasphoper:
2 points:
- CI is failing because of insufficient number of args provided in testing. Not sure where to set those
- there's a `package.json` change in this PR: `prepublish` -> `prepublishOnly`. Made it more intuitive for me to work with package management. I'd change it, but maybe I don't know all the implications

pxrl:
> 2 points:
> 
> * CI is failing because of insufficient number of args provided in testing. Not sure where to set those

See here: https://github.com/across-protocol/contracts/blob/8ef0a3a3e71aa9e7f9a4940c118873dee3b6b673/test/evm/hardhat/chain-adapters/Arbitrum_Adapter.ts#L65

> * there's a `package.json` change in this PR: `prepublish` -> `prepublishOnly`. Made it more intuitive for me to work with package management. I'd change it, but maybe I don't know all the implications

This just changes the target name for the yarn script, right?

contracts/libraries/OFTTransportAdapter.sol
```diff
@@ -0,0 +1,91 @@
...
+            dstEid,
+            to,
+            amountLD,
+            minAmountLD,
+            EMPTY_MSG_BYTES,
+            EMPTY_MSG_BYTES,
+            EMPTY_MSG_BYTES
+        );
+
+        MessagingFee memory fee = oftMessenger.quoteSend(sendParam, false);
```
nicholaspai, line 77: Why is false set as second param? Should add a comment

contracts/libraries/OFTTransportAdapter.sol
```diff
@@ -0,0 +1,91 @@
...
+
+        // todo: here, we're relying on `oftMessenger` to use the full allowance. Otherwise, next .safeApprove will revert.
+        usdt.safeApprove(address(this), _amount);
```
nicholaspai, line 80: Can't we check the .send() return value and make sure it uses exactly what we expect?

contracts/libraries/OFTTransportAdapter.sol
```diff
@@ -0,0 +1,91 @@
...
+        usdt = _usdt;
+        oftMessenger = _oftMessenger;
+        dstEid = _dstEid;
+    }
+
+    /**
+     * @notice Returns whether or not the OFT bridge is enabled.
+     * @dev If the IOFT is the zero address, OFT bridging is disabled.
+     */
+    function _isOFTEnabled(address _token) internal view returns (bool) {
```
pxrl, line 49: The USDT address is actually only used within this function, but it makes the implementation 1:1 coupled to a specific token. If we can store supported tokens in a mapping, or implement that check externally to this adapter (as is done for CCTP) then the adapter should become a bit more generic, and wouldn't necessarily need to store the USDT address as a variable.

I _think_ this would require a mapping of token -> oftMessenger, which is admittedly a bit more complex to implement. It could potentially be done via an admin function from the SpokePool. I'm ATM not sure how it could work when bridging from the HubPool.

contracts/libraries/OFTTransportAdapter.sol
```diff
@@ -0,0 +1,91 @@
...
+        // todo: here, we're relying on `oftMessenger` to use the full allowance. Otherwise, next .safeApprove will revert.
+        usdt.safeApprove(address(this), _amount);
+
+        // todo: not really sure if we should blindly trust the fee.nativeFee here and just send it away. Should we check it against some sane cap?
+        // setting refundAddress to zero addr here, because we calculate the fees precicely and we can save gas this way
+        oftMessenger.send{ value: fee.nativeFee }(sendParam, fee, ZERO_ADDRESS);
+
+        // todo: OFTAdapter enforces this, but should we check anyway?
+        // return vals from .send: (MessagingReceipt memory msgReceipt, OFTReceipt memory oftReceipt)
+        // require(amount == oftReceipt.amountSentLD);
```
nicholaspai, line 88: We should check. Ideally in a single requires statement

contracts/libraries/OFTTransportAdapter.sol
```diff
@@ -0,0 +1,91 @@
...
+    IERC20 public immutable usdt;
+    // Mailbox address for OFT cross-chain transfers.
+    IOFT public immutable oftMessenger;
+
+    /**
+     * @notice The destination endpoint id in the OFT messaging protocol.
+     * @dev Source https://docs.layerzero.network/v2/developers/evm/technical-reference/deployed-contracts.
+     * @dev Can also be found on target chain OFTAdapter -> endpoint() -> eid().
+     */
+    uint32 public immutable dstEid;
```
nicholaspai, line 31: We should hardcode this id mapping and set it based on the chainid of the spoke pool/adapter. See how we do so in Arbitrum Adapter for CCTP https://github.com/across-protocol/contracts/blob/master/contracts/chain-adapters/Arbitrum_Adapter.sol#L71C71-L71C79

package.json
```diff
@@ -38,7 +38,7 @@
     "test": "yarn test-evm && yarn test-svm",
     "test:report-gas": "IS_TEST=true REPORT_GAS=true hardhat test",
     "generate-evm-assets": "rm -rf typechain && TYPECHAIN=ethers yarn hardhat typechain",
-    "prepublish": "yarn build && hardhat export --export-all ./cache/massExport.json && ts-node ./scripts/processHardhatExport.ts && prettier --write ./deployments/deployments.json && yarn generate-evm-assets",
+    "prepublishOnly": "yarn build && hardhat export --export-all ./cache/massExport.json && ts-node ./scripts/processHardhatExport.ts && prettier --write ./deployments/deployments.json && yarn generate-evm-assets",
```
nicholaspai, line 41: Why this change?

package.json
```diff
@@ -137,5 +140,6 @@
   "resolutions": {
     "**/eccrypto/secp256k1": "3.8.1",
     "**/eth-crypto/secp256k1": "5.0.1"
-  }
+  },
+  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
```
nicholaspai, line 144: Why this change?

nicholaspai (commented):
Left some comments!

contracts/libraries/OFTTransportAdapter.sol
```diff
@@ -0,0 +1,91 @@
...
+            amountLD,
+            minAmountLD,
+            EMPTY_MSG_BYTES,
+            EMPTY_MSG_BYTES,
+            EMPTY_MSG_BYTES
+        );
+
+        MessagingFee memory fee = oftMessenger.quoteSend(sendParam, false);
+
+        // todo: here, we're relying on `oftMessenger` to use the full allowance. Otherwise, next .safeApprove will revert.
```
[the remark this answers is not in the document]
bmzig: Why not just use `forceApprove` here? Seems much safer.

contracts/libraries/OFTTransportAdapter.sol
```diff
@@ -0,0 +1,91 @@
...
+ */
+contract OFTTransportAdapter {
+    using SafeERC20 for IERC20;
+    using AddressToBytes32 for address;
+
+    bytes public constant EMPTY_MSG_BYTES = new bytes(0);
+    address public constant ZERO_ADDRESS = address(0);
+
+    // USDT address on current chain.
+    IERC20 public immutable usdt;
```
bmzig, line 22: nit: These immutable variables should be in all caps.

grasphoper:
@pxrl @nicholaspai 

more details on `prepublish` vs `prepublishOnly` change.

When I was fixing my `yarn install` problem, I initially got very confused as to what was happening. Turned out, `prepublish` script from `package.json` was run after every `yarn add` which was very counterintuitive.

I found [this](https://github.com/apiaryio/dredd/issues/945) as solution on the internet.

Wanted to start a discussion for if we need this to stay `prepublish` or if we're open to changing it. It might help someone new to the repo in the future.

This change is beyond the scope of the PR, but a one-liner so I decided to add it here and discuss

contracts/Arbitrum_SpokePool.sol
```diff
@@ -27,10 +28,14 @@ contract Arbitrum_SpokePool is SpokePool, CircleCCTPAdapter {
         uint32 _depositQuoteTimeBuffer,
         uint32 _fillDeadlineBuffer,
         IERC20 _l2Usdc,
-        ITokenMessenger _cctpTokenMessenger
+        ITokenMessenger _cctpTokenMessenger,
+        IERC20 _l2Usdt,
+        IOFT _oftMessenger,
+        uint32 _ethereumUsdtDstEid
```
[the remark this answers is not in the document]
grasphoper: Yeah, I like `oftDstEid`, thanks! That naming was off

contracts/Arbitrum_SpokePool.sol
```diff
@@ -86,6 +91,10 @@ contract Arbitrum_SpokePool is SpokePool, CircleCCTPAdapter {
         // If the l2TokenAddress is UDSC, we need to use the CCTP bridge.
         if (_isCCTPEnabled() && l2TokenAddress == address(usdcToken)) {
```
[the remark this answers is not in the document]
grasphoper: Good catch

contracts/chain-adapters/Arbitrum_Adapter.sol
```diff
@@ -61,14 +62,23 @@ contract Arbitrum_Adapter is AdapterInterface, CircleCCTPAdapter {
...
         ArbitrumL1InboxLike _l1ArbitrumInbox,
         ArbitrumL1ERC20GatewayLike _l1ERC20GatewayRouter,
         address _l2RefundL2Address,
         IERC20 _l1Usdc,
-        ITokenMessenger _cctpTokenMessenger
-    ) CircleCCTPAdapter(_l1Usdc, _cctpTokenMessenger, CircleDomainIds.Arbitrum) {
+        ITokenMessenger _cctpTokenMessenger,
+        IERC20 _l1Usdt,
+        IOFT _oftTransport,
+        uint32 _arbitrumUsdtDstEid
```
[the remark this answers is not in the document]
grasphoper: Agreed

contracts/libraries/OFTTransportAdapter.sol
```diff
@@ -0,0 +1,91 @@
...
+        oftMessenger = _oftMessenger;
+        dstEid = _dstEid;
+    }
+
+    /**
+     * @notice Returns whether or not the OFT bridge is enabled.
+     * @dev If the IOFT is the zero address, OFT bridging is disabled.
+     */
+    function _isOFTEnabled(address _token) internal view returns (bool) {
+        return address(oftMessenger) != address(0) && address(usdt) == _token;
```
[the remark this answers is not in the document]
grasphoper: This one is interesting, because the current logic in the PR is the one you suggested to change to in the CCTP branch.

We have 2 options:
1. check zero-addr first, then token
2. check token first, then zero-addr

For CCTP, OFT enabled chains:
2 is better, because the 2nd part of the branch will not get evaluated.

For chains that inherit CCTP, OFT, but not use it
1 is better, because the zero check will always succeed

Which one do we want you think?

contracts/libraries/OFTTransportAdapter.sol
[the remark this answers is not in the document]
grasphoper: > Is this RE the USDT approval constraint that requires approvals to be reset to 0 before being modified again?
Yes

> Why not just use forceApprove here? Seems much safer.
Is this an fn in SafeERC20?
Yeah, it's my first time using the Safe calls, I might be using the wrong one here.

contracts/libraries/OFTTransportAdapter.sol
grasphoper: > I think this would require a mapping of token -> oftMessenger

Yep.

Let's discuss this in the above thread

contracts/libraries/OFTTransportAdapter.sol
grasphoper: Agreed, thought about doing it also

contracts/libraries/OFTTransportAdapter.sol
grasphoper: Thanks! Will change

contracts/libraries/OFTTransportAdapter.sol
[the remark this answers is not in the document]
bmzig: It is. We use it throughout here because of USDT. Basically, it tries to approve, and if the internal txn reverts, it will approve to 0 and then re-approve.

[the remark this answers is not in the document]
grasphoper: I see, sounds like a good solution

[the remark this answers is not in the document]
grasphoper: Also, do we ever use raw ERC20 functions, e.g. `usdt.approve(..)` ?

contracts/libraries/OFTTransportAdapter.sol
```diff
@@ -0,0 +1,91 @@
...
+
+    /**
+     * @notice intiailizes the OFTTransportAdapter contract.
+     * @param _usdt USDT address on the current chain.
+     * @param _oftMessenger OFTAdapter contract to bridge to the other chain. If address is set to zero, OFT bridging will be disabled.
+     * @param _dstEid The endpoint ID that OFT will transfer funds to.
+     */
+    constructor(IERC20 _usdt, IOFT _oftMessenger, uint32 _dstEid) {
+        usdt = _usdt;
+        oftMessenger = _oftMessenger;
```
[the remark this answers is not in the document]
nicholaspai: I think we should only support USDT to start and if we want to use this adapter in the future to support multiple mappings we should update it. The thinking is we don't know yet who should set the mapping, should it be the HubPool.owne()--probably. But we've never had an Adapter be `Ownable` before and I don't think we should start now. We can instead deploy the tokens we support in the `constructor` but I don't want to overengineer anything and this is a simple feature we can add in the future

package.json
nicholaspai: which comment? can you link it

contracts/libraries/OFTTransportAdapter.sol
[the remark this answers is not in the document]
bmzig: I don't think we do. If we do, then we should probably replace it.

package.json
grasphoper: Oh, sry, [this one](https://github.com/across-protocol/contracts/pull/902#issuecomment-2685903839)

contracts/libraries/OFTTransportAdapter.sol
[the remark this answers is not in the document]
pxrl: USDT is only a one of several tokens that we support, so to me it makes most sense to filter on token address first. This means we only pay a penalty when evaluating USDT on chains that do not support OFT; but the benefit is that all other tokens we bridge back do not incur that additional comparison. The set of non-OFT USDT deployments should reduce over time as more chains migrate to OFT.

pxrl:
> @pxrl @nicholaspai
> 
> more details on `prepublish` vs `prepublishOnly` change.
> 
> When I was fixing my `yarn install` problem, I initially got very confused as to what was happening. Turned out, `prepublish` script from `package.json` was run after every `yarn add` which was very counterintuitive.
> 
> I found [this](https://github.com/apiaryio/dredd/issues/945) as solution on the internet.
> 
> Wanted to start a discussion for if we need this to stay `prepublish` or if we're open to changing it. It might help someone new to the repo in the future.
> 
> This change is beyond the scope of the PR, but a one-liner so I decided to add it here and discuss

In general it seems odd that we're using `prepublish` in this way - note that yarn's own documentation says it should not have any side-effects, so it seems like we're misusing that script target.

https://yarnpkg.com/advanced/lifecycle-scripts#prepublish
```
warning
Because it's only called on prepublish, the prepublish hook shouldn't have side effects. In particular don't transpile the package sources in prepublish, as people consuming directly your repository (such as through the [git: protocol](https://yarnpkg.com/protocol/git)) wouldn't be able to use your project. Instead, use prepack.
```

grasphoper:
> > @pxrl @nicholaspai
> > more details on `prepublish` vs `prepublishOnly` change.
> > When I was fixing my `yarn install` problem, I initially got very confused as to what was happening. Turned out, `prepublish` script from `package.json` was run after every `yarn add` which was very counterintuitive.
> > I found [this](https://github.com/apiaryio/dredd/issues/945) as solution on the internet.
> > Wanted to start a discussion for if we need this to stay `prepublish` or if we're open to changing it. It might help someone new to the repo in the future.
> > This change is beyond the scope of the PR, but a one-liner so I decided to add it here and discuss
> 
> In general it seems odd that we're using `prepublish` in this way - note that yarn's own documentation (admittedly for v4, whilst we're on v1) says it should not have any side-effects, so it seems like we're misusing that script target.
> 
> https://yarnpkg.com/advanced/lifecycle-scripts#prepublish `warning Because it's only called on prepublish, the prepublish hook shouldn't have side effects. In particular don't transpile the package sources in prepublish, as people consuming directly your repository (such as through the [git: protocol](https://yarnpkg.com/protocol/git)) wouldn't be able to use your project. Instead, use prepack.`
> 
> See here for the yarn v1.22 docs on these targets: https://classic.yarnpkg.com/en/docs/package-json#scripts-
> 
> Also based on this, `prepublishOnly` is npm-specific and yarn does not list any support for it, so I think renaming this here is likely to have side-effects for our build process and CI.

So, from what I see, our only alternatives would be these ? `install, postinstall, prepublish, and prepare`

```
Certain script names are special. If defined, the preinstall script is called by yarn before your package is installed. For compatibility reasons, scripts called install, postinstall, prepublish, and prepare will all be called after your package has finished installing.
```

contracts/libraries/OFTTransportAdapter.sol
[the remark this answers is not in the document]
nicholaspai: Has @grasphoper mentioned to you @pxrl that we can't support any ERC20's with decimal precision >6 ? That's another reason why I wanted this adapter to explicitly mention USDT.

nicholaspai:
> > > @pxrl @nicholaspai
> > > more details on `prepublish` vs `prepublishOnly` change.
> > > When I was fixing my `yarn install` problem, I initially got very confused as to what was happening. Turned out, `prepublish` script from `package.json` was run after every `yarn add` which was very counterintuitive.
> > > I found [this](https://github.com/apiaryio/dredd/issues/945) as solution on the internet.
> > > Wanted to start a discussion for if we need this to stay `prepublish` or if we're open to changing it. It might help someone new to the repo in the future.
> > > This change is beyond the scope of the PR, but a one-liner so I decided to add it here and discuss
> > 
> > 
> > In general it seems odd that we're using `prepublish` in this way - note that yarn's own documentation (admittedly for v4, whilst we're on v1) says it should not have any side-effects, so it seems like we're misusing that script target.
> > https://yarnpkg.com/advanced/lifecycle-scripts#prepublish `warning Because it's only called on prepublish, the prepublish hook shouldn't have side effects. In particular don't transpile the package sources in prepublish, as people consuming directly your repository (such as through the [git: protocol](https://yarnpkg.com/protocol/git)) wouldn't be able to use your project. Instead, use prepack.`
> > See here for the yarn v1.22 docs on these targets: https://classic.yarnpkg.com/en/docs/package-json#scripts-
> > Also based on this, `prepublishOnly` is npm-specific and yarn does not list any support for it, so I think renaming this here is likely to have side-effects for our build process and CI.
> 
> So, from what I see, our only alternatives would be these ? `install, postinstall, prepublish, and prepare`. Quote from v1 docs:
> 
> > Certain script names are special. If defined, the preinstall script is called by yarn before your package is installed. For compatibility reasons, scripts called install, postinstall, prepublish, and prepare will all be called after your package has finished installing.

Regardless of the outcome of this script name, we should change it in a separate PR and not bundle it into this PR

grasphoper:
@nicholaspai @pxrl  Perhaps this `prepublish` thing deserves a separate PR to discuss if we decide it's worth the time. For now though, I will just revert to using `prepublish` in this one

grasphoper:
Hi @pxrl @nicholaspai @bmzig

I updated this PR with these main changes:
- modified deploy scripts
- added basic tests for OFT functionality
- addressed many of the comments on prev. reviews

Main unresolved issues (they're also in "todo: ..." comments in the code):
- in `OFTTransportAdapter`, when we're calling this `OFT_MESSENGER.send{ value: fee.nativeFee }(...)`, we're trusting `OFT_MESSENGER.quoteSend(...)` to have returned a sane fee in the previous call. Should we check against some sane limit here?
- related to previous point: `_contractHasSufficientEthBalance`. When calling canonical L2 -> L1 bridge, `Arbitrum_Adapter` checks that we have enough eth on the contract before calling the function (I guess for better error reporting). Should we do the same here?
- this PR creates an apparent imbalance between AlephZero SpokePool and Adapter, where `Arbitrum_SpokePool` just inherits full `Arbitrum_SpookePool` functionality, but on the adapter side, I assume it's using `Arbitrum_CustomGasToken_Adapter`, which didn't have any functionality added. In practice though, we should be able to just set the `OFTTransportAdapter` to zero address on `AlephZero_SpokePool` and be fine, but I wanted to mention this

contracts/libraries/OFTTransportAdapter.sol
```diff
@@ -0,0 +1,114 @@
...
+        // `false` in the 2nd param here refers to `bool _payInLzToken`. We will pay in native token, so set to `false`
+        MessagingFee memory fee = OFT_MESSENGER.quoteSend(sendParam, false);
+
+        // approve the exact _amount for `OFT_MESSENGER` to spend. Fee will be paid in native token
+        USDT.forceApprove(address(OFT_MESSENGER), _amount);
+
+        // todo: not really sure if we should blindly trust the fee.nativeFee here and just send it away. Should we check it against some sane cap?
+        // todo: parent pool calls `_contractHasSufficientEthBalance` on canonical bridge transfer. Should we do the same?
+        // setting `refundAddress` to `ZERO_ADDRESS` here, because we calculate the fees precicely and we can save gas this way
+        (, OFTReceipt memory oftReceipt) = OFT_MESSENGER.send{ value: fee.nativeFee }(sendParam, fee, ZERO_ADDRESS);
```
nicholaspai, line 109: Regarding your comment: 

> // todo: not really sure if we should blindly trust the fee.nativeFee here and just send it away. Should we check it against some sane cap?

I'm not sure what you mean to check. The `value` is being set to `fee.nativeFee` so the OFT_MESSENGER can't take any more ETH from this contract. Is there something else to check?

contracts/libraries/OFTTransportAdapter.sol
```diff
@@ -0,0 +1,114 @@
...
+        );
+
+        // `false` in the 2nd param here refers to `bool _payInLzToken`. We will pay in native token, so set to `false`
+        MessagingFee memory fee = OFT_MESSENGER.quoteSend(sendParam, false);
+
+        // approve the exact _amount for `OFT_MESSENGER` to spend. Fee will be paid in native token
+        USDT.forceApprove(address(OFT_MESSENGER), _amount);
+
+        // todo: not really sure if we should blindly trust the fee.nativeFee here and just send it away. Should we check it against some sane cap?
+        // todo: parent pool calls `_contractHasSufficientEthBalance` on canonical bridge transfer. Should we do the same?
```
nicholaspai, line 107: >  // todo: parent pool calls `_contractHasSufficientEthBalance` on canonical bridge transfer. Should we do the same?

`_contractHasSufficientEthBalance` doesn't really work well for this use case since it computes total value based on gas price, gas cost, etc. I don't think its neccessary.

contracts/libraries/OFTTransportAdapter.sol
```diff
@@ -0,0 +1,114 @@
...
+            EMPTY_MSG_BYTES
+        );
+
+        // `false` in the 2nd param here refers to `bool _payInLzToken`. We will pay in native token, so set to `false`
+        MessagingFee memory fee = OFT_MESSENGER.quoteSend(sendParam, false);
+
+        // approve the exact _amount for `OFT_MESSENGER` to spend. Fee will be paid in native token
+        USDT.forceApprove(address(OFT_MESSENGER), _amount);
+
+        // todo: not really sure if we should blindly trust the fee.nativeFee here and just send it away. Should we check it against some sane cap?
```
nicholaspai, line 106: I think we can set a max CAP of 1 ether just as a safety check? WDYT?

nicholaspai:
> this PR creates an apparent imbalance between AlephZero SpokePool and Adapter, where Arbitrum_SpokePool just inherits full Arbitrum_SpookePool functionality, but on the adapter side, I assume it's using Arbitrum_CustomGasToken_Adapter, which didn't have any functionality added. In practice though, we should be able to just set the OFTTransportAdapter to zero address on AlephZero_SpokePool and be fine, but I wanted to mention this

@grasphoper I'm not sure I understand this point, can you explain more?

contracts/libraries/OFTTransportAdapter.sol
grasphoper: will do

contracts/libraries/OFTTransportAdapter.sol
grasphoper: Yeah I was thinking something like that. 1 ETH is kind of a lot, but maybe if we're sending a million?

So, I did a mini-investigation on fees. If I try to `quoteSend` even on a ridiculous amount (tested with 1M USDT) -- I get a micro-fee: got a `34088408170172` == ~0.000034 ETH fee. So, I'm thinking maybe 0.1 ETH as cap?

For reference:
Called `quote` on `0x1a44076050125825900e736c501f859c50fE728c` this address -- their EndpointV2 contract
With params
```
_params: [30110,"0x0000000000000000000000009A8f92a830A5cB89a3816e3D267CB7791c16b04D","0x9a8f92a830a5cb89a3816e3d267cb7791c16b04d000000000000000000000000000000000000000000000000000000e8d4a51000","0x00030100110100000000000000000000000000013880",false]
_sender:
0x9A8f92a830A5cB89a3816e3D267CB7791c16b04D
```

grasphoper: Also, if we're checking for this, maybe we want to spend some extra gas and set our sender address to a gas refund address in the `.send` call later?

grasphoper:
> > this PR creates an apparent imbalance between AlephZero SpokePool and Adapter, where Arbitrum_SpokePool just inherits full Arbitrum_SpookePool functionality, but on the adapter side, I assume it's using Arbitrum_CustomGasToken_Adapter, which didn't have any functionality added. In practice though, we should be able to just set the OFTTransportAdapter to zero address on AlephZero_SpokePool and be fine, but I wanted to mention this
> 
> @grasphoper I'm not sure I understand this point, can you explain more?

For reference to reviewers:

Talked about this off site, the gist of it:
- `AlephZero_SpookePool` will theoretically support USDT0 `OFT` transfers after this PR is merged. Which, USDT0 is not deployed on `AlephZero`, and this functionality would just be disabled by setting `OFTMessenger` address to zero addr if we're to redeploy
- `AlephZero_Adapter` (which is really `Arbitrum_CustomGasToken_Adapter`) will have to OFT support for now

contracts/libraries/OFTTransportAdapter.sol
grasphoper: Also, on the cap, can ask LZ guys if fee can ever reach > 0.1 ETH

nicholaspai: > Also, if we're checking for this, maybe we want to spend some extra gas and set our sender address to a gas refund address in the `.send` call later?

Not sure what you mean here

grasphoper: Currently, in `.send`, I set last param to `ZERO_ADDRESS`, which is `gasRefund` address. I do it mainly to save a bit of gas and not set that param to `address(this)`. But if we're not that strict about gas costs, we can as well be safer and set a gas refund to our address in case something somewhere messes up and `quoteSend` is not equal to what we actually pay but instead we pay less and would receive a refund if the refund address was set

nicholaspai: when does the gas refund happen? Wouldn't we always want the refund to be sent to an address we control?

contracts/Arbitrum_SpokePool.sol
```diff
@@ -3,34 +3,43 @@ pragma solidity ^0.8.0;
...
 
     /// @custom:oz-upgrades-unsafe-allow constructor
     constructor(
         address _wrappedNativeTokenAddress,
         uint32 _depositQuoteTimeBuffer,
         uint32 _fillDeadlineBuffer,
         IERC20 _l2Usdc,
-        ITokenMessenger _cctpTokenMessenger
+        ITokenMessenger _cctpTokenMessenger,
+        // _oftFeeCap can be set to 1 ether for arbitrum, but has to be custom-set for other chains that might use inherit this adapter, like AlephZero
```
nicholaspai, line 37:
```suggestion
        // _oftFeeCap can be set to 1 ether for Arbitrum, but has to be custom-set for other chains that might inherit this adapter, like AlephZero
```

contracts/libraries/OFTTransportAdapter.sol
```diff
@@ -0,0 +1,111 @@
...
+             * These can be set to empty bytes arrays for the purposes of sending a simple cross-chain transfer.
+             */
+            EMPTY_MSG_BYTES,
+            EMPTY_MSG_BYTES,
+            EMPTY_MSG_BYTES
+        );
+
+        // `false` in the 2nd param here refers to `bool _payInLzToken`. We will pay in native token, so set to `false`
+        MessagingFee memory fee = _messenger.quoteSend(sendParam, false);
+        require(fee.nativeFee <= FEE_CAP, "OFT FEE_CAP exceeded");
```
nicholaspai, line 100: nit prefer using custom [Errors](https://github.com/across-protocol/contracts/blob/555475cdee6109afc85065ca415c740d7f97b992/contracts/interfaces/V3SpokePoolInterface.sol#L326) to save gas costs

contracts/Arbitrum_SpokePool.sol
```diff
@@ -112,4 +136,19 @@ contract Arbitrum_SpokePool is SpokePool, CircleCCTPAdapter {
 
     // Apply AVM-specific transformation to cross domain admin address on L1.
     function _requireAdminSender() internal override onlyFromCrossDomainAdmin {}
+
+    function _setOftMessenger(IERC20 _token, IOFT _messenger) internal {
+        oftMessengers[_token] = _messenger;
+        emit OFTMessengerSet(_token, _messenger);
+    }
+
+    function _getOftMessenger(IERC20 _token) internal view returns (IOFT messenger, bool oftEnabled) {
```
nicholaspai, line 145: Instead of returning a tuple, could you just return IOFT and if this is equal to `address(0)` you can assume `oftEnabled == false`?

contracts/Arbitrum_SpokePool.sol
```diff
@@ -3,34 +3,43 @@ pragma solidity ^0.8.0;
...
     // Admin controlled mapping of arbitrum tokens to L1 counterpart. L1 counterpart addresses
     // are necessary params used when bridging tokens to L1.
     mapping(address => address) public whitelistedTokens;
 
+    // a map to store token -> IOFT relationships for OFT bridging.
+    mapping(IERC20 => IOFT) public oftMessengers;
+
     event SetL2GatewayRouter(address indexed newL2GatewayRouter);
     event WhitelistedTokens(address indexed l2Token, address indexed l1Token);
+    // todo: do we really need this event? Considering we're approaching contract size limits
```
nicholaspai, line 27: we probably don't need this event, though I also don't think adding it adds much bytecode size.

contracts/Arbitrum_SpokePool.sol
```diff
@@ -3,34 +3,43 @@ pragma solidity ^0.8.0;
...
-contract Arbitrum_SpokePool is SpokePool, CircleCCTPAdapter {
+contract Arbitrum_SpokePool is SpokePool, CircleCCTPAdapter, OFTTransportAdapter {
     // Address of the Arbitrum L2 token gateway to send funds to L1.
     address public l2GatewayRouter;
 
     // Admin controlled mapping of arbitrum tokens to L1 counterpart. L1 counterpart addresses
     // are necessary params used when bridging tokens to L1.
     mapping(address => address) public whitelistedTokens;
 
+    // a map to store token -> IOFT relationships for OFT bridging.
```
nicholaspai, line 22: Please start all comments with capital letters for consistency

contracts/Arbitrum_SpokePool.sol
```diff
@@ -78,14 +87,29 @@ contract Arbitrum_SpokePool is SpokePool, CircleCCTPAdapter {
...
      **************************************/
 
     function _bridgeTokensToHubPool(uint256 amountToReturn, address l2TokenAddress) internal override {
+        (IOFT oftMessenger, bool oftEnabled) = _getOftMessenger(IERC20(l2TokenAddress));
+
         // If the l2TokenAddress is UDSC, we need to use the CCTP bridge.
         if (_isCCTPEnabled() && l2TokenAddress == address(usdcToken)) {
             _transferUsdc(withdrawalRecipient, amountToReturn);
+        }
+        // If OFT is enabled for token, we need to use the OFT bridge.
```
nicholaspai, line 110: I think this comment is self-evident with the code and can be removed

contracts/chain-adapters/Arbitrum_Adapter.sol
```diff
@@ -54,24 +57,36 @@ contract Arbitrum_Adapter is AdapterInterface, CircleCCTPAdapter {
     // Generic gateway: https://github.com/OffchainLabs/token-bridge-contracts/blob/main/contracts/tokenbridge/ethereum/gateway/L1ArbitrumGateway.sol
     ArbitrumL1ERC20GatewayLike public immutable L1_ERC20_GATEWAY_ROUTER;
 
+    // we don't allow fees to go higher than `OFT_FEE_CAP` when calling .send in `OFTTransportAdapter`. It gives us some protection from potential bugs on OFT's end
```
nicholaspai, line 60: Please start all comments with capital letters for consistency

contracts/chain-adapters/Arbitrum_Adapter.sol
```diff
@@ -54,24 +57,36 @@ contract Arbitrum_Adapter is AdapterInterface, CircleCCTPAdapter {
...
+    uint256 public constant OFT_FEE_CAP = 1 ether;
+
+    // helper contract to help us map token -> oft messenger for OFT-enabled tokens
```
nicholaspai, line 63: Please start all comments with capital letters for consistency

```suggestion
    // Helper contract to help us map token -> OFT messenger for OFT-enabled tokens
```

contracts/chain-adapters/Arbitrum_Adapter.sol
```diff
@@ -113,9 +128,14 @@ contract Arbitrum_Adapter is AdapterInterface, CircleCCTPAdapter {
         uint256 amount,
         address to
     ) external payable override {
+        (IOFT oftMessenger, bool oftEnabled) = _getOftMessenger(IERC20(l1Token));
+
         // Check if this token is USDC, which requires a custom bridge via CCTP.
         if (_isCCTPEnabled() && l1Token == address(usdcToken)) {
             _transferUsdc(to, amount);
+            // If oft messaging is enabled for token, use it as a bridging mechanism
```
nicholaspai, line 136: I also think this comment is self-evident and can be removed

contracts/libraries/OFTAddressBook.sol
```diff
@@ -0,0 +1,21 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.0;
+
+import { Ownable } from "@openzeppelin/contracts/access/Ownable.sol";
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import { IOFT } from "@layerzerolabs/oft-evm/contracts/interfaces/IOFT.sol";
+
+/**
+ * @dev a helper contract for chain adapters that support OFT messaging from L1
```
nicholaspai, line 9:
```suggestion
 * @dev A helper contract for chain adapters that support OFT messaging from L1
```

contracts/libraries/OFTAddressBook.sol
```diff
@@ -0,0 +1,21 @@
...
+ * @dev handles OFT token -> messenger mapping storage, as adapters are called via delegatecall and don't have relevant storage space
```
nicholaspai, line 10:
```suggestion
 * @dev Handles OFT token -> messenger mapping storage, as adapters are called via delegatecall and don't have relevant storage space
```

contracts/libraries/OFTTransportAdapter.sol
```diff
@@ -0,0 +1,111 @@
...
+        address _to,
+        uint256 _amount
+    ) internal {
+        bytes32 to = _to.toBytes32();
+
+        SendParam memory sendParam = SendParam(
+            DST_EID,
+            to,
+            /**
+             * _amount, _amount here specify `amountLD` and `minAmountLD`. These 2 have a subtle relationship
```
nicholaspai, line 78: I think this comment can be simplified a bit, and it also is a bit misleading because this contract doesn't really add any enforcement to the dust that might get removed. I don't think it needs to do anything extra here, but I'd like to see this comment simplified.

contracts/libraries/OFTTransportAdapter.sol
```diff
@@ -0,0 +1,111 @@
...
+
+        // `false` in the 2nd param here refers to `bool _payInLzToken`. We will pay in native token, so set to `false`
+        MessagingFee memory fee = _messenger.quoteSend(sendParam, false);
+        require(fee.nativeFee <= FEE_CAP, "OFT FEE_CAP exceeded");
+        require(fee.nativeFee <= address(this).balance, "OFT insufficient funds for fee");
+
+        // approve the exact _amount for `_messenger` to spend. Fee will be paid in native token
+        _token.forceApprove(address(_messenger), _amount);
+
+        (, OFTReceipt memory oftReceipt) = _messenger.send{ value: fee.nativeFee }(sendParam, fee, address(this));
```
nicholaspai, line 106: Just curious, what is the first value in this tuple return value?

deploy/004_deploy_arbitrum_adapter.ts
```diff
@@ -27,6 +28,7 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
       l2RefundAddress,
       USDC[chainId],
       L1_ADDRESS_MAP[chainId].cctpTokenMessenger,
+      L1_ADDRESS_MAP[chainId].oftAddressBook, // todo: will this get populated once we deploy oftAddressBook via a deploy script #63?
```
nicholaspai, line 31: No, we need to manually add this value to the L1_ADDRESS_MAP mapping

test/evm/hardhat/MerkleLib.utils.ts
```diff
@@ -106,10 +106,25 @@ export function buildPoolRebalanceLeaves(
...
 }
 
-export async function constructSingleRelayerRefundTree(l2Token: Contract | String, destinationChainId: number) {
+export async function constructSingleRelayerRefundTree(
+  l2Token: Contract | String,
+  destinationChainId: number,
+  amount?: BigNumber
+) {
+  // Explicitly determine which amount to use based on whether amount parameter is provided
+  let amountToUse: BigNumber;
```
nicholaspai, line 115: Could you do `const amountToUse = isDefined(amount) ? amount : amountToReturn`?

test/evm/hardhat/chain-adapters/Arbitrum_Adapter.ts
```diff
@@ -55,14 +69,26 @@ describe("Arbitrum Chain Adapter", function () {
     cctpTokenMinter = await createFakeFromABI(CCTPTokenMinterInterface);
     cctpMessenger.localMinter.returns(cctpTokenMinter.address);
     cctpTokenMinter.burnLimitsPerMessage.returns(toWei("1000000"));
+
+    oftMessenger = await createTypedFakeFromABI([...IOFT__factory.abi] as any[]);
+    oftAddressBook = await createTypedFakeFromABI([...OFTAddressBook__factory.abi] as any[]);
```
nicholaspai, line 74: why can't you use the `createFakeFromABI` function we use above? In general, alerts go off when we have to use the type cast `any[]` anywhere

test/evm/hardhat/chain-specific-spokepools/Arbitrum_SpokePool.ts
```diff
@@ -40,19 +58,28 @@ describe("Arbitrum Spoke Pool", function () {
     cctpTokenMinter = await createFakeFromABI(CCTPTokenMinterInterface);
     l2CctpTokenMessenger.localMinter.returns(cctpTokenMinter.address);
     cctpTokenMinter.burnLimitsPerMessage.returns(toWei("1000000"));
+    l2OftMessenger = await createTypedFakeFromABI([...IOFT__factory.abi] as any[]);
```
nicholaspai, line 61: Same question about not using `createFakeFromABI`

test/evm/hardhat/chain-specific-spokepools/Arbitrum_SpokePool.ts
```diff
@@ -138,4 +172,46 @@ describe("Arbitrum Spoke Pool", function () {
...
+    l2OftMessenger.send.returns([msgReceipt, oftReceipt]);
+
+    await arbitrumSpokePool.connect(relayer).executeRelayerRefundLeaf(0, leaves[0], tree.getHexProof(leaves[0]));
+    // Adapter should have approved gateway to spend its ERC20.
+    expect(await l2UsdtContract.allowance(arbitrumSpokePool.address, l2OftMessenger.address)).to.equal(
+      l2UsdtSendAmount
+    );
+
+    // We should have called send on the l2OftMessenger
+    expect(l2OftMessenger.send).to.have.been.calledOnce;
```
nicholaspai, line 215: I like using the [calledWith](https://smock.readthedocs.io/en/latest/fakes.html#asserting-call-arguments-or-value) syntax to check the arguments passed to `send` are as expected

test/evm/hardhat/chain-adapters/Arbitrum_Adapter.ts
```diff
@@ -225,4 +252,42 @@ describe("Arbitrum Chain Adapter", function () {
...
+
+    oftMessenger.send.returns([msgReceipt, oftReceipt]);
+
+    await hubPool.connect(dataWorker).executeRootBundle(...Object.values(leaves[0]), tree.getHexProof(leaves[0]));
+
+    // Adapter should have approved gateway to spend its ERC20.
+    expect(await usdt.allowance(hubPool.address, oftMessenger.address)).to.equal(tokensSendToL2);
+
+    // We should have called send on the oftMessenger
+    expect(oftMessenger.send).to.have.been.calledOnce;
```
nicholaspai, line 291: I like using the [calledWith](https://smock.readthedocs.io/en/latest/fakes.html#asserting-call-arguments-or-value) syntax to check the arguments passed to `send` are as expected

nicholaspai (commented):
Lots of great work here, left many comments!

contracts/Arbitrum_SpokePool.sol
grasphoper: I'll leave it in then to be consistent with having `WhitelistedTokens` also. We could remove both if we feel like it in another PR

contracts/Arbitrum_SpokePool.sol
grasphoper: If I do this, do I still leave the function declaration? It's really just getting a value from map by key

contracts/libraries/OFTTransportAdapter.sol
grasphoper: I can leave out all of the "how to do it correctly" explanation there and just have something simple like:

```
            /**
             * _amount, _amount here specify `amountLD` and `minAmountLD`. Setting `minAmountLD` equal to `amountLD` protects us
             * from any changes to the sent amount due to internal OFT contract logic, e.g. `_removeDust`
             */
```

contracts/libraries/OFTTransportAdapter.sol
grasphoper:
```
struct MessagingReceipt {
    bytes32 guid;
    uint64 nonce;
    MessagingFee fee;
}
```

Returns this. I don't think we can use it in a meaningful way from our contract. guid and nonce are emitted as a part of `OFTSent` event on their contract

test/evm/hardhat/chain-adapters/Arbitrum_Adapter.ts
grasphoper: So, is the ` as any[]);` part problematic? Apparently I don't need it, `oftMessenger = await createTypedFakeFromABI([...IOFT__factory.abi]);` works just fine. Will change to that. I had to escape the `readonly` error (happens if I use raw `IOFT__factory.abi`)

Other than that, `createFakeFromABI` and `createTypedFakeFromABI` have the same argument types, `createTypedFakeFromABI` just gives your contract a defined interface, which is nice for e.g. autocompletion

nicholaspai: `as any[]` is just not good practice because it doesn't add any type safety

contracts/Arbitrum_SpokePool.sol
```diff
@@ -3,34 +3,42 @@ pragma solidity ^0.8.0;
...
     // Admin controlled mapping of arbitrum tokens to L1 counterpart. L1 counterpart addresses
     // are necessary params used when bridging tokens to L1.
     mapping(address => address) public whitelistedTokens;
 
+    // A map to store token -> IOFT relationships for OFT bridging.
+    mapping(IERC20 => IOFT) public oftMessengers;
+
     event SetL2GatewayRouter(address indexed newL2GatewayRouter);
     event WhitelistedTokens(address indexed l2Token, address indexed l1Token);
+    event OFTMessengerSet(IERC20 indexed token, IOFT indexed messenger);
```
nicholaspai, line 27: nit: to be consistent, name this `SetOFTMessenger` so the verb comes before the noun

contracts/chain-adapters/Arbitrum_Adapter.sol
```diff
@@ -54,24 +57,36 @@ contract Arbitrum_Adapter is AdapterInterface, CircleCCTPAdapter {
     // Generic gateway: https://github.com/OffchainLabs/token-bridge-contracts/blob/main/contracts/tokenbridge/ethereum/gateway/L1ArbitrumGateway.sol
     ArbitrumL1ERC20GatewayLike public immutable L1_ERC20_GATEWAY_ROUTER;
 
+    // We don't allow fees to go higher than `OFT_FEE_CAP` when calling .send in `OFTTransportAdapter`. It gives us some protection from potential bugs on OFT's end
```
nicholaspai, line 60: i'm surprised the linter doesn't split this comment onto two lines

contracts/libraries/OFTTransportAdapter.sol
```diff
@@ -0,0 +1,111 @@
...
+        address _to,
+        uint256 _amount
+    ) internal {
+        bytes32 to = _to.toBytes32();
+
+        SendParam memory sendParam = SendParam(
+            DST_EID,
+            to,
+            /**
+             * _amount, _amount here specify `amountLD` and `minAmountLD`. Setting `minAmountLD` equal to `amountLD` protects us
```
nicholaspai, line 82: What does this mean "protects us"? Do you mean if `_removeDust` were to change the amount sent value, then this call would revert?

contracts/libraries/OFTTransportAdapter.sol
```diff
@@ -0,0 +1,111 @@
...
+        if (fee.nativeFee > FEE_CAP) revert FeeCapExceeded(fee.nativeFee);
+        if (fee.nativeFee > address(this).balance)
+            revert InsufficientBalanceForFee(fee.nativeFee, address(this).balance);
+
+        // approve the exact _amount for `_messenger` to spend. Fee will be paid in native token
+        _token.forceApprove(address(_messenger), _amount);
+
+        (, OFTReceipt memory oftReceipt) = _messenger.send{ value: fee.nativeFee }(sendParam, fee, address(this));
+
+        // we require that received amount of this transfer at destination exactly matches the sent amount
```
nicholaspai, line 107:
```suggestion
        // The HubPool expects that the amount received by the SpokePool is exactly the sent amount
```

deploy/004_deploy_arbitrum_adapter.ts
```diff
@@ -1,4 +1,4 @@
-import { L1_ADDRESS_MAP, USDC } from "./consts";
+import { L1_ADDRESS_MAP, USDC, USDT } from "./consts";
```
nicholaspai, line 1: Does this import need to be here?

test/evm/hardhat/chain-specific-spokepools/Arbitrum_SpokePool.ts
```diff
@@ -40,19 +59,28 @@ describe("Arbitrum Spoke Pool", function () {
...
         kind: "uups",
         unsafeAllow: ["delegatecall"],
-        constructorArgs: [l2Weth, 60 * 60, 9 * 60 * 60, l2Usdc, l2CctpTokenMessenger.address],
+        constructorArgs: [
+          l2Weth,
+          60 * 60,
+          9 * 60 * 60,
+          l2Usdc,
+          l2CctpTokenMessenger.address,
+          ethers.utils.parseEther("0.1"),
```
nicholaspai, line 76: nit: use `toWei("0.1")`

test/evm/hardhat/chain-adapters/Arbitrum_Adapter.ts
```diff
@@ -225,4 +253,55 @@ describe("Arbitrum Chain Adapter", function () {
...
+      .connect(dataWorker)
+      .proposeRootBundle([3117], 1, tree.getHexRoot(), consts.mockRelayerRefundRoot, consts.mockSlowRelayRoot);
+    await timer.setCurrentTime(Number(await timer.getCurrentTime()) + consts.refundProposalLiveness + 1);
+
+    // set up correct messenger to be returned on a proper `oftMessengers` call
+    oftAddressBook.oftMessengers.whenCalledWith(usdt.address).returns(oftMessenger.address);
+
+    // set up `quoteSend` return val
+    const msgFeeStruct: MessagingFeeStructOutput = [
+      BigNumber.from(1e9 * 200_000), // nativeFee: 1 GWEI gas price * 200,000 gas cost
```
nicholaspai, line 270: These are big numbers and you're multiplying them as numbers, to be safer do: `toWeiWithDecimals("1", 9).mul(200_000)`

test/evm/hardhat/chain-adapters/Arbitrum_Adapter.ts
```diff
@@ -225,4 +253,55 @@ describe("Arbitrum Chain Adapter", function () {
...
+      BigNumber.from(0), // lzTokenFee: 0
```
nicholaspai, line 271: nit: use `bnZero`

test/evm/hardhat/chain-specific-spokepools/Arbitrum_SpokePool.ts
```diff
@@ -62,7 +90,14 @@ describe("Arbitrum Spoke Pool", function () {
...
```
nicholaspai, line 99: use `toWei("0.1")`

test/evm/hardhat/chain-specific-spokepools/Arbitrum_SpokePool.ts
```diff
@@ -138,4 +173,59 @@ describe("Arbitrum Spoke Pool", function () {
...
+  it("Bridge tokens to hub pool correctly using the OFT messaging for L2 USDT token", async function () {
+    const l2UsdtSendAmount = BigNumber.from("1234567");
+    const { leaves, tree } = await constructSingleRelayerRefundTree(
+      l2UsdtContract.address,
+      await arbitrumSpokePool.callStatic.chainId(),
+      l2UsdtSendAmount
+    );
+    await arbitrumSpokePool.connect(crossDomainAlias).relayRootBundle(tree.getHexRoot(), mockTreeRoot);
+
+    const oftNativeFee = BigNumber.from(1e9 * 200_000); // 1 GWEI gas price * 200,000 gas cost
```
nicholaspai, line 186: Same comment as above, this isn't neccessarily safe. Do `toWeiWithDecimals("1", 9).mul(200_000)`

test/evm/hardhat/chain-specific-spokepools/Arbitrum_SpokePool.ts
```diff
@@ -138,4 +173,59 @@ describe("Arbitrum Spoke Pool", function () {
...
+
+    // seed `arbitrumSpokePool` with some eth for native fee payment in OFT logic
+    await ethers.provider.send("hardhat_setBalance", [arbitrumSpokePool.address, oftNativeFee.toHexString()]);
+
+    // set up `quoteSend` return val
+    const msgFeeStruct: MessagingFeeStructOutput = [
+      oftNativeFee, // nativeFee
+      BigNumber.from(0), // lzTokenFee
```
nicholaspai, line 194: bnZero

contracts/chain-adapters/Arbitrum_Adapter.sol
grasphoper: I use prettier as my formatter and solhint as linter. Is this correct setup?

test/evm/hardhat/chain-specific-spokepools/Arbitrum_SpokePool.ts
grasphoper: I think it's only available in sdk, which is not used in contracts

contracts/chain-adapters/Arbitrum_Adapter.sol
nicholaspai: yep that's right for solidity files

nicholaspai: if it passes the linting CI step then its all good

test/evm/hardhat/chain-specific-spokepools/Arbitrum_SpokePool.ts
nicholaspai: ok thats fine

contracts/Arbitrum_SpokePool.sol
```diff
@@ -3,34 +3,42 @@ pragma solidity ^0.8.0;
...
-contract Arbitrum_SpokePool is SpokePool, CircleCCTPAdapter {
+contract Arbitrum_SpokePool is SpokePool, CircleCCTPAdapter, OFTTransportAdapter {
     // Address of the Arbitrum L2 token gateway to send funds to L1.
     address public l2GatewayRouter;
 
     // Admin controlled mapping of arbitrum tokens to L1 counterpart. L1 counterpart addresses
     // are necessary params used when bridging tokens to L1.
     mapping(address => address) public whitelistedTokens;
 
+    // A map to store token -> IOFT relationships for OFT bridging.
```
bmzig, line 22: With a new mapping, don't we need to bump down the `__gap` in `SpokePool.sol`?

bmzig: Actually, no, definitely not globally. Not sure what we want to do about this.

contracts/chain-adapters/Arbitrum_Adapter.sol
```diff
@@ -54,24 +57,36 @@ contract Arbitrum_Adapter is AdapterInterface, CircleCCTPAdapter {
...
         ArbitrumL1ERC20GatewayLike _l1ERC20GatewayRouter,
         address _l2RefundL2Address,
         IERC20 _l1Usdc,
-        ITokenMessenger _cctpTokenMessenger
-    ) CircleCCTPAdapter(_l1Usdc, _cctpTokenMessenger, CircleDomainIds.Arbitrum) {
+        ITokenMessenger _cctpTokenMessenger,
+        OFTAddressBook _oftAddressBook
+    )
+        CircleCCTPAdapter(_l1Usdc, _cctpTokenMessenger, CircleDomainIds.Arbitrum)
+        OFTTransportAdapter(OFTEIds.Arbitrum, OFT_FEE_CAP)
```
nicholaspai, line 84: I think its better to pas in this OFT_FEE_CAP as a constructor param and then update the deploy script to set this to 1 ether. Just in case we want to change it in the future then we wouldn't have to update the contract

nicholaspai (approved):
LGTM, I just left one comment!

contracts/interfaces/IOFT.sol
```diff
@@ -0,0 +1,83 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.0;
+// ! todo: should this file live under a MIT licence? It's MIT-licenced in LZ code
```
grasphoper, line 3: Please give licence advice!

contracts/Linea_SpokePool.sol
```diff
@@ -16,6 +16,9 @@ import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
 contract Linea_SpokePool is SpokePool {
     using SafeERC20 for IERC20;
 
+    // Linea_SpokePool does not use OFT messaging, setting the cap to 0
+    uint256 private constant OFT_FEE_CAP = 0;
```
nicholaspai, line 20: It might in the future, right? We should probably pass this in the constructor

contracts/Ovm_SpokePool.sol
```diff
@@ -34,6 +34,10 @@ interface IL2ERC20Bridge {
  */
 contract Ovm_SpokePool is SpokePool, CircleCCTPAdapter {
     using SafeERC20 for IERC20;
+
+    // Ovm_SpokePool does not use OFT messaging, setting the cap to 0
+    uint256 private constant OFT_FEE_CAP = 0;
```
nicholaspai, line 39: https://github.com/across-protocol/contracts/pull/902/files#r1991491081

contracts/PolygonZkEVM_SpokePool.sol
```diff
@@ -32,6 +32,9 @@ interface IBridgeMessageReceiver {
 contract PolygonZkEVM_SpokePool is SpokePool, IBridgeMessageReceiver {
     using SafeERC20 for IERC20;
 
+    // PolygonZkEVM_SpokePool does not use OFT messaging, setting the cap to 0
+    uint256 private constant OFT_FEE_CAP = 0;
```
nicholaspai, line 36: https://github.com/across-protocol/contracts/pull/902/files#r1991491081

contracts/ZkSync_SpokePool.sol
```diff
@@ -28,6 +28,9 @@ contract ZkSync_SpokePool is SpokePool {
     // while changing only constructor parameters can lead to substantial fee savings. So, the following params
     // are all set by passing in constructor params where possible.
 
+    // Scroll_SpokePool does not use OFT messaging, setting the cap to 0
+    uint256 private constant OFT_FEE_CAP = 0;
```
nicholaspai, line 32: https://github.com/across-protocol/contracts/pull/902/files#r1991491081

contracts/interfaces/IOFT.sol
nicholaspai: We should just use the same license as our other licenses

contracts/interfaces/IOFT.sol
```diff
@@ -0,0 +1,83 @@
...
+
+/**
+ * @notice This file contains minimal copies of relevant structs / interfaces for OFT bridging.
```
nicholaspai, line 6: nit: can you link the permalink of the source code you copied it from?

contracts/Linea_SpokePool.sol
grasphoper: The thing is, when we support this, we'll have to make changes to `_bridgeTokensToHubPool` function in this same contract. I think it makes sense to add it to constructor then, just not to add additional params to chains that will not support this prematurely. What do you think?

grasphoper: Another solution could be this:

I add OFT code to every contract that "can potentially support OFT" and then in `_bridgeTokensToHubPool` it's going to be behind a zero address check. When we decide to turn it on and support a token, the code will be ready. Hmm, perhaps do this?

grasphoper: Cause currently only Arbitrum_SpokePool / Arbitrum_Adapter are using the OFT logic fully, and we'd need to change code for others to do the same.

grasphoper: Talked off-site about this. Will leave code like it is for now to not grow this PR any more

grasphoper:
Context for @pxrl and @bmzig. Please review :)

### Summary of OFT changes:

- Add `OFTTransportAdapter` : a stateless contract that effectively acts as a library for OFT bridging
- Adapters:
    - modify `Arbitrum_Adapter` to include an OFT logic branch in `relayTokens` function
    - OFT requires token ⇒ IOFT relationship. Stored in a separate(new and simple) `AddressBook` contract because adapters are supposed to be stateless
- SpokePools:
    - Inherit `OFTTransportAdapter` from `SpokePool`; use a mapping for token ⇒ IOFT relationship. This lead to some storage layout changes, please take a look at those
    - Every contract that inherits `SpokePool` now can use `OFT` by adding a simple if branch into `_bridgeTokensToHubPool` function. For now though, only added to 2 pools: Arbitrum and AlephZero(cause it inherits Arbitrum). Aim to support USDT0 transfers Ethereum <> Arbitrum initially
- Modified deploy scripts as necessary:
    - Arbitrum_Adapter, Arbitrum_SpokePool
    - AlephZero_SpokePool
    - (new) AddressBook
- Added tests as necessary:
    - Arbitrum_SpokePool + Adapter. No other pools / adapters fully support OFT yet (do not include OFT branch in their token send logic)
- `contracts/interfaces/IOFT.sol` was copied from layerzero repo, but with different solidity version (we had an incompatibility otherwise)

bmzig (approved):

contracts/libraries/OFTTransportAdapter.sol
```diff
@@ -0,0 +1,110 @@
...
+             * EMPTY_MSG_BYTES, EMPTY_MSG_BYTES, EMPTY_MSG_BYTES here specify `extraOptions`, `composeMsg` and `oftCmd`.
+             * These can be set to empty bytes arrays for the purposes of sending a simple cross-chain transfer.
+             */
+            EMPTY_MSG_BYTES,
+            EMPTY_MSG_BYTES,
+            EMPTY_MSG_BYTES
+        );
+
+        // `false` in the 2nd param here refers to `bool _payInLzToken`. We will pay in native token, so set to `false`
+        MessagingFee memory fee = _messenger.quoteSend(sendParam, false);
```
bmzig, line 98: Very nitty, but we can save a small amount of gas if you load `fee.nativeFee` onto the stack, i.e. `uint256 nativeFee = fee.nativeFee`.

It will save a negligible amount of gas, but I think it doesn't hurt to do.

grasphoper: Ah, I totally missed this before merging. Will bundle this change into the another PR, thank you!

grasphoper: You mean that I later load from memory 3 times in the code below? Yeah, makes sense, I can load once. It's actually a very thorough observation 😄

PR #903, opened 2025-02-25
Fix `yarn build-evm` command on M4 mac

grasphoper:
I deleted hardhat entry from `yarn.lock` and ran `yarn install`. After that, `yarn build-evm` works successfully

It introduces potential compatibility issues, as well as produces contract size warnings which weren't there before.


grasphoper:
Found another solution by using terminal launched via Rosetta to make the master build work. Closing this

PR #906, opened 2025-03-03
Fix type extension issues, like with `HardhatRuntimeEnvironment`

grasphoper:
LSP is sad when we don't provide info about type extensions, as a result a lot of our deployment scripts and testing scripts had unpleasant errors reported by VSCode, or LSPs in general, even though the code is fine.

This PR fixes `HardhatRuntimeEnvironment`. Works on my machine, but please test it on yours too

### DIFF
--- a/tasks/testChainAdapter.ts
+++ b/tasks/testChainAdapter.ts
@@ -20,7 +20,7 @@ task("testChainAdapter", "Verify a chain adapter")
   .setAction(async function (args, hre: HardhatRuntimeEnvironment) {
     const { deployments, ethers, getChainId, network } = hre;
     const provider = new ethers.providers.StaticJsonRpcProvider(network.config.url);
-    const signer = new ethers.Wallet.fromMnemonic(getMnemonic()).connect(provider);
+    const signer = ethers.Wallet.fromMnemonic(getMnemonic()).connect(provider);
 
     const hubChainId = await getChainId();
     const spokeChainId = parseInt(args.chain);
--- a/tasks/testChainAdapter.ts
+++ b/tasks/testChainAdapter.ts
@@ -1,7 +1,7 @@
 import assert from "assert";
 import { getMnemonic } from "@uma/common";
 import { task } from "hardhat/config";
-import { HardhatRuntimeEnvironment } from "hardhat/types";
+import { HardhatRuntimeEnvironment, HttpNetworkConfig } from "hardhat/types";
 import { CHAIN_IDs, TOKEN_SYMBOLS_MAP } from "../utils/constants";
 import { askYesNoQuestion, resolveTokenOnChain } from "./utils";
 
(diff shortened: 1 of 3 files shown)


nicholaspai (approved):

type-extensions/hardhat-extensions.d.ts
```diff
@@ -0,0 +1,2 @@
+import "hardhat-deploy/src/type-extensions.ts";
+import "@nomiclabs/hardhat-ethers";
```
nicholaspai, line 2: Are there any useful docs or comments on why this file is necessary? Is it because `hardhat-deploy` extends `hardhat` and the types get messed up?

grasphoper: My understanding:
- different packages can extend the `HardhatRuntimeEnvironment` type with new fns and properties (originally defined in `hardhat/types`)
- `hardhat-deploy` and `hardhat-ethers` do
- LSP cannot infer the correct extended type for `HardhatRuntimeEnvironment` by itself and needs a hint in a special (declaration) `.d.ts` file

Do you want me to add some comments in code to indicate that?

grasphoper: Ah, apparently just doing the imports in the `type-extensions/hardhat-extensions.d.ts` doesn't work for me 100% of the time. Will investigate further later ...

tasks/testChainAdapter.ts
```diff
@@ -19,8 +19,8 @@ task("testChainAdapter", "Verify a chain adapter")
   .addParam("amount", "Amount to bridge to the destination chain")
   .setAction(async function (args, hre: HardhatRuntimeEnvironment) {
     const { deployments, ethers, getChainId, network } = hre;
-    const provider = new ethers.providers.StaticJsonRpcProvider(network.config.url);
-    const signer = new ethers.Wallet.fromMnemonic(getMnemonic()).connect(provider);
+    const provider = new ethers.providers.StaticJsonRpcProvider((network.config as any).url);
+    const signer = ethers.Wallet.fromMnemonic(getMnemonic()).connect(provider);
```
pxrl, line 23: Good catch, +1

tasks/testChainAdapter.ts
```diff
@@ -19,8 +19,8 @@ task("testChainAdapter", "Verify a chain adapter")
   .addParam("amount", "Amount to bridge to the destination chain")
   .setAction(async function (args, hre: HardhatRuntimeEnvironment) {
     const { deployments, ethers, getChainId, network } = hre;
-    const provider = new ethers.providers.StaticJsonRpcProvider(network.config.url);
-    const signer = new ethers.Wallet.fromMnemonic(getMnemonic()).connect(provider);
+    const provider = new ethers.providers.StaticJsonRpcProvider((network.config as any).url);
```
pxrl, line 22: This suppresses the warnings but it doesn't do anything to improve safety. I'm in two minds about whether we want this.

grasphoper:
Will close this for now, might create a new PR when I actually find a fix that works consistently and is code-editor agnostic

tasks/testChainAdapter.ts
grasphoper: Yeah, I only caught that thanks to the types working ! :)

PR #907, opened 2025-03-05
fix(ZkSync_SpokePool): Add __gap

nicholaspai:
This contract gets extended by the Lens_SpokePool which doesn't add any storage but we should add it in case a future variable gets added to the Lens_SpokePool


bmzig (dismissed):

grasphoper (approved):
LGTM

bmzig (approved):

PR #910, opened 2025-03-05
feat(Linea): Add CCTP V2 support to Linea

nicholaspai:
This PR adds a new CCTPV2 Adapter that we add to the Linea_Adapter and Linea_SpokePool.

CCTP V2 has a different interface than V1 but it has a similar on-chain flow--so we can re-use some of the code.


contracts/Linea_SpokePool.sol
```diff
@@ -29,7 +30,7 @@ contract Linea_SpokePool is SpokePool {
     /**
      * @notice Address of Linea's USDC Bridge contract on L2.
      */
-    IUSDCBridge public l2UsdcBridge;
+    IUSDCBridge private DEPRECATED_l2UsdcBridge;
```
nicholaspai, line 33: I think because this contract is a Child contract and won't get extended, we can delete this storage slot? However this is risky if this contract ever gets extended by a future Child contract so safer to just make it private and preserve the slot

test/evm/hardhat/chain-specific-spokepools/PolygonZkEVM_SpokePool.ts
```diff
@@ -72,11 +72,14 @@ describe("Polygon zkEVM Spoke Pool", function () {
   });
 
   it("Only cross domain owner upgrade logic contract", async function () {
-    const implementation = await hre.upgrades.deployImplementation(await getContractFactory("Linea_SpokePool", owner), {
-      kind: "uups",
-      unsafeAllow: ["delegatecall"],
-      constructorArgs: [weth.address, 60 * 60, 9 * 60 * 60],
-    });
+    const implementation = await hre.upgrades.deployImplementation(
+      await getContractFactory("PolygonZkEVM_SpokePool", owner),
```
nicholaspai, line 76: long-standing typo fixed here

contracts/chain-adapters/Linea_Adapter.sol
```diff
@@ -14,31 +15,29 @@ import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
...
      */
     constructor(
         WETH9Interface _l1Weth,
         IMessageService _l1MessageService,
         ITokenBridge _l1TokenBridge,
-        IUSDCBridge _l1UsdcBridge
-    ) {
+        IERC20 _l1Usdc,
+        ITokenMessenger _cctpTokenMessenger
+    ) CircleCCTPAdapter(_l1Usdc, _cctpTokenMessenger, CircleDomainIds.Linea) {
```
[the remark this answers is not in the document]
nicholaspai: hmm we did this to reduce deployment error chance, but i can see it being helpful. I'm also partial to keeping code consistent between different chain adapters. WDYT?

contracts/Linea_SpokePool.sol
nicholaspai: thats an interesting idea. Are there negative side-effects to keeping this as-is ? I can't think of any harm

pxrl: Actually not since you marked it private. I overlooked that on the first pass. So there's basically no benefit to changing it :+1:

contracts/chain-adapters/Linea_Adapter.sol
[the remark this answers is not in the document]
pxrl: Yeah; it should be consistent across all the CCTP implementations. I think we should change it, but it doesn't have to be done now because we're not actually wanting to do CCTP on testnet at the moment. We can still do it in a follow-up whenever we decide it's time.

pxrl (commented):
This looks fine to me :+1:

pxrl (dismissed):

contracts/Linea_SpokePool.sol
```diff
@@ -139,27 +133,32 @@ contract Linea_SpokePool is SpokePool {
...
             // before the execution of any wrapped token refund leaf. So it must be unwrapped before being sent as a
             // fee to the l2MessageService.
             WETH9Interface(l2TokenAddress).withdraw(amountToReturn + msg.value); // Unwrap into ETH.
             l2MessageService.sendMessage{ value: amountToReturn + msg.value }(withdrawalRecipient, msg.value, "");
         }
         // If the l1Token is USDC, then we need sent it via the USDC Bridge.
-        else if (l2TokenAddress == l2UsdcBridge.usdc()) {
-            IERC20(l2TokenAddress).safeIncreaseAllowance(address(l2UsdcBridge), amountToReturn);
-            l2UsdcBridge.depositTo{ value: msg.value }(amountToReturn, withdrawalRecipient);
+        else if (l2TokenAddress == address(usdcToken) && _isCCTPEnabled()) {
```
bmzig, line 152: Although I'm sure there are plenty of other gas optimizations we can do, wydt about removing the `_isCCTPEnabled` check here so we don't do the extra access to `cctpTokenMessenger`? Linea either has native USDC or USDC.e, but will never have both concurrently, so we may be able to just get rid of this check.

bmzig: I can see this being bad if there are other "Linea-like" L2s that we need to support.

nicholaspai: > I can see this being bad if there are other "Linea-like" L2s that we need to support.

This is one consideration yeah, i'd prefer the extra safety for now. The adapters don't need to be super gas efficient as they're not called that regularly.

contracts/libraries/CircleCCTPAdapter.sol
```diff
@@ -59,12 +67,23 @@ abstract contract CircleCCTPAdapter {
...
     ) {
         usdcToken = _usdcToken;
         cctpTokenMessenger = _cctpTokenMessenger;
         recipientCircleDomainId = _recipientCircleDomainId;
+
+        // Only the CCTP V2 TokenMessenger has a feeRecipient() function, so we use it to
+        // figure out if we are using CCTP V2 or V1. `success` can be true even if the contract doesn't
+        // implement feeRecipient but it has a fallback function so to be extra safe, we check the return value
+        // of feeRecipient() as well.
+        (bool success, bytes memory feeRecipient) = address(cctpTokenMessenger).call(
```
bmzig, line 83: nit: we can use `staticcall` instead.

bmzig (dismissed):

deploy/028_deploy_linea_adapter.ts
```diff
@@ -14,7 +14,9 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
       WETH[chainId],
       L1_ADDRESS_MAP[chainId].lineaMessageService,
       L1_ADDRESS_MAP[chainId].lineaTokenBridge,
-      L1_ADDRESS_MAP[chainId].lineaUsdcBridge,
+      // TODO: Requires USDC Linea address to be added, currently only USDC.e is set for Linea
```
pxrl, line 17: This should be the same address because Linea USDC is Circle's upgradable USDC standard.

contracts/libraries/CircleCCTPAdapter.sol
```diff
@@ -59,12 +67,23 @@ abstract contract CircleCCTPAdapter {
...
         usdcToken = _usdcToken;
         cctpTokenMessenger = _cctpTokenMessenger;
         recipientCircleDomainId = _recipientCircleDomainId;
+
+        // Only the CCTP V2 TokenMessenger has a feeRecipient() function, so we use it to
+        // figure out if we are using CCTP V2 or V1. `success` can be true even if the contract doesn't
+        // implement feeRecipient but it has a fallback function so to be extra safe, we check the return value
+        // of feeRecipient() as well.
+        (bool success, bytes memory feeRecipient) = address(cctpTokenMessenger).staticcall(
+            abi.encodeWithSignature("feeRecipient()")
```
mrice32, line 84: This is something I want to highlight to OZ.

We are doing this to avoid adding an extra constructor argument and, thus, create a larger diff for all the contracts that inherit from this CCTP contract. However, the downside is that this could be broken in the future if circle removes this function.

mrice32 (approved):
LGTM!

deploy/028_deploy_linea_adapter.ts
```diff
@@ -14,7 +14,9 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
...
+      USDC[chainId],
```
pxrl, line 18: Have you tested this via an actual deployment? I think it's dependent on us adding Linea USDC definition to the constants repo.

A short-term workaround would be to import `USDCe` instead; this is OK in this context because the Linea address will be unchanged through the migration.

deploy/029_deploy_linea_spokepool.ts
```diff
@@ -14,11 +14,16 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
...
-    L2_ADDRESS_MAP[chainId].lineaUsdcBridge,
     hubPool.address,
     hubPool.address,
   ];
-  const constructorArgs = [WETH[chainId], QUOTE_TIME_BUFFER, FILL_DEADLINE_BUFFER];
+  const constructorArgs = [
+    WETH[chainId],
+    QUOTE_TIME_BUFFER,
+    FILL_DEADLINE_BUFFER,
+    USDC[chainId],
```
pxrl, line 24: Same as above - I think _for now_ we need `USDCe` instead.

contracts/libraries/CircleCCTPAdapter.sol
```diff
@@ -50,6 +51,13 @@ abstract contract CircleCCTPAdapter {
     /// @custom:oz-upgrades-unsafe-allow state-variable-immutable
     ITokenMessenger public immutable cctpTokenMessenger;
```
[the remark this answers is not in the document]
nicholaspai: See my comment on line 70 why I chose not to do this--to reduce PR diff

deploy/028_deploy_linea_adapter.ts
nicholaspai: No that's why I added a TODO here. this won't work, but I also don't anticipate us using this deploy script until after we've updated const. For now I've updated to USDCe

deploy/029_deploy_linea_spokepool.ts
nicholaspai: f908e3888dd80f1f7c8621966edcbd5de9aae7fe

deploy/028_deploy_linea_adapter.ts
nicholaspai: f908e3888dd80f1f7c8621966edcbd5de9aae7fe

contracts/chain-adapters/Linea_Adapter.sol
```diff
@@ -67,17 +66,15 @@ contract Linea_Adapter is AdapterInterface {
         uint256 amount,
         address to
     ) external payable override {
+        if (l1Token == address(usdcToken) && _isCCTPEnabled()) {
+            _transferUsdc(to, amount);
+        }
         // If the l1Token is WETH then unwrap it to ETH then send the ETH directly
         // via the Canoncial Message Service.
         if (l1Token == address(L1_WETH)) {
```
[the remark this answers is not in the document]
nicholaspai: f908e3888dd80f1f7c8621966edcbd5de9aae7fe

contracts/libraries/CircleCCTPAdapter.sol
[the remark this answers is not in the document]
mrice32: Okay, I'm fine either way. I don't think it would be a huge diff increase, but also not a big inconsistency either.

mrice32 (dismissed):

PR #912, opened 2025-03-06
improve(ERC7683OriginSettler): Specify destinationSettler inside orderData

nicholaspai:
This way the OriginSettler doesn't need to store known DestinationSettler addresses.

Currently, the only way for the OriginSettler to produce a valid `FundsDeposited` event is if the user-specified `destinationSettler` is whitelisted on the origin settler.

### DIFF
--- a/contracts/SpokePool.sol
+++ b/contracts/SpokePool.sol
@@ -1122,12 +1122,21 @@ abstract contract SpokePool is
         bytes calldata originData,
         bytes calldata fillerData
     ) external {
-        if (keccak256(abi.encode(originData, chainId())) != orderId) {
+        if (keccak256(abi.encode(originData, address(this), chainId())) != orderId) {
             revert WrongERC7683OrderId();
         }
 
         // Ensure that the call is not malformed. If the call is malformed, abi.decode will fail.
-        V3SpokePoolInterface.V3RelayData memory relayData = abi.decode(originData, (V3SpokePoolInterface.V3RelayData));
+        (V3SpokePoolInterface.V3RelayData memory relayData, address destinationSettler) = abi.decode(
+            originData,
+            (V3SpokePoolInterface.V3RelayData, address)
+        );
+        if (destinationSettler != address(this)) {
+            // @todo I added in this check because I'm assuming that just because the FillInstructions emitted in the ResolvedCrossChainOrder specify a
+            // destinationSettler address, that doens't prevent the filler from being that fill to a different destinationSettler. So, we
+            // enforce that the correct destinationSettler is used here.
+            revert InvalidDestinationSettler();
+        }
         AcrossDestinationFillerData memory destinationFillerData = abi.decode(
             fillerData,
             (AcrossDestinationFillerData)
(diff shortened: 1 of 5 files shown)


contracts/erc7683/ERC7683OrderDepositor.sol
```diff
@@ -244,7 +244,7 @@ abstract contract ERC7683OrderDepositor is IOriginSettler {
         relayData.message = acrossOrderData.message;
         fillInstructions[0] = FillInstruction({
             destinationChainId: SafeCast.toUint64(acrossOrderData.destinationChainId),
-            destinationSettler: _destinationSettler(acrossOrderData.destinationChainId).toBytes32(),
+            destinationSettler: acrossOrderData.destinationSettler.toBytes32(),
```
nicholaspai, line 247: @mrice32 how do we enforce the correct destinationSettler is called with [fill()](https://github.com/across-protocol/contracts/blob/466127e11673fb662e6c2cf3a8436c1606269ed3/contracts/SpokePool.sol#L1131)? The fill function doesn't seem to check destinationSettler at all.

nicholaspai: here's my workaround for now: 87261babad57945fde7207b6b5efbaea3ba4c737

This should prevent the filler from calling fill() on the incorrect destinationSettler.

mrice32: Hmmmm, this raises an interesting point and may make this strategy not work.

If the user specifies a malicious destination settler contract, the filler could be exploited. Is it the origin settler's responsibility to validate that if the filler correctly follows the fill instructions that they will be repaid? I think yes, but if we allow the order to specify an arbitrary destination settler that is unverifiable, then it's impossible to do that verification...

Thoughts?

nicholaspai: I do think the filler shouldn't blindly submit calldata on the destination chain, right? They should probably be aware of known destination settlers, and then only fill deposits that target those.

contracts/SpokePool.sol
```diff
@@ -1122,12 +1122,21 @@ abstract contract SpokePool is
...
         // Ensure that the call is not malformed. If the call is malformed, abi.decode will fail.
-        V3SpokePoolInterface.V3RelayData memory relayData = abi.decode(originData, (V3SpokePoolInterface.V3RelayData));
+        (V3SpokePoolInterface.V3RelayData memory relayData, address destinationSettler) = abi.decode(
+            originData,
+            (V3SpokePoolInterface.V3RelayData, address)
+        );
+        if (destinationSettler != address(this)) {
+            // @todo I added in this check because I'm assuming that just because the FillInstructions emitted in the ResolvedCrossChainOrder specify a
+            // destinationSettler address, that doens't prevent the filler from being that fill to a different destinationSettler. So, we
+            // enforce that the correct destinationSettler is used here.
```
nicholaspai, line 1137: This check might be over-engineering. Currently we don't check that the `destinationSettler` is correct at all, so I think we can probably keep the same logic and assume the filler has done its job verifying the `destinationSettler` emitted in the `ResolvedOrder`'s `FillInstructions`

nicholaspai:
Closing this because we should eventually move this contract into the SpokePool so that the cross chain admin can just set the destinationSettlers to protect fillers a bit

PR #914, opened 2025-03-08
feat: add xERC20 standard support via Hyperlane

grasphoper:
### Summary

This PR adds support for xERC20 cross-chain messaging via Hyperlane. This PR is quite big, because we want to support multiple chains from the get-go: Arbitrum, Optimism, Base, Blast, Linea, Mode, Unichain. This is required to support all [ezETH](https://app.renzoprotocol.com/bridge) chain routes for the chains that we already support.

For every chain in the list, I introduced changes to both the chain adapter and the chain-specific spoke pool. However, more spoke pools were modified, because `xERC20` functionality was added to the base `SpokePool` class, which all spoke pools inherit.

A large chunk of the code is adding tests for the new changes (only chains listed above).

### Main changes:
  - `HypXERC20Adapter.sol`: main logic for sending cross-chain transfers via Hyperlane’s xERC20 router.
  - `SpokePool.sol`: `SpokePool is HypXERC20Adapter` and required storage logic: `mapping(L2 token => matching Hyperlane router)`.
  - chain-specific SpokePools: Every SpokePool had to change because of SpokePool constructor changes. Most SpokePools implemented a simple additional `if` branch in `_bridgeTokensToHubPool` function. Others just hardcoded non-support of xERC20 for now.
  - storage layouts of all spokepools changed because of this added mapping on base `SpokePool` class
  - chain-adapters: all adapters for chains listed abouve in `ezETH` list implemented a xERC20 if branch in `relayTokens` function.
  - Deployment scripts: a lot of them had to change because of constructor changes of spoke pools / adapters.
  - Tests: added xERC20 test for all spokes / adapters for ezETH chains. Adjusted multiple unrelated tests, that broke because of constructor changes.
     
### Quality of life improvements:
  - adjust `yarn test-evm` to include local `foundry` tests.
  - added `yarn test-evm-hardhat` and `yarn test-evm-foundry` commands. For a foundry command, created a new foundry profile that only runs local tests as opposed to both local and fork ones (mirror CI logic)
  - added `yarn evm-contract-sizes` script. I use the command and always forget the syntax. Now there’s no need to remember (also handy picking of optimizer runs).

Closes ACX-3889


contracts/Ovm_SpokePool.sol
```diff
@@ -32,7 +33,7 @@ interface IL2ERC20Bridge {
  * @notice OVM specific SpokePool. Uses OVM cross-domain-enabled logic to implement admin only access to functions. * Optimism, Base, and Boba each implement this spoke pool and set their chain specific contract addresses for l2Eth and l2Weth.
  * @custom:security-contact bugs@across.to
  */
-contract Ovm_SpokePool is SpokePool, CircleCCTPAdapter {
+contract Ovm_SpokePool is SpokePool, CircleCCTPAdapter, HypXERC20Adapter {
```
mrice32, line 36: @grasphoper @nicholaspai just out of curiosity, do you think it would be better if we inherited CCTP, xERC20, etc at SpokePool rather than in these derived classes? It might reduce boilerplate, but it would also mean it's impossible to have a SpokePool _without_ this functionality (that isn't strictly required for anything the SP does).

grasphoper: We discussed it with Nick a bit

I think we could go this route for HypXERC20Adapter, as not many pools are not using it (there are like 5)

As for the OFT for example, it's only there on 2 chains for now: Ethereum and Arbitrum. I think it may be overkill to add. Also, will make our lives harder regarding bytecode sizes

But for `HypXERC20Adapter` it might really be better to reduce boilerplate

utils/utils.hre.ts
```diff
@@ -37,7 +37,7 @@ export async function getSpokePoolDeploymentInfo(
   return { hubPool, hubChainId, spokeChainId };
 }
 
-type FnArgs = number | string;
+type FnArgs = number | string | BigNumber;
```
grasphoper, line 40: Checked that this works by deploying `ArbitrumAdapter` on Sepolia with current deploy script

contracts/libraries/AdapterStore.sol
```diff
@@ -3,38 +3,64 @@ pragma solidity ^0.8.0;
...
+import { IHypXERC20Router } from "../interfaces/IHypXERC20Router.sol";
 
 /**
- * @dev A helper contract for chain adapters that support OFT or XERC20 messaging on L1
+ * @dev A helper contract for chain adapters that support OFT or XERC20 messaging(via Hyperlane) on L1
  * @dev Handles token => messenger/router mapping storage, as adapters are called via delegatecall and don't have relevant storage space
  */
 contract AdapterStore is Ownable {
     mapping(uint256 => mapping(address => address)) public oftMessengers;
+    mapping(uint256 => mapping(address => address)) public hypXERC20Routers;
```
nicholaspai, line 14: Let's imagine a world where we want to add a mapping to this contract state. How do you think we should go about that?

Let's say that we have some new bridge system that we want to add to the Arbitrum_Adapter. We'd have to construct the Arbitrum_Adapter with a new AdapterStore and migrate all `hypXERC20Routers` over the new store. This seems super painful.

WDYT about making this contract upgradeable, or making the `mapping` more dynamic:
`mapping(bytes32 => mapping(uint256 => mapping(address => address)))`

This is very verbose, but we can then create new L1 mappings pretty easily:
`setMapping(bytes32("hypXERC20", l2ChainId, l1Token, router)`
`setMapping(bytes32("oftMessenger",l2ChainId,l1Token, messenger)`
`setMapping(bytes32("futureBridgingSystem", l2ChainId, l1Token, router)`
...

nicholaspai (commented):
All of the package.json/foundry.toml additions are great, and thanks for adding more unit tests!

I left a couple comments but this looks very close!

contracts/libraries/HypXERC20Adapter.sol
```diff
@@ -0,0 +1,94 @@
...
+        // Quote the gas payment required for the transfer
+        uint256 fee = _router.quoteGasPayment(HYP_DST_DOMAIN);
+        if (fee > HYP_FEE_CAP) revert HypFeeCapExceeded();
+        if (fee > address(this).balance) revert HypInsufficientBalanceForFee();
+
+        // Approve the exact _amount for `_router` to spend
+        _token.forceApprove(address(_router), _amount);
+
+        // Send the transfer via Hyperlane. Return value is not useful to check
+        _router.transferRemote{ value: fee }(HYP_DST_DOMAIN, to, _amount);
```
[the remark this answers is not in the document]
grasphoper: It would make sense if it reverted. Will look into their implementation just to be sure.

contracts/libraries/AdapterStore.sol
grasphoper: I'm actually totally wrong about a single layer map. It needs to be 2 layer still of course. `mapping(bytes32 => mapping(address => address)`. So hash(dstChainId, "oft"/etc) + token

With your solution(3 layer map) we get clearer logs.

contracts/libraries/HypXERC20Adapter.sol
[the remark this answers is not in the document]
grasphoper: I actually think it doesn't revert. It just checks the fee, and then sends the exact same amount. I don't think they subtract from the amount anywhere in their code (still looking).

contracts/libraries/AdapterStore.sol
grasphoper: Here's a little table for pros and cons.

| Option                          | Pros                                                                                         | Cons                                              |
|---------------------------------|----------------------------------------------------------------------------------------------|---------------------------------------------------|
| **Upgradable contract**         | ✅ Clear logs <br> ✅ Human error protection                                                 | ❌ Harder to manage than a regular contract      |
| **3-layer mapping**             | ✅ Clear logs <br> ✅ Deploy once and “forget”                                              | ❌ No human error protection                      |
| _Example:_                      | _`mapping(bytes32 => mapping(uint256 => mapping(address => address)))`_ <br> _E.g. `"hypXERC20" => l2ChainId => l1Token => messengerAddr`_ | |
| **2-layer mapping**             | ✅ Less verbose mapping <br> ✅ Deploy once and “forget”                                    | ❌ Less clear logs <br> ❌ No human error protection |
| _Example:_                      | _`mapping(bytes32 => mapping(address => address))`_ <br> _E.g. `keccak256("oftMessenger", adapterChainId) => l1Token => messengerAddr`_ | |

What do you think? @nicholaspai @pxrl @bmzig @mrice32 
I am kind of leaning towards having human error protection with an upgradable contract

contracts/AdapterStore.sol
```diff
@@ -20,6 +22,11 @@ contract AdapterStore is Ownable {
     error HypTokenMismatch();
     error ArrayLengthMismatch();
 
+    function initialize() public initializer {
+        __Ownable_init();
+        __UUPSUpgradeable_init();
+    }
```
grasphoper, line 28: Honestly, I am still not sure whether I need to `_disableInitializers` in constructor. I don't think so, but after reading docs about it for some time I still don't 100% understand what it does

test/evm/foundry/local/Forwarder.t.sol
```diff
@@ -59,6 +59,7 @@ contract ForwarderTest is Test {
         crossDomainMessenger = new MockBedrockCrossDomainMessenger();
         standardBridge = new MockBedrockL1StandardBridge();
 
+        // A placeholder. Not initialized properly as a proxy contract. Not relevant for this test
         adapterStore = new AdapterStore();
```
grasphoper, line 63: I think it's good enough for this test, right?

utils/utils.hre.ts
```diff
@@ -53,8 +53,8 @@ export async function deployNewProxy(
     unsafeAllowArgs.push("state-variable-immutable");
   }
 
-  // If a SpokePool can be found in deployments/deployments.json, then only deploy an implementation contract.
-  const proxy = getDeployedAddress("SpokePool", chainId, false);
+  // Look for an already deployed proxy with `name` in deployments/deployments.json. If it's found, only deploy an implementation contract.
+  const proxy = getDeployedAddress(name, chainId, false);
```
grasphoper, line 57: Not sure why we were checking "SpokePool" here. Changed it to check `name`. Am I missing something?

contracts/libraries/AdapterStore.sol
grasphoper: @nicholaspai commited the upgradeable contract here [7ef6170](https://github.com/across-protocol/contracts/pull/914/commits/7ef61700d1bda1ea03601fdcb36efc63770ebee2)

contracts/AdapterStore.sol
```diff
@@ -0,0 +1,100 @@
...
+
+    function setOFTMessenger(
+        uint256 adapterChainId,
+        address l1Token,
+        address oftMessenger
+    ) external onlyOwner {
+        _setOFTMessenger(adapterChainId, l1Token, oftMessenger);
+    }
+
+    function batchSetOFTMessengers(
```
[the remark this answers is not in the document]
grasphoper: Just the lower gas cost. I was deciding between the two. There was 'bytecode space' to make these fns explicit, that's why I chose to do it like this. Also, it's L1 gas

contracts/libraries/AdapterStore.sol
nicholaspai: I actually think the upgradeable version is too complex and we only need to change the mapping struct. I like the three layer mapping personally because its easier to reason about.

contracts/AdapterStore.sol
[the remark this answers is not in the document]
nicholaspai: yeah i like adding a gas optimized batch function. this contract is expected to be simple with just two functions so i like this

utils/utils.hre.ts
nicholaspai: you should revert this this will break a lot of things. Look at [here](https://github.com/across-protocol/contracts/blob/95c4f923932d597d3e63449718bba5c674b084eb/deployments/deployments.json#L10) where we name each spoke pool as just SpokePool for convenience

contracts/libraries/AdapterStore.sol
grasphoper: I'll revert and add a 3-layer mapping instead then

grasphoper: added in [b3face7](https://github.com/across-protocol/contracts/pull/914/commits/b3face70b1c4cbd11e088095c30d1c39c8b32f45)

contracts/libraries/HypXERC20Adapter.sol
```diff
@@ -0,0 +1,94 @@
...
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+
+/**
+ * @notice List of Hyperlane domain ids for different chains.
+ * @dev source https://github.com/hyperlane-xyz/hyperlane-registry
+ * @dev they are mostly the same as chain ids, but not always. So double-check in the repo.
+ */
+library HyperlaneDomainIds {
+    uint32 public constant Ethereum = 1;
```
mrice32, line 16: @nicholaspai I know this follows the same pattern as CCTP, but why do we hardcode these in the contracts rather than making them a constructor param, so we don't have to modify contracts to add new ones or reuse a contract on a diff chain?

At some point, we want to move away from making a fresh contract for each chain we deploy to, so we'll need to move away from hardcoding chain identifiers and other config params, no?

mrice32 (commented):
Looks great! One minor comment on domain ids.

contracts/libraries/HypXERC20Adapter.sol
grasphoper: These are used in 2 places:

1. Create chain adapter. (this one illustrates a point so i skip 2.)

Currently, we have 1 adapter per supported chain.
Let's say we move away from this and have instead a generic adapter that we can deploy on many different chains, like the SP1Helios one.

We can expose `_hyperlaneDomainId` as a constructor param there. This doesn't really contradict the `library` we have here. This library is like a "constants helper lookbook" rather than something we hardcode into our generic contracts.

I.e. our `Base_Adapter` has `HyperlaneDomainIds.Base` hardcoded in constructor, but it makes sense, cause that adapter is not meant to be generic. I don't think we want to expose `_hyperlaneDomainId` as a constructor param there.

IMO, using such enums is just "a better way to specify what this param means"(compared to say, using a constant literal and a comment.

Wdythink?

nicholaspai: I'm not sure, I see Matt's point that we want to make this more generic and not require changes to this library, but I also want to prevent programmer error. in general though, I think we should be ok with assuming the programmer won't deploy the contract with an error since we (1) have peer-reviewed deploy scripts and (2) rarely deploy these contracts. So I lean towards removing these libraries.

nicholaspai (commented):
In the future I would recommend adding the deploy script changes AFTER the contract changes, to make reviewing easier.

scripts/evm-contract-sizes.sh
```diff
@@ -0,0 +1,32 @@
...
+# Function to restore the original file
+restore_config() {
+    mv foundry.toml.bak foundry.toml
+}
+
+# Set up trap to restore file on script exit (success or failure)
+trap restore_config EXIT
+
+# Modify the optimizer runs in foundry.toml
+sed -i '' "s/optimizer_runs = .*/optimizer_runs = $OPTIMIZER_RUNS/" foundry.toml
```
nicholaspai, line 26: Why don't you use the `--optimize` flag that is [native to `forge build`](https://book.getfoundry.sh/reference/forge/forge-build#the-optimizer) instead of modifying the file?

deploy/028_deploy_linea_adapter.ts
```diff
@@ -1,11 +1,25 @@
...
   const { deployer } = await hre.getNamedAccounts();
   const chainId = parseInt(await hre.getChainId());
 
+  assert(
+    chainId == CHAIN_IDs.MAINNET,
+    "We only support deploying Linea Adapter on Mainnet for now. To deploy on testnet, update consts and configs."
+  );
+
+  // Pick correct destination chain id to set based on deployment network
+  const dstChainId = CHAIN_IDs.MAINNET ? CHAIN_IDs.LINEA : undefined;
```
nicholaspai, line 18: this seems like a typo since `CHAIN_IDs.MAINNET` is guaranteed to exist. Should this just be `dstChainId = CHAIN_IDs.LINEA`?

contracts/chain-adapters/Mode_Adapter.sol
```diff
@@ -53,9 +67,12 @@ contract Mode_Adapter is CrossDomainEnabled, AdapterInterface, CircleCCTPAdapter
             ITokenMessenger(address(0)),
             CircleDomainIds.UNINITIALIZED
         )
+        HypXERC20Adapter(34443, _hypXERC20FeeCap)
```
nicholaspai, line 70:
```suggestion
        HypXERC20Adapter(_dstChainId, _hypXERC20FeeCap)
```

contracts/chain-adapters/Optimism_Adapter.sol
```diff
@@ -61,19 +69,28 @@ contract Optimism_Adapter is CrossDomainEnabled, AdapterInterface, CircleCCTPAda
...
         IERC20 _l1Usdc,
-        ITokenMessenger _cctpTokenMessenger
+        ITokenMessenger _cctpTokenMessenger,
+        uint256 _dstChainId,
+        AdapterStore _adapterStore,
+        uint256 _hypXERC20FeeCap
     )
         CrossDomainEnabled(_crossDomainMessenger)
         CircleCCTPAdapter(_l1Usdc, _cctpTokenMessenger, CircleDomainIds.Optimism)
+        HypXERC20Adapter(10, _hypXERC20FeeCap)
```
nicholaspai, line 88: Is it 10 even for testnets?

nicholaspai (commented):
left comments

contracts/libraries/HypXERC20Adapter.sol
nicholaspai: The Ethereum ID is still useful to keep hardcoded here, by the way. We always will use the Ethereum ID when creating an L2 SpokePool.

contracts/chain-adapters/Arbitrum_Adapter.sol
```diff
@@ -57,7 +58,7 @@ contract Arbitrum_Adapter is AdapterInterface, CircleCCTPAdapter, OFTTransportAd
     // Generic gateway: https://github.com/OffchainLabs/token-bridge-contracts/blob/main/contracts/tokenbridge/ethereum/gateway/L1ArbitrumGateway.sol
     ArbitrumL1ERC20GatewayLike public immutable L1_ERC20_GATEWAY_ROUTER;
 
-    // Helper storage contract to help this Adapter map token => OFT messenger for OFT bridging.
+    // Helper storage contract to support bridging via differnt token standards: OFT, XERC20
     AdapterStore public immutable ADAPTER_STORE;
```
nicholaspai, line 62: curious, why isn't ADAPTER_STORE part of the OFT_Adapter and HYP_Adapter contracts?

grasphoper: Because SpokePools also inherit those adapters, and they have a differnt storage mechanism

nicholaspai: but this variable is immutable wouldn't that be fine to include in the OFT/HYP adapters?

deploy/028_deploy_linea_adapter.ts
grasphoper: wow, good catch! Was meant to be this `const dstChainId = chainId == CHAIN_IDs.MAINNET ? CHAIN_IDs.LINEA : undefined;`

scripts/evm-contract-sizes.sh
grasphoper: Didn't know about that flag. I wonder if it'll override the config file, or the other way around. I'll try it

contracts/chain-adapters/Arbitrum_Adapter.sol
grasphoper: Well, we wouldn't be reading storage vals from it anyway in e.g. `_transferViaOFT`, we'd have to still read oftMessenger value externally from each adapter / spoke.
For spokes we'd have to set it to zeroaddr

Having it fully external feels marginally better to me

nicholaspai: by default its equal to zero address, and one benefit of having it in the OFTAdapter is we don't need to re-implement `_getOftMessenger()` in every single adapter that uses OFTs. so I lean the other way

nicholaspai: can you do:
- set ADAPTER_STORE in the OFTTransport_Adapter
- In `_transferViaOFT` read `address messenger = ADAPTER_STORE.crossChainMessengers(keccak256("oft"), DST_EID, token)`

This way you don't need to the child contract (e.g. UniversalAdapter or ArbitrumAdapter) to have to implement `_getOftMessenger()` and you don't need the child to have an ADAPTER_STORE and DESTINATION_CHAIN_ID value.

grasphoper: My biggest concern is the changes to `SpokePool` mechanics. We'd have

1.
```
mapping(bytes32 => mapping(uint256 => mapping(address => address))) public crossChainMessengers;
// instead of
mapping(address => address) public oftMessengers;
mapping(address => address) public hypXERC20Routers;
```
makes the code less clear but that's ~okay

2.
implement `setMessenger` function similar to `AdapterStore`

3.
On constructor, we'd pass in `address(this)` to `OFTTransportAdapter` to "act as AdapterStore"

Meaning we're introducing this "call to self" in the `_transferViaOft functionality`

It would be smth like
```
Arbitrum_SpokePool._bridgeTokensToHubPool => OFTTransportAdapter._transferViaOFT => SpokePool.crossChainMessengers(...)
```

Meaning we'd get crossChainMessenger through an ~external?/public call to self. I'm not sure how exactly that'd resolve in Solidity, I'm guessting that would count for a cross-cotract call even though it's to the same address

This flow feels a bit ugly to me, but if we're okay with this than I can change to it yeah. We'd reduce some boilerplate immutable vars like this for sure

contracts/AdapterStore.sol
```diff
@@ -0,0 +1,61 @@
...
+    bytes32 public constant OFT_MESSENGER = bytes32("OFT_MESSENGER");
+    bytes32 public constant HYP_XERC20_ROUTER = bytes32("HYP_XERC20_ROUTER");
+}
+
+/**
+ * @dev A helper contract for chain adapters that support OFT or XERC20 messaging(via Hyperlane) on L1
+ * @dev Handles token => messenger/router mapping storage, as adapters are called via delegatecall and don't have relevant storage space
+ */
+contract AdapterStore is Ownable {
+    // (messengerType, l2ChainId, l1Token) => messenger address
```
nicholaspai, line 16:
```suggestion
    // (messengerType, dstDomainId, l1Token) => messenger address
```

nicholaspai: Also, is the `l1Token` param always an L1 token? Could it be an L2 token when going from L2 to L1?

contracts/AdapterStore.sol
```diff
@@ -0,0 +1,61 @@
...
+    mapping(bytes32 => mapping(uint256 => mapping(address => address))) public crossChainMessengers;
+
+    event MessengerSet(
+        bytes32 indexed messengerType,
+        uint256 indexed l2ChainId,
+        address indexed l1Token,
```
nicholaspai, line 22: Same comment for [l1Token](https://github.com/across-protocol/contracts/pull/914/files#r2012272807)

deploy/002_deploy_optimism_adapter.ts
```diff
@@ -10,12 +11,19 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
   const chainId = parseInt(await hre.getChainId());
   const opStack = OP_STACK_ADDRESS_MAP[chainId][SPOKE_CHAIN_ID];
 
+  // Set the Hyperlane xERC20 destination domain based on the chain https://github.com/hyperlane-xyz/hyperlane-registry/tree/main/chains
+  const hypXERC20DstDomain = chainId == CHAIN_IDs.MAINNET ? 10 : 11155420;
```
nicholaspai, line 15: these domain ID's are so ugly

deploy/028_deploy_linea_adapter.ts
```diff
@@ -1,11 +1,23 @@
...
   const { deployer } = await hre.getNamedAccounts();
   const chainId = parseInt(await hre.getChainId());
 
+  assert(
+    chainId == CHAIN_IDs.MAINNET,
+    "We only support deploying Linea Adapter on Mainnet for now. To deploy on testnet, update consts and configs."
+  );
+
+  // Set the Hyperlane xERC20 destination domain based on the chain https://github.com/hyperlane-xyz/hyperlane-registry/tree/main/chains
+  const hypXERC20DstDomain = chainId == CHAIN_IDs.MAINNET ? 59144 : undefined;
```
nicholaspai, line 18: this should never be able to be undefined, because of the assert right? If this is because typescript is complaining then we should fix it.

deploy/051_deploy_worldchain_spokepool.ts
```diff
@@ -2,16 +2,26 @@ import { DeployFunction } from "hardhat-deploy/types";
...
   const initArgs = [
     1,
     // Set hub pool as cross domain admin since it delegatecalls the Adapter logic.
     hubPool.address,
     hubPool.address,
   ];
+
+  // Set the Hyperlane xERC20 destination domain based on the chain
+  // https://github.com/hyperlane-xyz/hyperlane-registry/tree/main/chains
+  const hypXERC20DstDomain = hubChainId == CHAIN_IDs.MAINNET ? 1 : 11155111;
```
nicholaspai, line 20: I just realized how often you're repeating this entire line. I recommend exposing a public `getHypXERC20DstDomain()` utility function, and also exposing a constants library (perhaps in deploy/consts) for the Hyperlane Destination domains.

Same suggestions for OFT's of course

test/evm/hardhat/chain-adapters/Arbitrum_Adapter.ts
```diff
@@ -311,4 +337,43 @@ describe("Arbitrum Chain Adapter", function () {
     expect(oftMessenger.send).to.have.been.calledOnce;
     expect(oftMessenger.send).to.have.been.calledWith(sendParam, msgFeeStruct, hubPool.address);
   });
+
+  it("Correctly calls Hyperlane XERC20 bridge", async function () {
+    // Set hyperlane router in adapter store
+    hypXERC20Router.wrappedToken.returns(ezETH.address);
```
nicholaspai, line 343: this is just a nit but I'm pretty sure this test would work if you pretend `weth` is "ezEth". In other words, I don't think you need all that setup code to deploy ezEth and configure it.

contracts/AdapterStore.sol
```diff
@@ -0,0 +1,61 @@
...
+ * @dev A helper contract for chain adapters that support OFT or XERC20 messaging(via Hyperlane) on L1
+ * @dev Handles token => messenger/router mapping storage, as adapters are called via delegatecall and don't have relevant storage space
+ */
+contract AdapterStore is Ownable {
+    // (messengerType, l2ChainId, l1Token) => messenger address
+    mapping(bytes32 => mapping(uint256 => mapping(address => address))) public crossChainMessengers;
+
+    event MessengerSet(
+        bytes32 indexed messengerType,
+        uint256 indexed l2ChainId,
```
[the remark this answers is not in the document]
grasphoper: Yeah, good point, `dstDomainId` is more fitting now

deploy/002_deploy_optimism_adapter.ts
grasphoper: What do you mean exactly? 😄

deploy/028_deploy_linea_adapter.ts
grasphoper: I did it like this so it's clear what should happen when we actually wanna enable the testnet. The person who's enabling it would just substitute `undefined` for a number. But I can just have `const hypXERC20DstDomain = 59144`

test/evm/hardhat/chain-adapters/Arbitrum_Adapter.ts
grasphoper: Yeah, I did something like that initially. I was just worried that my test becomes entangled with the token I'm using as "fake ezETH"

Let's say down the line we add some if branch:

```
if token == ETH {
bridge_in_a_special_way_for_ETH();
} else if token == ezETH {
bridge_via_xERC20()
}
```

Setting it up like this avoids that entanglement. Although adds a bit more code

test/evm/hardhat/chain-specific-spokepools/Optimism_SpokePool.ts
```diff
@@ -53,7 +66,7 @@ describe("Optimism Spoke Pool", function () {
       { kind: "uups", unsafeAllow: ["delegatecall"], constructorArgs: [weth.address] }
     );
 
-    await seedContract(optimismSpokePool, relayer, [dai], weth, amountHeldByPool);
+    await seedContract(optimismSpokePool, relayer, [dai, l2EzETH], weth, amountHeldByPool);
```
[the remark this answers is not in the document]
grasphoper: Yeah, I did it initially. [this](https://github.com/across-protocol/contracts/pull/914#discussion_r2012760450) is why I didn't keep it

contracts/AdapterStore.sol
```diff
@@ -0,0 +1,62 @@
...
+    ) external onlyOwner {
+        _setMessenger(messengerType, dstDomainId, srcChainToken, srcChainMessenger);
+    }
+
+    function batchSetMessengers(
+        bytes32 messengerType,
+        uint256[] calldata dstDomainIds,
+        address[] calldata srcChainTokens,
+        address[] calldata srcChainMessengers
+    ) external onlyOwner {
```
grasphoper, line 43: something I thought about and left like this:

we could be accepting `bytes32[] messengerType` instead of `bytes32 messengerType`. It would allow us to set multiple messengers from different messaging protocols together. Having it set like this creates a bit of logical separation from setting messangers of differnt types. E.g. you have to make 2 separate calls to set OFT, xERC20 messengers. I think that's okay, but open to changing it

deploy/002_deploy_optimism_adapter.ts
nicholaspai: ha not your fault, just that they're very long and i wish they [hyperlane/lz] just used the destination chain ID...

deploy/028_deploy_linea_adapter.ts
nicholaspai: it's just weird to me to have a ternary that will never have the second condition ever yet. In this case I'd usually leave a comment like "TODO: change this line to .... when testnet is available"

test/evm/hardhat/chain-adapters/Arbitrum_Adapter.ts
nicholaspai: ok doesn't matter to me really

contracts/AdapterStore.sol
nicholaspai: i think you should accept `bytes32[] messengerType` for the reasons you mentioned

package.json
```diff
@@ -33,16 +33,19 @@
...
     "test": "yarn test-evm && yarn test-svm",
     "test:report-gas": "IS_TEST=true REPORT_GAS=true hardhat test",
+    "evm-contract-sizes": "sh scripts/evm-contract-sizes.sh",
     "generate-evm-assets": "rm -rf typechain && TYPECHAIN=ethers yarn hardhat typechain",
     "prepublish": "yarn build && hardhat export --export-all ./cache/massExport.json && ts-node ./scripts/processHardhatExport.ts && prettier --write ./deployments/deployments.json && yarn generate-evm-assets",
     "pre-commit-hook": "sh scripts/preCommitHook.sh"
   },
   "dependencies": {
-    "@across-protocol/constants": "^3.1.37",
+    "@across-protocol/constants": "3.1.43",
```
nicholaspai, line 48:
```suggestion
    "@across-protocol/constants": "^3.1.43",
```

why did you remove the ^?

nicholaspai (commented):
I resolved all addressed comments. There a few outstanding, but also left a new one on the package.json change.

package.json
grasphoper: Ah, it's because of the command I called `yarn add @across-protocol/constants@3.1.43`. Should have done this instead `yarn add @across-protocol/constants@^3.1.43`

nicholaspai (approved):
LGTM

PR #916, opened 2025-03-11
feat: Add universal adapters and spoke pools

nicholaspai:
Can be used to relay messages from L1 to L2 spoke pools using verification + light client contracts on the L2. This PR supports two different L1 state validity proving tech stacks:
- SP1 + Helios light client
- RiscZero + Helios light client


contracts/SP1_SpokePool.sol
```diff
@@ -0,0 +1,168 @@
...
+    address public immutable hubPool;
+
+    /// @notice The address of the SP1 verifier contract.
+    address public verifier;
+
+    /// @notice The address of the Helios light client contract.
+    address public helios;
+
+    /// @notice The verification key for the acrossCall program.
+    bytes32 public acrossCallProgramVKey;
```
mrice32, line 39: Should these be immutable? Not a leading question -- just leaving it here to make sure we consider whether we need these to be updatable without a redeploy.

contracts/chain-adapters/SP1_Adapter.sol
```diff
@@ -0,0 +1,59 @@
...
+     * @notice No-op relay tokens method.
+     */
+    function relayTokens(
+        address,
+        address,
+        uint256,
+        address
+    ) external payable override {
+        // This method is intentionally left as a no-op.
+        // If the adapter is intended to be able to relay tokens, this method should be overridden.
```
mrice32, line 57: +1 great point on overriding

We may want to include CCTP, OFTs, xERC20, etc (all the token standards we support) in some derived class in case we want to enable some token rebalancing routes to these chains where canonical bridges exist.

contracts/chain-adapters/SP1_Adapter.sol
```diff
@@ -0,0 +1,59 @@
...
+        DATA_STORE = _store;
+    }
+
+    /**
+     * @notice Saves root bundle data in a simple storage contract that can be proven and relayed to L2.
+     * @param target Contract on the destination that will receive the message.
+     * @param message Data to send to target.
+     */
+    function relayMessage(address target, bytes calldata message) external payable override {
+        DATA_STORE.storeData(target, abi.encode(address(this), target, message));
```
mrice32, line 43: Something we should discuss: to keep the cost of sending roots low as we scale to many more chains, we may want to specifically store roots in a way that is target-agnostic (but we would need to differentiate this from admin messages) to avoid storing it for each chain.

One way this could work is that we check the function signature of the admin message. If it's relayRootBundle, we store it without the target, and then each call for other succinct chains will not need to store this value.

Thoughts?

contracts/SP1_SpokePool.sol
```diff
@@ -0,0 +1,168 @@
...
+        verifier = _verifier;
+        helios = _helios;
+        acrossCallProgramVKey = _acrossCallProgramVKey;
+        hubPool = _hubPool;
+    }
+
+    /**
+     * @notice This can be called by an EOA to relay call data stored on the HubPool on L1 into this contract.
+     * @dev Consider making this an onlyOwner function so that only a priveleged EOA can relay state. This EOA
+     * wouldn't be able to tamper with the _publicValues.
```
mrice32, line 98: Just OOC, why do you think this should be privileged?

contracts/SP1_SpokePool.sol
```diff
@@ -0,0 +1,168 @@
...
+        ContractPublicValues memory publicValues = abi.decode(_publicValues, (ContractPublicValues));
+
+        // Verify Helios light client is aware of the state root containing the public values:
+        bytes32 executionStateRoot = IHelios(helios).executionStateRoots(_head);
+        if (executionStateRoot != publicValues.stateRoot) {
+            revert StateRootMismatch();
+        }
+
+        // Validate state is intended to be sent to this contract:
+        (address _hubPool, address _target, bytes memory _message) = abi.decode(
```
mrice32, line 120: Would it be the `DataStore` address that we should be validating against since that's where the data is stored on mainnet (not the `HubPool`)?

contracts/chain-adapters/SP1_Adapter.sol
```diff
@@ -0,0 +1,59 @@
...
+
+    /**
+     * @notice No-op relay tokens method.
+     */
+    function relayTokens(
+        address,
+        address,
+        uint256,
+        address
+    ) external payable override {
```
mrice32, line 55: To make sure this can't mess up the HubPool accounting, should this revert rather than no-op to ensure even an erroneous bundle can't attempt to tokens through this implementation?

contracts/SP1_SpokePool.sol
```diff
@@ -0,0 +1,168 @@
...
+
+    function initialize(
+        uint32 _initialDepositId,
+        address _crossDomainAdmin,
+        address _withdrawalRecipient
+    ) public initializer {
+        __SpokePool_init(_initialDepositId, _crossDomainAdmin, _withdrawalRecipient);
+    }
+
+    function _bridgeTokensToHubPool(uint256, address) internal override {
```
mrice32, line 156: Same question as below: maybe this should revert just to protect against this being called erroneously?

Probably matters less here, since the `SpokePool` doesn't do any accounting on balances.

mrice32 (commented):
This is awesome -- just a few minor comments.

contracts/chain-adapters/SP1_Adapter.sol
```diff
@@ -0,0 +1,59 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.0;
+
+import "./interfaces/AdapterInterface.sol";
+
+contract DataStore {
+    error NotHubPool();
+    mapping(address => bytes) public dataForTarget;
```
nicholaspai, line 8: Should change this to bytes[] so we don't overwrite the calldata for each target address each time

contracts/SP1_SpokePool.sol
```diff
@@ -0,0 +1,168 @@
...
+contract SP1_SpokePool is SpokePool {
+    // The public values stored on L1 that can be relayed into this contract when accompanied with an SP1 proof.
+    struct ContractPublicValues {
+        bytes32 stateRoot;
+        bytes contractCalldata;
+        bytes contractOutput;
+    }
+    /// @notice The address of the HubPool contract. Checked against public values to ensure only state
+    /// stored by HubPool is relayed.
+    address public immutable hubPool;
```
nicholaspai, line 30: Should change this to hubPoolStore and assume that it's a contract that only the HubPool can write to

contracts/SP1_SpokePool.sol
```diff
@@ -0,0 +1,168 @@
...
+    ) external view;
+}
+
+/**
+ * @notice SP1 Spoke pool capable of receiving data stored in L1 state via SP1 + Helios light clients.
+ */
+contract SP1_SpokePool is SpokePool {
+    // The public values stored on L1 that can be relayed into this contract when accompanied with an SP1 proof.
+    struct ContractPublicValues {
+        bytes32 stateRoot;
```
nicholaspai, line 24: Should add contract address to this and compare with hubPoolStore

contracts/chain-adapters/SP1_Adapter.sol
nicholaspai: I thought about this but this only can work if we're ok overwriting the stored value every time. I'm not sure if this is how the Helios light clients work, and whether this forces the EOA publishing the messages on the L2 side to publish state in some order. 

For example, imagine we update the store to store value A on block 1 and then update it again to store value B on block 2. If we publish the state B on the L2 SpokePool before we publish state A, then is it possible to still relate A and B over to the SpokePool?

If not, then we would have to store data in the Store using an [array](https://github.com/across-protocol/contracts/pull/916#discussion_r1990499915) and keep pushing to this array. And then there aren't really benefits to storing data target-agnostic

nicholaspai (commented):
Reminders to self and responses to @mrice32

contracts/SP1_SpokePool.sol
nicholaspai: I don't think it need to be but I'm flagging we should consider it

contracts/SP1_SpokePool.sol
nicholaspai: Yeah same page. See https://github.com/across-protocol/contracts/pull/916#discussion_r1990501728

contracts/chain-adapters/SP1_Adapter.sol
nicholaspai: Related reasoning why storing data target-agnostic could be dangerous https://github.com/across-protocol/contracts/pull/916#discussion_r1990505774

contracts/SP1_SpokePool.sol
```diff
@@ -0,0 +1,167 @@
...
```
nicholaspai, line 24: I think the `stateRoot` existing in ContractPublicValues prevents a lot of replay-attack vectors but still allows for the case where the `message` destined for this SpokePool is legimately the same as some previous message. For example, the `pauseFills()` calldata would be the same multiple times if wanted to pause-unpause-pause-unpause for some reason over the course of X months.

nicholaspai:
@mrice32 please re-review, I've added some features:
- SP1_Adapter should re-use slots when trying to call relayRootBundle to all SpokePools
- SP1_Adapter will encode the target's address when trying to call any other function besides relayRootBundle on the SpokePool

There are open questions:
- How does the SP1_SpokePool prevent the EOA calling receiveL1State() from replaying messages? A message is stored in perpetuity on the HubPoolStore so technically it can be replayed many times via receiveL1State() with different state roots

contracts/chain-adapters/SP1_Adapter.sol
```diff
@@ -0,0 +1,112 @@
...
+            // This is a gas optimization so that we only update a
+            // storage slot in the HubPoolStore once per root bundle execution, since the data passed to relayRootBundle
+            // will be the same for all chains. We are assuming therefore that the relayerRefundRoot and slowRelayRoot
+            // combination are never repeated for a root bundle.
+            DATA_STORE.storeDataForTargetWithNonce(address(0), message, 0);
+        } else {
+            // Because we do not have the chain ID where the target is deployed, we can only associate this message
+            // with the target address. Therefore we are assuming that target spoke pool addresses are unique across
+            // chains.
+            DATA_STORE.storeDataForTarget(target, message);
```
grasphoper, line 94: > target spoke pool addresses are unique across chains.
This has to be true for SP1 spoke pools on different chains, right?

Overall, `SP1_Adapter` breaks the semantics of a "Chain Adapter", cause it's intended to relay messages to multiple chains.

Something to be mindful of, especially since we discussed potentially deploying SpokePools at deterministic addrs in the past. Also, from experience, deployment addrs on different EVM chains can 'accidentally' happen to be the same even with regular `create` opcode, depending on the nonce(and bytecode?) I guess

deploy/111_deploy_sp1_spokepool.ts
```diff
@@ -0,0 +1,28 @@
...
+    L2_ADDRESS_MAP[spokeChainId].sp1Verifier,
+    L2_ADDRESS_MAP[spokeChainId].helios,
+    "1234", // across SP1 program key
+    L1_ADDRESS_MAP[CHAIN_IDs.MAINNET].hubPoolStore,
+    WETH[spokeChainId],
+    QUOTE_TIME_BUFFER,
+    FILL_DEADLINE_BUFFER,
+  ];
+
+  // @dev DO NOT DEPLOY using create2. The SP1 Adapter writes calldata to be relayed to L2 by associating it with the
```
grasphoper, line 21: Like in my comment above, have to also double-check that the resulting addr is not the same, because it's not guaranteed with `create` either

contracts/chain-adapters/SP1_Adapter.sol
nicholaspai: > Overall, SP1_Adapter breaks the semantics of a "Chain Adapter", cause it's intended to relay messages to multiple chains.

This is how we intended to use the Succinct_Adapter, though. I think where possible, the ideal is to have as few adapters service as many chains as possible. This wiill reduce dev deployment time.

> Just something to be mindful of, especially since we discussed potentially deploying SpokePools at deterministic addrs in the past. Also, from experience, deployment addrs on different EVM chains can 'accidentally' happen to be the same even with regular create opcode, depending on the nonce(and bytecode?) I guess

This is easy to avoid if we churn the `nonce` enough. However, there's an easy way around this which is either:
- deploy a different `HubPoolStore` for each chain, or
- deploy a different SP1_Adapter for each chain and set a `DESTINATION_CHAIN_ID` on construction

deploy/111_deploy_sp1_spokepool.ts
nicholaspai: Good points. What do we think about deploying a different HubPoolStore per chain? Or adding a DESTINATION_CHAIN_ID constructor var to the SP1_Adapter? These both have rough edges where we'd need to deploy a new contract for each chain. I think that's still OK.

contracts/external/interfaces/ISP1Helios.sol
andy8052, line 1: Not a big deal but would we want to rename this file just IHelios not ISP1Helios?

contracts/chain-adapters/SP1_Adapter.sol
mrice32: Yeah, I think it would be good to have the capability to use any token standard we already support by default. Would that work?

contracts/R0_SpokePool.sol
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    /**
+     * @notice This can be called by an EOA to relay events that the HubPool emits on L1.
+     * @param journalData The public data written by the guest program
+     * @param seal The encoded cryptographic proof (i.e. SNARK).
+     */
+    function receiveL1State(bytes calldata journalData, bytes calldata seal) external validateInternalCalls {
+        // Decode and validate the journal data
+        Journal memory journal = abi.decode(journalData, (Journal));
```
mrice32, line 139: nit: why not just have the Journal struct passed in as a struct rather than encoded bytes?

contracts/R0_SpokePool.sol
```diff
@@ -0,0 +1,183 @@
...
+        if (journal.contractAddress != hubPoolStore) {
+            revert NotHubPoolStore();
+        }
+        if (journal.eventParams_target != address(this)) {
+            revert NotTarget();
+        }
+        if (IHeliosSteelValidator(steel).validateCommitment(journal.commitment)) {
```
mrice32, line 146: Does this commitment contain both the block info you're proving _and_ the proof? Or are we assuming the block info has already been proven?

(Sorry I may be misunderstanding the flow)

contracts/R0_SpokePool.sol
```diff
@@ -0,0 +1,183 @@
...
+            revert InvalidSteelCommitment();
+        }
+
+        // Verify the proof
+        bytes32 journalHash = sha256(journalData);
+        IRiscZeroVerifier(verifier).verify(seal, imageId, journalHash);
+
+        // Prevent replay attacks by using storage key which includes a nonce. The only way for someone to re-execute
```
mrice32, line 154: Sorry, are you saying the message includes a nonce? Or if not, how is the nonce included in the dataHash?

contracts/SP1_SpokePool.sol
```diff
@@ -0,0 +1,167 @@
...
+     * wouldn't be able to tamper with the _publicValues but we can reduce the chance of replay-attacks this way if
+     * we set the EOA to a trusted actor.
+     * @param _publicValues L1 contract state we want to relay into this contract. Contains a message that will
+     * be treated as calldata for a delegatecall into this contract.
+     * @param _proofBytes Proof bytes for the public values.
+     * @param _head Head number related to _publicValues.
+     */
+    function receiveL1State(
+        bytes calldata _publicValues,
+        bytes calldata _proofBytes,
```
mrice32, line 118: Because this function _verifies the proof_ (rather than asking for the slot from the verifier contract), we need to be sure the helios contract allows the same proof to be replayed. For instance, we wouldn't want an attacker to be able to frontrun the proof on the verifier contract and cause our verification in this function to fail (DoS).

contracts/chain-adapters/UniversalStorageProof_Adapter.sol
```diff
@@ -0,0 +1,113 @@
...
+     * @param message Data to send to target.
+     */
+    function relayMessage(address target, bytes calldata message) external payable override {
+        bytes4 selector = bytes4(message[:4]);
+        if (selector == SpokePoolInterface.relayRootBundle.selector) {
+            // Check if the message contains a relayRootBundle() call for the target SpokePool. If so, then
+            // store the data without a specific target in-mind and hardcode nonce.
+            // This is a gas optimization so that we only update a
+            // storage slot in the HubPoolStore once per root bundle execution, since the data passed to relayRootBundle
+            // will be the same for all chains. We are assuming therefore that the relayerRefundRoot and slowRelayRoot
```
mrice32, line 88: I think this is probably safe, but might be worth brainstorming if there's a way to keep the gas optimization without making this assumption, since it would lead to a post-liveness bundle execution failure if this doesn't hold true.

contracts/libraries/HeliosSteelValidator.sol
```diff
@@ -0,0 +1,76 @@
...
+            bytes32 headerRoot = helios.headers(slot);
+            // Skip missed slots (empty roots)
+            if (headerRoot == bytes32(0)) {
+                continue;
+            }
+            return (headerRoot == parentRoot);
+        }
+
+        return false;
+    }
```
mrice32, line 75: I'm a little confused about this loop.

The goal is to find the hash for the most recently proven block in helios _before_ the blockhash that the event was in?

This feels potentially attackable if someone proves a block between the parent root and the root of the block where the event was. This would cause the loop to stop early and return false, no?

Why not have the caller pass the parent slot in to avoid the loop altogether? Or am I missing something?

mrice32 (commented):
Looks great! A few more comments

contracts/chain-adapters/SP1_Adapter.sol
nicholaspai: i don't see how that can be done cleanly for any token standard. We can have this adapter extend the OFTAdapter and/or the HypeXERC20Adapter but each new standard we'll have to update this contract. WDYT?

contracts/R0_SpokePool.sol
nicholaspai: We can do that, though we might get some gas savings if we don't. It'll be slightly easier for the off-chain state to receive as a Journal so i'll change optimistically

contracts/R0_SpokePool.sol
nicholaspai: the `commitment` includes information about the block. See [here](https://github.com/risc0/risc0-ethereum/blob/95c71d5247216c80f95bf4d2cbf4408d7b384d1f/contracts/src/steel/Steel.sol#L41) how it works for the same-chain state

contracts/SP1_SpokePool.sol
nicholaspai: Is there anything to be done in this contract to support this use case? 

> we need to be sure the helios contract allows the same proof to be replayed

It seems like something we need to ensure in our fork of Helios

contracts/chain-adapters/UniversalStorageProof_Adapter.sol
nicholaspai: How would you keep track of which "time" we are seeing this exact `refundRoot` x `slowRelayRoot` combination?

contracts/libraries/HeliosSteelValidator.sol
nicholaspai: We discussed this with R0:

> Steel supports multiple commitment versions. Version 0 encodes a block number and block hash. However, since the current IHelios interface only stores the state root and the beacon block root (and not the execution block hash), I've proposed using Version 1. Version 1 is designed as an EIP-4788 Beacon commitment, where a timestamp is paired with the beacon block root.

In other words, we can only pass in the `timestamp` if we're trying to use the Helios light client since it only stores state roots.

However,  on line 61, we directly derive the `slot` from timestamp. So really, this loop is only used if the state root of the derived slot is empty (i.e. the slot is "empty" or was missed by the validator). So, this while loop in the happy case should iterate maximum once, and in all cases, should never iterate more than once if the `headerRoot != 0x`. So I don't think this DDOS vector is possible, wdyt?

contracts/R0_SpokePool.sol
nicholaspai: https://github.com/across-protocol/contracts/pull/916/commits/74685ce8b2b9cdf3bef4faf69069cae5459cdf29

contracts/R0_SpokePool.sol
nicholaspai: https://github.com/across-protocol/contracts/pull/916/commits/74685ce8b2b9cdf3bef4faf69069cae5459cdf29

contracts/chain-adapters/SP1_Adapter.sol
nicholaspai: We can add in a follow up PR, this PR is too large

nicholaspai: and the HypERC20 code isn't complete yet

nicholaspai: The main consideration is that this PR then depends on the OFT/HypXERC20 PR's being merged and therefore reviewed along with this PR, right?

nicholaspai: To remove dependency on OFT/HypXERC20 unaudited PR's, we've decided to extend CCTP adapter with this PR and then follow up in a separate PR that adds OFT/HypXERC20 support to this PR once merged into master

contracts/chain-adapters/UniversalStorageProof_Adapter.sol
nicholaspai: I ended up adding a global variable to keep track of the relayRootBundle calldata. I think this maintains the gas optimization while also protecting against the case where the calldata is repeated intra bundles

contracts/SP1_SpokePool.sol
nicholaspai: > (rather than asking for the slot from the verifier contract)

Do you think we should skip the `verifyProof` part and only call `IHelios(helios).getStorageSlot(_head, publicValues.contractAddress, publicValues.slotKey)`?

contracts/UniversalStorageProof_SpokePool.sol
```diff
@@ -0,0 +1,143 @@
...
+
+        // Validate state is intended to be sent to this contract. The target could have been set to the zero address
+        // which is used by the StorageProof_Adapter to denote messages that can be sent to any target.
+        (address target, bytes memory message) = abi.decode(_value, (address, bytes));
+        if (target != address(0) && target != address(this)) {
+            revert NotTarget();
+        }
+
+        // Prevent replay attacks.
+        bytes32 dataHash = keccak256(abi.encode(_slotKey, _blockNumber));
```
nicholaspai, line 114:
```suggestion
        bytes32 dataHash = keccak256(abi.encode(_slotKey));
```

contracts/UniversalStorageProof_SpokePool.sol
```diff
@@ -0,0 +1,159 @@
...
+        emit VerifiedProof(dataHash, msg.sender);
+
+        _executeMessage(message);
+    }
+
+    /**
+     * @notice This function is only callable by the owner and is used as an emergency fallback to relay
+     * a message to this SpokePool in the case where the light-client is not functioning correctly.
+     */
+    function adminReceiveL1State(bytes memory message) external onlyOwner validateInternalCalls {
```
nicholaspai, line 133: Open question: do we need this fallback function to circumvent the light client (in case it breaks for some reason), and if so, is this is the right implementation?

contracts/chain-adapters/utilities/HubPoolStore.sol
```diff
@@ -0,0 +1,128 @@
...
+        HubPoolInterface.RootBundle memory pendingRootBundle = IHubPool(hubPool).rootBundleProposal();
+        // We need to know whether the data contains an admin or a normal relayRootBundle call in order to set the
+        // target and the nonce correctly in the data hash. The target cannot be set to address(0) for an admin bundle
+        // because then the admin bundle could be replayed on any spoke pool, and we should use the global nonce
+        // like other admin functions. Therefore, we check whether there is a pending bundle that hasn't passed
+        // liveness yet, or there is no bundle proposed. In these two cases, it is impossible for the HubPool
+        // to attempt to relay a root bundle unless it is relaying an admin bundle. In the third case where
+        // the pending bundle has passed liveness, we check whether the data is the same as the pending bundle
+        // to determine whether it is an admin bundle or not, as a final check.
+        bool isAdminRootBundle = pendingRootBundle.challengePeriodEndTimestamp > block.timestamp ||
```
nicholaspai, line 84: This is a pretty complex `bool` but it should protect us from sending admin relayRootBundle() calls that (1) can be replay attacked on different `UniversalStorageProof_SpokePools` and (2) block non-admin relayRootBundle() calls with identical calldata

contracts/chain-adapters/utilities/HubPoolStore.sol
```diff
@@ -0,0 +1,128 @@
...
+ * @dev Designed to be used with UniversalStorageProof_Adapter and UniversalStorageProof_SpokePool.
+ * @dev This contract DOES NOT prevent replay attacks of storage proofs on the L2 spoke pool if the
+ * UniversalStorageProof_Adapters using this contract are mapped to spokepools with the same address on different
+ * L2 chains. See comment in storeRelayAdminFunctionCalldata() for more details.
+ */
+contract HubPoolStore {
+    error NotHubPool();
+
+    // Maps unique calldata hashes to calldata.
+    mapping(bytes32 => bytes) public relayAdminFunctionCalldata;
```
mrice32, line 23: I think this is inconsistent with the `SpokePool` verification logic, right?

In the SP, we assume a hash is stored in this slot, right, but here we store the full bytes value, which would require us to verify multiple slots, right?

nicholaspai: oh true, in the SP we should assume the slot is stored as a `bytes` not a `bytes32` i'll correct that

nicholaspai: 6f3f9177c0cc3337cd4ac3dc563a5f8746397ca4

contracts/UniversalStorageProof_SpokePool.sol
mrice32: I think we do. I think the open question is who the owner is and if there should be any verification about the light client being up to date.

contracts/UniversalStorageProof_SpokePool.sol
```diff
@@ -0,0 +1,160 @@
...
+     * wouldn't be able to tamper with the _publicValues but we can reduce the chance of replay-attacks this way if
+     * we set the EOA to a trusted actor. Replay attacks are possible if this contract has the same address
+     * on multiple chains.
+     * @param _slotKey Slot storage hash
+     * @param _value Slot storage value
+     * @param _blockNumber Block number in light client we want to check slot value of slot key
+     */
+    function receiveL1State(
+        bytes32 _slotKey,
+        bytes calldata _value,
```
nicholaspai, line 101: I can change this to only taking in the message and using address(this) as the target

contracts/UniversalStorageProof_SpokePool.sol
```diff
@@ -0,0 +1,160 @@
...
+        // Verify Helios light client is aware of the storage slot:
+        bytes32 slotValueHash = IHelios(helios).getStorageSlot(_blockNumber, hubPoolStore, _slotKey);
+        if (expectedSlotValueHash != slotValueHash) {
+            revert SlotValueMismatch();
+        }
+
+        // Validate state is intended to be sent to this contract. The target could have been set to the zero address
+        // which is used by the StorageProof_Adapter to denote messages that can be sent to any target.
+        (address target, bytes memory message) = abi.decode(_value, (address, bytes));
+        if (target != address(0) && target != address(this)) {
```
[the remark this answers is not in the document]
nicholaspai: Yes I thought about this and commented it on the adapter code. I think the cleanest way to do this is to deploy a new adapter for chain and set a l2 chain id on construction. However that does mean we need to redeploy the adapter for each chain and if we ever upgrade this code we'd have to upgrade for every single chain.

contracts/UniversalStorageProof_SpokePool.sol
nicholaspai: > if there should be any verification about the light client being up to date (i.e. should we not allow the owner to override unless the light client is < n blocks behind the current timestamp).

WDYT about a check like `if (block.number - BUFFER <= helios.head()) revert AdminCallTooCloseToHeliosHead` ?

contracts/chain-adapters/utilities/HubPoolStore.sol
```diff
@@ -0,0 +1,128 @@
...
+ * @dev Designed to be used with UniversalStorageProof_Adapter and UniversalStorageProof_SpokePool.
+ * @dev This contract DOES NOT prevent replay attacks of storage proofs on the L2 spoke pool if the
+ * UniversalStorageProof_Adapters using this contract are mapped to spokepools with the same address on different
+ * L2 chains. See comment in storeRelayAdminFunctionCalldata() for more details.
+ */
+contract HubPoolStore {
+    error NotHubPool();
+
+    // Maps unique calldata hashes to hash of calldata.
+    mapping(bytes32 => bytes32) public relayAdminFunctionCalldata;
```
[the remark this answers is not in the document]
nicholaspai: 9175ed5e7818f6bf03e3c1d82f603e74f90617e1

contracts/chain-adapters/utilities/HubPoolStore.sol
mrice32: Yeah, they can, but we're checking that the multisig made the call to the HubPool, this logic wouldn't care who called the multisig.

contracts/UniversalStorageProof_SpokePool.sol
mrice32: Yeah, something like that would make sense. But we need to relate block.timestamp to helios slots -- I'll check on if/how we can do this. We can't use the local block.number value since that's the block number on the local chain, not mainnet. The block numbers in helios are mainnet block numbers.

contracts/chain-adapters/UniversalStorageProof_Adapter.sol
```diff
@@ -0,0 +1,72 @@
...
+     * @param target Contract on the destination that will receive the message.
+     * @param message Data to send to target.
+     */
+    function relayMessage(address target, bytes calldata message) external payable override {
+        bytes4 selector = bytes4(message[:4]);
+        if (selector == SpokePoolInterface.relayRootBundle.selector) {
+            DATA_STORE.storeRelayRootsCalldata(target, message);
+        } else {
+            DATA_STORE.storeRelayAdminFunctionCalldata(target, message);
+        }
```
grasphoper, line 46: This if makes it so we have 2 places where we decide if it's an admin bundle: here and in `HubPoolStore`. If we move this logic to the HubPoolStore, we can decide admin bundle or not all in one place

grasphoper: e.g. call `DATA_STORE.storeCrossChainMessage` and decide admin bundle or not there

nicholaspai: good point

contracts/chain-adapters/utilities/HubPoolStore.sol
nicholaspai: also this means we can't really handle the case well where the hub pool owner executes a root bundle...seems unlikely though

contracts/chain-adapters/UniversalStorageProof_Adapter.sol
nicholaspai: addressed this here https://github.com/across-protocol/contracts/pull/916/commits/130bb2fcd610d8b096182b5e1c7ee910278afda4

contracts/chain-adapters/utilities/HubPoolStore.sol
```diff
@@ -0,0 +1,90 @@
...
+        }
+    }
+
+    function _storeRelayMessageCalldataForAnyTarget(bytes calldata data) internal {
+        // When the data can be sent to any target, we assume that the data contains a relayRootBundleCall as
+        // constructed by an executeRootBundle() call, therefore this data will be identical for all spoke pools
+        // in this bundle. We can use the current hub pool's challengePeriodEndTimestamp as the nonce for this data
+        // so that all relayRootBundle calldata for this bundle gets stored to the same slot and we only write to
+        // this slot once.
+        _storeData(address(0), IHubPool(hubPool).rootBundleProposal().challengePeriodEndTimestamp, data);
```
grasphoper, line 75: Trying to understand what we're protecting ourselves against here by setting nonce to `challengePeriodEndTimestamp` instead of just using `dataUuid++`

`nonce` only gets checked in the `_storeData` function. So we're protected from accidentally calling `_storeData` twice for the same bundle. But doesn't this get enforced at the HubPool level? E.g. for other Adapters, they can relay the same bundle twice with no protection, i.e. it's enforced by HubPool

nicholaspai: >  But doesn't this get enforced at the HubPool level? E.g. for other Adapters, they can relay the same bundle twice with no protection, i.e. it's enforced by HubPool

The HubPool will call `relayMessage(target, message)` once for each chain, using the `crossChainContracts[chainId].spokePool` as the target. The message will be the same each time. So if multiple chains use this same adapter, then this HubPoolStore will get called once per each of these chains.

grasphoper: Ah, I see. Makes sense

contracts/chain-adapters/utilities/HubPoolStore.sol
```diff
@@ -0,0 +1,90 @@
...
+     * @param data Data to send to Universal SpokePool.
+     * @param isAdminSender True if the data is an admin function call, false otherwise.
+     */
+    function storeRelayMessageCalldata(
+        address target,
+        bytes calldata data,
+        bool isAdminSender
+    ) external onlyHubPool {
+        if (isAdminSender) {
+            _storeData(target, dataUuid++, data);
```
grasphoper, line 63: nit: do we wanna change it to? 
```
_storeData(target, dataUuid, data);
dataUuid += 1;
```
IMO makes code easier to read

grasphoper: Ah, this is 2 storage reads in a row. Need a stack var first then

contracts/chain-adapters/utilities/HubPoolStore.sol
nicholaspai: https://github.com/across-protocol/contracts/pull/916/commits/130bb2fcd610d8b096182b5e1c7ee910278afda4

contracts/UniversalStorageProof_SpokePool.sol
```diff
@@ -0,0 +1,157 @@
...
+
+    /**
+     * @notice This can be called by an EOA to relay call data stored on the HubPool on L1 into this contract.
+     * @dev Replay attacks are possible with this _message if this contract has the same address on another chain.
+     * @param _slotKey Slot storage hash
+     * @param _value Slot storage value
+     * @param _blockNumber Block number in light client we want to check slot value of slot key
+     */
+    function receiveL1State(
+        bytes32 _slotKey,
```
mrice32, line 97: Just wanted to call this out: I DON'T think this is a bug/exploitable in the current contracts, but it is possible to prove _any_ key in the mainnet contract. This means that storage slots, even those outside the mapping could be treated as hashed messages.

AFAICT, nothing else in the `HubPoolStore` storage could be manipulated to be a valid message hash, but wanted to mention it. If were concerned about this, we could have the caller pass in the nonce, and we could generate the key from that to ensure it was a member of the mapping we expect.

contracts/UniversalStorageProof_SpokePool.sol
mrice32: Helios PR should be updated to provide the `headTimestamp()` function.

mrice32 (approved):
Looks great!

contracts/UniversalStorageProof_SpokePool.sol
nicholaspai: yea i think this is a good callout, i'll add a comment

nicholaspai: ended up computing slot key in contract

mrice32 (approved):

contracts/UniversalStorageProof_SpokePool.sol
```diff
@@ -0,0 +1,188 @@
...
+    /// data that has been written to this store.
+    address public immutable hubPoolStore;
+
+    /// @notice Slot index of the HubPoolStore's relayMessageCallData mapping.
+    uint256 public constant HUB_POOL_STORE_CALLDATA_MAPPING_SLOT_INDEX = 0;
+
+    /// @notice The address of the Helios L1 light client contract.
+    address public immutable helios;
+
+    /// @notice The owner of this contract must wait until this amount of seconds have passed since the latest
```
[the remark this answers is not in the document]
grasphoper: Helios actually calls `owner` `guardian` instead. That's good optics LOL

grasphoper (approved):

PR #917, opened 2025-03-12
feat(svm): deposit tests with codama and solana kit

md0x:
Changes proposed in this PR:
- Add deposit tests with codama client and solana kit


Reinis-FRP (approved):
nice

PR #919, opened 2025-03-13
Single AddressBook for all adapters

grasphoper:
With current design, we need to use 1 `AddressBook` per chain adapter, because we have no way to distinguish which adapter is using the `AddressBook`. This PR changes that, indexing the storage data additionally by `ADAPTER_ID`


contracts/libraries/AddressBook.sol
```diff
@@ -0,0 +1,26 @@
...
+
+import { Ownable } from "@openzeppelin/contracts/access/Ownable.sol";
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import { MultiCaller } from "@uma/core/contracts/common/implementation/MultiCaller.sol";
+
+/**
+ * @dev A helper contract for chain adapters that support OFT or XERC20 messaging on L1
+ * @dev Handles token => messenger/router mapping storage, as adapters are called via delegatecall and don't have relevant storage space
+ */
+// todo: MultiCaller or dedicated batchSetOFTMessenger function? Or both? Multicaller looks cleaner, but batch function is more gas-efficient
```
grasphoper, line 12: I am leaning towards `batchSetOFTMessenger` fn for gas efficiency on an L1

contracts/chain-adapters/Arbitrum_Adapter.sol
```diff
@@ -57,8 +57,11 @@ contract Arbitrum_Adapter is AdapterInterface, CircleCCTPAdapter, OFTTransportAd
     // Generic gateway: https://github.com/OffchainLabs/token-bridge-contracts/blob/main/contracts/tokenbridge/ethereum/gateway/L1ArbitrumGateway.sol
     ArbitrumL1ERC20GatewayLike public immutable L1_ERC20_GATEWAY_ROUTER;
 
-    // Helper contract to help us map token -> OFT messenger for OFT-enabled tokens
-    OFTAddressBook public immutable OFT_ADDRESS_BOOK;
+    // Helper contract to help us map token => OFT messenger for OFT bridging
+    AddressBook public immutable ADDRESS_BOOK;
+
+    // Uniquely identifies adapter type (Aritrum, Optimism etc.) for calls to ADDRESS_BOOK
+    uint256 public immutable ADAPTER_ID;
```
nicholaspai, line 64: I'd prefer naming this CHAIN_ID since that's the term we use throughout the rest of the stack. I assume you're going to set this to the chain ID?

contracts/libraries/AddressBook.sol
```diff
@@ -0,0 +1,26 @@
...
+contract AddressBook is Ownable, MultiCaller {
```
nicholaspai, line 13: What about naming this AdapterStore (we use Store a lot to denote a contract that is just designed to store data)? AddressBook seems too generic of a name.

deploy/004_deploy_arbitrum_adapter.ts
```diff
@@ -11,6 +11,8 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
   // set to the Risk Labs relayer address. The deployer should change this if necessary.
   const l2RefundAddress = "0x07aE8551Be970cB1cCa11Dd7a11F47Ae82e70E67";
 
+  // id that uniquely identifies a chain adapter type. Set to closely match chain ids
+  const adapterId = 42161;
```
nicholaspai, line 15:
```suggestion
  const adapterId = chainId;
```

nicholaspai (commented):
First review

contracts/chain-adapters/Arbitrum_Adapter.sol
grasphoper: I think `CHAIN_ID` is fine, but a little misleading. `DESTINATION_CHAIN_ID` ? 

Also, what about solana chain id? I'm finding differnt ids for it from different sources. We can figure something out

contracts/libraries/AddressBook.sol
grasphoper: I like it, will use `AdapterStore`

contracts/libraries/AddressBook.sol
grasphoper: I think it's going to be cheaper "by design"
Multicall calls multiple delegate calls to self. Calling a regular function in a loop should be cheaper than that, right?

contracts/chain-adapters/Arbitrum_Adapter.sol
grasphoper: 1 more thing: if we commit to this being chainId, would it make sense to just make this ID a constant, not a constructor param?

deploy/004_deploy_arbitrum_adapter.ts
grasphoper: This `chainId` would be mainnet though, right? We need arbitrumChainId

grasphoper: Will add `const dstChainId = CHAIN_IDs.ARBITRUM;`

contracts/SpokePool.sol
```diff
@@ -1757,6 +1759,9 @@ abstract contract SpokePool is
     }
 
     function _setOftMessenger(address _token, address _messenger) internal {
+        if (IOFT(_messenger).token() != _token) {
+            revert OFTTokenMismatch();
+        }
```
grasphoper, line 1764: Added this to help prevent human errors. Do you think it's worth the gas cost? I think so

contracts/libraries/AdapterStore.sol
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    function _setOFTMessenger(
+        uint256 _adapterChainId,
+        address _l1Token,
+        address _oftMessenger
+    ) internal {
+        if (IOFT(_oftMessenger).token() != _l1Token) {
+            revert OFTTokenMismatch();
+        }
```
grasphoper, line 48: Same as https://github.com/across-protocol/contracts/pull/919/files#r1994184061

contracts/SpokePool.sol
nicholaspai: sure, this function is rarely called

deploy/004_deploy_arbitrum_adapter.ts
```diff
@@ -11,6 +12,8 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
   // set to the Risk Labs relayer address. The deployer should change this if necessary.
   const l2RefundAddress = "0x07aE8551Be970cB1cCa11Dd7a11F47Ae82e70E67";
 
+  // id that uniquely identifies a chain adapter type. Set to closely match chain ids
+  const dstChainId = CHAIN_IDs.ARBITRUM;
```
nicholaspai, line 16: why not use `chainId`? We already load that on line 8, and it helps support testnets also

test/evm/hardhat/chain-adapters/Arbitrum_Adapter.ts
```diff
@@ -49,9 +49,10 @@ let l1ERC20GatewayRouter: FakeContract,
   cctpMessenger: FakeContract,
   cctpTokenMinter: FakeContract,
   oftMessenger: FakeContract<IOFT>,
-  oftAddressBook: FakeContract<OFTAddressBook>;
+  adapterStore: FakeContract<AdapterStore>;
 
 const arbitrumChainId = 42161;
+const adapterId = arbitrumChainId;
```
nicholaspai, line 55: why distinguish between adapter ID's and arbitrum chain ID's at all? I think its more intuitive if you just use chain ID's everywhere and have no notion of an "adapter ID"

deploy/004_deploy_arbitrum_adapter.ts
grasphoper: https://github.com/across-protocol/contracts/pull/919#discussion_r1994009169

grasphoper: I did this to support testnet deployment `const dstChainId = chainId == MAINNET_CHAIN_IDs.MAINNET ? CHAIN_IDs.ARBITRUM : CHAIN_IDs.ARBITRUM_SEPOLIA;`

deploy/004_deploy_arbitrum_adapter.ts
nicholaspai: True never mind

nicholaspai (approved):

PR #922, opened 2025-03-14
feat(svm): fill tests with codama and solana kit

md0x:
Changes proposed in this PR:
- Add simple happy path Fill tests with Codama and solana kit


test/svm/SvmSpoke.Fill.ts
```diff
@@ -631,4 +649,107 @@ describe("svm_spoke.fill", () => {
...
+        relayerTokenAccount: address(relayerTA.toString()),
+        recipientTokenAccount: address(recipientTA.toString()),
+        fillStatus: address(accounts.fillStatus.toString()),
+        tokenProgram: address(TOKEN_PROGRAM_ID.toString()),
+        associatedTokenProgram: address(ASSOCIATED_TOKEN_PROGRAM_ID.toString()),
+        systemProgram: address(anchor.web3.SystemProgram.programId.toString()),
+        program: address(program.programId.toString()),
+        eventAuthority,
+        signer,
+        remainingAccounts: fillRemainingAccounts,
```
Reinis-FRP, line 684: this is not in `FillRelayAsyncInput` props

test/svm/SvmSpoke.Fill.ts
```diff
@@ -631,4 +649,107 @@ describe("svm_spoke.fill", () => {
...
+      });
+
+      const fillRelayInput: FillRelayAsyncInput = {
+        ...formattedRelayData,
+        ...formattedAccounts,
+      };
+
+      const fillRelayIx = {
+        ...(await SvmSpokeClient.getFillRelayInstructionAsync(fillRelayInput)),
+        accounts: (await SvmSpokeClient.getFillRelayInstructionAsync(fillRelayInput)).accounts.map((account) =>
```
Reinis-FRP, line 723: would we better have single call to `getFillRelayInstructionAsync` and store result in a variable rather than call it twice here

test/svm/SvmSpoke.Fill.ts
md0x: You are right, remaining accounts are passed here:
https://github.com/across-protocol/contracts/blob/583d129037f51f5f3558fa00966b88a99399b2b7/test/svm/SvmSpoke.Fill.ts#L726-L730

And I forgot to remove it from this obj.

Reinis-FRP (approved):

PR #926, opened 2025-03-17
feat(SpokePool): Remove enabledDepositRoutes check for unsafeDeposit

nicholaspai:
Allows Across to support inputTokens without enabling deposit routes for them, by forcing filler to take repayment of the token on the origin chain.


contracts/SpokePool.sol
```diff
@@ -1330,10 +1341,6 @@ abstract contract SpokePool is
         // Verify depositor is a valid EVM address.
         params.depositor.checkAddress();
 
-        // Check that deposit route is enabled for the input token. There are no checks required for the output token
-        // which is pulled from the relayer at fill time and passed through this contract atomically to the recipient.
-        if (!enabledDepositRoutes[params.inputToken.toAddress()][params.destinationChainId]) revert DisabledRoute();
```
nicholaspai, line 1335: Second open question: Should we remove this check in the Solana SpokePool as well? I assume no, its not needed for now.

contracts/SpokePool.sol
```diff
@@ -1330,10 +1343,6 @@ abstract contract SpokePool is
...
```
nicholaspai, line 1335: third open question: removing this check now allows depositor to set any value for `destinationChainId` they want to. This opens up griefing vectors against offchain components that we need to make sure relayers and dataworkers protect against

contracts/SpokePool.sol
```diff
@@ -557,6 +562,9 @@ abstract contract SpokePool is
             exclusivityParameter: exclusivityParameter,
             message: message
         });
+        // Check that deposit route is enabled for the input token. There are no checks required for the output token
+        // which is pulled from the relayer at fill time and passed through this contract atomically to the recipient.
+        if (!enabledDepositRoutes[inputToken.toAddress()][destinationChainId]) revert DisabledRoute();
```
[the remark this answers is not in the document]
bmzig: I'm of the opinion that if we are getting this audited, then we might as well make it across all deposits. I think that if integrators do want to use this feature, then it is a little annoying to have to make an unsafe deposit, which forces more responsibility on them to make "correctly structured" deposits (i.e. no duplicates). This fixes the griefing vector, and I'd think that if we modify the dataworker to force origin repayment on these token types when a pool rebalance root is not set, then it seems like it would be OK to just fully support anyways.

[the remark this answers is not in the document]
nicholaspai: https://github.com/across-protocol/contracts/pull/926/commits/d94d98554241e460fafec4865027efd69536adcf

contracts/chain-adapters/Solana_Adapter.sol
```diff
@@ -113,7 +113,7 @@ contract Solana_Adapter is AdapterInterface, CircleCCTPAdapter {
         }
 
         bytes4 selector = bytes4(message[:4]);
-        if (selector == SpokePoolInterface.setEnableRoute.selector) {
```
nicholaspai, line 116: Note to self: can delete this section if we choose to remove deposit routes in Solana spoke

contracts/SpokePool.sol
mrice32: Yes, but we'll do it in a separate PR.

contracts/SpokePool.sol
mrice32: What griefing vectors? Just that people can send completely invalid deposits?

nicholaspai: yeah, basically you can set the destination chain ID so high that validators can't handle it. Shouldn't be an issue if they use BN.js

contracts/SpokePool.sol
```diff
@@ -1519,7 +1498,7 @@ abstract contract SpokePool is
         uint256 returnValue;
         bytes memory data = abi.encodeCall(IERC20Upgradeable.transfer, (to, amount));
         assembly {
-            success := call(gas(), token, 0, add(data, 0x20), mload(data), 0, 0x20)
+            success := call(MAX_ERC20_TRANSFER_GAS_COST, token, 0, add(data, 0x20), mload(data), 0, 0x20)
```
mrice32, line 1501: Since each leaf is a separate token, shouldn't we just assume that the executor will protect themselves against this and not execute leaves where the token is expensive to transfer? You could even imagine the executor having a token whitelist where it refuses to pay out tokens that are not on that list.

mrice32 (commented):
Looks great! I want @Reinis-FRP to provide a review if he hasn't already.

contracts/SpokePool.sol
nicholaspai: thats a good point too, and reduces contract complexity +1

mrice32 (approved):

contracts/SpokePool.sol
nicholaspai: https://github.com/across-protocol/contracts/pull/926/commits/263dcd915b9a6d2e9aba7b88f6691d7be020e1d6

PR #933, opened 2025-03-20
fix(tasks): fix usage comment + typo in `evm-relay-message-withdrawal`

grasphoper:
When trying to run the version in `master`, I get `Error HH310: Invalid param --spokePool. Command line params must be lowercase.`

This is why I changed params to lowercase.

FYI @bmzig


grasphoper:
> This isn't an issue per se. You are supposed to run hardhat tasks with camel case like `yarn evm-relay-message-withdrawal --spoke-pool [spoke_pool_address] --adapter [adapter_address] --l2-token [l2_token_address] --amount-to-return [amount_to_return]`. With that being said, the comment on how to run it was indeed incorrect.

Right, that's weird. Reverted back the name changes. Changed the comment to reflect the correct command

bmzig (approved):

PR #939, opened 2025-03-27
feat(svm): remove enabled deposit route check

Reinis-FRP:
This implements removing enabled deposit route check in SVM, same as for EVM in #926


nicholaspai:
@Reinis-FRP can you target the [march-evm-audit-universal-adapter](https://github.com/across-protocol/contracts/tree/march-evm-audit-universal-adapter) branch pls?

programs/svm-spoke/src/instructions/admin.rs
```diff
@@ -136,67 +136,31 @@ pub fn set_cross_domain_admin(ctx: Context<SetCrossDomainAdmin>, cross_domain_ad
...
 }
 
-#[event_cpi]
 #[derive(Accounts)]
-#[instruction(origin_token: Pubkey, destination_chain_id: u64)]
-pub struct SetEnableRoute<'info> {
-    #[account(constraint = is_local_or_remote_owner(&signer, &state) @ SvmError::NotOwner)]
-    pub signer: Signer<'info>,
-
+pub struct CreateVault<'info> {
```
chrismaree, line 140: I'm a bit confused by this change. Was the point of this PR not to just remove enable deposit route check?

Reinis-FRP: yes, alternative would be to remove SetEnableRoute altogether and let the operator create vault PDA manually.

Reinis-FRP: I did end up deleting the instruction altogether in the last commit, since vault ATA can be created by anyone directly via associated token program.

scripts/svm/createVault.ts
```diff
@@ -0,0 +1,66 @@
+// This script is used by a chain admin to create a vault for a token on the Solana Spoke Pool.
```
chrismaree, line 1: this can, in theory, be called by anyone other than just the admin, right? as we are simply creating a ATA for the spokepool PDA to store a given token?

Im thinking through the equivalent functionality we have on other chains now that we no longer have a deposit route check: in theory, someone else could call use this script (or otherwise) to create PDAs so that you can deposit and fill non-whitelisted tokens on behalf of the protocol, just like we can on other chains (by just depositing without using a verified token to a given spoke). 

point being, this is not used by an admin, right? there is no special permissioning to create one of these vaults?

chrismaree (approved):
this looks good to me! its simple and clean. 

I have two open comments on naming but other than that this looks good!

scripts/svm/createVault.ts
Reinis-FRP: Indeed, anyone can create the vault. normally we expect admin to do this as that requires spending rent-exempt deposit on vault ATA creation. But any other user is free to do this if they want to deposit funds on a non-whitelisted token.

Reinis-FRP: updated the comment to reflect that

PR #940, opened 2025-03-27
improve: support upgradeable USDC in zkstack adapters

bmzig:
Upgradeable USDC interfaces with a separate second bridge contract for zk-stack. This PR makes the adapter aware of this edge case.

Note: If there is no support for upgradeable USDC, then `_usdcSharedBridge` should just be the shared bridge address.


contracts/chain-adapters/ZkStack_Adapter.sol
```diff
@@ -49,9 +49,15 @@ contract ZkStack_Adapter is AdapterInterface {
     // Set l1Weth at construction time to make testing easier.
     WETH9Interface public immutable L1_WETH;
 
+    // Set l1Usdc at construction time to help determine which shared bridge to use.
+    address public immutable L1_USDC;
```
nicholaspai, line 53: If ZkStack eventually extends the CCTPAdapter then how do you think we should handle the latter's usdc variable?

nicholaspai (approved):
Just one q!

contracts/chain-adapters/ZkStack_Adapter.sol
bmzig: Good point. I agree. It is best if we conform to that interface. https://github.com/across-protocol/contracts/pull/940/commits/52f517495aa1052870306397d1fae410b275a688

contracts/chain-adapters/ZkStack_CustomGasToken_Adapter.sol
```diff
@@ -201,8 +215,28 @@ contract ZkStack_CustomGasToken_Adapter is AdapterInterface {
                     refundRecipient: L2_REFUND_ADDRESS
                 })
             );
+        } else if (l1Token == address(usdcToken)) {
+            if (_isCCTPEnabled()) {
+                _transferUsdc(to, amount);
+            } else {
+                IERC20(CUSTOM_GAS_TOKEN).forceApprove(USDC_SHARED_BRIDGE, txBaseCost);
+                IERC20(l1Token).forceApprove(USDC_SHARED_BRIDGE, amount);
+                txHash = BRIDGE_HUB.requestL2TransactionTwoBridges(
```
nicholaspai, line 224: nit: can probably refactor this request call with the below one into an internal function, where the input is the "bridge" contract to call `requestL2TransactionTwoBridges` on. This way there's less reused code

nicholaspai (commented):
One sggestion

contracts/chain-adapters/ZkStack_Adapter.sol
```diff
@@ -158,8 +172,28 @@ contract ZkStack_Adapter is AdapterInterface {
                     refundRecipient: L2_REFUND_ADDRESS
                 })
             );
+        } else if (l1Token == address(usdcToken)) {
+            // Either use CCTP or the custom shared bridge when bridging USDC.
+            if (_isCCTPEnabled()) {
+                _transferUsdc(to, amount);
+            } else {
```
nicholaspai, line 179: Should we be checking that `USDC_SHARED_BRIDGE` is not zero address?

mrice32:
@nicholaspai @bmzig I merged this into @pxrl's PR to combine them. Feel free to move any unresolved threads there.

PR #942, opened 2025-03-28
feat(svm): test native sol deposits

Reinis-FRP:
This PR demonstrates its possible to support native token deposits on solana with the current program implementation by prepending transfer/wrapping instructions to the deposit transaction.


chrismaree (approved):
This is great! If we order the prs correctly you can delete the route pda logic after rebasing so we can have it match real world context.

PR #944, opened 2025-03-31
improve: query shared bridge address dynamically

bmzig:
OZ recommended we do this in the past. I was hesitant due to minimizing the number of calls we need to make (and therefore lowering L1 gas costs), but after seeing how disruptive a change in this address can be, my opinion changed.


contracts/chain-adapters/ZkStack_Adapter.sol
```diff
@@ -139,6 +135,7 @@ contract ZkStack_Adapter is AdapterInterface {
         // A bypass proxy seems to no longer be needed to avoid deposit limits. The tracking of these limits seems to be deprecated.
         // See: https://github.com/matter-labs/era-contracts/blob/bce4b2d0f34bd87f1aaadd291772935afb1c3bd6/l1-contracts/contracts/bridge/L1ERC20Bridge.sol#L54-L55
         uint256 txBaseCost = _computeETHTxCost(L2_GAS_LIMIT);
+        address sharedBridge = BRIDGE_HUB.sharedBridge();
```
nicholaspai, line 138: nit: can move this into the `else` statement on line 158 because we don't need it in other branches

nicholaspai (commented):
one suggestion

mrice32 (approved):
LGTM!

PR #948, opened 2025-04-03
chore: Tweak Solana SpokePool name in deployments

pxrl:
Avoid the name "SvmSpoke" because it references the architecture, not the chain itself. Revert to "SpokePool", since that's consistent with all other chains.


pxrl:
@Reinis-FRP Do you anticipcate the need for any other changes here? i.e. are there any other dependencies on the name "SvmSpoke" ?

pxrl:
Thinking about this a bit more; this change might be misguided...

PR #957, opened 2025-04-16
fix(svm): L-01 create new vault on deposit if needed

Reinis-FRP:

> [Pull request #939](https://github.com/across-protocol/contracts/pull/939) removes route-enabling functionality in the SVM Spoke, allowing depositors to provide any input token for any destination chain, not only those with the route enabled. Previously, when a route was enabled, the [Associated Token Account (ATA) of the vault for user deposits](https://github.com/across-protocol/contracts/pull/939/files#diff-f52705e50e7f3f740fed5f68d5e5807eccc7354ed50fe83a3ac43c9c8942dad9L173) would be created. The [DisabledRoute](https://github.com/across-protocol/contracts/pull/939/files#diff-cba2b3c01b372a1d1cec7767a7fb0fd415844f25d8c5e673cf85838db2e6df8eL42) error would also be generated if a user tried to deposit to a disabled route. Since, by this point, all routes would be enabled, a user may try to submit a deposit where the vault ATA has not been initialized. This will cause the deposit transaction to fail without a descriptive error message.
>
> Thus, if the vault has not been created, the depositor would have to initialize the PDA themselves, for example, using the [createVault.ts](https://github.com/across-protocol/contracts/blob/71e990b3f908ecf994be8723e041b584c7d49318/scripts/svm/createVault.ts) script. However, this is not a simple process for the majority of users and it may be unclear why the deposit failed. This could cause users to abandon their deposit attempt and consider alternative bridging solutions. A more straightforward solution would be to initialize the vault during `deposit` using the `init_if_needed` flag as was [previously done](https://github.com/across-protocol/contracts/pull/939/files#diff-f52705e50e7f3f740fed5f68d5e5807eccc7354ed50fe83a3ac43c9c8942dad9L167) when the admin enabled a route.
>
> Consider adding the `init_if_needed` flag or clearly documenting the desired vault creation process within the docstring for the `deposit` function.

This PR addresses the issue by adding the `init_if_needed` flag to the `vault` account in the `deposit` method.


md0x (approved):
Looks great!

PR #958, opened 2025-04-16
fix(svm): M-01 Deposit Tokens Transfers

md0x:


src/svm/web3-v1/helpers.ts
```diff
@@ -20,3 +22,32 @@ export const isSolanaDevnet = (provider: AnchorProvider): boolean => {
   else if (solanaRpcEndpoint.includes("mainnet")) return false;
   else throw new Error(`Unsupported solanaCluster endpoint: ${solanaRpcEndpoint}`);
 };
+
+/**
+ * Returns the delegate PDA for a deposit.
+ */
+export const getDepositDelegatePda = (depositData: DepositData, stateSeed: BN, programId: PublicKey) => {
```
md0x, line 29: To keep this PR focused, let’s tackle updates to scripts/svm in a follow‑up PR.

programs/svm-spoke/src/utils/deposit_utils.rs
```diff
@@ -9,3 +9,19 @@ pub fn get_unsafe_deposit_id(msg_sender: Pubkey, depositor: Pubkey, deposit_nonc
...
+    input_amount: u64,
+    output_amount: u64,
+    destination_chain_id: u64,
+) -> [u8; 32] {
+    let mut data = Vec::with_capacity(32 + 32 + 8 + 8 + 8);
+    data.extend_from_slice(input_token.as_ref());
+    data.extend_from_slice(output_token.as_ref());
+    data.extend_from_slice(&input_amount.to_le_bytes());
+    data.extend_from_slice(&output_amount.to_le_bytes());
+    data.extend_from_slice(&destination_chain_id.to_le_bytes());
```
Reinis-FRP, line 25: maybe use `AnchorSerialize::serialize` or `try_to_vec`, so that we don't need to be explicit on individual field encoding

programs/svm-spoke/src/instructions/fill.rs
```diff
@@ -32,6 +32,10 @@ pub struct FillRelay<'info> {
     )]
     pub state: Account<'info, State>,
 
+    #[account(seeds = [b"delegate", state.seed.to_le_bytes().as_ref(), relay_hash.as_ref()], bump)]
```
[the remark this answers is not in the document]
Reinis-FRP: do tests depend on `state.seed`? if not, we better drop the state seed here as its 0 on public networks

programs/svm-spoke/src/instructions/deposit.rs
```diff
@@ -101,21 +102,39 @@ pub fn _deposit(
...
         }
-
         if exclusive_relayer == Pubkey::default() {
             return err!(CommonError::InvalidExclusiveRelayer);
         }
     }
 
-    // Depositor must have delegated input_amount to the state PDA.
+    // Verify delegate PDA
+    let seed_hash = derive_deposit_delegate_seed_hash(
```
Reinis-FRP, line 111: As we don't include `quote_timestamp` and `fill_deadline` this would still leave room for some griefing attacks where a deposit could become practically unfillable if someone changed those parameters using a delegation from a separate transaction.

Since the above params could be hard to predict in `deposit_now` one alternative would be to calculate `seed_hash` at the entry of the handler for each of the deposit flavors and pass it as parameter to the common implementation.

programs/svm-spoke/src/utils/deposit_utils.rs
```diff
@@ -9,3 +9,46 @@ pub fn get_unsafe_deposit_id(msg_sender: Pubkey, depositor: Pubkey, deposit_nonc
 
     keccak::hash(&data).to_bytes()
 }
+
+#[derive(AnchorSerialize)]
+pub struct DepositDelegateSeedData {
```
Reinis-FRP, line 14: nit: does this need to be pub? we're using it only in this module

programs/svm-spoke/src/utils/fill_utils.rs
```diff
@@ -0,0 +1,19 @@
+use anchor_lang::{prelude::*, solana_program::keccak};
+
+#[derive(AnchorSerialize)]
+pub struct FillDelegateSeedData {
```
Reinis-FRP, line 4: nit: does this need to be pub? we're using it only in this module

programs/svm-spoke/src/instructions/deposit.rs
md0x: Yeah, we’d still have to calculate it off-chain and pass in the delegate PDA, which depends on current_time in deposit_now, and that’s the tricky part, right? So I don’t think that really solves the problem.

programs/svm-spoke/src/instructions/deposit.rs
```diff
@@ -36,6 +38,9 @@ pub struct Deposit<'info> {
     )]
     pub state: Account<'info, State>,
 
+    /// CHECK: PDA derived with seeds ["delegate", seed_hash]; used as a CPI signer.
+    pub delegate: UncheckedAccount<'info>,
```
Reinis-FRP, line 42: would need to update the docs in lib.rs

programs/svm-spoke/src/instructions/fill.rs
```diff
@@ -32,6 +32,9 @@ pub struct FillRelay<'info> {
     )]
     pub state: Account<'info, State>,
 
+    /// CHECK: PDA derived with seeds ["delegate", seed_hash]; used as a CPI signer.
+    pub delegate: UncheckedAccount<'info>,
```
Reinis-FRP, line 36: would need to update the docs in lib.rs

programs/svm-spoke/src/instructions/deposit.rs
```diff
@@ -159,6 +169,22 @@ pub fn deposit(
...
+            input_token,
+            output_token,
+            input_amount,
+            output_amount,
+            destination_chain_id,
+            exclusive_relayer,
+            quote_timestamp,
+            fill_deadline,
+            exclusivity_parameter,
+            message: message.clone(),
```
Reinis-FRP, line 185:
```suggestion
            message: &message,
```
better avoid cloning

programs/svm-spoke/src/instructions/deposit.rs
```diff
@@ -195,7 +222,22 @@ pub fn deposit_now(
...
+            recipient,
+            input_token,
+            output_token,
+            input_amount,
+            output_amount,
+            destination_chain_id,
+            exclusive_relayer,
+            fill_deadline_offset,
+            exclusivity_period,
+            message: message.clone(),
```
Reinis-FRP, line 237:
```suggestion
            message: &message,
```
better avoid cloning

programs/svm-spoke/src/instructions/deposit.rs
```diff
@@ -232,6 +276,22 @@ pub fn unsafe_deposit(
...
+            input_token,
+            output_token,
+            input_amount,
+            output_amount,
+            destination_chain_id,
+            exclusive_relayer,
+            quote_timestamp,
+            fill_deadline,
+            exclusivity_parameter,
+            message: message.clone(),
```
Reinis-FRP, line 292:
```suggestion
            message: &message,
```
better avoid cloning

programs/svm-spoke/src/utils/delegate_utils.rs
```diff
@@ -0,0 +1,45 @@
...
+    pub input_token: Pubkey,
+    pub output_token: Pubkey,
+    pub input_amount: u64,
+    pub output_amount: u64,
+    pub destination_chain_id: u64,
+    pub exclusive_relayer: Pubkey,
+    pub quote_timestamp: u32,
+    pub fill_deadline: u32,
+    pub exclusivity_parameter: u32,
+    pub message: Vec<u8>,
```
Reinis-FRP, line 22:
```suggestion
pub struct DepositSeedData<'a> {
    pub depositor: Pubkey,
    pub recipient: Pubkey,
    pub input_token: Pubkey,
    pub output_token: Pubkey,
    pub input_amount: u64,
    pub output_amount: u64,
    pub destination_chain_id: u64,
    pub exclusive_relayer: Pubkey,
    pub quote_timestamp: u32,
    pub fill_deadline: u32,
    pub exclusivity_parameter: u32,
    pub message: &'a Vec<u8>,
```
avoids cloning

programs/svm-spoke/src/utils/delegate_utils.rs
```diff
@@ -0,0 +1,45 @@
...
+    pub recipient: Pubkey,
+    pub input_token: Pubkey,
+    pub output_token: Pubkey,
+    pub input_amount: u64,
+    pub output_amount: u64,
+    pub destination_chain_id: u64,
+    pub exclusive_relayer: Pubkey,
+    pub fill_deadline_offset: u32,
+    pub exclusivity_period: u32,
+    pub message: Vec<u8>,
```
Reinis-FRP, line 37:
```suggestion
pub struct DepositNowSeedData<'a> {
    pub depositor: Pubkey,
    pub recipient: Pubkey,
    pub input_token: Pubkey,
    pub output_token: Pubkey,
    pub input_amount: u64,
    pub output_amount: u64,
    pub destination_chain_id: u64,
    pub exclusive_relayer: Pubkey,
    pub fill_deadline_offset: u32,
    pub exclusivity_period: u32,
    pub message: &'a Vec<u8>,
```
avoids cloning

programs/svm-spoke/src/instructions/deposit.rs
```diff
@@ -101,21 +105,27 @@ pub fn _deposit(
...
             return err!(CommonError::InvalidExclusiveRelayer);
         }
     }
 
-    // Depositor must have delegated input_amount to the state PDA.
+    // Verify delegate PDA
+    let (pda, bump) = Pubkey::find_program_address(&[b"delegate", &delegate_seed_hash], &ctx.program_id);
+    if pda != ctx.accounts.delegate.key() {
+        return err!(SvmError::InvalidDelegatePda);
+    }
```
Reinis-FRP, line 117: better move this logic inside `transfer_from` by passing `delegate_seed_hash` as we have the same logic for fills

programs/svm-spoke/src/instructions/fill.rs
```diff
@@ -114,15 +118,23 @@ pub fn fill_relay<'info>(
         _ => FillType::FastFill,
     };
 
-    // Relayer must have delegated output_amount to the state PDA
+    // Verify delegate PDA
+    let seed_hash = derive_seed_hash(&(FillSeedData { relay_hash, repayment_chain_id, repayment_address }));
+    let (pda, bump) = Pubkey::find_program_address(&[b"delegate", &seed_hash], &ctx.program_id);
+    if ctx.accounts.delegate.key() != pda {
+        return err!(SvmError::InvalidDelegatePda);
+    }
```
Reinis-FRP, line 126: better move this logic inside `transfer_from` by passing `seed_hash` as we have the same logic for deposits

programs/svm-spoke/src/instructions/deposit.rs
```diff
@@ -125,7 +119,7 @@ pub fn _deposit(
         &ctx.accounts.mint,
         &ctx.accounts.token_program,
         delegate_seed_hash,
-        bump,
+        &ctx.program_id,
```
Reinis-FRP, line 122: do we need to pass it as param? can we use &SvmSpoke::id() instead?

Reinis-FRP (approved):
nice!

mrice32 (approved):

PR #962, opened 2025-04-17
fix(svm): N-03 use consistent variable names in the instruction constraint

Reinis-FRP:

> The `fill_relay` function [accepts a `_relay_hash` parameter](https://github.com/across-protocol/contracts/blob/71e990b3f908ecf994be8723e041b584c7d49318/programs/svm-spoke/src/lib.rs#L409) but the `FillRelay` struct constraint [calls it `relay_hash`](https://github.com/across-protocol/contracts/blob/71e990b3f908ecf994be8723e041b584c7d49318/programs/svm-spoke/src/instructions/fill.rs#L19) (without the leading underscore).
>
> To avoid confusion, consider maintaining consistent variable names when using the `instruction` constraint.

This PR addresses the issue by using the same variable names when using the `instruction` constraint.


md0x (approved):
Looks good!

PR #968, opened 2025-04-22
feat(svm): deterministic anchor builds

md0x:


scripts/anchor-test.sh
```diff
@@ -1,19 +1,24 @@
...
+    anchor idl build \
+      -p "$dir_name" \
+      -o "target/idl/$program_name.json" \
+      -t "target/types/$program_name.ts"
 done
 
-echo "🧪 Running anchor tests..."
-anchor test --skip-build
+echo "🧪 Running deterministic tests (test feature, no IDL)…"
+anchor test --verifiable --no-idl -- --features test
```
Reinis-FRP, line 24: does test has these flags? I think here we should just leave `--skip-build` as built should have been performed above

PR #969, opened 2025-04-22
chore: update scroll adapter to interface with v2 message queue

bmzig:
It's probably a good idea to update the scroll gas price oracle contract to the V2 message queue so that our estimated fees to send to L2 match the fees the messenger relay expects to receive.


nicholaspai (commented):
I think you need to update deployments/deployments.json

though I wonder if we even use the Adapter addresses in that file anywhere and whether we can remove it.

nicholaspai (approved):

PR #972, opened 2025-04-23
fix: upgrade anchor to 0.31.1

Reinis-FRP:
#960 was a workaround for dealing with a breaking `proc-macro2` change. Now that Anchor has fixed it [here](https://github.com/solana-foundation/anchor/pull/3663) and included in the [latest `0.31.1` release](https://github.com/solana-foundation/anchor/releases/tag/v0.31.1), we can revert the pinning of toolchain and custom anchor scripts and use the released Anchor instead.


.github/workflows/pr.yml
```diff
@@ -16,23 +16,23 @@ jobs:
         with:
           node-version: "${{ matrix.node }}"
           cache: yarn
-      - name: Setup Anchor & Build
-        uses: metadaoproject/setup-anchor@v2
```
Reinis-FRP, line 20: this used `@coral-xyz/anchor-cli` npm package which has not been updated for a while, hence switching to `solana-developers/github-actions` which autodetects the used versions and installs them.

md0x (approved):

PR #976, opened 2025-04-25
N-02 Relay Tokens From L1 Emits Empty Transaction Hash When Relaying USDC through CCTP

bmzig:
> The relayTokens functions [[1]](https://github.com/across-protocol/contracts/blob/77761d74152e319e6edd6d4d766cb46c39e6f38d/contracts/chain-adapters/ZkStack_Adapter.sol#L154) [[2]](https://github.com/across-protocol/contracts/blob/77761d74152e319e6edd6d4d766cb46c39e6f38d/contracts/chain-adapters/ZkStack_CustomGasToken_Adapter.sol#L178) are used to bridge tokens from L1 to ZK Stack. Both instances define a txHash variable [[1]](https://github.com/across-protocol/contracts/blob/77761d74152e319e6edd6d4d766cb46c39e6f38d/contracts/chain-adapters/ZkStack_Adapter.sol#L164) [[2]](https://github.com/across-protocol/contracts/blob/77761d74152e319e6edd6d4d766cb46c39e6f38d/contracts/chain-adapters/ZkStack_CustomGasToken_Adapter.sol#L189) which will later be assigned to the transaction hash value returned by the BRIDGE_HUB when initiating a bridging transaction. However, in both cases [[1]](https://github.com/across-protocol/contracts/blob/77761d74152e319e6edd6d4d766cb46c39e6f38d/contracts/chain-adapters/ZkStack_Adapter.sol#L182-L185) [[2]](https://github.com/across-protocol/contracts/blob/77761d74152e319e6edd6d4d766cb46c39e6f38d/contracts/chain-adapters/ZkStack_CustomGasToken_Adapter.sol#L226-L228) when CCTP is enabled, the _transferUsdc function is used, bypassing the BRIDGE_HUB. This function does not return a transaction hash, however, causing the relayTokens functions to emit an empty ZkStackMessageRelayed event [[1]](https://github.com/across-protocol/contracts/blob/77761d74152e319e6edd6d4d766cb46c39e6f38d/contracts/chain-adapters/ZkStack_Adapter.sol#L222) [[2]](https://github.com/across-protocol/contracts/blob/77761d74152e319e6edd6d4d766cb46c39e6f38d/contracts/chain-adapters/ZkStack_CustomGasToken_Adapter.sol#L266). This behavior can be confusing for users, especially when trying to index the event based on the emitted hash.

> In case no off-chain components rely on the emitted event, consider removing the emitted event to avoid confusion. Otherwise, consider adding thorough documentation for the ZkStackMessageRelayed event, outlining its expected behavior.

I don't believe we use `ZkStackMessageRelayed` anywhere off-chain, so I agree with the suggestion to remove it if it is unnecessary.

### DIFF
--- a/contracts/chain-adapters/ZkStack_Adapter.sol
+++ b/contracts/chain-adapters/ZkStack_Adapter.sol
@@ -58,8 +58,6 @@ contract ZkStack_Adapter is AdapterInterface, CircleCCTPAdapter {
     // when calling a hub pool message relay, which would otherwise cause a large amount of ETH to be sent to L2.
     uint256 private immutable MAX_TX_GASPRICE;
 
-    event ZkStackMessageRelayed(bytes32 indexed canonicalTxHash);
-
     error ETHGasTokenRequired();
     error TransactionFeeTooHigh();
     error InvalidBridgeConfig();
@@ -140,7 +138,6 @@ contract ZkStack_Adapter is AdapterInterface, CircleCCTPAdapter {
         );
 
         emit MessageRelayed(target, message);
-        emit ZkStackMessageRelayed(canonicalTxHash);
     }
 
     /**
@@ -219,7 +216,6 @@ contract ZkStack_Adapter is AdapterInterface, CircleCCTPAdapter {
         }
 
         emit TokensRelayed(l1Token, l2Token, amount, to);
-        emit ZkStackMessageRelayed(txHash);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


nicholaspai (approved):

pxrl (approved):
+1 no instances of `ZkStackMessageRelayed` are found in the relayer repository.

PR #985, opened 2025-04-30
feat(svm): fill with across plus and codama test

md0x:
Changes proposed in this PR:
- Add test to send fill with across plus message and codama client


Reinis-FRP (approved):
great! we still might want to figure out how to do encoding properly as currently we rely on some of anchor classes and types that they don't export globally, e.g.:
```
import { IdlCoder } from "@coral-xyz/anchor/dist/cjs/coder/borsh/idl";
import { IdlTypeDef } from "@coral-xyz/anchor/dist/cjs/idl";
```

PR #990, opened 2025-05-05
redeploy spoke pool verifier

nicholaspai:
- feat(SpokePoolVerifier): Redeploy with outputToken as parameter
- feat: Deploy SpokePoolVerifier with outputToken as parameter


dohaki (approved):
LGTM

PR #997, opened 2025-05-09
fix: verified svm builds

Reinis-FRP:


.github/actions/generate-evm-artifacts/action.yml
```diff
@@ -0,0 +1,19 @@
...
+  using: "composite"
+  steps:
+    - name: "Use Node ${{ inputs.node_version }}"
+      uses: actions/setup-node@v3
+      with:
+        node-version: "${{ inputs.node_version }}"
+        cache: yarn
+    - name: Install packages
+      shell: bash
+      run: yarn install --frozen-lockfile
```
Reinis-FRP, line 16: This still have below errors here when building optional `node-hid` and `cpu-features`. We cannot do `--ignore-optional` here like in other places as `hardhat compile` depends on other optional dependencies being installed.

```
warning Error running install script for optional dependency: "/home/runner/work/contracts/contracts/node_modules/node-hid: Command failed.
info This module is OPTIONAL, you can safely ignore this error

warning Error running install script for optional dependency: "/home/runner/work/contracts/contracts/node_modules/cpu-features: Command failed.
info This module is OPTIONAL, you can safely ignore this error
```

.github/workflows/pr.yml
```diff
@@ -1,38 +1,88 @@
 name: Checks
 on: [push, pull_request]
+env:
+  NODE_VERSION: 20
```
Reinis-FRP, line 4: Before we had matrix strategy, but that got overly repetitive when splitting to multiple jobs. And we were not using the matrix functionality anyways as all jobs were run against a single node version.

.github/workflows/pr.yml
```diff
@@ -1,38 +1,88 @@
...
           cache: yarn
-      - name: Extract Solana versions
-        uses: solana-developers/github-actions/extract-versions@v0.2.5
-        id: versions
-      - name: Setup Anchor & Solana
-        uses: solana-developers/github-actions/setup-all@v0.2.5
-        with:
-          anchor_version: ${{ steps.versions.outputs.anchor_version }}
-          solana_version: ${{ steps.versions.outputs.solana_version }}
-          node_version: 20
```
Reinis-FRP, line 27: solana and anchor binaries were not needed for lint

.github/workflows/publish.yml
```diff
@@ -40,8 +40,8 @@ jobs:
           always-auth: true
           registry-url: "https://registry.npmjs.org"
 
-      # Perform installs, run tests, run a build step, etc. here, as needed.
-      - run: yarn
+      # Perform installs and run a build step.
+      - run: yarn install --frozen-lockfile && yarn build
```
Reinis-FRP, line 44: With removed `prepublish` script we have to build explicitly. For now this includes regular SVM build without caching, but would need to create a follow-up PR that covers verified builds.

package.json
```diff
@@ -26,19 +26,23 @@
...
-    "generate-svm-assets": "sh ./scripts/generate-svm-assets.sh && yarn generate-svm-clients",
-    "generate-svm-clients": "yarn ts-node ./scripts/svm/utils/generate-svm-clients.ts && yarn ts-node ./scripts/svm/utils/rename-clients-imports.ts",
+    "generate-svm-artifacts": "bash ./scripts/svm/buildHelpers/buildIdl.sh && sh ./scripts/svm/buildHelpers/generateSvmAssets.sh && yarn generate-svm-clients",
+    "generate-svm-clients": "yarn ts-node ./scripts/svm/buildHelpers/generateSvmClients.ts && yarn ts-node ./scripts/svm/buildHelpers/renameClientsImports.ts",
     "build-evm": "hardhat compile",
-    "build-svm": "echo 'Generating IDLs...' && anchor build > /dev/null 2>&1 || true && anchor run generateExternalTypes && anchor build",
-    "build-ts": "rm -rf ./dist && tsc && rsync -a --include '*/' --include '*.d.ts' --exclude '*' ./typechain ./dist/",
-    "copy-idl": "mkdir -p dist/src/svm/assets/idl && cp src/svm/assets/idl/*.json dist/src/svm/assets/idl/",
-    "build": "yarn build-evm && yarn build-svm && yarn generate-svm-assets && yarn build-ts && yarn copy-idl",
+    "build-svm": "bash ./scripts/svm/buildHelpers/buildSvmLocalToolchain.sh",
```
Reinis-FRP, line 32: this was unnecessarily duplicating `anchor build`.

package.json
```diff
@@ -26,19 +26,23 @@
...
+    "build-svm-solana-verify": "bash ./scripts/svm/buildHelpers/buildSolanaVerify.sh",
+    "build-ts": "rm -rf ./dist && tsc && rsync -a --include '*/' --include '*.d.ts' --exclude '*' ./typechain ./dist/ && yarn export-idl",
+    "export-idl": "mkdir -p dist/src/svm/assets/idl && cp src/svm/assets/idl/*.json dist/src/svm/assets/idl/",
```
Reinis-FRP, line 35: moved `export-idl` to `build-ts` (renamed from `copy-idl` before) as it is relevant for `dist` files.

package.json
```diff
@@ -26,19 +26,23 @@
...
-    "test-svm": "anchor test -- --features test",
+    "test-svm": "IS_TEST=true yarn build-svm && yarn generate-svm-artifacts && anchor test --skip-build",
+    "test-svm-solana-verify": "IS_TEST=true yarn build-svm-solana-verify && yarn generate-svm-artifacts && anchor test --skip-build",
     "test": "yarn test-evm && yarn test-svm",
+    "test-verified": "yarn test-evm && yarn test-svm-solana-verify",
     "test:report-gas": "IS_TEST=true REPORT_GAS=true hardhat test",
-    "generate-evm-assets": "rm -rf typechain && TYPECHAIN=ethers yarn hardhat typechain",
-    "prepublish": "yarn build && hardhat export --export-all ./cache/massExport.json && ts-node ./scripts/processHardhatExport.ts && prettier --write ./deployments/deployments.json && yarn generate-evm-assets",
+    "generate-evm-artifacts": "rm -rf typechain && TYPECHAIN=ethers yarn hardhat typechain",
+    "process-hardhat-export": "hardhat export --export-all ./cache/massExport.json && ts-node ./scripts/processHardhatExport.ts && prettier --write ./deployments/deployments.json",
```
Reinis-FRP, line 45: We had this as part of `prepublish` script, but these commands were ran after `yarn build` which already had populated the `dist` files, so this could have potentially left local tree inconsistent with dist. Hence, these steps are moved to a dedicated script if anyone still needs them.

Also removed typechain generation completely as they are already generated in `hardhat compile` with `@typechain/hardhat` plugin.

.github/workflows/pr.yml
```diff
@@ -1,38 +1,88 @@
...
+    runs-on: ubuntu-latest
+    env:
+      ARTIFACTS_PATHS: |
+        target/idl
+        target/types
+        src/svm/assets
+        src/svm/clients
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v3
```
md0x, line 18: Now that we’ve got more jobs, it might be worth caching `actions/checkout@v3` too, using something like `hashFiles('**/.git/**')`. We’re doing a v3 checkout in 7 different jobs, plus a v2 checkout in the publish step.

.github/actions/setup-solana-anchor/action.yml
```diff
@@ -0,0 +1,22 @@
...
+    description: "Solana verify version to install (leave empty to skip)"
+    required: false
+    default: ""
+  node_version:
+    description: "Node version to install"
+    required: true
+runs:
+  using: "composite"
+  steps:
+    - name: Extract Solana versions
```
md0x, line 13: Nit: think we could reuse this in the publish workflow too?

Reinis-FRP: yes, indeed - I planned handling publish workflow in a follow-up

package.json
```diff
@@ -26,19 +26,23 @@
...
+    "build-svm": "bash ./scripts/svm/buildHelpers/buildSvmLocalToolchain.sh",
+    "build-svm-solana-verify": "bash ./scripts/svm/buildHelpers/buildSolanaVerify.sh",
+    "build-ts": "rm -rf ./dist && tsc && rsync -a --include '*/' --include '*.d.ts' --exclude '*' ./typechain ./dist/ && yarn export-idl",
+    "export-idl": "mkdir -p dist/src/svm/assets/idl && cp src/svm/assets/idl/*.json dist/src/svm/assets/idl/",
+    "build": "yarn build-evm && yarn build-svm && yarn generate-svm-artifacts && yarn build-ts",
+    "build-verified": "yarn build-evm && yarn build-svm-solana-verify && yarn generate-svm-artifacts && yarn build-ts",
     "test-evm": "IS_TEST=true hardhat test",
-    "test-svm": "anchor test -- --features test",
+    "test-svm": "IS_TEST=true yarn build-svm && yarn generate-svm-artifacts && anchor test --skip-build",
+    "test-svm-solana-verify": "IS_TEST=true yarn build-svm-solana-verify && yarn generate-svm-artifacts && anchor test --skip-build",
```
md0x, line 40: On question: do we need to do `generate-svm-artifacts` here given that in the CI we download the cached ones?
https://github.com/across-protocol/contracts/blob/66e3f2b7d28a249f26af94920519bc643188950b/.github/workflows/pr.yml#L142

Reinis-FRP, line 40: In CI we invoke `anchor test --skip-build` directly in order to reuse build that was potentially grabbed from cache, but `test-svm` and `test-svm-solana-verify` scripts are only meant for running locally

md0x (dismissed):
Great work!

PR #998, opened 2025-05-13
feat: optionally adjust output amount by swap price improvement over min

mrice32:
This PR adds an option so that callers can have their bridge output increased proportionally to the improvement the swap returns over the min allowed swap output.

This essentially assumes the bridge quote was generated based on the min and that the percentage should stay fixed.

### DIFF
--- a/contracts/SpokePoolV3Periphery.sol
+++ b/contracts/SpokePoolV3Periphery.sol
@@ -142,11 +142,9 @@ contract SpokePoolV3Periphery is SpokePoolV3PeripheryInterface, Lockable, MultiC
         uint256 acrossOutputAmount
     );
 
-    /**
-     *
+    /****************************************
      *                ERRORS                *
-     *
-     */
+     ****************************************/
     error InvalidPermit2();
     error ContractInitialized();
     error InvalidSignatureLength();
--- a/contracts/SpokePoolV3Periphery.sol
+++ b/contracts/SpokePoolV3Periphery.sol
@@ -641,9 +641,12 @@ contract SpokePoolV3Periphery is SpokePoolV3PeripheryInterface, Lockable, MultiC
         // that we weren't partial filled).
         if (srcBalanceBefore - _swapToken.balanceOf(address(this)) != _swapTokenAmount) revert LeftoverSrcTokens();
 
-        // Calculate adjusted output amount proportionally when we receive more than expected
+        // Calculate adjusted output amount proportionally when we receive more than expected and adjustment is enabled
         uint256 adjustedOutputAmount = swapAndDepositData.depositData.outputAmount;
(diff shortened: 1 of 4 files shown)


contracts/SpokePoolV3Periphery.sol
```diff
@@ -533,8 +533,9 @@ contract SpokePoolV3Periphery is SpokePoolV3PeripheryInterface, Lockable, MultiC
         bytes32 typedDataHash,
         bytes calldata signature
     ) private view {
-        if (!SignatureChecker.isValidSignatureNow(signatureOwner, _hashTypedDataV4(typedDataHash), signature))
+        if (!SignatureChecker.isValidSignatureNow(signatureOwner, _hashTypedDataV4(typedDataHash), signature)) {
```
mrice32, line 536: formatting

contracts/SpokePoolV3Periphery.sol
```diff
@@ -602,9 +603,11 @@ contract SpokePoolV3Periphery is SpokePoolV3PeripheryInterface, Lockable, MultiC
 
         // The exchange will either receive funds from this contract via a direct transfer, an approval to spend funds on this contract, or via an
         // EIP1271 permit2 signature.
-        if (_transferType == TransferType.Approval) _swapToken.forceApprove(_exchange, _swapTokenAmount);
-        else if (_transferType == TransferType.Transfer) _swapToken.transfer(_exchange, _swapTokenAmount);
-        else {
+        if (_transferType == TransferType.Approval) {
+            _swapToken.forceApprove(_exchange, _swapTokenAmount);
+        } else if (_transferType == TransferType.Transfer) {
+            _swapToken.transfer(_exchange, _swapTokenAmount);
```
mrice32, line 609: formatting

contracts/SpokePoolV3Periphery.sol
```diff
@@ -636,6 +639,17 @@ contract SpokePoolV3Periphery is SpokePoolV3PeripheryInterface, Lockable, MultiC
...
 
+        // Calculate adjusted output amount proportionally when we receive more than expected and adjustment is enabled
+        uint256 adjustedOutputAmount = swapAndDepositData.depositData.outputAmount;
+        if (
+            returnAmount > swapAndDepositData.minExpectedInputTokenAmount &&
+            swapAndDepositData.enableProportionalAdjustment
+        ) {
+            adjustedOutputAmount =
+                (swapAndDepositData.depositData.outputAmount * returnAmount) /
+                swapAndDepositData.minExpectedInputTokenAmount;
```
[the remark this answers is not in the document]
mrice32: I think it could be. From a practical perspective that seems like a really bad idea (allowing slippage to send you to zero, but we may want to protect against it.

PR #999, opened 2025-05-13
fix: verified svm build in release

Reinis-FRP:
This continues from #997 to add verified SVM binaries building to release workflow. These are published separate from NPM package and use GitHub release assets instead.


.github/actions/cache-evm-artifacts/action.yml
```diff
@@ -11,14 +11,19 @@ outputs:
...
+        node_version: ${{ inputs.node_version }}
     - name: Restore cached EVM artifacts
       id: evm-artifacts-cache
       uses: actions/cache@v4
       with:
         # The job that generates the artifacts is responsible for archiving them to the cache tarball. This avoids any
         # conflicts with other caching actions that might have cleaned some of cached contents.
         path: evm-artifacts.tar
-        key: evm-artifacts-${{ runner.os }}-node-${{ inputs.node_version }}-${{ hashFiles('yarn.lock', 'hardhat.config.ts', 'contracts/**/*.sol') }}
+        key: evm-artifacts-${{ runner.os }}-node-${{ steps.resolved-node.outputs.version }}-${{ hashFiles('yarn.lock', 'hardhat.config.ts', 'contracts/**/*.sol') }}
```
Reinis-FRP, line 26: This makes it possible to reuse cache on release when it was generated during the latest push to master

PR #1000, opened 2025-05-13
feat: add balance injection to multicall handler calls

mrice32:
This is a bit clunky, but it adds a way to do balance injection for calls to multicall handler, so the calldata can use the updated balance.

It's done as a nested call to keep the MulticallHandler backwards compatible.

### DIFF
--- a/contracts/handlers/MulticallHandler.sol
+++ b/contracts/handlers/MulticallHandler.sol
@@ -22,6 +22,11 @@ contract MulticallHandler is AcrossMessageHandler, ReentrancyGuard {
         uint256 value;
     }
 
+    struct Replacement {
+        address token;
+        uint256 offset;
+    }
+
     struct Instructions {
         //  Calls that will be attempted.
         Call[] calls;
@@ -40,6 +45,9 @@ contract MulticallHandler is AcrossMessageHandler, ReentrancyGuard {
     error CallReverted(uint256 index, Call[] calls);
     error NotSelf();
     error InvalidCall(uint256 index, Call[] calls);
+    error ReplacementCallFailed(bytes calldData);
+    error CalldataTooShort(uint256 callDataLength, uint256 offset);
+
 
     modifier onlySelf() {
         _requireSelf();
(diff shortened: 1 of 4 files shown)


contracts/handlers/MulticallHandler.sol
```diff
@@ -112,6 +120,44 @@ contract MulticallHandler is AcrossMessageHandler, ReentrancyGuard {
...
+                // Get the pointer to the offset that the caller wants to overwrite.
+                let ptr := add(callData, offset)
+                // Get the current value at the offset.
+                let current := mload(ptr)
+                // Or the current value with the new value.
+                // Reasoning:
+                // - caller should 0-out any portion that they want overwritten.
+                // - if the caller is storing the balance in a smaller integer, like a uint160 or uint128,
+                //   the higher bits will be 0 and not overwrite any other data in the calldata
+                // - The catch: the smaller integer where they want to store the balance must end no
```
bmzig, line 149: I don't understand this comment. Is this saying that it might not overwrite the calldata properly if `bal` exceeds the maximum value for the function input we are trying to call and our current calldata is not inputting the max? e.g. for calling a function with a u160 input, we could have `bal=2^160` while the calldata is specifying `2^120` or something, and we will just end up calling the function with input `2^120` and not `2^160-1`?

mrice32: Yeah, maybe this comment could be improved.

If the balance is too large, that's one possible issue. That would result in just overwriting bytes that you didn't intend to overwrite (the bytes for the value that precedes it).

The comment is getting at something different:
- Let's say I want to replace a `uint160` (20 bytes) with the balance.
- Let's say I have a function that takes a single packed arg. Something like `myFun(bytes32 packedData)`.
- If the `uint160` balance is encoded in the 20 bytes of the `bytes32`, then what value do I use as my offset?
- I want the balance value to _end_ at the 24th byte (4 byte selector + 20 bytes of integer).
- That would mean I need an offset of -8, which I can't express.

There may be a workaround to make this work... we could allow negative offsets and perform a `shl` on the balance by that number of bytes and replacing at offset 0.

mrice32: https://github.com/across-protocol/contracts/pull/1000/commits/b9b093e62a994a3ea304241834a2ac46da756c8f

package.json
```diff
@@ -113,7 +113,7 @@
     "prettier": "^2.3.2",
     "prettier-plugin-solidity": "^1.0.0-beta.13",
     "pretty-quick": "^2.0.1",
-    "solhint": "^3.3.6",
+    "solhint": "^3.6.2",
```
mrice32, line 116: Solhint couldn't handle "memory-safe" assembly syntax.

PR #1010, opened 2025-05-29
fix(svm): deposit scripts

md0x:
Changes proposed in this PR:
- Update deposit scripts to work with latest SvmSpoke updates

### DIFF
--- a/scripts/svm/simpleDeposit.ts
+++ b/scripts/svm/simpleDeposit.ts
@@ -9,10 +9,16 @@ import {
   getAssociatedTokenAddressSync,
   getMint,
 } from "@solana/spl-token";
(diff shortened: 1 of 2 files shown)


scripts/svm/simpleDeposit.ts
```diff
@@ -90,20 +96,49 @@ async function deposit(): Promise<void> {
...
+    message,
+  };
+  const delegatePda = getDepositPda(depositData, program.programId);
+
   // Delegate state PDA to pull depositor tokens.
   const approveIx = await createApproveCheckedInstruction(
     userTokenAccount,
     inputToken,
-    statePda,
+    delegatePda,
```
Reinis-FRP, line 119: I think we should also fix this in `simpleFill` script as well

Reinis-FRP (approved):

PR #1014, opened 2025-06-03
chore: Update & deploy new WorldChain implementation

pxrl:
Updated for CCTP.


deploy/050_deploy_worldchain_adapter.ts
pxrl, line 1: nb. I wasn't successful in verifying via this adapter due to the Etherscan v2 migration. I ended up having to deploy & verify with foundry instead, so this deployment script is mostly for documentation.

bmzig (approved):
It doesn't look like the world chain spoke pool was verified https://worldscan.org/address/0x1771c470d41b8c39338450C380bf2C080a2CEdD8

pxrl:
> It doesn't look like the world chain spoke pool was verified https://worldscan.org/address/0x1771c470d41b8c39338450C380bf2C080a2CEdD8

We're setup to verify via BlockScout, so it's done there: https://worldchain-mainnet.explorer.alchemy.com/address/0x1771c470d41b8c39338450C380bf2C080a2CEdD8?tab=contract

I think we might have a couple of chains in this same situation because Blockscout typically seems to be day-0 and Etherscan clones come in the days/weeks after. I do think WorldScan is preferable to use, especially given the weird long hostname for BlockScout on WorldChain.

deployments/worldchain/WorldChain_SpokePool.json
```diff
@@ -1,5 +1,5 @@
 {
-  "address": "0xAcDE84590A2ba1aCfEbB220997Eb41B7A01f271f",
+  "address": "0x1771c470d41b8c39338450C380bf2C080a2CEdD8",
```
pxrl, line 2: nb. Address matches the current implementation address: https://worldchain-mainnet.explorer.alchemy.com/address/0x09aea4b2242abC8bb4BB78D537A67a245A7bEC64

There's an inconsistency here in that we tend to track implementation addresses when we'd probably rather track proxy deployments, but the current address 0xAcDE... is a stale implementation, so may as well update it (and pull in the updated ABI).

PR #1045, opened 2025-06-11
fix(svm): use bytes32 for foreign amounts

Reinis-FRP:
Across on SVM should support foreign token amounts that could potentially not fit in the `u64` type. This changes the type to `[u8; 32]` which would now apply to `output_amount` for deposits and `input_amount` for fills and slow fills. The proposed implementation uses these same raw values in event logs and it is expected that Across protocol UMIP would dictate to use big-endianess encoding for these types when matching deposits/fills on EVM chains.


programs/svm-spoke/src/instructions/deposit.rs
```diff
@@ -21,15 +21,7 @@ use crate::{
 
 #[event_cpi]
 #[derive(Accounts)]
-#[instruction(
-    depositor: Pubkey,
-    recipient: Pubkey,
-    input_token: Pubkey,
-    output_token: Pubkey,
-    input_amount: u64,
-    output_amount: u64,
```
Reinis-FRP, line 30: I was going to change `output_amount` type, but realized we only need args up to `input_token` here

mrice32 (approved):

md0x (approved):

PR #1065, opened 2025-08-14
chore: Bump node 16 (20) -> 22

pxrl:
16 is long EoL but we were in practice using 20 in CI.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@across-protocol/contracts",
-  "version": "4.1.4",
+  "version": "4.1.3",
   "author": "UMA Team",
   "license": "AGPL-3.0-only",
   "repository": {


.github/actions/setup-node-if-needed/action.yml
```diff
@@ -15,7 +15,7 @@ runs:
       id: check-node
       shell: bash
       run: |
-        if command -v node >/dev/null 2>&1; then
+        if node --version 2>/dev/null | grep -q "v${{ inputs.node_version }}"; then
```
pxrl, line 18: This pre-existing check would resolve true if the installed version of node was different to what was specified via `inputs.node_version`. Harden it a bit by verifying that the actual installed version of node reports something that gives a strong hint at being the right version. This _should_ be compatible with specifying additional constraints on the version - i.e. minor and patch levels.

I wonder if there's a cleaner way to do this - i.e. I'd suspect `setup-node` has some built-in version checker, but since action was only recently by @Reinis-FRP I'm assuming there was some blocker there.

nicholaspai (dismissed):

PR #1066, opened 2025-08-14
chore: Bump kit 2.1.0 -> 2.3.0

pxrl:


pxrl:
nb. there might be showstoppers here - really not sure, so this PR is exploratory.

PR #1044, opened 2025-06-11
feat: Added foundry deploy script

fusmanii:
First attempt at deploy script migration to foundry. Just did one deploy script and looking for feedback.

List of changes:
- Created `DeployHubPool.s.sol` and `script/DeployEthereumSpokePool.s.sol` inside `script` folder
- Created `DeployConstants.sol` inside `script` folder that works similar to `deploy/consts.ts` by providing third party smart contract addresses
- Created `DeploymentUtils.sol` inside `script` that works similar to `utils/utils.hre.ts`
- Contract can be deployed using `forge script script/DeployHubPool.s.sol:DeployHubPool --rpc-url $RPC_URL --broadcast --verify -vvvv` which deploys and verifies the contract and adds the deploy info in the `broadcast` folder
- Created a script `script/extract_foundry_addresses.sh` that extracts all the deployed addresses from the `broadcast` folder and puts it in a json, an md file and a sol file (similar to `deployments.json`)

Looking for feedback on the proposed flow and how I can make this better and address some of the short comings of the current deployment flow 

Fixes https://linear.app/uma/issue/UMA-2906/contracts-migrate-smart-contract-deploy-to-foundry

PR #1068, opened 2025-08-14
feat(Foundry): Part 1 - Added address extraction script

fusmanii:
Splitting this https://github.com/across-protocol/contracts/pull/1044 into multiple PRs

Part 1: Extract addresses script that generates `DeployedAddresses.sol` that can be used in foundry deploy scripts

After every deploy, just need to run `yarn extract-addresses` and deployed addresses contract will get updated with the latest addresses from `broadcast` folder and `deployments.json` file

More info in the attached EXTRACT_ADDRESSES.md

Diff is large because of the `broadcast/deployed-addresses.json` and `broadcast/deployed-addresses.md` files


script/utils/ExtractDeployedFoundryAddresses.ts
```diff
@@ -0,0 +1,589 @@
...
+      return contracts;
+    } catch (error) {
+      console.error(`Error reading ${broadcastFile.filePath}: ${error}`);
+      return [];
+    }
+  }
+}
+
+function getChainName(chainId: number): string {
+  const chainNames: { [id: number]: string } = {
```
nicholaspai: We're going to have to keep this in sync with other constants from other repos, like [this one in Constants](https://github.com/across-protocol/constants/blob/master/src/networks.ts) can you import from there instead so we can reduce the number of constants we need to keep updated?

script/utils/ExtractDeployedFoundryAddresses.ts
```diff
@@ -0,0 +1,589 @@
...
+  // Remove leading/trailing underscores
+  sanitized = sanitized.replace(/^_+|_+$/g, "");
+  // Ensure it starts with a letter
+  if (sanitized && /^\d/.test(sanitized)) {
+    sanitized = "CONTRACT_" + sanitized;
+  }
+  return sanitized.toUpperCase();
+}
+
+function generateFoundryScript(broadcastFiles: BroadcastFile[], outputFile: string): void {
```
nicholaspai: what is the purpose of auto generating this contract?

script/utils/ExtractDeployedFoundryAddresses.ts
fusmanii: good point, will update it

script/utils/ExtractDeployedFoundryAddresses.ts
fusmanii: Originally I was planning to hardcode all the deploy addresses into that contract but after feedback went with having this contract read from a json file. Now that you mention it, I dont think I need to generate it every time

script/utils/ExtractDeployedFoundryAddresses.ts
fusmanii: done https://github.com/across-protocol/contracts/pull/1068/commits/31853e3ce0e3fad25e98a29723ac76c16758211a

script/utils/ExtractDeployedFoundryAddresses.ts
fusmanii: https://github.com/across-protocol/contracts/pull/1068/commits/ba64e013eb5b9dee3c0f2e0885f19fcf313ef0a3

pxrl:
@fusmanii Did we look at `forge-deploy` for this purpose? It's made by the same developer as `hardhat-deploy`, so I wonder if it might be more of a drop-in replacement for use with foundry.

https://github.com/wighawag/forge-deploy

script/utils/EXTRACT_ADDRESSES.md
```diff
@@ -0,0 +1,199 @@
...
+- Base (Chain ID: 8453)
+- Base Sepolia (Chain ID: 84532)
+- BSC (Chain ID: 56)
+- Lens (Chain ID: 232)
+- Boba (Chain ID: 288)
+- zkSync Era (Chain ID: 324)
+- World Chain (Chain ID: 480)
+- Redstone (Chain ID: 690)
+- Lisk (Chain ID: 1135)
+- List Sepolia (Chain ID: 4202)
```
pxrl:
```suggestion
- Lisk Sepolia (Chain ID: 4202)
```

script/utils/ExtractDeployedFoundryAddresses.ts
```diff
@@ -0,0 +1,386 @@
+#!/usr/bin/env node
```
pxrl: Does it work to call `node` on a ts file?

script/utils/EXTRACT_ADDRESSES.md
fusmanii: https://github.com/across-protocol/contracts/pull/1068/commits/037f8ffa0529df6eda4379176fa8369bb8869279

script/utils/ExtractDeployedFoundryAddresses.ts
fusmanii: forgot to remove, done https://github.com/across-protocol/contracts/pull/1068/commits/1db9230ffdeaa121b8241b89bef53559ac9a6553#diff-175744961939b2f9d0f6f4d483df761c3f3ed16129129e21842eb59dfb4dd1c0L1

fusmanii:
@pxrl first time hearing about https://github.com/wighawag/forge-deploy, taking a look

pxrl:
> @pxrl @fusmanii For forge-deploy, I see that the last commit to it was 9 months ago. Do you think it's worth adding a dependency that's not actively maintained? IMO we should avoid using deps like these if feasible. Thoughts?

Ah, I hadn't spotted that. 9 months with no activity is a bit of a red flag, though the author does otherwise have a really solid track record. If it worked and solved our use case then I think it'd at least be worth evaluating further, but if it doesn't work that's a showstopper.

In general I just want to be sure that we've researched and evaluated any off-the-shelf tooling for this - we're not the only project with this use-case so it seems really odd that it's so under-served.

broadcast/deployed-addresses.md
pxrl, line 1: @fusmanii We have chainIds organised by testnet & prod up in the `constants` repo - wdyt about splitting the output accordingly here in deployed-addresses.md and deployed-addresses.json?

script/utils/EXTRACT_ADDRESSES.md
```diff
@@ -0,0 +1,199 @@
...
+3. It extracts contract addresses from each file's transaction data
+4. It organizes the data by chain ID and contract name
+5. It generates the three output files with the extracted information
+6. The Solidity contract uses Foundry's `parseJson` functionality to read addresses dynamically from the JSON file
+7. All addresses are properly formatted using EIP-55 checksum for Solidity compatibility
+
+## Important Notes
+
+- Run the extraction script after each deployment to keep the addresses up to date
+- The script only processes the latest deployment for each script/chain combination
```
pxrl, line 195: ooc, how does this handle deploying new implementation contracts? That was one area that hardhat didn't handle very well - it apparently couldn't distinguish between proxy & implementation deploments and the recorded metadata always corresponded to the latest implementation deployment, so that typically had to be dropped from the resulting commit. (nb. this might also just be due to a deficiency in our deployment scripts or a misunderstanding of how to track this properly).

broadcast/deployed-addresses.md
fusmanii: good idea, update the md file here https://github.com/across-protocol/contracts/pull/1068/commits/ebc36ce9ee714b16a2d95ef2ca8b7662cf52ea3e 
left json file as is since it would make the DeployedAddresses.sol contract that reads that json file very convoluted to check if its mainnet vs testnet. Foundry doesn't have a good way to check if a json path exists in order to fetch addresses for correct chain id

script/utils/EXTRACT_ADDRESSES.md
fusmanii: In my script [here](https://github.com/across-protocol/contracts/pull/1070/files#diff-1eb2fa597ba6985a52da1944ead06204769d9688901083065a5e6308c3d7278aR80) I added a flag to handle implementations only deployments

nicholaspai (approved):

PR #1071, opened 2025-08-16
feat(Foundry): Added hubpool, ethereum and arbitrum spoke pool scripts

fusmanii:


broadcast/deployed-addresses.json
```diff
@@ -500,9 +500,9 @@
       "chain_name": "Soneium",
       "contracts": {
         "SpokePool": {
-          "address": "0x3baD7AD0728f9917d1Bf08af5782dCbD516cDd96",
-          "transaction_hash": "Unknown",
-          "block_number": 1709997
+          "address": "0x48e687205d3962c43891b8cde5a4fe75fa6c8d7a",
```
nicholaspai: Why are the files in here diff?

nicholaspai (approved):

broadcast/deployed-addresses.json
fusmanii: nice catch, the script was updating the proxy address

PR #1072, opened 2025-08-18
feat: a custom `AdminRelayTokensAdapter` that allows admin to call `relayTokens` via an admin action

grasphoper:
Closes ACX-4313


contracts/chain-adapters/AdminRelayTokensAdapter.sol
```diff
@@ -0,0 +1,47 @@
...
+        (address l1Token, address l2Token, uint256 amount, address spokePool) = abi.decode(
+            message,
+            (address, address, uint256, address)
+        );
+        require(target == spokePool, "target / spokepool mismatch");
+
+        // We are ok with this low-level call since the adapter address is set by the admin and we've
+        // already checked that its not the zero address.
+        // solhint-disable-next-line avoid-low-level-calls
+        (bool success, ) = targetAdapter.delegatecall(
```
nicholaspai: This method is `payable`, should we revert if there is any msg.value?

grasphoper: I don't think you can attach any `value` with a `delegatecall`, and that's the only way the HubPool should be able to call this contract, so I don't think there's any risk in not checking. What do you think?

grasphoper: Honestly, I'll just add it

contracts/chain-adapters/AdminRelayTokensAdapter.sol
```diff
@@ -0,0 +1,51 @@
...
+            (address, address, uint256, address)
+        );
+        require(target == spokePool, "target / spokepool mismatch");
+
+        // We are ok with this low-level call since the adapter address is set by the admin and we've
+        // already checked that its not the zero address.
+        // solhint-disable-next-line avoid-low-level-calls
+        (bool success, ) = UNDERLYING_ADAPTER.delegatecall(
+            abi.encodeWithSignature(
+                "relayTokens(address,address,uint256,address)",
```
fusmanii:
```suggestion
            abi.encodeWithSelector(
                AdapterInterface.relayTokens.selector,
```

contracts/chain-adapters/AdminRelayTokensAdapter.sol
```diff
@@ -0,0 +1,51 @@
...
+                l1Token, // l1Token.
+                l2Token, // l2Token.
+                uint256(amount), // amount.
```
fusmanii:
```suggestion
                amount, // amount.
```
also minor but I think amount is already `uint256`

contracts/chain-adapters/AdminRelayTokensAdapter.sol
grasphoper: [77dab61](https://github.com/across-protocol/contracts/pull/1072/commits/77dab6118f6291ff8155dff9f5ce8a434a03d72e)

contracts/chain-adapters/AdminRelayTokensAdapter.sol
grasphoper: Agree, [1b15167](https://github.com/across-protocol/contracts/pull/1072/commits/1b1516729f7c8a37ad43039af189ac25b6c305e6)

contracts/chain-adapters/AdminRelayTokensAdapter.sol
grasphoper: [1b15167](https://github.com/across-protocol/contracts/pull/1072/commits/1b1516729f7c8a37ad43039af189ac25b6c305e6)

contracts/chain-adapters/AdminRelayTokensAdapter.sol
```diff
@@ -0,0 +1,51 @@
...
+    constructor(address underlyingAdapter) {
+        UNDERLYING_ADAPTER = underlyingAdapter;
+    }
+
+    /**
+     * @param target Receiver of tokens on the destination chain. SpokePool address passed in by the HubPool
+     * @param message Abi-encoded arguments params for the `relayTokens` function call: (address l1Token,
+     * address l2Token, uint256 amount, address spokePool)
+     */
+    function relayMessage(address target, bytes calldata message) external payable {
```
fusmanii, line 26: does this function need some kind of guard to check if caller is admin?

grasphoper: @fusmanii 

It doesn't because this function is being `delegatecall`ed into, basically acting as a library that HubPool will use.

Without HubPools state (e.g. token balances) this function call is not meaningful.

Basically the following flow will happen:

1. admin calls HubPool.relaySpokePoolAdminFunction
2. (now we're executing in the context of HubPool) HubPool delegatecalls `AdminRelayTokensAdapter.relayMessage`
3. (still in the context of HubPool because of delegatecall) HubPool* delegatecalls `UNDERLYING_ADAPTER.relayTokens` -- this step is exactly the same as what HubPool does when sending tokens to Spokes normally during a Rebalance leaf execution (if `UNDERLYING_ADAPTER` is, say, `Arbitrum_Adapter`)

fusmanii (dismissed):

contracts/chain-adapters/AdminRelayTokensAdapter.sol
```diff
@@ -0,0 +1,51 @@
...
+        // This adapter is not intended to accept any value as a part of `relayMessage`. It's intended usecase is to be
+        // `deletagecall`ed into, and delegatecall does not allow for attaching the `value`
+        require(msg.value == 0, "value transfer not supported");
+
+        (address l1Token, address l2Token, uint256 amount, address spokePool) = abi.decode(
```
mrice32, line 27: Why encode the spokePool address at all? Why not just use the target?

grasphoper: My thinking is, this adapter is going to be set as an ad-hoc adapter for non-existing chainId, so by default the spokePool for that is just zero address, but we could e.g. do:
- setCrossChainContracts(thisadapter, arbitrum-spoke-addr)

and then HubPool would send in the `arbitrum-spoke-addr` as `target`, just felt easier to pass in as part of encoded arguments

mrice32: If we used it that way, wouldn't the check that requires target == spokePool below fail?

contracts/chain-adapters/AdminRelayTokensAdapter.sol
```diff
@@ -0,0 +1,51 @@
...
+
+    /**
+     * @param target Receiver of tokens on the destination chain. SpokePool address passed in by the HubPool
+     * @param message Abi-encoded arguments params for the `relayTokens` function call: (address l1Token,
+     * address l2Token, uint256 amount, address spokePool)
+     */
+    function relayMessage(address target, bytes calldata message) external payable {
+        // This adapter is not intended to accept any value as a part of `relayMessage`. It's intended usecase is to be
+        // `deletagecall`ed into, and delegatecall does not allow for attaching the `value`
+        require(msg.value == 0, "value transfer not supported");
```
bmzig: I'm a nit and I don't like require statements. To me, if we are going to make this contract (and hopefully not modify it ever), we might as well use the ever-so-slightly more gas-optimal custom errors. 

On another note, we probably don't need this check either way. None of our other adapters check for empty msg.value since (as far as I know) nothing happens when you attach msg.value to these txns (except the eth just goes to the hub pool). Also, letting this be payable might be useful if we do want to check that this adapter works for e.g. the arbitrum or zksync adapters since we can call this adapter via an EOA and attach value to the txn.

grasphoper: Will add custom errors.

I agree on the `msg.value`, mentioned it [here](https://github.com/across-protocol/contracts/pull/1072#discussion_r2283654228) too

@nicholaspai I think we should be able to skip this msg.value == 0 check as Bennett mentions above

contracts/chain-adapters/AdminRelayTokensAdapter.sol
grasphoper: Discussed offline that having both `target` and encoded `spokePool` agree is a good safety net

PR #1074, opened 2025-08-20
chore: prepare Polygon contracts for USDT0 rollout

grasphoper:


nicholaspai (commented):
We should point any unaudited contract changes to a feature branch so we can audit that branch

grasphoper:
> We should point any unaudited contract changes to a feature branch so we can audit that branch

Do we want to combine this change with some other changes? E.g. can't `if/prepare-polygon-for-usdt0` be the audit branch for this?

grasphoper:
@nicholaspai changed base to `audit-aug20`

nicholaspai (approved):

PR #1081, opened 2025-08-25
feat: deploy scroll, blast, mode, redstone, zora, ink, and soneium spoke pools

bmzig:


nicholaspai (approved):

script/utils/ExtractDeployedFoundryAddresses.ts
```diff
@@ -373,7 +373,7 @@ function generateAddressesFile(broadcastFiles: BroadcastFile[], outputFile: stri
       }
     }
 
-    content.push("");
+    // content.push("");
```
nicholaspai: Left over comment

grasphoper:
Do we want to put these new addresses into `deployments.json` for backward compatibility for now? AFAIK, bits of relayer code rely on `deployments.json` being a source of truth.

Otherwise when bumping the contracts version in the `relayer`, we need to handle these changes in `ContractUtils.ts`

fusmanii:
@grasphoper I was hoping to make the new `broadcast/deployed-addresses.json` as source of truth since it has all the addresses from `deployments.json` and all new deploys from `broadcast` 

also its automatically kept up to date by with `extract-addresses` script

script/utils/ExtractDeployedFoundryAddresses.ts
fusmanii: https://github.com/across-protocol/contracts/pull/1081/commits/6c1bfd2ad979abb301ca868b20e49b157c8a31ad

pxrl:
> Do we want to put these new addresses into `deployments.json` for backward compatibility for now? AFAIK, bits of relayer code rely on `deployments.json` being a source of truth.
> 
> Otherwise when bumping the contracts version in the `relayer`, we need to handle these changes in `ContractUtils.ts`

These are only implementation contracts so we don't actually want them in deployments.json - if we updated them there then it'd break the bots.

fusmanii (approved):

broadcast/deployed-addresses.json
```diff
@@ -358,6 +358,11 @@
           "address": "0x8A8cA9c4112c67b7Dae7dF7E89EA45D592362107",
           "transaction_hash": "Unknown",
           "block_number": 2884429
+        },
+        "Lens_SpokePool": {
+          "address": "0xc7772ce23a3ed7f87fe51b87617c7c7d21f15d39",
+          "transaction_hash": "0xa4938f34425c997b4686b5bad4f66b04bfbc2270dfe9c00b8f134da6a7b6edf5",
+          "block_number": 4092265
         }
```
pxrl: Do we want to include this deployment?

fusmanii: done https://github.com/across-protocol/contracts/pull/1081/commits/9fe5920d39c54ffd41b42cf1192b668d185e7f72

foundry.lock
```diff
@@ -0,0 +1,5 @@
+{
+  "lib/forge-std": {
+    "rev": "6bce1540c7a5d1c40eec032a1ae16f0e01f82b92"
+  }
+}
```
grasphoper, line 5: @fusmanii could you explain why we need this?

scripts/veryfyBytecode.sh
```diff
@@ -12,27 +20,18 @@ RPC=${args[1]}
...
 
-if [[ $CODE_ONCHAIN == $CREATION ]]; then 
-    echo "✅ Code match"; 
-else 
-    echo "❌ Code mismatch"; 
+if [[ $CODE_ONCHAIN == $CREATION ]]; then
+    echo "✅ Code match"
+else
+    echo "❌ Code mismatch"
 fi
```
grasphoper, line 37: @fusmanii For the future, we should probably standardize the way we do this "veryfying of bytecode" action.

We've got this script, and 2 more: `tasks/verifyBytecode.ts` , `tasks/verifySpokePool.ts`

The latter script seems unused currently (am I wrong here @pxrl?) I like the idea of it: sanity checking some params for the spoke. It'd be nice to keep that up-to-date and potentially also create a foundry version.

It'd be nice if we consolidated all this functionality into some single foundry script that could work for any deployment: either foundry or hardhat.

These are a little vague thoughts and only tangentially relevant

foundry.lock
fusmanii: this is just pinning `lib/forge-std` version, not sure why it hasn't been added to the repo already

scripts/veryfyBytecode.sh
fusmanii: I added this as a quick way to check bytecode of a deployed contract using foundry but I agree, ideally there would be one script that can verify bytecode/params of a spoke pool regardless if it was deployed using foundry or hardhat

pxrl: > The latter script seems unused currently (am I wrong here @pxrl?) I like the idea of it: sanity checking some params for the spoke. It'd be nice to keep that up-to-date and potentially also create a foundry version.

Yeah; it's only sporadically used, but it's a fairly inexpensive way of checking for any configuration issues in the SpokePool.  There's value in checking that the onchain values match what was (intended to be) configured by the deployment script, but there's secondary value in dumping the values so that people can sanity check the results for anything that might be misconfigured in the deployment script. The visual inspection might catch issues that we are prone to overlook if the error is buried in a config file.

Ideally it could use executed post-deployment and the output could form part of the eventual PR to add the SpokePool artefacts.

PR #1082, opened 2025-08-25
improve: enforce OFT fee payment by caller

grasphoper:


contracts/Polygon_SpokePool.sol
```diff
@@ -1,9 +1,10 @@
 // SPDX-License-Identifier: BUSL-1.1
-pragma solidity ^0.8.0;
+pragma solidity ^0.8.18;
```
nicholaspai: Why?

nicholaspai (approved):

contracts/Polygon_SpokePool.sol
grasphoper: Reverted [be34a77](https://github.com/across-protocol/contracts/pull/1082/commits/be34a7740d5492460972b0ce552c982ad884afba)

PR #1087, opened 2025-08-28
chore: Consolidated OP chain spoke pools

fusmanii:
Removed: 
- Base Spoke Pool
- Mode Spoke Pool
- Redstone Spoke Pool
- Zora Spoke Pool
- Bob Spoke Pool
- DoctorWho Spoke Pool

Created `OP_SpokePool.sol` that can be used for all of them since they all share the same code


hardhat.config.ts
```diff
@@ -97,6 +97,7 @@ const config: HardhatUserConfig = {
       "contracts/Zora_SpokePool.sol": LARGE_CONTRACT_COMPILER_SETTINGS,
       "contracts/Mode_SpokePool.sol": LARGE_CONTRACT_COMPILER_SETTINGS,
       "contracts/Base_SpokePool.sol": LARGE_CONTRACT_COMPILER_SETTINGS,
```
pxrl: Since these are gone we should be able to clean up a bunch here.

contracts/OP_SpokePool.sol
pxrl, line 1: This one probably isn't the best candidate here, since it supports neither bridged nor native USDC. I think the Base SpokePool probably would be a better choice.

All of this is subject to whether there's a larger overhaul on the SpokePools wrt. bridge adapter modularity and configuration. But I think it makes sense to have a PR that handles the consolidation anyway, since it's basically inevitable that we delete these.

bmzig: Aren't all the spoke pools being deleted here the same up to comments?

script/010DeployOPSpokePool.s.sol
bmzig, line 1: On a similar vein as https://github.com/across-protocol/contracts/pull/1087/files#r2355754767, I think you can refactor this entire scripts directory and just make a 001_DeployOpSpokePool.s.sol or something (and delete the other duplicate scripts). That would obviously require some knowledge of the chain ID, but I think you can get that, right?

contracts/OP_SpokePool.sol
fusmanii: yes its just comment diff on all these spoke pool, in this case I think github randomly picked Bob to show as name change

script/010DeployOPSpokePool.s.sol
fusmanii: yup wanted to get an initial feedback on the direction I am taking before I put more effort into it. Going to work on the deploy script next

nicholaspai (dismissed):

deploy/consts.ts
```diff
@@ -293,11 +301,23 @@ export const L2_ADDRESS_MAP: { [key: number]: { [contractName: string]: string }
...
     permit2: "0x000000000022D473030F116dDEE9F6B43aC78BA3",
   },
+  [CHAIN_IDs.MODE_SEPOLIA]: {
+    cctpTokenMessenger: "0x0000000000000000000000000000000000000000",
+    cctpMessageTransmitter: "0x0000000000000000000000000000000000000000",
+  },
+  [CHAIN_IDs.ZORA]: {
+    cctpTokenMessenger: "0x0000000000000000000000000000000000000000",
+    cctpMessageTransmitter: "0x0000000000000000000000000000000000000000",
+  },
```
grasphoper: Is our deploy script supposed to handle these cases instead? Storing zero addrs in a mapping like this seems a bit odd.

We might want to instead check for existence of a param from the deploy script

scripts/checkStorageLayout.sh
```diff
@@ -1,14 +1,14 @@
 #!/bin/bash
 
-CONTRACTS=("Universal_SpokePool" "Arbitrum_SpokePool" "Optimism_SpokePool" "Polygon_SpokePool" "Linea_SpokePool" "ZkSync_SpokePool" "Ethereum_SpokePool" "Base_SpokePool" "Mode_SpokePool" "Blast_SpokePool" "AlephZero_SpokePool" "Redstone_SpokePool" "Scroll_SpokePool" "WorldChain_SpokePool" "Zora_SpokePool" "PolygonZkEVM_SpokePool")
+CONTRACTS=("Universal_SpokePool" "Arbitrum_SpokePool" "Optimism_SpokePool" "Polygon_SpokePool" "Linea_SpokePool" "ZkSync_SpokePool" "Ethereum_SpokePool" "Blast_SpokePool" "AlephZero_SpokePool" "Scroll_SpokePool" "WorldChain_SpokePool" "PolygonZkEVM_SpokePool")
```
grasphoper: We should add `OP_SpokePool` here, right? Also, remove the existing files e.g. `Zora_SpokePool.json` from `storage-layouts` folder

deploy/consts.ts
fusmanii: this is actually limitation of foundry json parsing, if a value is not present then it complains 
I couldn't really find a work around so for now we have to add these here

scripts/checkStorageLayout.sh
fusmanii: yes good point

deploy/consts.ts
grasphoper: Hm, sadge. So this is specifically to support using this constants file in the solidity deployment scripts? Could you please point to where we are using this currently?

fusmanii: It's currently being used [here](https://github.com/across-protocol/contracts/blob/master/script/utils/Constants.sol#L129) 

and get read like [this](https://github.com/across-protocol/contracts/blob/master/script/059DeployLensSpokePool.s.sol#L34)

grasphoper: Thanks. Btw what happens on a non-existing value? The .sol script just crashes?

scripts/checkStorageLayout.sh
fusmanii: https://github.com/across-protocol/contracts/pull/1087/commits/51d739166d615b79ede52c43ef08abf0912793c5

deploy/consts.ts
fusmanii: yup you get this error

```
Error: script failed: vm.parseJsonAddress: path ".L2_ADDRESS_MAP.34443.cctpV2TokenMessenger" must return exactly one JSON value
```

generated/constants.json
```diff
@@ -683,8 +691,11 @@
       "scrollMessenger": "0x781e90f1c8Fc4611c9b7497C3B47F99Ef6969CbC",
       "permit2": "0x000000000022D473030F116dDEE9F6B43aC78BA3"
     },
+    "808813": {
+      "cctpV2TokenMessenger": "0x0000000000000000000000000000000000000000"
+    },
     "7777777": {
-      "permit2": "0x000000000022D473030F116dDEE9F6B43aC78BA3"
+      "cctpV2TokenMessenger": "0x0000000000000000000000000000000000000000"
```
grasphoper: This override happened because in `deploy/consts.ts` we included a duplicate entry for ZORA

grasphoper: Should fix

lib/forge-std
grasphoper, line 1: Just bumping the `forge-std` here right?

generated/constants.json
fusmanii: nice catch, fixed http://github.com/across-protocol/contracts/pull/1087/commits/026624fef78e3dfd157360ca2ca3b8017bfbb911

lib/forge-std
fusmanii: actually I am undoing changes from before
I accidentally downgraded

grasphoper: Hmm, actually, in `master` we're currently at `8bbcf6e3f8f62f419e5429a0bd89331c85c37824`, which is the latest forge-std release, shouldn't we stay on that one? https://github.com/foundry-rs/forge-std/releases/tag/v1.10.0

PR #1091, opened 2025-08-30
chore: deploy new BSC contracts

grasphoper:
This deployment is required to update us to using a new `SP1Helios` deployment from [here](https://github.com/across-protocol/sp1-helios/pull/37), but also bumps BSC deployment with OFT functionality.

This PR also deploys `Universal_Adapter` for BSC in order to keep parity of functionality between Adapter and Spoke for BSC (i.e. both support OFT now)

### DIFF
--- a/deploy/111_deploy_universal_spokepool.ts
+++ b/deploy/111_deploy_universal_spokepool.ts
@@ -12,7 +12,6 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
     // Set these in hardhat.config.ts networks.bsc
     // gas: "auto",
     // gasPrice: 3e8, // 0.3 GWEI
-    // allowUnlimitedContractSize: true,
     // gasMultiplier: 4.0,
   }
 
--- a/hardhat.config.ts
+++ b/hardhat.config.ts
@@ -201,7 +201,6 @@ const config: HardhatUserConfig = {
       // ! Notice. Params below helped deploy Universal_Spoke on BSC, but might not be desirable always
       // gas: "auto",
       // gasPrice: 3e8, // 0.3 GWEI
-      // allowUnlimitedContractSize: true,
       // gasMultiplier: 4.0,
       chainId: CHAIN_IDs.BSC,
       url: getNodeUrl(CHAIN_IDs.BSC),


hardhat.config.ts
```diff
@@ -199,6 +198,11 @@ const config: HardhatUserConfig = {
       companionNetworks: { l1: "mainnet" },
     },
     bsc: {
+      // ! Notice. Params below helped deploy Universal_Spoke on BSC, but might not be desirable always
+      // gas: "auto",
+      // gasPrice: 3e8, // 0.3 GWEI
+      // allowUnlimitedContractSize: true,
```
bmzig: Do you know why `allowUnlimitedContractSize` was needed here? That seems like it specifically _should_ cause the deployment to fail when actually sending the create transaction.

deploy/111_deploy_universal_spokepool.ts
```diff
@@ -18,6 +27,10 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
     throw new Error(`Wrapped native token not found for ${wrappedNativeSymbol} on chainId ${spokeChainId}`);
   }
 
+  const oftEid = getOftEid(hubChainId);
+  // ! Notice. Deployed has to adjust this fee cap based on dst chain's native token. 4.4 BNB for BSC
+  const oftFeeCap = toWei(4.4); // ~1 ETH fee cap
```
nicholaspai, line 31: What would happen if the BNB:ETH exchange rate moved sharply? Would we need to redeploy or does our fee cap give us a wide room for error?

hardhat.config.ts
nicholaspai: Yea it might be a local error...

grasphoper: > Do you know why allowUnlimitedContractSize was needed here? 

Hmm, perhaps it wasn't needed actually. I was trying to throw everything at this config to make it not revert with the "deployment out of gas" error

deploy/111_deploy_universal_spokepool.ts
grasphoper: This is extremely wide. A normal quote is never > $10. We give a HUGE buffer here

The fee cap is meant to prevent a single incorreect quote from draining our (potentially big?) native token balance, we cap it at like $3-4K. IMO 4.4 BNB is fine here

hardhat.config.ts
grasphoper: I'm pretty sure it wasn't needed. Removed here [ca5816e](https://github.com/across-protocol/contracts/pull/1091/commits/ca5816eac0c852576d18b2f0997e990c2e8e4273)

bmzig (approved):

deployments/deployments.json
```diff
@@ -22,7 +22,7 @@
     "Mode_Adapter": { "address": "0xf1B59868697f3925b72889ede818B9E7ba0316d0", "blockNumber": 19914094 },
     "MulticallHandler": { "address": "0x924a9f036260DdD5808007E1AA95f08eD08aA569", "blockNumber": 20277013 },
     "Lisk_Adapter": { "address": "0xF039AdCC74936F90fE175e8b3FE0FdC8b8E0c73b", "blockNumber": 22474211 },
-    "Universal_Adapter": { "address": "0x22001f37B586792F25Ef9d19d99537C6446e0833", "blockNumber": 22368546 },
+    "Universal_Adapter": { "address": "0x6f1C9d3bcDF51316E7b515a62C02F601500b084b", "blockNumber": 23251254 },
```
pxrl, line 25: @grasphoper Driveby query: Any thoughts on how we manage this moving forward, w > 1 Universal Adapter chain?

This looks like another instance of the issue to what we have where the same contract is used for multiple deployments - we end up having to manually rename them (i.e. OP stack deployments of the OP Adapter are not named as OP Adapter). It's then not easy to resolve the contract code/ABI via the name alone, because they don't match anymore.

The default pattern has typically been to copy the contract, but that's unsustainable and bad practice. You can see this mostly in the SpokePool contracts.

pxrl (approved):

PR #1095, opened 2025-09-02
feat: Added upgradeToAndCall to Spoke Pool Upgrade task

fusmanii:
We want to ensure that the new implementation is working as expected by toggling `pauseDeposits` function post upgrade


tasks/upgradeSpokePool.ts
```diff
@@ -17,13 +18,75 @@ task("upgrade-spokepool", "Generate calldata to upgrade a SpokePool deployment")
...
-    const upgradeTo = spokePool.interface.encodeFunctionData("upgradeTo", [implementation]);
-    console.log(`upgradeTo bytes: `, upgradeTo);
+    let calldata = "";
+    if (upgradeOnly) {
+      calldata = spokePool.interface.encodeFunctionData("upgradeTo", [implementation]);
+      console.log(`upgradeTo bytes: `, calldata);
+    } else {
+      /**
+       * We perform this seemingly unnecessary pause/unpause sequence because we want to ensure that the
+       * upgrade is successful and the new implementation is functioning correctly.
```
nicholaspai:
```suggestion
       * upgrade is successful and the new implementation gets forwarded calls by the proxy contract as expected.
```

nicholaspai (approved):
+1

tasks/upgradeSpokePool.ts
```diff
@@ -17,13 +18,75 @@ task("upgrade-spokepool", "Generate calldata to upgrade a SpokePool deployment")
     }
 
     // @dev Any spoke pool's interface can be used here since they all should have the same upgradeTo function signature.
-    const abi = ["function upgradeTo(address newImplementation) external"];
+    const abi = [
```
pxrl: ooc, can we draw on the ABI via typechain?

I'm generally pretty wary of using typechain, but given that the ABI is local to this repo it feels like we should be able to source it from somewhere dynamically, rather than have to redefine it.

tasks/upgradeSpokePool.ts
```diff
@@ -3,8 +3,9 @@ import { HardhatRuntimeEnvironment } from "hardhat/types";
 
 task("upgrade-spokepool", "Generate calldata to upgrade a SpokePool deployment")
   .addParam("implementation", "New SpokePool implementation address")
+  .addOptionalParam("upgradeOnly", "Upgrade only, do not pause/unpause deposits")
```
pxrl: Do you think there's a scenario where we wouldn't want the atomic pause/unpause?

tasks/upgradeSpokePool.ts
```diff
@@ -17,13 +18,75 @@ task("upgrade-spokepool", "Generate calldata to upgrade a SpokePool deployment")
...
+       *
+       * Since the upgrade and call happens atomically, the upgrade will revert if the new implementation
+       * is not functioning correctly.
+       */
+      const data = spokePool.interface.encodeFunctionData("multicall", [
+        [
+          spokePool.interface.encodeFunctionData("pauseDeposits", [true]),
+          spokePool.interface.encodeFunctionData("pauseDeposits", [false]),
+        ],
+      ]);
```
pxrl: Nice; I like this!

tasks/upgradeSpokePool.ts
nicholaspai: On EVM chains we are absolutely sure there is no difference from the L1 EVM

fusmanii: I just kept it in case someone is looking for the old upgrade only behaviour. @pxrl do you think its better (and safer) to only have atomic pause/unpause option?

tasks/upgradeSpokePool.ts
fusmanii: @pxrl good point, I was able to get the abi directly from the artifact instead

tasks/upgradeSpokePool.ts
pxrl: Yeah; I feel like we should just keep the interface simple and omit this option, since we don't actually have any concrete scenarios where we'd want to skip it. That feels like it'd be an exceptional circumstance.

pxrl: @fusmanii This is my only remaining comment btw - OK to drop the arg and then merge it?

fusmanii: done https://github.com/across-protocol/contracts/pull/1095/commits/1c11d8bb3938fdf449ede76450c02dd19eb41fcc

fusmanii: @pxrl dropped the param

tasks/upgradeSpokePool.ts
```diff
@@ -4,22 +4,40 @@ import { HardhatRuntimeEnvironment } from "hardhat/types";
 task("upgrade-spokepool", "Generate calldata to upgrade a SpokePool deployment")
   .addParam("implementation", "New SpokePool implementation address")
   .setAction(async function (args, hre: HardhatRuntimeEnvironment) {
-    const { implementation } = args;
+    const { implementation, upgradeOnly } = args;
```
nicholaspai: You can remove this now right?

nicholaspai (approved):

PR #1096, opened 2025-09-03
feat: Generated the constants file

fusmanii:
All constants are taken from constants repo so that there is only one source of truth


script/utils/GenerateConstantsJson.ts
```diff
@@ -0,0 +1,255 @@
...
+import {
+  ZERO_ADDRESS,
+  USDC,
+  USDCe,
+  WETH,
+  WGHO,
+  QUOTE_TIME_BUFFER,
+  FILL_DEADLINE_BUFFER,
+  ARBITRUM_MAX_SUBMISSION_COST,
+  AZERO_GAS_PRICE,
```
pxrl: We can probably drop references to Aleph Zero because the Aleph Zero deployment is EoL now.

script/utils/GenerateConstantsJson.ts
```diff
@@ -0,0 +1,255 @@
...
+  for (const [key, value] of Object.entries(chainIds)) {
+    if (typeof value === "number") {
+      result[key] = value;
+    }
+  }
+  return result;
+}
+
+// Helper function to extract wrapped native tokens from TOKEN_SYMBOLS_MAP
+function extractWrappedNativeTokens(): { [key: string]: string } {
```
pxrl: In the constants repo we define the native token symbol for each chain - feels like we could use that here to do something a bit more dynamic.

script/utils/GenerateConstantsJson.ts
```diff
@@ -0,0 +1,255 @@
...
+  }
+
+  return result;
+}
+
+// Helper function to extract USDCe addresses
+function extractUsdceAddresses(): { [key: string]: string } {
+  const result: { [key: string]: string } = {};
+
+  for (const [chainId, address] of Object.entries(USDCe)) {
```
pxrl: I don't think this is super reliable because the USDC symbol itself can be quite random. Zora has USDzC for example.

script/utils/GenerateConstantsJson.ts
```diff
@@ -0,0 +1,255 @@
...
+    if (chainName) {
+      result[chainName] = address;
+    }
+  }
+
+  return result;
+}
+
+// Helper function to extract WGHO addresses
+function extractWghoAddresses(): { [key: string]: string } {
```
pxrl: Integrating nativeToken from the constants repo would probably also help to remove a bespoke function like this one.

script/utils/GenerateConstantsJson.ts
```diff
@@ -0,0 +1,255 @@
...
+      result[hubChainName] = {};
+      for (const [spokeChainId, addresses] of Object.entries(spokeChains)) {
+        const spokeChainName = Object.keys(CHAIN_IDs).find(
+          (key) => CHAIN_IDs[key as keyof typeof CHAIN_IDs] === Number(spokeChainId)
+        );
+        if (spokeChainName) {
+          result[hubChainName][spokeChainName] = addresses;
+        }
+      }
+    }
```
pxrl: Thoughts on dropping the indentation by returning early if `hubChainName` is not defined?

script/utils/GenerateConstantsJson.ts
```diff
@@ -0,0 +1,255 @@
...
+  const result: { [key: string]: number } = {};
+
+  for (const [chainId, domainId] of Object.entries(CIRCLE_DOMAIN_IDs)) {
+    const chainName = Object.keys(CHAIN_IDs).find(
+      (key) => CHAIN_IDs[key as keyof typeof CHAIN_IDs] === Number(chainId)
+    );
+    if (chainName) {
+      result[chainName] = domainId;
+    }
+  }
```
pxrl: This pattern is a bit repetitive now - is there maybe some way to refactor such that the core functionality is generic? Or are the upstream objects themselves too bespoke for that?

script/utils/GenerateConstantsJson.ts
```diff
@@ -0,0 +1,255 @@
...
+    chainIds: convertChainIdsToObject(CHAIN_IDs),
+    oftEids: convertOftEids(),
+    wrappedNativeTokens: extractWrappedNativeTokens(),
+    l2Addresses: convertL2Addresses(),
+    l1Addresses: convertL1Addresses(),
+    opStackAddresses: convertOpStackAddresses(),
+    circleDomainIds: convertCircleDomainIds(),
+    timeConstants: {
+      QUOTE_TIME_BUFFER: QUOTE_TIME_BUFFER,
+      FILL_DEADLINE_BUFFER: FILL_DEADLINE_BUFFER,
```
pxrl:
```suggestion
      QUOTE_TIME_BUFFER,
      FILL_DEADLINE_BUFFER,
```

script/utils/constants.json
```diff
@@ -1,181 +1,245 @@
...
-    "BLAST": 30243,
-    "BSC": 30102,
     "MAINNET": 30101,
     "OPTIMISM": 30111,
-    "OPTIMISM_SEPOLIA": 40232,
+    "BSC": 30102,
+    "UNICHAIN": 30320,
     "POLYGON": 30109,
+    "LENS": -1,
+    "BOBA": -1,
```
pxrl: ooc, rather than enumerating each chain w/ -1 for no OFT EID, could we just omit chains without an OFT EID?

script/utils/constants.json
```diff
@@ -1,181 +1,245 @@
...
-    "LINEA_SEPOLIA": "0x10253594A832f967994b44f33411940533302ACb",
-    "SCROLL_SEPOLIA": "0x5300000000000000000000000000000000000004",
-    "SCROLL": "0x5300000000000000000000000000000000000004",
+    "BSC": "0x2170Ed0880ac9A755fd29B2688956BD959F933F8",
     "UNICHAIN": "0x4200000000000000000000000000000000000006",
-    "UNICHAIN_SEPOLIA": "0x4200000000000000000000000000000000000006",
-    "ALEPH_ZERO": "0xB3f0eE446723f4258862D949B4c9688e7e7d35d3",
-    "BLAST": "0x4300000000000000000000000000000000000004",
-    "BLAST_SEPOLIA": "0x4200000000000000000000000000000000000023",
+    "POLYGON": "0x7ceB23fD6bC0adD59E62ac25578270cFf1b9f619",
```
pxrl: We have different addresses for Polygon here - the new address is incorrect because it's WETH.

script/utils/constants.json
```diff
@@ -190,7 +254,7 @@
       "polygonDepositManager": "0x401F6c983eA34274ec46f84D70b31C151321188b",
       "cctpTokenMessenger": "0xBd3fa81B58Ba92a82136038B25aDec7066af3155",
       "cctpV2TokenMessenger": "0x28b5a0e9C621a5BadaA536219b3a228C8168cf5d",
-      "cctpMessageTransmitter": "0x0a992d191DEeC32aFe36203Ad87D7d289a738F81",
+      "cctpMessageTransmitter": "0x0a992d191deec32afe36203ad87d7d289a738f81",
```
pxrl: It could be a good idea to split simple changes like this out to a separate PR - that'll help reduce the diff here and make it easier to review the subsequent address changes.

script/utils/GenerateConstantsJson.ts
```diff
@@ -0,0 +1,255 @@
...
+    if (chainName) {
+      result[chainName] = addresses;
+    }
+  }
+
+  return result;
+}
+
+// Helper function to extract USDC addresses
+function extractUsdcAddresses(): { [key: string]: string } {
```
bmzig: Two questions for this entire file:
1. Why are we setting the keys as the chain name? I would think it's a lot easier for foundry to know the network's chainId over the name of the network itself?
2. Maybe a stupid question, but these functions seem really complex for what they are doing. Why can't we just call `JSON.stringify(USDC)` and similar things for all of these other functions? I think adding this much code to just extract values from a map makes the script itself pretty rigid/hard to maintain.

script/utils/GenerateConstantsJson.ts
fusmanii: done https://github.com/across-protocol/contracts/pull/1096/commits/791d7e72f32f0428f0bd73ed3b0a7e3935cc58e4

script/utils/GenerateConstantsJson.ts
fusmanii: I was able to remove this function all together: https://github.com/across-protocol/contracts/pull/1096/commits/791d7e72f32f0428f0bd73ed3b0a7e3935cc58e4

script/utils/GenerateConstantsJson.ts
fusmanii: done: https://github.com/across-protocol/contracts/pull/1096/commits/791d7e72f32f0428f0bd73ed3b0a7e3935cc58e4

script/utils/constants.json
fusmanii: good point, done: https://github.com/across-protocol/contracts/pull/1096/commits/791d7e72f32f0428f0bd73ed3b0a7e3935cc58e4

script/utils/GenerateConstantsJson.ts
fusmanii: Good idea, I updated the scripts to fetch by chainid instead https://github.com/across-protocol/contracts/pull/1096/commits/46770666fc32c6563194bf7f4288f764ec67f531

make every a lot cleaner

script/utils/GenerateConstantsJson.ts
```diff
@@ -0,0 +1,111 @@
...
+    result[chainId] = eid;
+  }
+
+  return result;
+}
+
+// Generate the constants.json structure
+function generateConstantsJson() {
+  const constants = {
+    PUBLIC_NETWORKS,
```
fusmanii: @pxrl I imported the entire `PUBLIC_NETWORKS` which will allow me to get CCTP domain and OFT ids as well

PR #1098, opened 2025-09-04
chore: Update sha.js version

fusmanii:
Vulnerability in sha.js <= 2.4.11 https://github.com/advisories/GHSA-95m3-7q98-8xr5

### DIFF
--- a/package.json
+++ b/package.json
@@ -146,6 +146,8 @@
   },
   "resolutions": {
     "**/eccrypto/secp256k1": "3.8.1",
-    "**/eth-crypto/secp256k1": "5.0.1"
+    "**/eth-crypto/secp256k1": "5.0.1",
+    "**/create-hash/sha.js": "2.4.12",
+    "sha.js": "2.4.12"
   }
 }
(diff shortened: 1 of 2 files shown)


nicholaspai (approved):

package.json
```diff
@@ -146,6 +146,8 @@
   },
   "resolutions": {
     "**/eccrypto/secp256k1": "3.8.1",
-    "**/eth-crypto/secp256k1": "5.0.1"
+    "**/eth-crypto/secp256k1": "5.0.1",
+    "**/create-hash/sha.js": "2.4.12",
+    "sha.js": "2.4.12"
```
pxrl, line 151: Do we know where this outdated version comes from? If it's in the uma repo we should probably go upstream and fix it there so we don't need to override the resolution.

fusmanii: It looks like its coming from `ethereum-cryptography` which `ethereumjs-util` depends on

fusmanii: we are at latest version of `ethereumjs-util` already, so I think we need the resolution until they patch on their end

PR #1102, opened 2025-09-10
feat(Foundry): Generate Constants script

fusmanii:
This contains just the script that will generate the `constants.json` file, the rest of the implementation is in https://github.com/across-protocol/contracts/pull/1096 so that its easier to review

### DIFF
--- a/script/utils/GenerateConstantsJson.ts
+++ b/script/utils/GenerateConstantsJson.ts
@@ -1,0 +1,97 @@
+#!/usr/bin/env ts-node
+
+import * as fs from "fs";
+import * as path from "path";
+
+// Import the constants from the TypeScript files
+import { CHAIN_IDs, PUBLIC_NETWORKS } from "../../utils/constants";
+import {
+  ZERO_ADDRESS,
+  USDC,
+  USDCe,
+  WETH,
+  WGHO,
+  QUOTE_TIME_BUFFER,
+  FILL_DEADLINE_BUFFER,
+  ARBITRUM_MAX_SUBMISSION_COST,
+  CIRCLE_UNINITIALIZED_DOMAIN_ID,
+  ZK_L1_GAS_TO_L2_GAS_PER_PUBDATA_LIMIT,
+  ZK_L2_GAS_LIMIT,
+  ZK_MAX_GASPRICE,
+  L1_ADDRESS_MAP,
+  OP_STACK_ADDRESS_MAP,
+  L2_ADDRESS_MAP,
+} from "../../deploy/consts";
+
+// Helper function to convert chain IDs object to the expected format
+function convertChainIdsToObject(chainIds: any): { [key: string]: number } {
+  const result: { [key: string]: number } = {};
+  for (const [key, value] of Object.entries(chainIds)) {
+    if (typeof value === "number") {
+      result[key] = value;
+    }
+  }
+  return result;
+}
+
+// Generate the constants.json structure
+function generateConstantsJson() {
+  const constants = {
+    PUBLIC_NETWORKS,
+    CHAIN_IDs: convertChainIdsToObject(CHAIN_IDs),
+    WETH,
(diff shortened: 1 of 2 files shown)


script/utils/GenerateConstantsJson.ts
```diff
@@ -0,0 +1,140 @@
...
+ */
+function convertChainIdsToObject(): { [key: string]: number } {
+  const result: { [key: string]: number } = {};
+  for (const [key, value] of Object.entries(CHAIN_IDs)) {
+    if (typeof value === "number") {
+      result[key] = value;
+    }
+  }
+  return result;
+}
```
pxrl, line 45: Minor simplification you could consider here:

```suggestion
const convertChainIdsToObject = (): { [key: string]: number } => Object.fromEntries(
  Object.values(CHAIN_IDs).map((chainId) => [chainId, chainId]
);
```

script/utils/GenerateConstantsJson.ts
```diff
@@ -0,0 +1,140 @@
...
+ *   "56": "0xbb4CdB9CBd36B01bD1cBaEBF2De08d9173bc095c",
+ *   "137": "0x0d500B1d8E8eF31E21C99d1Db9A6444d3ADf1270",
+ * }
+ */
+function generateWrappedNativeTokens(): { [key: string]: string } {
+  const result: { [key: string]: string } = {};
+  for (const [key, value] of Object.entries(PUBLIC_NETWORKS)) {
+    const nativeToken = value.nativeToken;
+    const wrappedPrefix = "W";
+    const wrappedNativeSymbol = `${wrappedPrefix}${nativeToken}`;
```
pxrl, line 63: Something to think about (nb. probably not something that we can solve immediately): this strategy of prepending W and assuming it's the wrapped symbol is done in a few places throughout our repos codebase, but it's basically tech debt. Sooner or later we'll hit a case where it breaks for us (...and tbh, it kinda did with the LGHO/WGHO complexity on Lens). It might be interesting to figure out a more sustainable / sensible solution.

script/utils/GenerateConstantsJson.ts
```diff
@@ -0,0 +1,140 @@
...
+
+    // Check if the wrapped token symbol exists in TOKEN_SYMBOLS_MAP
+    if (wrappedNativeSymbol in TOKEN_SYMBOLS_MAP) {
+      const tokenInfo = TOKEN_SYMBOLS_MAP[wrappedNativeSymbol as keyof typeof TOKEN_SYMBOLS_MAP];
+      const wrappedNativeToken = tokenInfo.addresses[Number(key)];
```
pxrl, line 68: Objects are always keyed by strings even if they were generated w/ numbers, so you should be fine to drop this `Number` conversion.

```suggestion
      const wrappedNativeToken = tokenInfo.addresses[key];
```

fusmanii: typescript actually complains here 
```
script/utils/GenerateConstantsJson.ts:61:54 - error TS7015: Element implicitly has an 'any' type because index expression is not of type 'number'.

61       const wrappedNativeToken = tokenInfo.addresses[key];
```
maybe its because I am running it as a script with `ts-node`?

script/utils/GenerateConstantsJson.ts
fusmanii: maybe we can add a `wrappedNativeToken` entry [here](https://github.com/across-protocol/constants/blob/master/src/networks.ts#L105) ?

script/utils/GenerateConstantsJson.ts
fusmanii: merging this pr, addresses this in the other [pr](https://github.com/across-protocol/contracts/pull/1096/commits/61301b70cc4f0a6f92aa6c87f40185fb1e702d6a)

PR #1105, opened 2025-09-12
chore: Generate constants.json file

fusmanii:
This is just running the script from https://github.com/across-protocol/contracts/pull/1102 that generates the constants.json that will be used on deploy scripts

### DIFF
--- a/script/utils/constants.json
+++ b/script/utils/constants.json
@@ -1,185 +1,631 @@
 {
-  "chainIds": {
-    "MAINNET": 1,
-    "SEPOLIA": 11155111,
-    "ARBITRUM": 42161,
-    "ARBITRUM_SEPOLIA": 421614,
-    "BSC": 56,
-    "POLYGON": 137,
-    "POLYGON_AMOY": 80002,
-    "ZK_SYNC": 324,
-    "OPTIMISM": 10,
-    "OPTIMISM_SEPOLIA": 11155420,
-    "BASE": 8453,
-    "BASE_SEPOLIA": 84532,
-    "LENS": 232,
-    "LENS_TESTNET": 37111,
-    "LINEA": 59144,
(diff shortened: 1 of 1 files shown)


script/utils/constants.json
```diff
@@ -1,185 +1,631 @@
...
-    "REDSTONE": "0x4200000000000000000000000000000000000006",
-    "SONEIUM": "0x4200000000000000000000000000000000000006",
-    "WORLD_CHAIN": "0x4200000000000000000000000000000000000006",
-    "ZORA": "0x4200000000000000000000000000000000000006"
+  "WRAPPED_NATIVE_TOKENS": {
+    "1": "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2",
+    "10": "0x4200000000000000000000000000000000000006",
+    "56": "0xbb4CdB9CBd36B01bD1cBaEBF2De08d9173bc095c",
+    "130": "0x4200000000000000000000000000000000000006",
+    "137": "0x0d500B1d8E8eF31E21C99d1Db9A6444d3ADf1270",
```
bmzig, line 468: Lens isn't here. Is that intentional?

fusmanii: yeah because the way these are generated is by appending `W` to the token symbol which doesn't work for Lens, that is why I has this [pr](https://github.com/across-protocol/constants/pull/150)

But should be fine as Lens script uses a different [function](https://github.com/across-protocol/contracts/blob/master/script/059DeployLensSpokePool.s.sol#L31) to fetch the wrapped native token address

fusmanii: from [here](https://github.com/across-protocol/contracts/pull/1105/files#diff-ef1dd9528bc8758097779c73515333e20607331a0f5033cb73c5f1c6e6b50501R798)

PR #1058, opened 2025-07-21
chore: new `MulticallHandler` deployments

dohaki:
New `MulticallHandler` deployments:
- Linea: `0xdF1C940487574EEfa79989a79a4936A0F979cDa2`
- zkSync: `0x68d3806E57148D6c6793C78EbDDbc272fE605dbf`
- Lens: `0x1Ed0D59019a52870337b51DEe8190486a8663037`
- HyperEVM: [`0x5e7840e06faccb6d1c3b5f5e0d1d3d07f2829bba`](https://hyperevmscan.io/address/0x5e7840e06faccb6d1c3b5f5e0d1d3d07f2829bba)
- Others: `0x0F7Ae28dE1C8532170AD4ee566B5801485c13a0E`

Closes ACX-4208

PR #1110, opened 2025-09-16
feat: Added constants & address generation scripts to pre-commit hook

fusmanii:
It also contains changes from https://github.com/across-protocol/contracts/pull/1058, but I can move to another pr if its easier to review like that

### DIFF
--- a/broadcast/deployed-addresses.json
+++ b/broadcast/deployed-addresses.json
@@ -1,5 +1,4 @@
 {
-  "generated_at": "2025-09-15T23:00:31.247Z",
   "chains": {
     "1": {
       "chain_name": "Mainnet",
@@ -110,9 +109,9 @@
           "block_number": 19914094
         },
         "MulticallHandler": {
-          "address": "0x924a9f036260DdD5808007E1AA95f08eD08aA569",
+          "address": "0x0F7Ae28dE1C8532170AD4ee566B5801485c13a0E",
           "transaction_hash": "Unknown",
-          "block_number": 20277013
+          "block_number": 22967417
         },
         "Lisk_Adapter": {
           "address": "0xF039AdCC74936F90fE175e8b3FE0FdC8b8E0c73b",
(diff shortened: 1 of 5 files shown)


script/utils/extract_foundry_addresses.sh
```diff
@@ -14,28 +14,20 @@ echo "Extracting deployed contract addresses using TypeScript..."
...
     exit 1
 fi
-
-echo ""
-echo "Generated files:"
-echo "  - $PROJECT_ROOT/broadcast/deployed-addresses.md (Markdown format)"
-echo "  - $PROJECT_ROOT/broadcast/deployed-addresses.json (JSON format)"
-echo "  - $PROJECT_ROOT/script/DeployedAddresses.sol (Foundry smart contract with all addresses)"
-echo ""
-echo "You can now import DeployedAddresses.sol in your other Foundry scripts to use the deployed addresses."
```
fusmanii, line 41: Removed these looks to make it less noisy

script/utils/ExtractDeployedFoundryAddresses.ts
```diff
@@ -300,25 +295,6 @@ function generateAddressesFile(broadcastFiles: BroadcastFile[], outputFile: stri
...
-    `  Mainnet chains (${mainnetChains.length}): ${mainnetChains.map((id) => `${getChainName(id)}(${id})`).join(", ")}`
-  );
-  console.log(
-    `  Testnet chains (${testnetChains.length}): ${testnetChains.map((id) => `${getChainName(id)}(${id})`).join(", ")}`
-  );
-  console.log(
-    `  Other chains (${otherChains.length}): ${otherChains.map((id) => `${getChainName(id)}(${id})`).join(", ")}`
-  );
-  console.log("");
-
```
fusmanii, line 321: same here, all the changes in this file are removing logs

broadcast/deployed-addresses.json
```diff
@@ -110,9 +109,9 @@
           "block_number": 19914094
         },
         "MulticallHandler": {
-          "address": "0x924a9f036260DdD5808007E1AA95f08eD08aA569",
+          "address": "0x0F7Ae28dE1C8532170AD4ee566B5801485c13a0E",
```
nicholaspai, line 112: Is this right? This contract isn't verified

nicholaspai (commented):
The MUlticallHandler from this PR doesn't seem to be verified on Mainnet at least

broadcast/deployed-addresses.json
fusmanii: nice catch, looks like non of the `0x0F7Ae28dE1C8532170AD4ee566B5801485c13a0E` are verified, will confirm if its the right address

fusmanii: The addresses are correct but looks like they were never verified. I verified all of the unverified contracts

PR #1111, opened 2025-09-17
feat(Foundry): Updated constants usage

fusmanii:
With the new constants JSON file from https://github.com/across-protocol/contracts/pull/1105, need to update its usage in deploy scripts

### DIFF
--- a/generated/constants.json
+++ b/generated/constants.json
@@ -2,7 +2,7 @@
   "PUBLIC_NETWORKS": {
     "1": {
       "name": "Mainnet",
-      "family": 0,
+      "family": "NONE",
       "nativeToken": "ETH",
       "publicRPC": "https://eth.llamarpc.com",
       "blockExplorer": "https://etherscan.io",
@@ -12,7 +12,7 @@
     },
     "10": {
       "name": "Optimism",
-      "family": 1,
+      "family": "OP_STACK",
       "nativeToken": "ETH",
       "publicRPC": "https://mainnet.optimism.io",
       "blockExplorer": "https://optimistic.etherscan.io",
@@ -22,7 +22,7 @@
     },
     "56": {
       "name": "BNB Smart Chain",
-      "family": 0,
+      "family": "NONE",
       "nativeToken": "BNB",
       "publicRPC": "https://bsc-dataseed1.binance.org",
       "blockExplorer": "https://bscscan.com",
@@ -32,7 +32,7 @@
     },
     "130": {
       "name": "Unichain",
-      "family": 1,
+      "family": "OP_STACK",
       "nativeToken": "ETH",
       "publicRPC": "https://mainnet.unichain.org/",
       "blockExplorer": "https://uniscan.xyz",
@@ -42,7 +42,7 @@
     },
     "137": {
       "name": "Polygon",
-      "family": 0,
+      "family": "NONE",
       "nativeToken": "MATIC",
       "publicRPC": "https://polygon-rpc.com",
       "blockExplorer": "https://polygonscan.com",
@@ -52,7 +52,7 @@
     },
     "232": {
       "name": "Lens",
-      "family": 3,
+      "family": "ZK_STACK",
       "nativeToken": "WGHO",
       "publicRPC": "https://api.lens.matterhosted.dev",
       "blockExplorer": "https://explorer.lens.xyz",
@@ -62,7 +62,7 @@
     },
     "288": {
       "name": "Boba",
-      "family": 1,
+      "family": "OP_STACK",
       "nativeToken": "ETH",
       "publicRPC": "https://mainnet.boba.network",
       "blockExplorer": "https://blockexplorer.boba.network",
@@ -72,7 +72,7 @@
     },
     "300": {
       "name": "zkSync Sepolia",
-      "family": 0,
+      "family": "NONE",
       "nativeToken": "ETH",
       "publicRPC": "https://sepolia.era.zksync.dev",
       "blockExplorer": "https://sepolia-era.zksync.network",
@@ -82,7 +82,7 @@
     },
     "324": {
       "name": "zkSync",
-      "family": 0,
+      "family": "NONE",
       "nativeToken": "ETH",
       "publicRPC": "https://mainnet.era.zksync.io",
       "blockExplorer": "https://era.zksync.network",
@@ -92,7 +92,7 @@
     },
     "480": {
       "name": "World Chain",
-      "family": 1,
+      "family": "OP_STACK",
       "nativeToken": "ETH",
       "publicRPC": "https://worldchain-mainnet.g.alchemy.com/public",
       "blockExplorer": "https://worldchain-mainnet-explorer.alchemy.com",
@@ -102,7 +102,7 @@
     },
     "690": {
       "name": "Redstone",
-      "family": 1,
+      "family": "OP_STACK",
       "nativeToken": "ETH",
       "publicRPC": "https://rpc.redstonechain.com",
       "blockExplorer": "https://explorer.redstone.xyz",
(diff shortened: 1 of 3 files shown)


script/002DeployOptimismAdapter.s.sol
```diff
@@ -31,6 +32,8 @@ contract DeployOptimismAdapter is Script, Test, Constants {
             "Optimism_Adapter should only be deployed on Ethereum mainnet or Sepolia"
         );
 
+        address weth = getWrappedNativeToken(chainId);
+
         // Get OP Stack addresses for this chain and Optimism
         uint256 optimismChainId = getChainId("OPTIMISM");
```
pxrl, line 38: Only tangentially related to this change, but I feel like we should be reading the spoke chain ID from the environment and then using it to resolve all the correct addresses. With that approach I think we we remove a barrier to consolidating a lot of scripts. wdyt?

(This approach worked well for the OP Adapter deployment script [here](https://github.com/across-protocol/contracts/blob/f722857a9cd5edefb8fb33c70b2904a8822725e1/deploy/056_deploy_op_adapter.ts#L27))

script/utils/DeploymentUtils.sol
```diff
@@ -180,19 +180,21 @@ contract DeploymentUtils is Script, Test, Constants, DeployedAddresses {
      */
     function isTestnet(uint256 chainId) internal view returns (bool) {
```
pxrl, line 181: Testnet chain IDs are identified separately via `TESTNET_CHAIN_IDs` in constants. Could we check for `chainId` inclusion in that mapping instead of enumerating these here separately?

script/utils/Constants.sol
```diff
@@ -1,11 +1,7 @@
 // SPDX-License-Identifier: BUSL-1.1
```
fusmanii, line 1: Most of the changes in this file is just updating the JSON keys to uppercase

script/utils/constants.json
pxrl, line 1: Given that this is generated, wdyt about storing it separately from code? i.e. under an `artefacts` directory or something?

script/utils/Constants.sol
```diff
@@ -212,53 +218,37 @@ contract Constants is Script {
...
-        string memory chainName = _getChainName(chainId);
-        return uint32(vm.parseJsonUint(file, string.concat(".circleDomainIds.", chainName)));
+    function getCircleDomainId(uint256 chainId) public view returns (uint256) {
+        int256 cctpDomain = vm.parseJsonInt(
+            file,
+            string.concat(".PUBLIC_NETWORKS.", vm.toString(chainId), ".cctpDomain")
+        );
+        if (cctpDomain == -1) {
+            revert("Circle domain ID not found");
+        }
```
pxrl, line 229: Does this impliy that we know CCTP is supported for the chain in advance; should we have a separate function that determines whether it's supported? Or should the revert be caught to infer that CCTP is not supported?

Background: For a generic OP adapter that supports all variety of configurations, we basically need to configure for native or bridged USDC dynamically based on the domain ID.

pxrl (commented):
Nice - this is going to be really useful. I have a couple of broader comments but I like the direction this is going.

script/002DeployOptimismAdapter.s.sol
fusmanii: Thats good idea, I was thinking of using this strategy in the spokepool consolidation [pr](https://github.com/across-protocol/contracts/pull/1087), wanted to get some initial feedback on the direction first before I put any more effort

script/utils/DeploymentUtils.sol
```diff
@@ -201,10 +203,7 @@ contract DeploymentUtils is Script, Test, Constants, DeployedAddresses {
      * @param chainId Chain ID to check
      */
     function checkZkSyncChain(uint256 chainId) internal view {
```
grasphoper: Should we call this `isChainZkStack`? (as we're checking for `ZK_STACK` family here)

grasphoper (commented):
Looks good. It's be nice to have https://github.com/across-protocol/contracts/pull/1111/files#r2356459637 before we merge if possible

script/utils/Constants.sol
```diff
@@ -212,53 +218,37 @@ contract Constants is Script {
...
+        int256 oftEid = vm.parseJsonInt(file, string.concat(".PUBLIC_NETWORKS.", vm.toString(chainId), ".oftEid"));
+        if (oftEid == -1) {
+            revert("OFT EID not found");
+        }
+        return uint256(oftEid);
+    }
+
+    function getChainFamily(uint256 chainId) public view returns (string memory) {
+        uint256 family = vm.parseJsonUint(file, string.concat(".PUBLIC_NETWORKS.", vm.toString(chainId), ".family"));
+        return vm.parseJsonUint(file, string.concat(".CHAIN_FAMILIES.", vm.toString(family)));
```
grasphoper: This gives me LSP errors. Seems like we want to return a `string`, but returning a uint?

script/utils/DeploymentUtils.sol
fusmanii: done https://github.com/across-protocol/contracts/pull/1111/commits/45a034ac169786cc6982f947c4326049e418fe37

script/utils/constants.json
fusmanii: My concern with this is that the file can get out of date. From what I can tell, foundry doesn't have a like a pre-script hook where we can generate this file. 
I think in general better approach would be to have this file generated in the constants repo and we just import it here

script/utils/Constants.sol
fusmanii: done https://github.com/across-protocol/contracts/pull/1111/commits/7b16cb5ac73f2f9890b55dffe26e4b55fe2ee3c1

script/utils/DeploymentUtils.sol
fusmanii: done https://github.com/across-protocol/contracts/pull/1111/commits/7b16cb5ac73f2f9890b55dffe26e4b55fe2ee3c1

script/utils/Constants.sol
fusmanii: my idea here was that we dont want to have a script accidentally set `-1` as domain id but I think that would fail because of `uint256` on the contract.

added `hasCctpDomain` that can be used upstream  https://github.com/across-protocol/contracts/pull/1111/commits/9daffacbdf023e2269ad91ebdec2ca3dd5e5dd58

script/utils/constants.json
pxrl: Yeah; I like proding it as a build artefact over there and importing it here.

Given the above it's a moot point, but I don't fwiw understand what you mean about the file getting out of date; I'm just suggesting that it be stored separately from code in the repo.

script/utils/DeploymentUtils.sol
pxrl: It took a while for me to figure this out, but Matter Labs call it Elastic Chain. ZK_STACK is unfortunately a pretty heavily overloaded term so I am wondering whether we should just rename it everywhere.

script/002DeployOptimismAdapter.s.sol
pxrl: Yeah I left some comments on that one; we have an unresolved question about the exact strategy to take, but in general it's unavoiable that we consolidate these contracts & scripts so I'm fully in favour of frontloading as much of that as possible, including setting SPOKE_CHAIN_ID in the env and auto-configuring everything based on that.

script/utils/constants.json
fusmanii: moved it to `generated` folder (didn't go with artefacts since hardhat already uses that)
https://github.com/across-protocol/contracts/pull/1111/commits/822fd1d4f39f8399a09471055da9cb953631de90

generated/constants.json
pxrl: @fusmanii I noticed that we pull in the numeric value/index for ts enums. Is it possible to pull in the symbolic representation instead?

fusmanii: done https://github.com/across-protocol/contracts/pull/1111/commits/c14c3040134e1e546c68dc578b3748ac243a7577

PR #1113, opened 2025-09-19
chore: Support Etherscan V2 verification

fusmanii:
Any chain that is in [this](https://api.etherscan.io/v2/chainlist) list doesn't need an api key or `customChains` definition

All other chains that are not in that list will need to go in `blockscout` section


utils/utils.hre.ts
```diff
@@ -92,7 +92,11 @@ export async function deployNewProxy(
   // to the implementation's ABI on etherscan.
   // https://docs.openzeppelin.com/upgrades-plugins/1.x/api-hardhat-upgrades#verify
   const contract = `contracts/${name}.sol:${name}`;
-  await run("verify:verify", { address: instance, constructorArguments: constructorArgs, contract });
+  if (hre.config.blockscout.enabled && hre.config.blockscout.customChains.some((chain) => chain.chainId === chainId)) {
+    await run("verify:blockscout", { address: instance, constructorArguments: constructorArgs, contract });
+  } else {
+    await run("verify:verify", { address: instance, constructorArguments: constructorArgs, contract });
+  }
```
fusmanii: will need to abstract this away since `"verify:verify"` is used in other scripts as well

grasphoper:
@fusmanii this is very interesting! You didn't need to bump `hardhat` for this to work?

PR #1114, opened 2025-09-22
feat: Plasma Deployments

fusmanii:
- [x] SpokePool
- [x] UniversalAdapter
- [x] SpokePoolPeriphery
- [x] SpokePoolVerifier
- [x] MulticallHandler


broadcast/deployed-addresses.json
```diff
@@ -600,6 +600,26 @@
...
     },
+    "9745": {
+      "chain_name": "Plasma",
+      "contracts": {
+        "Helios": {
+          "address": "0x7e63a5f1a8f0b4d0934b2f2327daed3f6bb2ee75",
+          "transaction_hash": "Unknown",
+          "block_number": 1552582
+        },
+        "SpokePoolVerifier": {
```
bmzig, line 621: 2 small questions:
- Why does the `transaction_hash` resolve as unknown? If it's not important, then why even include it as part of this `deployed-addresses.json`?
- Should we be deploying the spoke pool verifier or the spoke pool periphery, which should include the spoke pool verifier logic?

fusmanii: - The  `transaction_hash` get populated if the contract has been deployed with foundry, I will have a follow up PR to populate all the missing fields from `deployments` 
- Good point, I wasn't familiar with SpokePoolPeriphery contract, I can deploy that as well

bmzig (dismissed):

hardhat.config.ts
```diff
@@ -250,6 +250,13 @@ const config: HardhatUserConfig = {
       accounts: { mnemonic },
       companionNetworks: { l1: "mainnet" },
     },
+    plasma: {
+      chainId: CHAIN_IDs.PLASMA,
+      url: getNodeUrl(CHAIN_IDs.PLASMA),
+      saveDeployments: true,
```
nicholaspai, line 256: Just a thought: I wonder if we can clean up this file by making a default config object function:

```ts
defaultConfig(chainId: number): {
   return {
       chainId,
       url: getNodeUrl(chainId),
       saveDeployments: true,
       accounts: { mnemonic },
       companionNetworks: { l1: "mainnet" },
    }
}
```

nicholaspai (approved):

pxrl (approved):
Updates look good to me; OK pending the tsc fix.

PR #1116, opened 2025-09-22
refactor(hardhat.config): Use default hardhat config where appropriate 

nicholaspai:
I didn't want to make the diff too large to review so I'll save testnet/zksync configs for follow up

### DIFF
--- a/hardhat.config.ts
+++ b/hardhat.config.ts
@@ -33,6 +33,16 @@ const getMnemonic = () => {
 };
 const mnemonic = getMnemonic();
 
+const getDefaultHardhatConfig = (chainId: number): any => {
+  return {
+    chainId,
+    url: getNodeUrl(chainId),
+    accounts: { mnemonic },
+    saveDeployments: true,
+    companionNetworks: { l1: "mainnet" },
+  };
+};
+
 // Custom tasks to add to HRE.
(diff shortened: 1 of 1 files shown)


hardhat.config.ts
```diff
@@ -155,27 +159,15 @@ const config: HardhatUserConfig = {
...
-      saveDeployments: true,
-      companionNetworks: { l1: "mainnet" },
-    },
+    optimism: getDefaultHardhatConfig(CHAIN_IDs.OPTIMISM),
     "optimism-sepolia": {
       chainId: CHAIN_IDs.OPTIMISM_SEPOLIA,
       url: getNodeUrl(CHAIN_IDs.OPTIMISM_SEPOLIA),
       accounts: { mnemonic },
       saveDeployments: true,
       companionNetworks: { l1: "sepolia" },
```
fusmanii, line 168:
```suggestion
      ...getDefaultHardhatConfig(CHAIN_IDs.OPTIMISM_SEPOLIA)
      companionNetworks: { l1: "sepolia" },
```
or would that be too convoluted?

nicholaspai: I think i'll define a helper function to do this, but i want to keep the diff low for now

PR #1118, opened 2025-09-23
chore: redeploy plasma

bmzig:


deployments/plasma/Universal_SpokePool.json
```diff
@@ -3278,9 +3278,9 @@
     "0x0000000000000000000000000000000000000000",
     "0x0000000000000000000000000000000000000000",
     30101,
-    "6000000000000000000000"
+    "600000000000000000000"
```
bmzig: Dang. Probably want this to be 6000, not 600. Will have to redeploy again.

PR #1120, opened 2025-09-23
feat: Add deploy Safe script

nicholaspai:
Required to have safe deployed when using the UniversalSpokePool which is an ownable contract.

Note that this script will deploy the safe at a new address than the existing safe's because the salt nonce is set differently from whatever the Safe UI uses.

### DIFF
--- a/scripts/deployMultisig.ts
+++ b/scripts/deployMultisig.ts
@@ -1,0 +1,97 @@
+import { getNodeUrl } from "../utils";
+import { ethers } from "../utils/utils";
+import { hre } from "../utils/utils.hre";
+import Safe, { SafeAccountConfig, PredictedSafeProps } from "@safe-global/protocol-kit";
+
+const safeAccountConfig: SafeAccountConfig = {
+  owners: [
+    "0x868CF19464e17F76D6419ACC802B122c22D2FD34",
+    "0xcc400c09ecBAC3e0033e4587BdFAABB26223e37d",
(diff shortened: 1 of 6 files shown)


scripts/deployMultisig.ts
```diff
@@ -0,0 +1,97 @@
...
+  owners: [
+    "0x868CF19464e17F76D6419ACC802B122c22D2FD34",
+    "0xcc400c09ecBAC3e0033e4587BdFAABB26223e37d",
+    "0x837219D7a9C666F5542c4559Bf17D7B804E5c5fe",
+    "0x1d933Fd71FF07E69f066d50B39a7C34EB3b69F05",
+    "0x996267d7d1B7f5046543feDe2c2Db473Ed4f65e9",
+  ],
+  threshold: 2,
+};
+const EXPECTED_SAFE_ADDRESS = "0x0Fc8E2BB9bEd4FDb51a0d36f2415c4C7F9e75F6e";
```
fusmanii, line 16: shouldn't this be `0xB524735356985D2f267FA010D681f061DfF03715`?

nicholaspai: nope, because we use a different salt than the one in Gnosis Safe UI: 

> Note that this script will deploy the safe at a new address than the existing safe's because the salt nonce is set differently from whatever the Safe UI uses.

nicholaspai: Oh also 0xB5 was created with a different set of signers

PR #1121, opened 2025-09-24
feat: Transfer Ownership on Universal SpokePool Deployment

fusmanii:
Closes https://linear.app/uma/issue/ACX-4470/contracts-support-ownership-transfer-to-new-safe-address-post

### DIFF
--- a/deploy/111_deploy_universal_spokepool.ts
+++ b/deploy/111_deploy_universal_spokepool.ts
@@ -65,6 +65,10 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
   // then a message designed to be sent to one chain could be sent to another's SpokePool.
   const { proxyAddress } = await deployNewProxy("Universal_SpokePool", constructorArgs, initArgs);
 
+  if (!proxyAddress) {
+    return;
+  }
+
   const nodeUrl = getNodeUrl(spokeChainId);
 
   const protocolKit = await Safe.init({
@@ -76,21 +80,20 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
     safeAddress: EXPECTED_SAFE_ADDRESS,
   });
   const isDeployed = await existingProtocolKit.isSafeDeployed();
-  if (proxyAddress) {
-    if (isDeployed) {
-      const factory = await hre.ethers.getContractFactory("Universal_SpokePool");
-      const contract = factory.attach(proxyAddress);
 
-      const owner = await contract.owner();
-      if (owner !== EXPECTED_SAFE_ADDRESS) {
-        await (await contract.transferOwnership(EXPECTED_SAFE_ADDRESS)).wait();
-        console.log("Transferred ownership to Expected Safe address:", await contract.owner());
-      } else {
-        console.log("Expected Safe address is already the owner of the Universal SpokePool");
-      }
-    } else {
-      console.log("Safe is not deployed, skipping ownership transfer");
-    }
+  if (!isDeployed) {
+    throw new Error("Expected Safe address is not deployed, please deploy it first");
+  }
+
+  const factory = await hre.ethers.getContractFactory("Universal_SpokePool");
+  const contract = factory.attach(proxyAddress);
+
+  const owner = await contract.owner();
+  if (owner !== EXPECTED_SAFE_ADDRESS) {
+    await (await contract.transferOwnership(EXPECTED_SAFE_ADDRESS)).wait();
+    console.log("Transferred ownership to Expected Safe address:", await contract.owner());
+  } else {
+    console.log("Expected Safe address is already the owner of the Universal SpokePool");
   }
 };
 module.exports = func;


deploy/111_deploy_universal_spokepool.ts
```diff
@@ -53,7 +61,22 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
   // target address of the spoke pool. This is because the HubPool does not pass in the chainId when calling
   // relayMessage() on the Adapter. Therefore, if Universal SpokePools share the same address,
   // then a message designed to be sent to one chain could be sent to another's SpokePool.
-  await deployNewProxy("Universal_SpokePool", constructorArgs, initArgs);
+  const { proxyAddress } = await deployNewProxy("Universal_SpokePool", constructorArgs, initArgs);
+
+  const provider = hre.ethers.provider;
+  const safeCode = await provider.getCode(EXPECTED_SAFE_ADDRESS);
+  if (safeCode !== "0x" && proxyAddress) {
```
pxrl: Also worth checking the owners of the safe deployment here?

deploy/111_deploy_universal_spokepool.ts
```diff
@@ -53,7 +61,22 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
...
+    const factory = await hre.ethers.getContractFactory("Universal_SpokePool");
+    const contract = factory.attach(proxyAddress);
+
+    const owner = await contract.owner();
+    if (owner !== EXPECTED_SAFE_ADDRESS) {
+      await contract.transferOwnership(EXPECTED_SAFE_ADDRESS);
+      console.log("Transferred ownership to Expected Safe address:", EXPECTED_SAFE_ADDRESS);
```
pxrl: wdyt about waiting for transaction confirmation and then reading the updated value here?

pxrl (commented):
LGTM; two queries.

deploy/111_deploy_universal_spokepool.ts
fusmanii: used safe kit to match the deployed mutisig https://github.com/across-protocol/contracts/pull/1121/commits/8c03b88657efb2ce280483bc20b3b18d8e2de02d

deploy/111_deploy_universal_spokepool.ts
fusmanii: done https://github.com/across-protocol/contracts/pull/1121/commits/8c03b88657efb2ce280483bc20b3b18d8e2de02d

deploy/111_deploy_universal_spokepool.ts
```diff
@@ -53,7 +63,35 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
...
+      if (owner !== EXPECTED_SAFE_ADDRESS) {
+        await (await contract.transferOwnership(EXPECTED_SAFE_ADDRESS)).wait();
+        console.log("Transferred ownership to Expected Safe address:", await contract.owner());
+      } else {
+        console.log("Expected Safe address is already the owner of the Universal SpokePool");
+      }
+    } else {
+      console.log("Safe is not deployed, skipping ownership transfer");
+    }
+  }
```
pxrl, line 97: I don't quite follow the logic here. If there's no `proxyAddress` is there any point to the code between lines 66 and 78?

It also seems possible to squash a couple of levels of indentation by inverting the logic and returning early on various conditions, i.e.:

```
if (!proxyAddress) {
  return;
}

if (!isDeployed) {
  console.log(...);
  return;
}

// ...
```

deploy/111_deploy_universal_spokepool.ts
```diff
@@ -53,7 +63,35 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
...
+
+      const owner = await contract.owner();
+      if (owner !== EXPECTED_SAFE_ADDRESS) {
+        await (await contract.transferOwnership(EXPECTED_SAFE_ADDRESS)).wait();
+        console.log("Transferred ownership to Expected Safe address:", await contract.owner());
+      } else {
+        console.log("Expected Safe address is already the owner of the Universal SpokePool");
+      }
+    } else {
+      console.log("Safe is not deployed, skipping ownership transfer");
```
pxrl: Actually one additional thing - shouldn't we consider this to be an error condition? We basically require a safe to own each Universal SpokePool; if for whatever reason we end up on this code path, I feel like it should be obvious to the deployer that something is wrong.

deploy/111_deploy_universal_spokepool.ts
fusmanii: looks much cleaner now, https://github.com/across-protocol/contracts/pull/1121/commits/9bd543b25e829cdd62c6b84acdb648cce2258742

deploy/111_deploy_universal_spokepool.ts
fusmanii: Converted the log to throw an error instead https://github.com/across-protocol/contracts/pull/1121/commits/9bd543b25e829cdd62c6b84acdb648cce2258742

PR #1123, opened 2025-09-26
feat: Upgrade Solana CCTP IDL's to V2

nicholaspai:
Question for reviewers: What else do I need to change so that I can import the new IDL's [here](https://github.com/across-protocol/relayer/blob/80b333c90da203bdedd12551d02b7195ef6985ba/src/adapter/bridges/SolanaUsdcCCTPBridge.ts#L26) ?

Source: https://developers.circle.com/cctp/solana-programs

### DIFF
--- a/idls/message_transmitter.json
+++ b/idls/message_transmitter.json
@@ -1,11 +1,1655 @@
 {
-  "address": "CCTPmbSD7gX1bxKPAmg77w8oFzNFpaQiQUWD43TKaecd",
+  "address": "CCTPV2Sm4AdWt5296sk4P66VBZ7bEhcARwFaaS9YPbeC",
   "metadata": {
     "name": "message_transmitter",
-    "version": "0.1.0",
-    "spec": "0.1.0"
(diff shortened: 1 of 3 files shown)


broadcast/deployed-addresses.json
```diff
@@ -1042,6 +1042,16 @@
           "address": "CCTPiPYPc6AsJuwueEnWgSgucamXDZwBd53dQ11YiKX3",
           "transaction_hash": "Unknown",
           "block_number": 277864039
+        },
+        "MessageTransmitterV2": {
+          "address": "CCTPV2Sm4AdWt5296sk4P66VBZ7bEhcARwFaaS9YPbeC",
+          "transaction_hash": "Unknown",
+          "block_number": 383716739
```
nicholaspai, line 1049: Should we hardcode these to the actual deployment blocks or does it not matter?

Reinis-FRP: I took it these from the last deployment slots as shown in explorer.

PR #1127, opened 2025-10-03
feat: Upgrade CCTP V1 contracts to V2

nicholaspai:
Deployed new adapters with CCTP V2 contracts enabled for: Optimism, Arbitrum, Unichain, Base, Polygon


bmzig (dismissed):

script/utils/GenerateConstantsJson.ts
```diff
@@ -101,6 +102,7 @@ function generateConstantsJson() {
     USDC,
     USDCe,
     WGHO,
+    WMATIC,
```
fusmanii, line 105: dont think you need to add this since `WRAPPED_NATIVE_TOKENS` already has `WMATIC`

fusmanii (commented):
Deployed contracts look good, just one comment

script/utils/GenerateConstantsJson.ts
nicholaspai: oh does it??

nicholaspai: oh but we need the L1 Matic address somehow

fusmanii (dismissed):

bmzig (approved):

PR #1131, opened 2025-10-07
feat: Added OP foundry deploy script

fusmanii:


nicholaspai (dismissed):

script/025DeployOPSpokePool.s.sol
```diff
@@ -0,0 +1,72 @@
...
+            info.hubPool, // _crossDomainAdmin
+            info.hubPool // _withdrawalRecipient
+        );
+
+        // Deploy the proxy
+        DeploymentResult memory result = deployNewProxy(
+            "OP_SpokePool",
+            constructorArgs,
+            initArgs,
+            true // implementationOnly
```
pxrl: We might want to read this from the environment, or try to infer it from existing config. If they find a proxy then they skip deploying a new one, and otherwise they deploy both implementation and proxy.

script/025DeployOPSpokePool.s.sol
```diff
@@ -0,0 +1,72 @@
...
+import { console } from "forge-std/console.sol";
+import { OP_SpokePool } from "../contracts/OP_SpokePool.sol";
+import { DeploymentUtils } from "./utils/DeploymentUtils.sol";
+
+// How to run:
+// 1. `source .env` where `.env` has MNEMONIC="x x x ... x"
+// 2. forge script script/039DeployOPSpokePool.s.sol:DeployOPSpokePool --rpc-url $NODE_URL_1 -vvvv
+// 3. Verify the above works in simulation mode.
+// 4. Deploy with:
+//        forge script script/039DeployOPSpokePool.s.sol:DeployOPSpokePool --rpc-url \
```
pxrl: Minor:

```suggestion
// 2. forge script script/025DeployOPSpokePool.s.sol:DeployOPSpokePool --rpc-url $NODE_URL_1 -vvvv
// 3. Verify the above works in simulation mode.
// 4. Deploy with:
//        forge script script/025DeployOPSpokePool.s.sol:DeployOPSpokePool --rpc-url \
```

pxrl (commented):
LGTM - one question on proxy deployment.

script/025DeployOPSpokePool.s.sol
fusmanii: that is already being done [here](https://github.com/across-protocol/contracts/blob/master/script/utils/DeploymentUtils.sol#L94) upstream, this flag is just for when we want to forcefully deploy the implementation only

script/025DeployOPSpokePool.s.sol
fusmanii: done https://github.com/across-protocol/contracts/pull/1131/commits/0625b2f24f0496804e33a02ca987a978fcf062b2

script/025DeployOPSpokePool.s.sol
fusmanii: set the flag to false https://github.com/across-protocol/contracts/pull/1131/commits/0625b2f24f0496804e33a02ca987a978fcf062b2

script/025DeployOPSpokePool.s.sol
```diff
@@ -0,0 +1,72 @@
...
+            FILL_DEADLINE_BUFFER(), // _fillDeadlineBuffer
+            l2Usdc, // _l2Usdc
+            cctpTokenMessenger // _cctpTokenMessenger
+        );
+
+        // Initialize deposit counter to 1
+        // Set hub pool as cross domain admin since it delegatecalls the Adapter logic.
+        bytes memory initArgs = abi.encodeWithSelector(
+            OP_SpokePool.initialize.selector,
+            1, // _initialDepositId
```
nicholaspai: Why 1? Why not 1_000_000 as in the other spoke pool deploy scripts?

fusmanii: Some of the deploy scripts have 1 and others have 1_000_000. Changed it to `1_000_000` and added a comment

nicholaspai (approved):

PR #1132, opened 2025-10-08
feat: OP Adapter update

fusmanii:
Adding similar logic that exists in https://github.com/across-protocol/contracts/blob/master/contracts/chain-adapters/ZkStack_Adapter.sol#L102 that checks that only one of Native or CCTP is set for bridging USDC

### DIFF
--- a/test/evm/hardhat/chain-adapters/OP_Adapter.ts
+++ b/test/evm/hardhat/chain-adapters/OP_Adapter.ts
@@ -12,9 +12,11 @@ import {
   toWei,
   getContractFactory,
   seedWallet,
+  toBN,
 } from "../../../../utils/utils";
 import { hubPoolFixture, enableTokensForLP } from "../fixtures/HubPool.Fixture";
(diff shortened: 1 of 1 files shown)


contracts/chain-adapters/OP_Adapter.sol
```diff
@@ -79,12 +91,7 @@ contract OP_Adapter is CrossDomainEnabled, AdapterInterface, CircleCCTPAdapter {
      * @param amount Amount of L1 tokens to deposit and L2 tokens to receive.
      * @param to Bridge recipient.
      */
-    function relayTokens(
-        address l1Token,
-        address l2Token,
-        uint256 amount,
-        address to
-    ) external payable override {
+    function relayTokens(address l1Token, address l2Token, uint256 amount, address to) external payable override {
```
pxrl, line 94: This looks like a formatting change; for the sake of minimising the diff it might make sense to back it out.

fusmanii: The pre-commit hook formatting it not letting me undo this

PR #1135, opened 2025-10-11
feat: Sponsored Bridging - CCTP

fusmanii:
Closes ACX-4537

### DIFF
--- a/contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
+++ b/contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
@@ -43,10 +43,9 @@ contract HyperCoreFlowExecutor is AccessControl {
     /// @notice All operations performed in this contract are relative to this baseToken
     address immutable baseToken;
 
-    /// @notice The minimum delay between finalizations of pending swaps.
-    uint256 public minDelayBetweenFinalizations = 1 minutes;
-    /// @notice The time of the last finalization of pending swaps per final token.
-    mapping(address finalToken => uint256 lastFinalizationTime) lastFinalizationTime;
+    /// @notice The block number of the last funds pull action per final token: either as a part of finalizing pending swaps,
+    /// or an admin funds pull
+    mapping(address finalToken => uint256 lastPullFundsBlock) lastPullFundsBlock;
 
     /// @notice A struct used for storing state of a swap flow that has been initialized, but not yet finished
     struct PendingSwap {
@@ -245,20 +244,6 @@ contract HyperCoreFlowExecutor is AccessControl {
      **************************************/
 
     /**
-     *
-     * @notice Set a minimum safe delay in seconds between consecutive calls to finalize pending swap flows. The reason
-     * for this delay is that the transfers we're sending out of the SwapHandler's account to the user take a couple of
-     * seconds to land on chain + balance read from the finalization function is of stale data (also a couple seconds stale).
-     * If the delay is too small, a collision could happen where we would try to send out funds to the user that are already
-     * "in-flight"(pending orders are about to be executed) to fill another user's order. This could be mitigated if we had
-     * a unique account per order (in the next implementation iteration).
-     * @param minDelay minimum delay to set, in seconds
-     */
-    function setMinDelayBetweenFinalizations(uint256 minDelay) external onlyDefaultAdmin {
-        minDelayBetweenFinalizations = minDelay;
-    }
-
-    /**
      * @notice Set or update information for the token to use it in this contract
      * @dev To be able to use the token in the swap flow, FinalTokenInfo has to be set as well
      * @dev Setting core token info to incorrect values can lead to loss of funds. Should NEVER be unset while the
@@ -312,6 +297,7 @@ contract HyperCoreFlowExecutor is AccessControl {
             suggestedDiscountBps: suggestedDiscountBps
         });
 
+        // TODO: this activation flow seems broken. Can a smart contract activate its core account by doing an EVM -> Core transfer?
         uint256 accountActivationFee = _getAccountActivationFeeEVM(accountActivationFeeToken, address(swapHandler));
         if (accountActivationFee > 0) {
             CoreTokenInfo memory accountActivationTokenInfo = coreTokenInfos[accountActivationFeeToken];
@@ -638,11 +624,8 @@ contract HyperCoreFlowExecutor is AccessControl {
         FinalTokenInfo memory finalTokenInfo = finalTokenInfos[finalToken];
 
         require(address(finalTokenInfo.swapHandler) != address(0), "Final token not registered");
-        require(
-            block.timestamp >= lastFinalizationTime[finalToken] + minDelayBetweenFinalizations,
-            "Min delay not reached"
-        );
-        lastFinalizationTime[finalToken] = block.timestamp;
+        require(lastPullFundsBlock[finalToken] < block.number, "Can't finalize twice in the same block");
+        lastPullFundsBlock[finalToken] = block.number;
 
         uint256 head = pendingQueueHead[finalToken];
         bytes32[] storage queue = pendingQueue[finalToken];
@@ -791,7 +774,7 @@ contract HyperCoreFlowExecutor is AccessControl {
             IERC20(finalToken).safeTransfer(address(swapHandler), sponsorDeltaEvm);
             cumulativeSponsoredAmount[finalToken] += sponsorDeltaEvm;
             if (
-                HyperCoreLib.isCoreAmountSafeToBridge(
+                !HyperCoreLib.isCoreAmountSafeToBridge(
                     finalTokenCoreInfo.coreIndex,
                     sponsorDeltaCore,
                     finalTokenCoreInfo.bridgeSafetyBufferCore
@@ -959,8 +942,9 @@ contract HyperCoreFlowExecutor is AccessControl {
             revert("Cannot sweep on core if there are pending limit orders");
         }
 
-        //  We also want to make sure the min delay between finalizations has been reached
-        require(block.timestamp >= lastFinalizationTime[token] + minDelayBetweenFinalizations, "Min delay not reached");
+        // Prevent pulling fantom funds (e.g. if finalizePendingSwaps reads stale balance because of this fund pull)
+        require(lastPullFundsBlock[token] < block.number, "Can't pull funds twice in the same block");
+        lastPullFundsBlock[token] = block.number;
 
         SwapHandler swapHandler = finalTokenInfos[token].swapHandler;
         swapHandler.transferFundsToUserOnCore(finalTokenInfos[token].assetIndex, msg.sender, amount);


contracts/periphery/mintburn/sponsored-cctp/SponsoredCCTPSrcPeriphery.sol
```diff
@@ -0,0 +1,65 @@
...
+        signer = _signer;
+    }
+
+    using SponsoredCCTPQuoteLib for bytes;
+
+    function depositForBurn(SponsoredCCTPInterface.SponsoredCCTPQuote memory quote, bytes memory signature) external {
+        if (!SponsoredCCTPQuoteLib.validateSignature(signer, quote, signature)) revert InvalidSignature();
+        if (usedNonces[quote.nonce]) revert InvalidNonce();
+        if (quote.deadline < block.timestamp) revert InvalidDeadline();
+
```
Reinis-FRP, line 36: I think we also need to validate `sourceDomain` matches the chain where this contract is deployed (could be stored as immutable set in constructor)

contracts/periphery/mintburn/SwapHandler.sol
```diff
@@ -0,0 +1,62 @@
...
+    function transferFundsToSelfOnCore(
+        address erc20EVMAddress,
+        uint64 erc20CoreIndex,
+        uint256 amountEVM,
+        int8 decimalDiff
+    ) external onlyParentHandler {
+        HyperCoreLib.transferERC20EVMToSelfOnCore(erc20EVMAddress, erc20CoreIndex, amountEVM, decimalDiff);
+    }
+
+    function transferFundsToUserOnCore(uint64 erc20CoreIndex, address to, uint64 amountCore) external {
```
tbwebb22: should this have the `onlyParentHandler` modifier?

contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
```diff
@@ -0,0 +1,701 @@
...
+
+        if (accountActivationFee > 0 && coreTokenInfo.canBeUsedForAccountActivation) {
+            _getFromDonationBox(finalToken, accountActivationFee);
+            IERC20(finalToken).safeTransfer(address(swapHandler), accountActivationFee);
+            swapHandler.activateCoreAccount(
+                finalToken,
+                coreTokenInfo.coreIndex,
+                accountActivationFee,
+                coreTokenInfo.tokenInfo.evmExtraWeiDecimals
+            );
```
tbwebb22: My understanding is that it doesn't matter where the 1 USDT, USDC, or USDH comes from for account activation. So from here we could just send directly to `SwapHandler` on Core, and then be able remove that `activateCoreAccount` function from `SwapHandler`

contracts/periphery/mintburn/SwapHandler.sol
grasphoper: 100%, great catch! Fixed

contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
grasphoper: The reason to do it like this is to not commingle this 1 dollar with the potential user sponsored payout. There's not an obvious problem from it if we calculate all amounts correctly. But this is just a bit safer I think (e.g. no L1Reads of Handler @ core balance that return the amount that's going to be sent away (1 dollar))

tbwebb22: makes sense!

contracts/libraries/BytesLib.sol
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    /**
+     * @notice Reads a bytes array from a bytes array at a given start index and length
+     * @param _bytes The bytes array to convert
+     * @param _start The start index of the bytes array
+     * @param _end The end index of the bytes array
+     * @return result The bytes array result
+     */
+    function slice(bytes memory _bytes, uint256 _start, uint256 _end) internal pure returns (bytes memory result) {
```
mrice32, line 74: Can we import OZ's function instead of homerolling? https://github.com/OpenZeppelin/openzeppelin-contracts/blob/1cf13771092c83a060eaef0f8809493fb4c04eb1/contracts/utils/Bytes.sol#L80-L98

contracts/libraries/SponsoredCCTPQuoteLib.sol
```diff
@@ -0,0 +1,122 @@
...
+import { SponsoredCCTPInterface } from "../interfaces/SponsoredCCTPInterface.sol";
+import { BytesLib } from "./BytesLib.sol";
+import { Bytes32ToAddress } from "./AddressConverters.sol";
+
+library SponsoredCCTPQuoteLib {
+    using BytesLib for bytes;
+    using Bytes32ToAddress for bytes32;
+
+    // Indices of each field in message
+    uint256 private constant VERSION_INDEX = 0;
```
mrice32, line 15: Why not use abi encoding to reduce the amount of manual encode/decode?

contracts/periphery/mintburn/sponsored-cctp/SponsoredCCTPDstPeriphery.sol
```diff
@@ -0,0 +1,106 @@
...
+        return quote.burnToken != quote.finalToken && _isQuoteValid(quote, signature);
+    }
+
+    function _isQuoteValid(
+        SponsoredCCTPInterface.SponsoredCCTPQuote memory quote,
+        bytes memory signature
+    ) internal view returns (bool) {
+        return
+            SponsoredCCTPQuoteLib.validateSignature(signer, quote, signature) &&
+            !usedNonces[quote.nonce] &&
```
mrice32, line 106: We need a `usedNonces[quote.nonce] = true` somewhere (either in this function or in the calling fn).

contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
```diff
@@ -0,0 +1,1074 @@
...
+        }
+
+        // Record `accountCreationFee` as zero if we can't use final token for account activation
+        uint256 accountCreationFee = coreUserAccountExists ? 0 : coreTokenInfo.accountActivationFeeEVM;
+        uint256 maxEvmAmountToSponsor = ((amount + extraFeesToSponsor) * maxBpsToSponsor) / BPS_SCALAR;
+        uint256 totalUserFeesEvm = extraFeesToSponsor + accountCreationFee;
+        uint256 amountToSponsor = totalUserFeesEvm;
+        if (amountToSponsor > maxEvmAmountToSponsor) {
+            amountToSponsor = maxEvmAmountToSponsor;
+        }
```
mrice32, line 379: I think this fallback logic makes sense, but just to be clear, this would be an error on the API side, correct?

Because at the API level, we should always ensure the maxFee is less than the maxAmountToSponsor - accountInitiationFee, right?

contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
```diff
@@ -0,0 +1,1074 @@
...
+        );
+        // If there are no funds left on the destination side of the bridge, the funds will be lost in the
+        // bridge. We check send safety via `isCoreAmountSafeToBridge`
+        bool isSafe = HyperCoreLib.isCoreAmountSafeToBridge(
+            coreTokenInfo.coreIndex,
+            quotedCoreAmount,
+            coreTokenInfo.bridgeSafetyBufferCore
+        );
+
+        // If the amount is not safe to bridge or the user has no HyperCore account and we can't sponsor its creation;
```
mrice32: We handle the Hypercore account sponsorship check in logic higher up in the function I think.
```suggestion
        // If the amount is not safe to bridge because the bridge doesn't have enough liquidity;
```

contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
```diff
@@ -0,0 +1,1074 @@
...
+        }
+
+        if (amountToSponsor > 0) {
+            // This will succeed because we checked the balance earlier
+            _getFromDonationBox(coreTokenInfo.tokenInfo.evmContract, amountToSponsor);
+        }
+
+        cumulativeSponsoredAmount[finalToken] += amountToSponsor;
+        // Record a propotional amount used to sponsor account creation
+        cumulativeSponsoredActivationFee[finalToken] += (amountToSponsor * accountCreationFee) / totalUserFeesEvm;
```
mrice32: Slight preference to avoid rounding here and assume first dollars go to account creation.
```suggestion
        cumulativeSponsoredActivationFee[finalToken] += amountToSponsor < accountCreationFee ? amountToSponsor : accountCreationFee;
```

mrice32 (commented):
Looks good! I got through most of the swap flow (although, I need to dig more into the utilities).

A few small comments, but overall looks solid!

contracts/libraries/BytesLib.sol
fusmanii: that does look much cleaner, will update it

contracts/libraries/SponsoredCCTPQuoteLib.sol
fusmanii: The message gets encoded by CCTP which we fetch from their API and they are using `encodePacked` instead

contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
fusmanii: There is a very slim chance that CCTP fees increase mid flight and we want to cover that case but for the most part it would be API side error. The API can also add a bit of a buffer to the maxFee to ensure we don't hit any fee spikes

contracts/periphery/mintburn/sponsored-cctp/SponsoredCCTPDstPeriphery.sol
fusmanii: done https://github.com/across-protocol/contracts/pull/1135/commits/5ff83c7c2cc8e80a320abc8ecf54dd4fbd006e48

contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
fusmanii: done https://github.com/across-protocol/contracts/pull/1135/commits/5ff83c7c2cc8e80a320abc8ecf54dd4fbd006e48

contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
fusmanii: done https://github.com/across-protocol/contracts/pull/1135/commits/5ff83c7c2cc8e80a320abc8ecf54dd4fbd006e48

contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
mrice32: Don't we set maxFee on origin and pass it to CCTP? Wouldn't the transfer have to be rejected if the fee can go above the maxFee?

contracts/libraries/SponsoredCCTPQuoteLib.sol
mrice32: Right, makes sense. I realized this later and forgot to delete my comment. +1

contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
```diff
@@ -0,0 +1,1070 @@
...
+                feelessAmountCoreInitialToken,
+                initialCoreTokenInfo.tokenInfo.weiDecimals,
+                finalCoreTokenInfo.tokenInfo.weiDecimals
+            );
+            maxAmountToSponsorCore = uint64((feelessAmountCoreFinalToken * maxBpsToSponsor) / BPS_SCALAR);
+            minAllowableAmountToForwardCore = feelessAmountCoreFinalToken + accountActivationFeeCore;
+        } else {
+            // toCore(amountInEquivalentCore) - slippage + coreAccountActivationFee
+            maxAmountToSponsorCore = 0;
+            minAllowableAmountToForwardCore =
```
mrice32: Don't we need a decimal conversion here between initial token and final token?

i.e. shouldn't this be converted to the final token decimals by calling `convertCoreDecimalsSimple`?

grasphoper: Yeah 100%. Let me fix

contracts/interfaces/SponsoredCCTPInterface.sol
```diff
@@ -0,0 +1,83 @@
...
+        // Maximum fee to pay on the destination domain, specified in units of burnToken
+        uint256 maxFee;
+        // Minimum finality threshold before allowed to attest
+        uint32 minFinalityThreshold;
+        // Nonce is used to prevent replay attacks.
+        bytes32 nonce;
+        // Timestamp of the quote after which it can no longer be used.
+        uint256 deadline;
+        // The maximum basis points of the amount that can be sponsored.
+        uint256 maxBpsToSponsor;
```
mrice32, line 74: I'm a little worried about slight desyncs between the API and the contracts wrt math, causing totally normal transfers to fail contract checks.

Do you think it would make the math simpler (and, therefore, fewer opportunities for rounding problems) if this was communicated in raw amount rather than bps?

mrice32: I'm specifically thinking about rounding issues with things like the account initialization fee and gas fees that are not correlated with transfer size. It could be the same either way, though, since CCTP fees and swap limits are denominated in bps.

contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
grasphoper: Fixed [fa0b1e7](https://github.com/across-protocol/contracts/pull/1135/commits/fa0b1e735b595434c1f4db72a85caf9450fbd6f3)

contracts/interfaces/SponsoredCCTPInterface.sol
grasphoper: I was thinking that this makes the flow easier because we don't need to worry about the destination token decimals then.

Otherwise we'd have to specify the `maxSponsoredAmount` in the final Token. Also, if a swap flow falls back to the simple transfer flow, we'd ned to convert from `maxSponsoredAmount` to the other token sponsored amount

For simple flow I think it's actually pretty hard for the API <> contract math to lead to a failed flow, the API should just round the bps amount up, wdyt?

contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
```diff
@@ -0,0 +1,1091 @@
...
+            (sponsorDeltaEvm, sponsorDeltaCore) = HyperCoreLib.minimumCoreReceiveAmountToAmounts(
+                sponsorDeltaCore,
+                finalTokenCoreInfo.tokenInfo.evmExtraWeiDecimals
+            );
+
+            _getFromDonationBox(finalToken, sponsorDeltaEvm);
+            IERC20(finalToken).safeTransfer(address(swapHandler), sponsorDeltaEvm);
+            cumulativeSponsoredAmount[finalToken] += sponsorDeltaEvm;
+            if (
+                HyperCoreLib.isCoreAmountSafeToBridge(
```
mrice32: Should this be `!HyperCoreLib.isCoreAmountSafeToBridge`?

grasphoper: It should :)

grasphoper: Fixed [d1a39d2](https://github.com/across-protocol/contracts/pull/1135/commits/d1a39d264a0eb9d9f05b59c8ad5b8db672ec2d0b)

PR #1136, opened 2025-10-14
feat(svm): cctp sponsored src

Reinis-FRP:
Fixes: [ACX-4550](https://linear.app/uma/issue/ACX-4550/srcchaincctpperipherysvm)

### DIFF
--- a/programs/sponsored-cctp-src-periphery/src/utils/sponsored_cctp_quote.rs
+++ b/programs/sponsored-cctp-src-periphery/src/utils/sponsored_cctp_quote.rs
@@ -2,8 +2,8 @@ use anchor_lang::{prelude::*, solana_program::keccak};
 
 use crate::error::{DataDecodingError, SvmError};
 
-// Macro to define the SponsoredCCTPQuote fields as an enum with associated constants for ordinal, start, count, and
-// minimum total bytes.
+// Macro to define the SponsoredCCTPQuote fields as an enum with associated constants for ordinal, start, end, count,
+// and minimum total bytes.
 macro_rules! define_quote_fields {
     ($($V:ident),+ $(,)?) => {
         #[derive(Copy, Clone)]


programs/sponsored-cctp-src-periphery/src/instructions/deposit.rs
```diff
@@ -0,0 +1,302 @@
...
+
+    #[account(mut, seeds = [b"rent_fund"], bump)]
+    pub rent_fund: SystemAccount<'info>,
+
+    #[account(mut,close = rent_fund, seeds = [b"used_nonce", &params.nonce.as_ref()], bump)]
+    pub used_nonce: Account<'info, UsedNonce>,
+}
+
+#[derive(AnchorSerialize, AnchorDeserialize)]
+pub struct UsedNonceAccountParams {
```
md0x, line 261: Nit: `UsedNonceAccountParams` struct has only one field we could use that field instead to simplify the `reclaim_used_nonce_account` instruction.

Reinis-FRP: kept this more for the sake of consistency with other instruction params that are structs. there should not be any onchain overhead for this as this serializes to the same bytes on the wire as plain nonce.

grasphoper (commented):
The code looks good to me. There were two last-second changes to our EVM-side code that we have to reflect here too:
- when taking a hash of a quote structure for checking a signature, we're splitting it up into two parts to avoid a stack-too-deep error: https://github.com/across-protocol/contracts/blob/c402e1f5a70e9c9ec206519516c96672bd2ad892/contracts/libraries/SponsoredCCTPQuoteLib.sol#L130. My intuition is that this is not very annoying to fix, but @Reinis-FRP please let us know. CC @fusmanii 
- two extra fields were added to the Quote struct: `executionMode` and `actionData`: https://github.com/across-protocol/contracts/blob/c402e1f5a70e9c9ec206519516c96672bd2ad892/contracts/interfaces/SponsoredCCTPInterface.sol#L94. `executionMode` is probably easy to add. Were there problems, though, to adding a `bytes` array like `actionData`? As a last resort, we could hardcode actionData to be empty from src Solana, which seems guaranteed to be doable

md0x (approved):

programs/sponsored-cctp-src-periphery/src/lib.rs
```diff
@@ -0,0 +1,209 @@
...
+declare_id!("CPr4bRvkVKcSCLyrQpkZrRrwGzQeVAXutFU8WupuBLXq");
+
+// External programs from idls directory (requires anchor run generateExternalTypes).
+declare_program!(message_transmitter_v2);
+declare_program!(token_messenger_minter_v2);
+
+/// # Across Sponsored CCTP Source Periphery
+///
+/// Source chain periphery program for users to interact with to start a sponsored or a non-sponsored flow that allows
+/// custom Accross-supported flows on destination chain. Uses Circle's CCTPv2 as an underlying bridge
```
md0x, line 34:
```suggestion
/// custom Across-supported flows on destination chain. Uses Circle's CCTPv2 as an underlying bridge
```

PR #1137, opened 2025-10-14
feat: Hypercorelib

tbwebb22:
Closes ACX-4548

### DIFF
--- a/contracts/libraries/HyperCoreLib.sol
+++ b/contracts/libraries/HyperCoreLib.sol
@@ -178,7 +178,7 @@ library HyperCoreLib {
         // Basic sanity checks
         if (limitPriceX1e8 == 0) revert LimitPxIsZero();
         if (sizeX1e8 == 0) revert OrderSizeIsZero();
-        if (tif == Tif.None || uint8(tif) > uint8(Tif.IOC)) revert InvalidTif();
+        if (tif == Tif.None || uint8(tif) > uint8(type(Tif).max)) revert InvalidTif();
 
         // Encode the action
         bytes memory encodedAction = abi.encode(asset, isBuy, limitPriceX1e8, sizeX1e8, reduceOnly, uint8(tif), cloid);


contracts/libraries/HyperCoreLib.sol
```diff
@@ -0,0 +1,164 @@
...
+        return 0;
+    }
+
+    /**
+     * @notice Transfers tokens from this contract on HyperCore to the `to` address on HyperCore
+     * @param erc20CoreIndex The HyperCore index id of the token
+     * @param to The address to receive tokens on HyperCore
+     * @param amountCore The amount to transfer on HyperCore
+     */
+    function transferERC20CoreToCore(uint64 erc20CoreIndex, address to, uint64 amountCore) internal {
```
tbwebb22, line 151: This function could check this contract's spot balance on HyperCore and revert if it is less than `amountCore`. Or this check could be handled at the higher level handler contracts. Thoughts on this?

grasphoper: My intuition is that we treat this lib as a rather low-level primitive. Handler contracts are the ones to make sure our interactions with HCore make sense in terms of failures. I'd leave this responsibility to them

contracts/libraries/HyperCoreLib.sol
```diff
@@ -0,0 +1,366 @@
...
+        TokenInfo memory _tokenInfo = abi.decode(result, (TokenInfo));
+        return _tokenInfo;
+    }
+
+    /**
+     * @notice Quotes the conversion of evm tokens to hypercore tokens
+     * @param erc20CoreIndex The HyperCore index id of the token to transfer
+     * @param decimalDiff The decimal difference of evmDecimals - coreDecimals
+     * @param bridgeAddress The asset bridge address of the token to transfer
+     * @param amountEVM The number of tokens that (pre-dusted) that we are trying to send
```
fusmanii, line 251:
```suggestion
     * @param amountEVM The number of tokens (pre-dusted) that we are trying to send
```
also what do you mean by `pre-dusted` here?

contracts/libraries/HyperCoreLib.sol
```diff
@@ -0,0 +1,366 @@
...
+    function into_assetBridgeAddress(uint64 erc20CoreIndex) internal pure returns (address) {
+        return address(uint160(BASE_ASSET_BRIDGE_ADDRESS_UINT256 + erc20CoreIndex));
+    }
+
+    /**
+     * @notice Converts an asset bridge address to a core index id
+     * @param assetBridgeAddress The asset bridge address to convert
+     * @return erc20CoreIndex The core token index id
+     */
+    function into_tokenId(address assetBridgeAddress) internal pure returns (uint64) {
```
fusmanii: let use camelCase for this and the rest of the functions below

contracts/libraries/HyperCoreLib.sol
```diff
@@ -0,0 +1,366 @@
...
+    ) internal pure returns (uint256 amountEVM, uint64 amountCore) {
+        uint256 scale = 10 ** decimalDiff;
+        uint256 maxTransferableAmountEVM = maxTransferableAmountCore * scale;
+
+        unchecked {
+            /// @dev Strip out dust from _amount so that _amount and maxEvmAmountFromCoreMax have a maximum of _decimalDiff starting 0s
+            amountEVM = amountEVMPreDusted - (amountEVMPreDusted % scale); // Safe: dustAmount = amountEVMPreDusted % scale, so dust <= amountEVMPreDusted
+
+            if (amountEVM > maxTransferableAmountEVM)
+                revert TransferAmtExceedsAssetBridgeBalance(amountEVM, maxTransferableAmountEVM);
```
fusmanii, line 338: nice check!!

contracts/libraries/HyperCoreLib.sol
```diff
@@ -0,0 +1,366 @@
...
+        uint64 limitPriceX1e8,
+        uint64 sizeX1e8,
+        bool reduceOnly,
+        uint8 encodedTif,
+        uint128 cloid
+    ) internal {
+        // Basic sanity checks
+        if (limitPriceX1e8 == 0) revert LimitPxIsZero();
+        if (sizeX1e8 == 0) revert OrderSizeIsZero();
+        if (!(encodedTif == 1 || encodedTif == 2 || encodedTif == 3)) revert InvalidTif();
```
fusmanii: should we create an enum for these?

contracts/libraries/HyperCoreLib.sol
```diff
@@ -0,0 +1,366 @@
...
+     * @param asset The asset index of the order
+     * @param isBuy Whether the order is a buy order
+     * @param limitPriceX1e8 The limit price of the order scaled by 1e8
+     * @param sizeX1e8 The size of the order scaled by 1e8
+     * @param reduceOnly If true, only reduce existing position rather than opening a new opposing order
+     * @param encodedTif Time-in-Force: 1 = ALO, 2 = GTC, 3 = IOC
+     * @param cloid The client order id of the order, 0 means no cloid
+     */
+    function submitLimitOrder(
+        uint32 asset,
```
fusmanii, line 170: is `asset` here different than the `erc20CoreIndex` you have in the functions above?

contracts/libraries/HyperCoreLib.sol
tbwebb22: so since the EVM and HyperCore typically have different decimals for the same token, when bridging across you can end up losing this dust amount that results from the decimal difference. This function takes the amount we want to send (EVM decimals, this is the pre-dusted amount) and removes that small amount that we would have lost from dust. So we send a little less, but don't lose any amount to dust.

contracts/libraries/HyperCoreLib.sol
tbwebb22: yes, `asset` refers to a spot market index which is different from the `erc20CoreIndex` 

https://hyperliquid.gitbook.io/hyperliquid-docs/for-developers/api/asset-ids

contracts/libraries/HyperCoreLib.sol
tbwebb22: good call, done!

contracts/libraries/HyperCoreLib.sol
tbwebb22: I like that, enum added

contracts/libraries/HyperCoreLib.sol
fusmanii: ok makes sense, and that would only be true in the case that evm decimal > core decimal

contracts/libraries/HyperCoreLib.sol
```diff
@@ -0,0 +1,374 @@
...
+        uint64 limitPriceX1e8,
+        uint64 sizeX1e8,
+        bool reduceOnly,
+        Tif tif,
+        uint128 cloid
+    ) internal {
+        // Basic sanity checks
+        if (limitPriceX1e8 == 0) revert LimitPxIsZero();
+        if (sizeX1e8 == 0) revert OrderSizeIsZero();
+        if (tif == Tif.None || uint8(tif) > uint8(Tif.IOC)) revert InvalidTif();
```
fusmanii: since `tif` will never be None lets remove it

also you can just do 
```suggestion
        if (uint8(tif) > uint8(type(Tif).max)) revert InvalidTif();
```

PR #1138, opened 2025-10-15
feat: add metadata event emitter contract

amateima:
This PR introduces a new `AcrossEventEmitter` contract that provides a dedicated mechanism for emitting on-chain metadata events. 

### Changes

`AcrossEventEmitter` is a simple contract for emitting arbitrary bytes-encoded metadata as on-chain events. The usecase behind this contract is to emit metadata without modifying Across core contracts. This will allow the indexer to enrich Across protocol data with additional info that cannot be grabbed easily from onchain events (e.g swaps performed as destination actions)

### DIFF
--- a/contracts/AcrossEventEmitter.sol
+++ b/contracts/AcrossEventEmitter.sol
@@ -1,13 +1,11 @@
 // SPDX-License-Identifier: BUSL-1.1
 pragma solidity ^0.8.0;
 
-import "@openzeppelin/contracts/security/ReentrancyGuard.sol";
-
 /**
  * @title AcrossEventEmitter
  * @notice A simple contract that emits events with bytes encoded metadata
  */
-contract AcrossEventEmitter is ReentrancyGuard {
+contract AcrossEventEmitter {
     /**
      * @notice Emitted when metadata is stored
      * @param data The metadata bytes emitted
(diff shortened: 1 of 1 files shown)


contracts/AcrossEventEmitter.sol
```diff
@@ -0,0 +1,40 @@
...
+    receive() external payable {
+        revert("Contract does not accept native token");
+    }
+
+    /**
+     * @notice Prevents native token from being sent to this contract via fallback
+     */
+    fallback() external payable {
+        revert("Contract doesn't accept native token");
+    }
```
mrice32: nit: why implement these functions at all?

contracts/AcrossEventEmitter.sol
```diff
@@ -0,0 +1,40 @@
...
+
+    /**
+     * @notice Emits metadata as an event
+     * @param data The bytes data to emit
+     */
+    function emitData(bytes calldata data) external nonReentrant {
+        require(data.length > 0, "Data cannot be empty");
+        require(data.length <= 2048, "Data too large");
```
mrice32: Why limit the size of the data?

Would this protect us from anything?

contracts/AcrossEventEmitter.sol
amateima: I considered would be good practice to not allow funds being sent by mistake, but I can remove this

contracts/AcrossEventEmitter.sol
amateima: I removed the upper limit [here](https://github.com/across-protocol/contracts/pull/1138/commits/e934edd75d6a7a97584d16d828be9428a7a66d38), but I'd keep the > 0 validation, so that the indexer would not index events with no metadata at all

contracts/AcrossEventEmitter.sol
mrice32: If you don't implement them, they'll revert when users try to call them, so it would be the same outcome.

contracts/AcrossEventEmitter.sol
mrice32: Sounds good!

contracts/AcrossEventEmitter.sol
```diff
@@ -0,0 +1,39 @@
...
+     */
+    fallback() external payable {
+        revert("Contract doesn't accept native token");
+    }
+
+    /**
+     * @notice Emits metadata as an event
+     * @param data The bytes data to emit
+     */
+    function emitData(bytes calldata data) external nonReentrant {
```
mrice32: Re-entrancy guards are generally best practice, but you don't need them if your contract isn't making any external calls (which is the case in this contract).

I would suggest removing to save a little gas.

PR #1139, opened 2025-10-16
feat: Hypercorelib - clean up decimal conversion functions

tbwebb22:
Cleans up functionality around decimal conversion and EVM <> Core bridging dust calculations
Adds spotPx function for getting spot market price


contracts/libraries/HyperCoreLib.sol
```diff
@@ -78,34 +74,28 @@ library HyperCoreLib {
...
         // if the transfer amount exceeds the bridge balance, this wil revert
-        HyperAssetAmount memory amounts = quoteHyperCoreAmount(
-            erc20CoreIndex,
-            decimalDiff,
-            toAssetBridgeAddress(erc20CoreIndex),
-            amountEVM
-        );
-
-        if (amounts.evm != 0) {
+        (uint256 _amountEVMToSend, uint64 _amountCoreToReceive) = maximumEVMSendAmountToAmounts(amountEVM, decimalDiff);
```
tbwebb22, line 88: do we even want to modify the EVM send amount here? or rely on the higher level contract to do this?

contracts/libraries/HyperCoreLib.sol
```diff
@@ -116,30 +106,23 @@ library HyperCoreLib {
...
-            decimalDiff,
-            toAssetBridgeAddress(erc20CoreIndex),
-            amountEVM
-        );
-
-        if (amounts.evm != 0) {
-            // Transfer the tokens to this contract's address on HyperCore
-            IERC20(erc20EVMAddress).safeTransfer(toAssetBridgeAddress(erc20CoreIndex), amounts.evm);
+    ) internal returns (uint256 amountEVMSent, uint64 amountCoreToReceive) {
+        (uint256 _amountEVMToSend, uint64 _amountCoreToReceive) = maximumEVMSendAmountToAmounts(amountEVM, decimalDiff);
```
tbwebb22, line 118: do we even want to modify the EVM send amount here? or rely on the higher level contract to do this?

grasphoper: I honestly would expect the higher level caller to always use the `maximumEVMSendAmountToAmounts` themselves when sending.

If they don't, they're risking making an error in accounting. And then the `maximumEVMSendAmountToAmounts` won't save them anyway?

Actually, if we do leave these in, we can assert the returned values against the actual amounts that we're sending. This costs a bit of gas and is kind of a "hidden execution flow", but I'm leaning we leave these here I guess.

PR #1141, opened 2025-10-20
feat: sponsored periphery audit

grasphoper:


contracts/periphery/mintburn/sponsored-cctp/SponsoredCCTPSrcPeriphery.sol
```diff
@@ -0,0 +1,79 @@
...
+        (
+            uint256 amount,
+            uint32 destinationDomain,
+            bytes32 mintRecipient,
+            address burnToken,
+            bytes32 destinationCaller,
+            uint256 maxFee,
+            uint32 minFinalityThreshold,
+            bytes memory hookData
+        ) = SponsoredCCTPQuoteLib.getDespoitForBurnData(quote);
```
nicholaspai: typo here `getDespoit`

contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
```diff
@@ -0,0 +1,1195 @@
...
+import { DonationBox } from "../../chain-adapters/DonationBox.sol";
+import { HyperCoreLib } from "../../libraries/HyperCoreLib.sol";
+import { CoreTokenInfo } from "./Structs.sol";
+import { FinalTokenInfo } from "./Structs.sol";
+import { SwapHandler } from "./SwapHandler.sol";
+import { BPS_DECIMALS, BPS_SCALAR } from "./Constants.sol";
+
+/**
+ * @title HyperCoreFlowExecutor
+ * @notice Contract handling HyperCore interactions for trasnfer-to-core or swap-with-core actions after stablecoin bridge transactions
```
nicholaspai:
```suggestion
 * @notice Contract handling HyperCore interactions for transfer-to-core or swap-with-core actions after stablecoin bridge transactions
```

contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
```diff
@@ -0,0 +1,1195 @@
...
+            finalToken,
+            amountInEVM,
+            totalEVMAmountToSponsor,
+            finalCoreSendAmount,
+            finalTokenInfo.assetIndex,
+            cloid
+        );
+    }
+
+    /// @notice Finalizes pending queue of swaps for `finalToken` if a corresponsing SwapHandler has enough balance
```
nicholaspai:
```suggestion
    /// @notice Finalizes pending queue of swaps for `finalToken` if a corresponding SwapHandler has enough balance
```

contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
```diff
@@ -0,0 +1,1195 @@
...
+
+            emit SwapFlowCompleted(
+                nonce,
+                pendingSwap.finalRecipient,
+                pendingSwap.finalToken,
+                pendingSwap.sponsoredCoreAmountPreFunded,
+                totalAmountToForwardToUser
+            );
+
+            availableCore -= totalAmountToForwardToUser;
```
nicholaspai: Isn't there reentrancy risk from the `transferFundsToUserOnCore` call() that appears above these internal updates to vars?

contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
```diff
@@ -0,0 +1,1195 @@
...
+     *      CONFIGURATION FUNCTIONS       *
+     **************************************/
+
+    /**
+     * @notice Set or update information for the token to use it in this contract
+     * @dev To be able to use the token in the swap flow, FinalTokenInfo has to be set as well
+     * @dev Setting core token info to incorrect values can lead to loss of funds. Should NEVER be unset while the
+     * finalTokenParams are not unset
+     */
+    function setCoreTokenInfo(
```
nicholaspai, line 255: Shouldn't all of the public/external functions in this contract, and maybe the PR in general be `nonReentrant`?

contracts/periphery/mintburn/sponsored-cctp/SponsoredCCTPSrcPeriphery.sol
mrice32: fixed: 335b7bd040c91b91d64df23ac242c8e1377f83fc

contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
mrice32: For this specific case, I don't think so.

The CoreWriter is a trusted precompile contract and only emits an event (no call-outs). Still may make sense to have re-entrancy guards, though. I'm thinking through that now.

contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
mrice32: Good question

grasphoper: With the arbitrary execution, there might be a possibility to execute two different-quoteNonce `receiveMessage` calls on the CCTP dst handler an reenter like that.

For OFT track, `lzCompose` also might be reenterable if there's a failed `lzCompose` that can be retried.

So `lzCompose` / `receiveMessage` would be nice to add `nonReentrant` on for sure just for peace of mind.

Not sure if it's strictly required on e.g. admin functions

contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
```diff
@@ -0,0 +1,1195 @@
...
+        uint256 extraBridgingFeesEVM,
+        address finalToken
+    ) internal virtual {
+        CoreTokenInfo storage coreTokenInfo = coreTokenInfos[finalToken];
+
+        bool isSponsoredFlow = maxBpsToSponsor > 0;
+        bool userHasNoCoreAccount = !HyperCoreLib.coreUserExists(finalRecipient);
+        if (userHasNoCoreAccount) {
+            if (isSponsoredFlow) {
+                revert AccountNotActivated(finalRecipient);
```
tbwebb22: I assume this reverts because we're expecting the permissioned bot to have called `activateUserAccount` (if the user doesn't exist yet) before we get into this flow? and same for the swap flow?

contracts/periphery/mintburn/ArbitraryEVMFlowExecutor.sol
```diff
@@ -0,0 +1,239 @@
...
+        MulticallHandler(payable(multicallHandler)).handleV3AcrossMessage(
+            initialToken,
+            amount,
+            address(this),
+            instructions
+        );
+
+        uint256 finalAmount;
+
+        // This means the swap (if one was intended) didn't happen (action failed), so we use the initial token as the final token.
```
nicholaspai, line 83: Is this internal function only intended to be used for swaps?  I think maybe this comment is misleading as many types of arbitrary actions could lead to the initialBalance being unchanged after calling `handleV3AcrossMessage`

contracts/periphery/mintburn/sponsored-cctp/SponsoredCCTPSrcPeriphery.sol
```diff
@@ -0,0 +1,79 @@
...
+            bytes32 destinationCaller,
+            uint256 maxFee,
+            uint32 minFinalityThreshold,
+            bytes memory hookData
+        ) = SponsoredCCTPQuoteLib.getDepositForBurnData(quote);
+
+        IERC20(burnToken).safeTransferFrom(msg.sender, address(this), amount);
+        IERC20(burnToken).forceApprove(address(cctpTokenMessenger), amount);
+
+        usedNonces[quote.nonce] = true;
```
nicholaspai: Should you set this above the external `burnToken` calls?

fusmanii: done https://github.com/across-protocol/contracts/pull/1141/commits/5de7a4bc0dcd153ed47e68b76293f96f032ccad3

contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
nicholaspai: I think we should add it for peace of mind where possible, and then in external functions we don't do that, we need to follow the Check-Effects-interaction pattern which we don't seem to always be doing? This applies to all contracts in this PR

contracts/periphery/mintburn/ArbitraryActionFlowExecutor.sol
```diff
@@ -0,0 +1,239 @@
...
+        });
+
+        return abi.encode(instructions);
+    }
+
+    /**
+     * @notice Execute simple transfer flow to HyperCore with the final token
+     * @dev Must be implemented by contracts that inherit from this contract
+     */
+    function _executeSimpleTransferFlow(
```
nicholaspai: Why did you decide to keep `_executeSimpleTransferFlow` and `_fallbackHyperEVMFlow` in this contract if the child contracts `SponsoredCCTPDstPeriphery` and `DstOFTHandler` both just implement the same simple call to HyperCoreFlowExecutor?

Could you instead implement his virtual function in this contract and call HyperFlowExecutor library functions directly?

contracts/periphery/mintburn/ArbitraryActionFlowExecutor.sol
```diff
@@ -0,0 +1,239 @@
...
+import { BPS_SCALAR } from "./Constants.sol";
+
+/**
+ * @title ArbitraryActionFlowExecutor
+ * @notice Base contract for executing arbitrary action sequences using MulticallHandler
+ * @dev This contract provides shared functionality for both OFT and CCTP handlers to execute
+ * arbitrary actions on HyperEVM via MulticallHandler, with optional transfer to HyperCore.
+ * @custom:security-contact bugs@across.to
+ */
+abstract contract ArbitraryActionFlowExecutor {
```
nicholaspai: It took me a little time to understand the difference between this contract and HyperCoreFlowExecutor and I think it boils down to that HyperCoreFlowExecutor executes actions on HyperCore while this contract executes actions on HyperEVM, right?

Would it be clearer to rename this contract to HyperEvmFlowExecutor? I understand you might have considered this but chose not to do so because this theoretically could be used for other EVM chains too. But then again, there are specific references to HyperEVM in this contract.

contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
```diff
@@ -0,0 +1,1195 @@
...
+        swapHandler.submitLimitOrder(finalTokenInfo, limitPriceX1e8, sizeX1e8, cloid);
+
+        pendingSwaps[quoteNonce] = PendingSwap({
+            finalRecipient: finalRecipient,
+            finalToken: finalToken,
+            minCoreAmountFromLO: guaranteedLOOut,
+            sponsoredCoreAmountPreFunded: totalCoreAmountToSponsor,
+            limitOrderCloid: cloid
+        });
+        pendingQueue[finalToken].push(quoteNonce);
```
nicholaspai: you're setting state variables here after you make an external call `submitLimitOrder`, is that a potential reentrancy vector?

contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
nicholaspai: to be safe, why can't you move this above the external call?

contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
```diff
@@ -0,0 +1,1195 @@
...
+        // Handler @ evm -> Handler @ core -> finalRecipient @ core
+        HyperCoreLib.transferERC20EVMToCore(
+            fundingToken,
+            coreTokenInfo.coreIndex,
+            finalRecipient,
+            activationFeeEvm,
+            coreTokenInfo.tokenInfo.evmExtraWeiDecimals
+        );
+
+        cumulativeSponsoredActivationFee[fundingToken] += activationFeeEvm;
```
nicholaspai: this is more of a nit since this is a permissioned function but again you should probably follow CEI pattern here

contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
grasphoper: Exactly! Now that I think about it, this functionality can live in a common place, like `_executeFlow`, and we may put a comment there as well wrt this

contracts/periphery/mintburn/ArbitraryActionFlowExecutor.sol
grasphoper: > Could you instead implement his virtual function in this contract and call HyperFlowExecutor library functions directly?

I don't 100% understand the solution you're proposing.

I think one option for us would be to return `(finalToken, finalAmount)` from `ArbitraryActionFlowExecutor` and have the Child Handler contract take these values and pass directly to HyperCoreFlowExecutor as part of these flows.

This would disentangle `ArbitraryActionFlowExecutor` from the HyperCoreFlowExecutor-specific functions.

Wdyt @nicholaspai @mrice32 ?

contracts/periphery/mintburn/ArbitraryActionFlowExecutor.sol
grasphoper: I think if we do https://github.com/across-protocol/contracts/pull/1141#discussion_r2449091611 then we might avoid entangling this contract to the HyperCoreFlowExecutor.

`_getDonationBox` we could also remove and just pass into the `_executeArbitraryActionFlow` function perhaps?

Overall not attached to the naming. I think we just chose whatever seemed appropriate in the limited time that we had :D

If we disentangle fully though, I think it can remain the `ArbitraryActionFlowExecutor`?

contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
grasphoper: Sounds good. `cumulativeSponsoredActivationFee` is not something this contract relies on, more of a "data-for-the-outside-observer" type function, but it's good to follow the pattern anyway I think

contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
grasphoper: The actions called here before the state changes take effect are:
2 X ERC20 safeTransfer calls
1 X HyperLiquid precompile call to submit the limit order

Those should be safe, but I think we can easily move these state changes to be above the external calls

contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
grasphoper: To fully move this, we'd need to split it up into two loops then: one that's calculating the `totalAmountToForwardToUser` per user, and then the next that moves the funds right?

contracts/periphery/mintburn/ArbitraryActionFlowExecutor.sol
grasphoper: PTAL https://github.com/across-protocol/contracts/pull/1149

contracts/periphery/mintburn/ArbitraryActionFlowExecutor.sol
grasphoper: PTAL https://github.com/across-protocol/contracts/pull/1149

contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
grasphoper: PTAL https://github.com/across-protocol/contracts/pull/1148

contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
grasphoper: PTAL https://github.com/across-protocol/contracts/pull/1148

contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
nicholaspai: I don't think its ever safe to assume safeTransfers are safe. It could be called with any ERC20 which could be programmed maliciously

PR #1143, opened 2025-10-20
feat: add arbitrary actions execution to sponsored bridging

mrice32:


contracts/periphery/mintburn/sponsored-oft/ComposeMsgCodec.sol
```diff
@@ -10,17 +10,31 @@ library ComposeMsgCodec {
     uint256 internal constant MAX_USER_SLIPPAGE_BPS_OFFSET = 96;
     uint256 internal constant FINAL_RECIPIENT_OFFSET = 128;
     uint256 internal constant FINAL_TOKEN_OFFSET = 160;
-    uint256 internal constant COMPOSE_MSG_BYTE_LENGTH = 192;
+    uint256 internal constant EXECUTION_MODE_OFFSET = 192;
+    // Minimum length without actionData (192 bytes + 32 bytes for executionMode + 32 bytes for actionData length)
+    uint256 internal constant MIN_COMPOSE_MSG_BYTE_LENGTH = 192;
```
grasphoper: Is it supposed to be 192 + 32? Because the extra uint8 will at least be packed as 32 bytes?

PR #1145, opened 2025-10-21
feat: add swapsProcessed return values to finalizePendingsSwaps

tbwebb22:
Adds `swapsProcessed` and `totalSwapsProcessed` as return vars to `finalizePendingSwaps`

### DIFF
--- a/contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
+++ b/contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
@@ -703,7 +703,7 @@ contract HyperCoreFlowExecutor is AccessControl {
     /// @notice Finalizes pending queue of swaps for `finalToken` if a corresponding SwapHandler has enough balance
     function finalizePendingSwaps(
         address finalToken,
-        uint256 maxToProcess
+        uint256 maxSwapCountToFinalize
     ) external returns (uint256 finalizedSwapsCount, uint256 finalizedSwapsAmount, uint256 totalPendingSwapsRemaining) {
         FinalTokenInfo memory finalTokenInfo = _getExistingFinalTokenInfo(finalToken);
         CoreTokenInfo memory coreTokenInfo = coreTokenInfos[finalToken];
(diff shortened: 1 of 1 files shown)


contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
```diff
@@ -701,15 +701,18 @@ contract HyperCoreFlowExecutor is AccessControl {
     }
 
     /// @notice Finalizes pending queue of swaps for `finalToken` if a corresponding SwapHandler has enough balance
-    function finalizePendingSwaps(address finalToken, uint256 maxToProcess) external {
+    function finalizePendingSwaps(
+        address finalToken,
+        uint256 maxToProcess
+    ) external returns (uint256 swapsProcessed, uint256 totalSwapsProcessed) {
```
grasphoper: There seems to be 2 types of terminology we use in this function:
the name of the fn is `finalize..` and the name of internal var is `processed`

I'm leaning that we stick with the `finalizePendingSwaps` name for the fn, and change return var names to `..Finalized`. We could also change the local `processed` fn name to finalized (although maybe not because it increases the diff from the "audit commit")

Also, from the POV of the bot who's calling this, the bot may care about both the number  of processed and the amount (in tokens) of them. For example, the bot may want to finalize only if the total amount in simulated finalization is e.g. more than $1000 or something.

Also, from the POV of the bot `totalSwapsProcessed` doesn't tell it much (current-decision-wise). E.g. it probably won't have any logic branches based on `totalSwapsProcessed`(I think?)

What could be useful to return is something like `totalPendingSwapsRemaning`.
Then a bot e.g. could reason: 

if the totalAmountFinalized is little, but if it brings our queue down to zero, then I'll finalize anyway.

Wdyt?

tbwebb22: I'm in favor of changing return vars and local vars to `finalize..` for consistency. Doesn't change the audit diff all that much as its just renamings

Agree that `totalSwapsProcessed` isn't that useful for the bot. But `swapAmountsFinalized` and `pendingSwapsRemaining` definitely would be. I can update to this

PR #1153, opened 2025-10-22
improve: Move BytesLib to external folder

fusmanii:


tbwebb22 (approved):

contracts/external/libraries/BytesLib.sol
```diff
@@ -6,21 +6,25 @@ library BytesLib {
...
-    error InvalidBytes();
-    error InvalidStart();
 
     /**************************************
      *              FUNCTIONS              *
      **************************************/
 
+    // The following 4 functions are copied from solidity-bytes-utils library
+    // https://github.com/GNSPS/solidity-bytes-utils/blob/fc502455bb2a7e26a743378df042612dd50d1eb9/contracts/BytesLib.sol#L323C5-L398C6
+    // Code was copied, and slightly modified to use revert instead of require
```
grasphoper, line 16: Don't we want to just include unmodified versions?
Or is the thinking that those were not audited anyway so there's no difference?

fusmanii: Yeah that was my thinking, since its not audited might as well improve on error handling

grasphoper (approved):

PR #1156, opened 2025-10-22
chore: Update solidity and OZ versions

fusmanii:


grasphoper:
@fusmanii a few questions:
- do you think we need to add new remappings to the foundry.toml for this?
- do you think we add a new foundry profile, and have 0.8.23 be the default (just like the default for hardhat), and 0.8.25 would be the ~periphery profile
- our CI still uses hardhat to compile all contracts, right? Do you think we need to add some custom config to `hardhat.config.ts` for periphery-specific files that would tell hardhat to compile with this same solidity version?

fusmanii:
- No need to add remapping in foundry.toml as the we differentiate between v4 vs v5 by using `@openzeppelin/contracts` vs `@openzeppelin/contracts-v5` which the current mapping already handles
- I was thinking of upgrading everything to 0.8.25, all contract are set to ^0.8.0 which encompasses 0.8.25. Do you think there are any risks?
- Yes I can update hardhat config. I hope was to do the same there and upgrade everything to 0.8.25 (apart from a few contracts that are pinned to 0.8.19)

wdyt?

grasphoper:
>I was thinking of upgrading everything to 0.8.25, all contract are set to ^0.8.0 which encompasses 0.8.25. Do you think there are any risks?

Hmm, sure I think we're fine to do it. Maybe worth reading release notes at least though for .24, .25

Sg about the remappings

PR #1168, opened 2025-10-27
fix: contract sizes

grasphoper:
Make `HyperCoreFlowExecutor` into a `delegatecall`able library contract that also has some state management by utilizing 7201 storage slots.

### DIFF
--- a/contracts/periphery/mintburn/BaseModuleHandler.sol
+++ b/contracts/periphery/mintburn/BaseModuleHandler.sol
@@ -31,8 +31,8 @@ abstract contract BaseModuleHandler is
 
     /// @notice Fallback function to proxy all calls to the HyperCore module via delegatecall
     /// @dev Permissioning is enforced by the delegated function's own modifiers (e.g. onlyPermissionedBot)
-    fallback() external payable nonReentrant {
-        _delegateToHyperCore(msg.data);
+    fallback(bytes calldata data) external payable nonReentrant returns (bytes memory) {
+        return _delegateToHyperCore(data);
     }
 
     /// @notice Internal delegatecall helper


contracts/periphery/mintburn/sponsored-cctp/SponsoredCCTPDstPeriphery.sol
```diff
@@ -46,24 +59,32 @@ contract SponsoredCCTPDstPeriphery is SponsoredCCTPInterface, HyperCoreFlowExecu
         address _donationBox,
         address _baseToken,
         address _multicallHandler
-    ) HyperCoreFlowExecutor(_donationBox, _baseToken) ArbitraryEVMFlowExecutor(_multicallHandler) {
+    )
+        BaseHyperCoreModuleHandler(_donationBox, _baseToken, DEFAULT_ADMIN_ROLE)
+        ArbitraryEVMFlowExecutor(_multicallHandler)
+    {
+        baseToken = _baseToken;
+        hyperCoreModule = address(new HyperCoreFlowExecutor(_donationBox, _baseToken));
```
fusmanii:
```suggestion
```
Is this still needed because you have the same line in `BaseModuleHandler`?

contracts/periphery/mintburn/AuthorizedFundedFlow.sol
```diff
@@ -0,0 +1,35 @@
...
+        bytes32 slot = AUTHORIZED_FUNDED_FLOW_SLOT;
+        bool authorized;
+        assembly {
+            authorized := sload(slot)
+        }
+        if (!authorized) {
+            revert FundedFlowNotAuthorized();
+        }
+        _;
+    }
```
fusmanii, line 34: Is this needed so that `HyperCoreFlowExecutor` can't be called directly?

contracts/periphery/mintburn/BaseModuleHandler.sol
```diff
@@ -0,0 +1,43 @@
...
+    constructor(address _donationBox, address _baseToken, bytes32 _roleAdmin) {
+        hyperCoreModule = address(new HyperCoreFlowExecutor(_donationBox, _baseToken));
+
+        _setRoleAdmin(PERMISSIONED_BOT_ROLE, _roleAdmin);
+        _setRoleAdmin(FUNDS_SWEEPER_ROLE, _roleAdmin);
+    }
+
+    /// @notice External delegatecall entrypoint to the HyperCore module
+    /// @dev Permissioning is enforced by the delegated function's own modifiers (e.g. onlyPermissionedBot)
+    function callHyperCoreModule(bytes calldata data) external payable nonReentrant returns (bytes memory) {
```
mrice32: OOC, why not make this the fallback function? My thinking is that this would simplify things for an offchain bot, as it would mostly look like a proxy.

contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
```diff
@@ -184,35 +193,30 @@ contract HyperCoreFlowExecutor is AccessControl, Lockable {
...
+        _getExistingCoreTokenInfo(evmTokenAddress);
         _;
     }
 
-    modifier onlyExistingCoreToken(address evmTokenAddress) {
-        _getExistingCoreTokenInfo(evmTokenAddress);
+    modifier onlyHandlerDelegatecall() {
+        if (address(this) != handler) {
+            revert NotHandler();
+        }
```
mrice32: Why is this necessary?

contracts/periphery/mintburn/BaseModuleHandler.sol
```diff
@@ -0,0 +1,43 @@
...
+
+import { AuthorizedFundedFlow } from "./AuthorizedFundedFlow.sol";
+import { HyperCoreFlowExecutor } from "./HyperCoreFlowExecutor.sol";
+import { HyperCoreFlowRoles } from "./HyperCoreFlowRoles.sol";
+
+import { AccessControl } from "@openzeppelin/contracts/access/AccessControl.sol";
+import { ReentrancyGuard } from "@openzeppelin/contracts/security/ReentrancyGuard.sol";
+
+/**
+ * @dev IMPORTANT. The storage layout of this contract is meant to be in sync with `HyperCoreFlowExeutor` in terms of
```
mrice32: Can we use https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/master/contracts/access/AccessControlUpgradeable.sol to avoid this requirement?

contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
grasphoper: I guess if not the handler is calling, then it's not handler's money that's being sent away. Yeah, I think we can remove this

contracts/periphery/mintburn/BaseModuleHandler.sol
grasphoper: Sure, I was gonna see how the proxy contract does this, but didn't get to it. 100% on board for making it look more like a proxy

contracts/periphery/mintburn/sponsored-cctp/SponsoredCCTPDstPeriphery.sol
grasphoper: Agree, this is not needed

contracts/periphery/mintburn/AuthorizedFundedFlow.sol
grasphoper: It's a way to authorize the codepath that's being executed. `executeMessage` function checks sigs + mints tokens. `_executeFlow` and similar funcitons expects to only be called in that trusted context. E.g. if you were able to call it outside of the `executeMessage`(or `lzCompose`) codepath, you'd be able to just send away money at will

contracts/periphery/mintburn/BaseModuleHandler.sol
grasphoper: 100%, that'd be great

contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
```diff
@@ -29,25 +32,14 @@ contract HyperCoreFlowExecutor is AccessControl, Lockable {
...
-    mapping(address => FinalTokenInfo) public finalTokenInfos;
-
     /// @notice All operations performed in this contract are relative to this baseToken
     address public immutable baseToken;
 
-    /// @notice The block number of the last funds pull action per final token: either as a part of finalizing pending swaps,
-    /// or an admin funds pull
-    mapping(address finalToken => uint256 lastPullFundsBlock) public lastPullFundsBlock;
+    /// @notice Handler that's delegatecalling into this contract
+    address public immutable handler;
```
tbwebb22: is this necessary? seems like its not being used since we removed the `onlyHandlerDelegatecall()` modifier

mrice32: Good point.

remappings.txt
```diff
@@ -0,0 +1,2 @@
+@openzeppelin/=node_modules/@openzeppelin/
+@openzeppelin/contracts/access/IAccessControl.sol=node_modules/@openzeppelin/contracts-v5/access/IAccessControl.sol
```
grasphoper: Don't we want to change remappings in `foundry.toml` and use `forge remappings > remappings.txt`?

Curious that this works

mrice32: Oh yep, didn't even realize we had that. Will move.

mrice32: Fixed.

contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
mrice32: Fixed! Good callout!

PR #1170, opened 2025-10-28
fix(svm): use native svm inputs for sponsored cctp bridge

Reinis-FRP:
Using Solana native input types allows increasing `actionData` length in sponsored CCTP bridge.

### DIFF
--- a/programs/sponsored-cctp-src-periphery/src/utils/sponsored_cctp_quote.rs
+++ b/programs/sponsored-cctp-src-periphery/src/utils/sponsored_cctp_quote.rs
@@ -1,105 +1,30 @@
 use anchor_lang::{prelude::*, solana_program::keccak};
 
-use crate::error::{DataDecodingError, SvmError};
-
-// Macro to define the SponsoredCCTPQuote fields as an enum with associated constants for ordinal, start, end, count,
-// and minimum total bytes.
-macro_rules! define_quote_fields {
-    ($($V:ident),+ $(,)?) => {
-        #[derive(Copy, Clone)]
-        pub enum SponsoredCCTPQuoteFields { $($V),+ }
-
-        impl SponsoredCCTPQuoteFields {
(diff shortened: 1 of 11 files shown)


programs/sponsored-cctp-src-periphery/src/utils/sponsored_cctp_quote.rs
```diff
@@ -117,171 +42,96 @@ impl<'a> SponsoredCCTPQuote<'a> {
...
 
-    pub fn source_domain(&self) -> Result<u32> {
-        Self::decode_to_u32(self.get_field_word(SponsoredCCTPQuoteFields::SourceDomain))
-    }
-
-    pub fn destination_domain(&self) -> Result<u32> {
-        Self::decode_to_u32(self.get_field_word(SponsoredCCTPQuoteFields::DestinationDomain))
-    }
+    pub fn encode_hook_data(&self) -> Vec<u8> {
+        // ABI encoded hookData on EVM holds 7 static 32-byte words followed by the actionData offset that points to the
```
md0x, line 82: Comment says “7 static words,” but code uses 8 * 32 (including executionMode)

is this correct?

md0x (approved):
This looks solid. Maybe we could add tests to verify hash parity between EVM and SVM (especially with different executionMode values and actionData sizes) and to compare ethers abi encoding with encode_hook_data to confirm offsets/padding. Otherwise, LGTM.

programs/sponsored-cctp-src-periphery/src/utils/sponsored_cctp_quote.rs
Reinis-FRP: yes, actionData offset is positioned as the 8th word after the prior 7 static fields. as offsets start at 0 then `action_data_offset` value should be 8 * 32 pointing to the following word that holds the length.

PR #1171, opened 2025-10-28
add executeExternalCall function

tbwebb22:
This PR adds `onlyAdmin` function `executeExternal` to `SpokePool`. This function does an external call to the decoded target address.

### DIFF
--- a/contracts/SpokePool.sol
+++ b/contracts/SpokePool.sol
@@ -202,6 +202,9 @@ abstract contract SpokePool is
     event PausedFills(bool isPaused);
     event SetOFTMessenger(address indexed token, address indexed messenger);
 
+    /// @notice Emitted when the call to external contract is executed, triggered by an admin action
+    event AdminExternalCallExecuted(address indexed target, bytes data);
+
     error OFTTokenMismatch();
     /// @notice Thrown when the native fee sent by the caller is insufficient to cover the OFT transfer.
     error OFTFeeUnderpaid();
@@ -388,6 +391,7 @@ abstract contract SpokePool is
         (success, returnData) = target.call(data);
 
         if (!success) revert ExternalCallExecutionFailed();
+        emit AdminExternalCallExecuted(target, data);
     }
 
     /**************************************


contracts/SpokePool.sol
```diff
@@ -370,6 +370,30 @@ abstract contract SpokePool is
         _setOftMessenger(token, messenger);
     }
 
+    /**
+     * @notice Execute custom actions encoded in the message.
+     * @dev If target address is this contract, a delegatecall to this contract is executed.
+     * @dev If target address is not this contract, an external call to the target address is executed.
+     * @param message The message containing the target and custom actions to execute.
+     */
+    function executeCustomActions(bytes calldata message) external onlyAdmin {
```
nicholaspai: this should be nonReentrant, right? would we ever want to reenter?

contracts/SpokePool.sol
```diff
@@ -370,6 +370,30 @@ abstract contract SpokePool is
...
+        (address target, bytes memory data) = abi.decode(message, (address, bytes));
+
+        if (target == address(0)) revert ZeroAddressTarget();
+        if (data.length < 4) revert MessageTooShort(); // need at least a selector
+
+        bool success;
+        if (target == address(this)) {
```
nicholaspai: Hm... I'm not sure we'd ever want to delegatecall back to itself this seems potentially dangerous. Is there a use case?

nicholaspai (commented):
Left a question about delegatecall-ing self

contracts/SpokePool.sol
```diff
@@ -370,6 +370,30 @@ abstract contract SpokePool is
...
+            // delegatecall to self
+            (success, ) = address(this).delegatecall(data);
+        } else {
+            // external call to target
+            (success, ) = target.call(data);
+        }
+
+        if (!success) revert CustomActionExecutionFailed();
```
fusmanii: You think it makes sense to return the return result of the `call` from above?

nicholaspai: i think that's a good idea

contracts/SpokePool.sol
tbwebb22: My understanding was that this was desired for future proofing - giving SpokePool the ability to call any of its own public or external functions on itself. 

However, looking at the `HubPool`, `relaySpokePoolAdminFunction` already lets us pass in arbitrary function data, so maybe there isn't really a use case for this.

@grasphoper I know we talked about this pattern, is there a use case for it?

contracts/SpokePool.sol
tbwebb22: yeah if we want to keep the self delegatecall functionality, we would have to leave the nonReentrant modifier off - since the other external onlyAdmin functions on the contract are nonReentrant, we wouldn't be able to delegatcall into those functions from this one. But if we decide to remove the self delegatecall, we could definitely make this function nonReentrant

contracts/SpokePool.sol
tbwebb22: oh yeah I think we should, I'll update that

tbwebb22: done

contracts/SpokePool.sol
nicholaspai: yeah I don't see a need for this personally, the relaySpokePoolAdminFunction should be used to call functions on this spoke pool generally

contracts/SpokePool.sol
nicholaspai: You could still add a custom reentrancy protection logic to the non self delegatecall path though

contracts/SpokePool.sol
mrice32: I think this was a (bad) suggestion that I made.

I can't see how this allows us to do anything we couldn't already do, so let's remove.

tbwebb22: Removed the self delegatecall, and also renamed this function to `executeExternalCall`

contracts/SpokePool.sol
tbwebb22: Added the `nonReentrant` modifier since we decided to remove the self delegatecall

nicholaspai (dismissed):

fusmanii (dismissed):

grasphoper (dismissed):

PR #1174, opened 2025-10-29
fix: Limit order finalization call

fusmanii:
During the finalization of limit orders, funds are on the swap handler on the Core side. So we need to send the funds to user on from the swap handler instead of the periphery

### DIFF
--- a/contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
+++ b/contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
@@ -667,7 +667,11 @@ contract HyperCoreFlowExecutor is AccessControl, Lockable {
             finalCoreTokenInfo.tokenInfo.evmExtraWeiDecimals
         );
 
-        HyperCoreLib.transferERC20CoreToCore(finalCoreTokenInfo.coreIndex, swap.finalRecipient, totalToSend);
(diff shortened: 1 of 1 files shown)


contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
```diff
@@ -667,7 +667,11 @@ contract HyperCoreFlowExecutor is AccessControl, Lockable {
             finalCoreTokenInfo.tokenInfo.evmExtraWeiDecimals
         );
 
-        HyperCoreLib.transferERC20CoreToCore(finalCoreTokenInfo.coreIndex, swap.finalRecipient, totalToSend);
+        finalTokenInfos[swap.finalToken].swapHandler.transferFundsToUserOnCore(
+            finalCoreTokenInfo.coreIndex,
+            swap.finalRecipient,
+            totalToSend
+        );
```
grasphoper, line 674: Might be better to pass `swapHandler` as arg from the caller fn (`finalizeSwapFlows`), as that already loads `finalTokenInfo` into `memory`

grasphoper: Oh, also it would then only be loaded once once per `finalizeSwapFlows`, while now we would be loading from storage once per `_finalizeSingleSwap` (unless compiler optimizations fix this?)

PR #1175, opened 2025-11-02
feat: OpenZeppelin V5 upgrade

fusmanii:


foundry.toml
```diff
@@ -28,11 +28,9 @@ remappings = [
 ]
 via_ir = true
 optimizer_runs = 800
-solc_version = "0.8.23"
+auto_detect_solc = true
```
fusmanii: Since there are different contract with different versions (uma contract are on 0.8.16), need to just let foundry figure out the solidity version (foundry doesn't allow to configure multiple versions like hardhat)

hardhat.config.ts
```diff
@@ -92,18 +92,7 @@ const config: HardhatUserConfig = {
...
-        // NOTE: Linea only supports 0.8.19.
-        // See https://docs.linea.build/build-on-linea/ethereum-differences#evm-opcodes
-        version: "0.8.19",
-      },
-      "contracts/SpokePoolVerifier.sol": {
-        ...DEFAULT_CONTRACT_COMPILER_SETTINGS,
-        // NOTE: Linea only supports 0.8.19.
-        // See https://docs.linea.build/build-on-linea/ethereum-differences#evm-opcodes
-        version: "0.8.19",
-      },
```
fusmanii, line 106: Linea upgraded all the way to Prague in [Oct](https://docs.linea.build/release-notes#beta-v40)

nicholaspai (approved):

contracts/external/uma/core/contracts/common/implementation/ExpandedERC20.sol
bmzig, line 1: Do you know why these were generated?

fusmanii: these are all the contracts from `@uma/core` that were being used and since they are using OZ v4, I needed to copy them over and have it point to `@openzeppelin/contracts-v4`

tbwebb22 (approved):

grasphoper (approved):

PR #1181, opened 2025-11-17
feat: Add simple Hypercore Depositor handler

nicholaspai:
An alternative to the MulticallHandler designed to support permissioned accounts to deposit ERC20's from hyperevm into an end user's hypercore account

### DIFF
--- a/contracts/handlers/HypercoreDepositorHandler.sol
+++ b/contracts/handlers/HypercoreDepositorHandler.sol
@@ -9,23 +9,25 @@ import { CoreWriterLib, PrecompileLib } from "hyper-evm-lib/src/CoreWriterLib.so
 import { HLConversions } from "hyper-evm-lib/src/common/HLConversions.sol";
 
 /**
- * @title Allows caller to bridge tokens from HyperEVM to Hypercore and send them to the end user's account
+ * @title Allows caller to bridge tokens from HyperEVM to Hypercore and send them to an end user's account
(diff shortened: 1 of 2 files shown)


bmzig (dismissed):

contracts/handlers/HypercoreDepositorHandler.sol
```diff
@@ -0,0 +1,69 @@
...
+contract HypercoreDepositorHandler is AcrossMessageHandler, ReentrancyGuard {
+    using SafeERC20 for IERC20;
+
+    /**
+     * @notice Bridges tokens from HyperEVM to Hypercore and sends them to the end user's account on Hypercore.
+     * @param token The address of the token to deposit.
+     * @param amount The amount of tokens on HyperEVM to deposit.
+     * @param user The address of the user on Hypercore to send the tokens to.
+     */
+    function depositToHypercore(address token, uint256 amount, address user) external nonReentrant {
```
mrice32, line 29: This assumes that the tokens are dropped on the contract atomically, right?

What are the downsides of doing a `transferFrom` from `msg.sender` to pull the tokens in?

nicholaspai: that's a great call, I overlooked this flow. We definitely should pull funds from the msg.sender in `depositToHypercore`, unlike in `handleAcrossV3Message`

PR #1190, opened 2025-11-25
feat: Update ZkSync_SpokePool to support new L2 ZkStack L2AssetRouter

nicholaspai:
This contract is designed to be used with ZkStack chains that migrate to the new [ZkGateway](https://forum.zknation.io/t/zip-6-prepare-zksync-for-zk-gateway/490#p-1247-changes-to-bridging-architecture-9) system. For example, this will be required for ZkSync (chain ID 324) and Lens.

~I don't know yet whether currently supported  ZkStack chains like Lens will use this new architecture or the current one. Therefore, to deploy hot fixes to Lens we would need to checkout an older commit and deploy that version of the ZkSync_SpokePool.~

~The alternative is introducing a new contract, but I'd like to consolidate changes.~

Lens will be upgraded with ZkSync, according to the ZkSync team.

### DIFF
--- a/foundry.toml
+++ b/foundry.toml
@@ -25,7 +25,7 @@ src = "contracts/Lens_SpokePool.sol"
 compile = true
 fallback_oz = true
 mode = '3'
-zksolc = "1.5.7"
+zksolc = "1.5.15"
 
 [rpc_endpoints]
 arbitrum = "${NODE_URL_42161}"


contracts/ZkSync_SpokePool.sol
fusmanii, line 216: does this need to be updated given that new vars are added?

fusmanii (dismissed):
looks good, just one q

contracts/ZkSync_SpokePool.sol
nicholaspai: no because the new vars are constants or immutable so they don't add any storage slots

contracts/ZkSync_SpokePool.sol
```diff
@@ -31,23 +37,45 @@ contract ZkSync_SpokePool is SpokePool, CircleCCTPAdapter {
...
     ZkBridgeLike public immutable zkUSDCBridge;
 
+    /// @dev The offset from which the built-in, but user space contracts are located.
+    /// Source: https://github.com/matter-labs/era-contracts/blob/48e189814aabb43964ed29817a7f05aa36f09fd6/l1-contracts/contracts/common/l2-helpers/L2ContractAddresses.sol#L12C1-L13C58
+    uint160 constant USER_CONTRACTS_OFFSET = 0x10000; // 2^16
+
+    /// Contract used to withdraw ERC20's to L1.
+    /// Source: https://github.com/matter-labs/era-contracts/blob/48e189814aabb43964ed29817a7f05aa36f09fd6/l1-contracts/contracts/common/l2-helpers/L2ContractAddresses.sol#L68
+    address public constant L2_ASSET_ROUTER_ADDR = address(USER_CONTRACTS_OFFSET + 0x03);
+    IL2AssetRouter public constant l2AssetRouter = IL2AssetRouter(L2_ASSET_ROUTER_ADDR);
```
grasphoper, line 54: `zkErc20Bridge` was mutable. Is `l2AssetRouter` expected to stay constant forever?

contracts/ZkSync_SpokePool.sol
```diff
@@ -31,23 +37,45 @@ contract ZkSync_SpokePool is SpokePool, CircleCCTPAdapter {
...
+    uint256 public immutable l1ChainId;
+
     event SetZkBridge(address indexed erc20Bridge, address indexed oldErc20Bridge);
 
     error InvalidBridgeConfig();
 
     /**
      * @notice Constructor.
-     * @notice Circle bridged & native USDC are optionally supported via configuration, but are mutually exclusive.
+     * @param _zkUSDCBridge Legacy bridge used to withdraw USDC to L1, for withdrawing all other ERC20's we use `l2AssetRouter`.
```
grasphoper, line 72: Will `_zkUSDCBridge` have to go in the near future as well? In particular, you're referring to it as "legacy": is there a reason?

contracts/ZkSync_SpokePool.sol
nicholaspai: That's kind of what the matter-labs code suggests: https://github.com/matter-labs/era-contracts/blob/48e189814aabb43964ed29817a7f05aa36f09fd6/l1-contracts/contracts/common/l2-helpers/L2ContractAddresses.sol#L68

contracts/ZkSync_SpokePool.sol
nicholaspai: Its legacy because I believe from the zksync team that this will be a special case for a token where we continue to use the deprecated L2ERC20Gateway contract.

hardhat.config.ts
grasphoper, line 1: @fusmanii what's our current plan for deploying elastic chain (zksync-like) contracts? Do we need any config changes to foundry or?

fusmanii: we have elastic chain deployment set up already, just need to install [foundryup-zksync](https://foundry-book.zksync.io/introduction/installation/) and it will pick up [this](https://github.com/across-protocol/contracts/blob/master/foundry.toml#L24) config

PR #1199, opened 2025-11-27
feat: add `updatevkey` task

grasphoper:
A task that can generate data to pass into `multicall` on the HubPool to update `vkey` of the Universal chains (that are using ZK-messaging)

### DIFF
--- a/script/tasks/UpdateVkey.s.sol
+++ b/script/tasks/UpdateVkey.s.sol
@@ -4,6 +4,7 @@ pragma solidity ^0.8.0;
 import { Script } from "forge-std/Script.sol";
 import { console } from "forge-std/console.sol";
 import { Constants } from "../utils/Constants.sol";
+import { DeployedAddresses } from "../utils/DeployedAddresses.sol";
 
 /**
  * @title UpdateVkey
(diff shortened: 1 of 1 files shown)


tasks/updatevkey.ts
```diff
@@ -0,0 +1,267 @@
+import { task, types } from "hardhat/config";
+import type { HardhatRuntimeEnvironment } from "hardhat/types";
+import { ethers } from "ethers";
+import deploymentsJson from "../deployments/deployments.json";
```
fusmanii:
```suggestion
import deploymentsJson from "../broadcast/deployed-addresses.json";
```

fusmanii (commented):
Looks good, just one comment on using `deployed-addresses.json` instead of `deployments.json`

tasks/updatevkey.ts
grasphoper: Oh duh, you're right. Addressed here [f713555](https://github.com/across-protocol/contracts/pull/1199/commits/f713555f46547b771d39bc29b157cf628f35be3c)

script/tasks/UpdateVkey.s.sol
```diff
@@ -0,0 +1,305 @@
...
+            ".address"
+        );
+        
+        try vm.parseJsonAddress(json, path) returns (address addr) {
+            return addr;
+        } catch {
+            return address(0);
+        }
+    }
+}
```
grasphoper, line 266: Can we use `getAddress` from `utils/DeployedAddresses.sol` here?

tbwebb22: https://github.com/across-protocol/contracts/pull/1199/commits/c0b73a2e05bd2a4c3486ac6b5341a37c924dd009

PR #1201, opened 2025-12-01
fix: Support new Universal Spoke Pool Proxy deployments

fusmanii:
The deploy script now handles missing spoke pool proxy and deploys a new proxy contract along with deploying the implementation

Also the address extraction script renames `ERC1967Proxy` to `SpokePool`

### DIFF
--- a/script/111DeployUniversalSpokePool.s.sol
+++ b/script/111DeployUniversalSpokePool.s.sol
@@ -73,7 +73,7 @@ contract DeployUniversalSpokePool is Script, Test, DeploymentUtils {
             "Universal_SpokePool",
             constructorArgs,
             initArgs,
-            false // implementationOnly
+            true // implementationOnly
         );
 
         // Log the deployed addresses


script/111DeployUniversalSpokePool.s.sol
```diff
@@ -73,7 +73,7 @@ contract DeployUniversalSpokePool is Script, Test, DeploymentUtils {
             "Universal_SpokePool",
             constructorArgs,
             initArgs,
-            true // implementationOnly
+            false // implementationOnly
```
tbwebb22: this means a new proxy will get deployed on every deployment, right? Is that what we want?

deploy/consts.ts
```diff
@@ -211,6 +213,7 @@ export const L2_ADDRESS_MAP: { [key: number]: { [contractName: string]: string }
     permit2: "0x000000000022d473030f116ddee9f6b43ac78ba3",
   },
   [CHAIN_IDs.PLASMA]: {
+    helios: "0x6f1cd5f317a7228269eab2b496313862de712ccb",
```
grasphoper, line 216: Can we somehow just query `helios` from a SpokePool? I feel like we have helios addresses in many different places

script/111DeployUniversalSpokePool.s.sol
fusmanii: No, this will only deploy a new proxy if one is not already deployed. Setting it to `true` will skip checking for proxy and only deploy implementation

deploy/consts.ts
fusmanii: `helios` is also in `deployments.json` which I want to move away from, so I am adding here as well. I will remove `helios` from `deployments.json` once I delete the hardhat deployment scripts

script/111DeployUniversalSpokePool.s.sol
tbwebb22: to me it looks like the opposite - if `implementationOnly` is false, we'll always end up in the `else` block here

https://github.com/across-protocol/contracts/blob/82b27cc417dbb5e376e0ef4f9dbc4a818cb00c8e/script/utils/DeploymentUtils.sol#L103-L130

fusmanii: yes you are right, nice catch switched it back to `true` https://github.com/across-protocol/contracts/pull/1201/commits/f8fa0ecb167132cd525bb4cf092b538f2cfa2ad8

PR #1206, opened 2025-12-08
  Clarify token support for Base-Solana bridging.

ukml:
Suggesting more details on supported tokens, as I encountered issues with non-USDC/ETH coins.


grasphoper:
Thanks for opening a PR. Supported tokens can be queried from the API:

```
curl -s 'https://app.across.to/api/available-routes?originChainId=8453&destinationChainId=34268394551451'
```

This can be seen in Across docs

PR #1215, opened 2025-12-10
chore: Typechain generation with foundry & export foundry artifacts

fusmanii:
Removing hardhat typechain plugin and generating typechain types using foundry artifacts. Also exposing foundry artifacts that other repos (like sdk) can use instead of hardhat artifacts

Note: foundry and typechain wasn't playing nice with deploy script having leading number, so I renamed all deploy script 

Closes: https://linear.app/uma/issue/ACP-22/export-foundry-artifacts

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@across-protocol/contracts",
-  "version": "4.1.23",
+  "version": "4.1.22",
   "author": "UMA Team",
   "license": "AGPL-3.0-only",
   "repository": {


.github/actions/generate-evm-artifacts/action.yml
```diff
@@ -16,9 +16,20 @@ runs:
     - name: Install packages
       shell: bash
       run: yarn install --frozen-lockfile
-    - name: Build EVM
+    - name: Build Harhat
```
tbwebb22: nitpick: typo here

.github/actions/generate-evm-artifacts/action.yml
```diff
@@ -16,9 +16,20 @@ runs:
...
+      uses: foundry-rs/foundry-toolchain@v1
+    - name: Install forge dependencies
+      shell: bash
+      run: forge install
+    - name: Build Foundry
+      shell: bash
+      run: forge build --skip test script && rm out/build-info/*.json
+    - name: Generate Typechain
+      shell: bash
+      run: npx typechain --target ethers-v5 "out/**/*.json" --out-dir typechain-types
```
grasphoper: Are we producing both `typechain` and `typechain-types`?
What would it take for us to fully move to foundry-generated typechain only? And save that into `typechain`

E.g., we'd have to test consumer repos for not breaking as a cause. Can we produce a pre-release and ask ops / backend teams to test this?

fusmanii: We can fully go to foundry-generate typechain, I was just thinking having both in parallel is safer. But I like your idea if pre-release. I will update to just do that

.github/actions/generate-evm-artifacts/action.yml
fusmanii: https://github.com/across-protocol/contracts/pull/1215/changes/c1834041803cc751169b6e62575935aa98c69451

grasphoper:
Let's not merge it until we're certain that all consumer repos are compatible. Also, as discussed in Slack, we might want to remove `typechain` generation from the `conatracts` repo altogether and let consumers generate types from compilation artifacts exported from this repo

.github/actions/generate-evm-artifacts/action.yml
```diff
@@ -16,9 +16,23 @@ runs:
     - name: Install packages
       shell: bash
       run: yarn install --frozen-lockfile
-    - name: Build EVM
+    - name: Build Hardhat
       shell: bash
       run: yarn build-evm
```
grasphoper: 2 questions:
- do we need this step here if we're using foundry exclusively?
- should we rename `build-evm` into `build-hardhat`? Otherwise it's confusing

.github/workflows/pr.yml
```diff
@@ -222,5 +225,9 @@ jobs:
         run: forge install
       - name: Inspect storage layouts
         run: ./scripts/checkStorageLayout.sh
+      - name: Download EVM artifacts
+        uses: actions/download-artifact@v4
+        with:
+          name: evm-artifacts-${{ runner.os }}-node-${{ env.NODE_VERSION }}
```
grasphoper, line 231: Why did we not need this before? Is this just an optimization to not recompile?
Won't a different foundry profile force a recompile anyway (local). Esp. after Taylor's changes where we change a strip_strings settings between local and default

broadcast/deployed-addresses.json
```diff
@@ -173,10 +173,15 @@
...
-          "transaction_hash": "0x0b05b7bc28e6d8ea9d9c135bbada908dd2b971e8428054a62598c2cbf9e24baa"
+        "PermissionSplitterProxy": {
+          "address": "0x0Bf07B2e415F02711fFBB32491f8ec9e5489B2e7",
+          "block_number": 19084679,
+          "transaction_hash": "0xa2a7b2c6812fb8ae34539fb04cd5f2a9112da1c7f6ffce0ddcf1fee7e43acf48"
+        },
+        "PermissionedMulticallHandler": {
+          "address": "0x64a43393866dba0044879979faa7ad3d000622e9",
+          "block_number": 24021499,
+          "transaction_hash": "0x70b3edf99db66aac34338c3686cd708ddfc2b503c2f707f51ced67dc50e5022f"
```
grasphoper, line 184: Why are these changing? I see, this is just reshuffling? Is our generation deterministic? If so, let's just update `master` with it, wdyt? To remove these changes

broadcast/deployed-addresses.json
```diff
@@ -514,14 +514,9 @@
...
-          "address": "0x3d589d40312bf2d20f13cd0af26a11144a9ca844",
-          "block_number": 21341868,
-          "transaction_hash": "0xe617eb954e31dc4305f9ddcd9badd2e0e707e4b9c266e16fddaf8e7c23180fde"
-        },
-        "SponsoredCCTPDstPeriphery": {
-          "address": "0x1c709fd0db6a6b877ddb19ae3d485b7b4add879f",
+          "address": "0x039d62c549f27ead0eb9b567d8776289e5020583",
           "block_number": 21416410,
-          "transaction_hash": "0x9c7604f9254450f13b0941008f5272f68ec58db61f32e5ee6713e72b4b56de0b"
+          "transaction_hash": "0x6875ef9c89d6855fb5019ce5d20f1466ecdced640b3be009fe6de82932469664"
```
grasphoper, line 519: Related to https://github.com/across-protocol/contracts/pull/1215/changes#r2693003352, but it seems like DonationBox address actually changed here?

0x3d589d40312bf2d20f13cd0af26a11144a9ca844 -> 0x039d62c549f27ead0eb9b567d8776289e5020583

hardhat.config.ts
```diff
@@ -1,12 +1,14 @@
 const { subtask } = require("hardhat/config");
 const { TASK_COMPILE_SOLIDITY_GET_SOURCE_PATHS } = require("hardhat/builtin-tasks/task-names");
 
+const isTest = process.env.IS_TEST === "true" || process.env.CI === "true";
```
grasphoper: I remember there was a problem with defining `isTest` like this. It means that we don't export certain artifacts for consumers. With this PR we're moving to ONLY exporting foundry artifacts, so this is OK?

scripts/typechainArtifacts.ts
grasphoper, line 1: One thing I worry about here is the custom dedupe logic we're using here. Is there analogous logic in hardhat that we can use as reference?

On the other hand, even if the dedupe doesn't 100% work as expected, what's our biggest risk?

.github/actions/generate-evm-artifacts/action.yml
fusmanii: Good point, I am not removing this just yet because I dont want to force other repos to migrate. The plan is to still support hardhat artifacts and only remove once all other repos have migrated to foundry

fusmanii: +1 on `build-hardhat`

.github/workflows/pr.yml
fusmanii: Yeah my thinking was that this will re-use artifacts but you are right tests are run using a different profile. I think I can generate the artifacts in the same manner (with FOUNDRY_PROFILE=local) then it can be reused

broadcast/deployed-addresses.json
fusmanii: Typechain doesn't like deploy scripts having a leading number so I removed them which caused reshuffling. Yeah I can update master

broadcast/deployed-addresses.json
fusmanii: yeah might be some local changes I had, going to revert this

hardhat.config.ts
fusmanii: The issue was with hardhat and typechian, but since we are moving away from hardhat we longer need this. But I will add this back and remove in future pr once everything is migrated

.github/actions/generate-evm-artifacts/action.yml
```diff
@@ -16,9 +16,17 @@ runs:
...
-      run: yarn build-evm
+      run: yarn build-hardhat
+    - name: Install Foundry
+      uses: foundry-rs/foundry-toolchain@v1
+    - name: Install forge dependencies
+      shell: bash
+      run: forge install
+    - name: Build Foundry
+      shell: bash
+      run: FOUNDRY_PROFILE=local forge build
```
grasphoper: 2 Qs:
- should we use a `yarn` cmd here instead? For easy syncing if the cmd ever got updated? Like, `local` changed to `local-test` recently, this would break
- do we want to export artifacts from a test build here? For now, the only difference seems to be the handling of revert strings, but that might change in the future

.github/workflows/pr.yml
```diff
@@ -47,6 +47,8 @@ jobs:
         artifacts
         cache
         typechain
+        out
+        cache-foundry
```
grasphoper, line 51: Related to comment above, `cache-foundry` or `cache-foundry-local`?

.github/workflows/publish.yml
```diff
@@ -21,6 +21,8 @@ env:
     artifacts
     cache
     typechain
+    out
+    cache-foundry
```
grasphoper, line 25: same as https://github.com/across-protocol/contracts/pull/1215/changes#r2714608860

grasphoper (dismissed):
LGTM, left a few clarifying Qs

.github/actions/generate-evm-artifacts/action.yml
fusmanii: - Yeah using `yarn` here is better 
- I want build with test profile so that forge test step can re-use the already compiled artifacts. Other wise it will recompile again

.github/workflows/pr.yml
fusmanii: oh I didn't realize cache folder for test profile has been changed

.github/actions/generate-evm-artifacts/action.yml
fusmanii: Looks like more config has been added to local profile, so I revered back to using prod profile

.github/workflows/pr.yml
fusmanii: reverted to not using test build

grasphoper (approved):

PR #1217, opened 2025-12-11
chore: Delete Hardhat deployment scripts

fusmanii:
Deleting all the deploy scripts that have been migrate to foundry


deploy/054_deploy_alephzero_spokepool.ts
grasphoper, line 1: Did we disable support for Aleph Zero entierly? I think so right. So we might want to just remove the foundry deploy script and AlephZero contract and related deployment addresses. We can address that in a separate PR but wanted to mention here.

Let's discuss this in slack or on a sync

PR #1218, opened 2025-12-12
feat: audit follow up

fusmanii:
Added support for:
- any token `-> PERP`, not just `-> SPOT`
- account creation on the fly from user funds

### DIFF
--- a/script/mintburn/cctp/createLighterDeposit.s.sol
+++ b/script/mintburn/cctp/createLighterDeposit.s.sol
@@ -110,6 +110,8 @@ contract CreateLighterDeposit is Script, Config {
             maxUserSlippageBps: 0,
             finalRecipient: lighterRecipient.toBytes32(),
             finalToken: env.dstUsdc.toBytes32(),
+            destinationDex: 0, // Not used for ArbitraryActionsToEVM
+            accountCreationMode: 0, // Not used for ArbitraryActionsToEVM
             executionMode: uint8(SponsoredCCTPInterface.ExecutionMode.ArbitraryActionsToEVM),
             actionData: actionData
         });
@@ -206,6 +208,8 @@ contract CreateLighterDeposit is Script, Config {
                 quote.maxUserSlippageBps,
                 quote.finalRecipient,
                 quote.finalToken,
+                quote.destinationDex,
+                quote.accountCreationMode,
                 quote.executionMode,
                 keccak256(quote.actionData)
             )


contracts/libraries/HyperCoreLib.sol
```diff
@@ -89,15 +91,21 @@ library HyperCoreLib {
...
-            transferToCore(erc20EVMAddress, erc20CoreIndex, _amountEVMToSend);
-            // Transfer the tokens from this contract on HyperCore to the `to` address on HyperCore
-            transferERC20CoreToCore(erc20CoreIndex, to, _amountCoreToReceive);
+            if (erc20CoreIndex == USDC_CORE_INDEX) {
+                IERC20(erc20EVMAddress).forceApprove(USDC_CORE_DEPOSIT_WALLET_ADDRESS, _amountEVMToSend);
+                ICoreDepositWallet(USDC_CORE_DEPOSIT_WALLET_ADDRESS).depositFor(to, _amountEVMToSend, destinationDex);
+            } else {
+                IERC20(erc20EVMAddress).safeTransfer(toAssetBridgeAddress(erc20CoreIndex), _amountEVMToSend);
+                // Transfer the tokens from this contract on HyperCore to the `to` address on HyperCore
+                transferERC20CoreToCore(erc20CoreIndex, to, _amountCoreToReceive, destinationDex, destinationDex);
```
tbwebb22: why do we pass in the `destinationDex` to the `sourceDex` function arg here?

fusmanii: nice catch, fixed https://github.com/across-protocol/contracts/pull/1218/changes/c661b36c14dfef73e3e6aa4988987e995e9f1e35

contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
```diff
@@ -898,19 +908,7 @@ contract HyperCoreFlowExecutor is AccessControlUpgradeable, AuthorizedFundedFlow
         require(safeToBridge, "Not safe to bridge");
         _getMainStorage().cumulativeSponsoredActivationFee[fundingToken] += evmAmountToSend;
 
-        // donationBox @ evm -> Handler @ evm
-        donationBox.withdraw(IERC20(fundingToken), evmAmountToSend);
```
tbwebb22, line 902: why is the donation box withdrawal removed?

contracts/handlers/HyperliquidDepositHandler.sol
```diff
@@ -180,7 +180,13 @@ contract HyperliquidDepositHandler is AcrossMessageHandler, ReentrancyGuard, Own
...
     function sweepCoreFundsToUser(address token, uint64 coreAmount, address user) external onlyOwner nonReentrant {
         uint64 tokenIndex = _getTokenInfo(token).tokenId;
-        HyperCoreLib.transferERC20CoreToCore(tokenIndex, user, coreAmount);
+        HyperCoreLib.transferERC20CoreToCore(
+            tokenIndex,
+            user,
+            coreAmount,
+            HyperCoreLib.CORE_SPOT_DEX_ID,
+            HyperCoreLib.CORE_SPOT_DEX_ID
+        );
```
grasphoper: We use `..HyperCoreLib.CORE_SPOT_DEX_ID` a lot as an arg to functions sending to / between core accounts.
Do you think it makes sense to have fns like:

```
.transferERC20EVMToSelfOnSpot
.transferERC20SpotToSpot
```

to mimic old function behaviours? This might make reviewing easier

contracts/libraries/HyperCoreLib.sol
```diff
@@ -111,19 +119,26 @@ library HyperCoreLib {
...
+        int8 decimalDiff,
+        uint32 destinationDex
     ) internal returns (uint256 amountEVMSent, uint64 amountCoreToReceive) {
         (uint256 _amountEVMToSend, uint64 _amountCoreToReceive) = maximumEVMSendAmountToAmounts(amountEVM, decimalDiff);
 
         if (_amountEVMToSend != 0) {
-            transferToCore(erc20EVMAddress, erc20CoreIndex, _amountEVMToSend);
+            if (erc20CoreIndex == USDC_CORE_INDEX) {
+                IERC20(erc20EVMAddress).forceApprove(USDC_CORE_DEPOSIT_WALLET_ADDRESS, _amountEVMToSend);
+                ICoreDepositWallet(USDC_CORE_DEPOSIT_WALLET_ADDRESS).deposit(_amountEVMToSend, destinationDex);
```
grasphoper: Depending on the `destinationDex` here, it's either:
- 1 action: evm -> spot
- 2 actions: evm -> spot + spot -> destinationDex

This might potentially break our assumptions at the call sites where we use `transferERC20EVMToSelfOnCore` (e.g. assuming that evm -> core transfer happens first always regardless of the ordering). Do you think any such cases exist here?

contracts/periphery/mintburn/sponsored-oft/ComposeMsgCodec.sol
```diff
@@ -10,9 +10,10 @@ library ComposeMsgCodec {
     uint256 internal constant MAX_USER_SLIPPAGE_BPS_OFFSET = 96;
     uint256 internal constant FINAL_RECIPIENT_OFFSET = 128;
     uint256 internal constant FINAL_TOKEN_OFFSET = 160;
-    uint256 internal constant EXECUTION_MODE_OFFSET = 192;
+    uint256 internal constant DESTINATION_DEX_OFFSET = 164;
+    uint256 internal constant EXECUTION_MODE_OFFSET = 196;
```
grasphoper: Should these be 192, 196 instead?

contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
```diff
@@ -898,19 +908,7 @@ contract HyperCoreFlowExecutor is AccessControlUpgradeable, AuthorizedFundedFlow
...
-            fundingToken,
-            coreTokenInfo.coreIndex,
-            evmAmountToSend,
-            coreTokenInfo.tokenInfo.evmExtraWeiDecimals
-        );
-        // The total balance withdrawn from Handler @ Core for this operation is activationFee + amountSent, so we set
-        // amountSent to 1 wei to only activate the account
-        // Handler @ core -> finalRecipient @ core
-        HyperCoreLib.transferERC20CoreToCore(coreTokenInfo.coreIndex, finalRecipient, 1);
+        HyperCoreLib.activateCoreAccountFromEVM(fundingToken, coreTokenInfo.coreIndex, finalRecipient, evmAmountToSend);
```
grasphoper, line 972: Where does the money for activation come from? We should still withdraw from donationBox here right?

grasphoper: Like we do here https://github.com/across-protocol/contracts/pull/1220/changes

contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
```diff
@@ -1116,8 +1115,14 @@ contract HyperCoreFlowExecutor is AccessControlUpgradeable, AuthorizedFundedFlow
...
-    function sweepOnCore(address token, uint64 amount) external onlyRole(FUNDS_SWEEPER_ROLE) {
-        HyperCoreLib.transferERC20CoreToCore(_getMainStorage().coreTokenInfos[token].coreIndex, msg.sender, amount);
+    function sweepOnCore(address token, uint64 amount, uint32 destinationDex) external onlyRole(FUNDS_SWEEPER_ROLE) {
+        HyperCoreLib.transferERC20CoreToCore(
+            _getMainStorage().coreTokenInfos[token].coreIndex,
+            msg.sender,
+            amount,
+            destinationDex,
+            destinationDex
+        );
```
grasphoper, line 1194: Can the funds ever be on non-spot balance of dst handler?

I.e., do we ever have to send `->dst handler @ perp` or do we always send

`dst handler @ spot -> user @ perp` even if we're sending to perp?

contracts/periphery/mintburn/SwapHandler.sol
```diff
@@ -21,30 +21,35 @@ contract SwapHandler {
...
     }
 
-    function activateCoreAccount(
-        address erc20EVMAddress,
-        uint64 erc20CoreIndex,
-        uint256 amountEVM,
-        int8 decimalDiff
-    ) external onlyParentHandler {
-        HyperCoreLib.transferERC20EVMToSelfOnCore(erc20EVMAddress, erc20CoreIndex, amountEVM, decimalDiff);
-    }
```
grasphoper, line 31: Good call :)

contracts/periphery/mintburn/sponsored-oft/ComposeMsgCodec.sol
```diff
@@ -10,9 +10,10 @@ library ComposeMsgCodec {
...
     // Minimum length with empty actionData: 7 regular params (32 bytes each) and 1 dynamic byte array (minumum 64 bytes)
-    uint256 internal constant MIN_COMPOSE_MSG_BYTE_LENGTH = 288;
+    uint256 internal constant MIN_COMPOSE_MSG_BYTE_LENGTH = 292;
```
grasphoper: This should also be different, the uint8 above will take 32 bytes in the compose msg

contracts/periphery/mintburn/sponsored-oft/ComposeMsgCodec.sol
fusmanii: yes should be 192 and 196, nice catch

contracts/libraries/HyperCoreLib.sol
fusmanii: Good point, I think the problem might come when finalizing a swap where we queue the `core -> core` calls first and then send sponsored funds to swap handler. So this means that we need to change the order of CoreWriter calls in finalize swap function

contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
fusmanii: nice catch, I accidentally deleted that call

contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
fusmanii: It shouldn't be since all CoreWriter transfers first land on spot and then they can be moved to perp if need be, and we are not even doing `spot -> perp` to self for dst handler

contracts/periphery/mintburn/sponsored-oft/ComposeMsgCodec.sol
fusmanii: yes you are right

contracts/libraries/HyperCoreLib.sol
fusmanii: after off-line conversation, this current implantation of `finalizeSwapFlows` works since all swap handler balance will always be in spot

contracts/periphery/mintburn/sponsored-oft/ComposeMsgCodec.sol
grasphoper: Actually, my comment is wrong 🤦 

they should be 192, 224
Just because all simple vars take up 32 bytes

fusmanii: nice catch x 2, updated

contracts/periphery/mintburn/sponsored-oft/ComposeMsgCodec.sol
```diff
@@ -68,23 +68,18 @@ library ComposeMsgCodec {
     }
 
     function _getDestinationDex(bytes memory data) internal pure returns (uint32 v) {
-        return BytesLib.toUint32(data, DESTINATION_DEX_OFFSET);
+        // 28 is the offset to read uint32 at the end of the 32-byte long slot
+        return BytesLib.toUint32(data, DESTINATION_DEX_OFFSET + 28);
```
grasphoper: @fusmanii reading less-than-32-byte-long vars via `BytesLib` works something like this.

You have your OFFSET, but then you need to specify the exact starting byte index of your var data.

Just FYI

contracts/libraries/SponsoredCCTPQuoteLib.sol
```diff
@@ -38,7 +38,7 @@ library SponsoredCCTPQuoteLib {
     uint256 private constant HOOK_DATA_INDEX = 228;
 
     // Minimum length of the message body (can be longer due to variable actionData)
-    uint256 private constant MIN_MSG_BYTES_LENGTH = 664;
+    uint256 private constant MIN_MSG_BYTES_LENGTH = 696;
```
Reinis-FRP: adding 2 words to hook data should increase minimum length of the message body by 64 bytes

contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
```diff
@@ -499,18 +500,30 @@ contract HyperCoreFlowExecutor is AccessControlUpgradeable, AuthorizedFundedFlow
...
 
-        // Check account activation
-        if (!HyperCoreLib.coreUserExists(params.finalRecipient)) {
-            if (params.maxBpsToSponsor > 0) {
-                revert AccountNotActivatedError(params.finalRecipient);
+        bool userAccountExists = HyperCoreLib.coreUserExists(params.finalRecipient);
+        uint64 accountActivationFeeCore;
+        if (!userAccountExists) {
+            if (params.accountCreationMode == AccountCreationMode.Standard) {
+                if (!HyperCoreLib.coreUserExists(params.finalRecipient)) {
```
Reinis-FRP: why repeated call and `userAccountExists=false` in this branch already

contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
```diff
@@ -591,23 +606,36 @@ contract HyperCoreFlowExecutor is AccessControlUpgradeable, AuthorizedFundedFlow
...
-
         MainStorage storage $ = _getMainStorage();
         CoreTokenInfo memory initialCoreTokenInfo = $.coreTokenInfos[initialToken];
         CoreTokenInfo memory finalCoreTokenInfo = $.coreTokenInfos[params.finalToken];
         FinalTokenInfo memory finalTokenInfo = _getExistingFinalTokenInfo(params.finalToken);
 
+        bool userAccountExists = HyperCoreLib.coreUserExists(params.finalRecipient);
+        if (!userAccountExists) {
+            if (params.accountCreationMode == AccountCreationMode.Standard) {
+                if (!HyperCoreLib.coreUserExists(params.finalRecipient)) {
```
Reinis-FRP: `userAccountExists=false` in this branch already

contracts/interfaces/SponsoredCCTPInterface.sol
```diff
@@ -78,6 +83,10 @@ interface SponsoredCCTPInterface {
         // The final token that final recipient will receive. This is needed as it can be different from the burnToken
         // in which case we perform a swap on the destination chain.
         bytes32 finalToken;
+        // The destination DEX on HyperCore.
+        uint32 destinationDex;
+        // AccountCreationMode: Standard or FromUserFunds
+        uint8 accountCreationMode;
```
Reinis-FRP, line 86: just checking, we don't need any of these new fields added to `SponsoredDepositForBurn` or do we?

contracts/libraries/SponsoredCCTPQuoteLib.sol
fusmanii: nice catch, I think it was missed when merging the other pr, fixed https://github.com/across-protocol/contracts/pull/1218/changes/a393df4b57400d8e017023570a61b1db1d179724

contracts/interfaces/SponsoredCCTPInterface.sol
fusmanii: good idea, updated https://github.com/across-protocol/contracts/pull/1218/commits/7624b3c1b769187588478b8394a9ce5b37a89582

contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
grasphoper: Oh, great call, thank you!

For standard mode, I just copied the existing logic and didn't update it based on the new outter `if` I added. Fixed in [2e5bb7e](https://github.com/across-protocol/contracts/pull/1218/commits/2e5bb7e17f5c7d7f0767ab871702cda038dc4c96)

contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
```diff
@@ -591,23 +604,34 @@ contract HyperCoreFlowExecutor is AccessControlUpgradeable, AuthorizedFundedFlow
...
+                    return;
+                }
+            } else {
+                // In the FromUserFunds logic for the swap flow, we only allow account creation if the final token is
+                // usable for it. Finilazing the swap flow should handle account creation automatically when doing the
+                // final transfer of tokens to the user
+                if (!finalCoreTokenInfo.canBeUsedForAccountActivation) {
+                    emit AccountNotActivated(params.quoteNonce, params.finalRecipient);
+                    params.finalToken = initialToken;
+                    _fallbackHyperEVMFlow(params);
```
tbwebb22: should we `return;` after this line?

grasphoper: Great call, added here [ddf951a](https://github.com/across-protocol/contracts/pull/1218/commits/ddf951ac3b6766b6603a6874ddbef855468f8a4b)

contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
```diff
@@ -591,23 +604,35 @@ contract HyperCoreFlowExecutor is AccessControlUpgradeable, AuthorizedFundedFlow
...
+                    return;
+                }
+            }
+        }
```
tbwebb22: also this could be simplified to something like:

```
if (!HyperCoreLib.coreUserExists(params.finalRecipient)) {
    bool isStandard = params.accountCreationMode == AccountCreationMode.Standard;
    
    if (isStandard && params.maxBpsToSponsor > 0) {
        revert AccountNotActivatedError(params.finalRecipient);
    }
    
    // Fall back unless FromUserFunds mode with an eligible activation token
    if (isStandard || !finalCoreTokenInfo.canBeUsedForAccountActivation) {
        emit AccountNotActivated(params.quoteNonce, params.finalRecipient);
        params.finalToken = initialToken;
        _fallbackHyperEVMFlow(params);
        return;
    }
}
```

contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
```diff
@@ -499,18 +500,28 @@ contract HyperCoreFlowExecutor is AccessControlUpgradeable, AuthorizedFundedFlow
...
             } else {
-                emit AccountNotActivated(params.quoteNonce, params.finalRecipient);
-                _fallbackHyperEVMFlow(params);
-                return;
+                // Notice. If the amount is too small to be able to activate an account, the funds will land there, but
+                // the account won't show up as active until an eligible transfer activates it
+                if (!coreTokenInfo.canBeUsedForAccountActivation) {
+                    _fallbackHyperEVMFlow(params);
+                    return;
+                }
```
tbwebb22: Do we want to emit the `AccountNotActivated` event here?
If so, we could probably simplify this in a similar way to my other comment

grasphoper: Added events in [2c97a95](https://github.com/across-protocol/contracts/pull/1218/commits/2c97a9528cbc107a96ae4619c991afaca6a1e712)

Looking into simplification

PR #1222, opened 2025-12-16
chore: account creation using user funds

grasphoper:


contracts/libraries/HyperCoreLib.sol
```diff
@@ -94,17 +95,33 @@ library HyperCoreLib {
...
         uint32 destinationDex
     ) internal returns (uint256 amountEVMSent, uint64 amountCoreToReceive) {
-        // if the transfer amount exceeds the bridge balance, this wil revert
+        bool transferToSelf = to == address(this);
+
+        // If the transfer amount exceeds the bridge balance, the funds will get blackholed
         (uint256 _amountEVMToSend, uint64 _amountCoreToReceive) = maximumEVMSendAmountToAmounts(amountEVM, decimalDiff);
 
-        if (_amountEVMToSend != 0) {
+        if (_amountEVMToSend > 0) {
```
tbwebb22, line 103: nitpick: 
!= 0 is a bit cheaper and functionally the same here since we're comparing a uint

contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
```diff
@@ -913,6 +904,45 @@ contract HyperCoreFlowExecutor is AccessControlUpgradeable, AuthorizedFundedFlow
...
+        // tokenOnHand is `baseToken` in the swap flow and is `finalToken` in the other flows
+        address tokenOnHandAddress,
+        CoreTokenInfo memory tokenOnHandInfo
+    ) internal returns (bool evmFallbackRequired) {
+        if (HyperCoreLib.coreUserExists(params.finalRecipient)) {
+            return false;
+        }
+
+        if (params.accountCreationMode == AccountCreationMode.FromUserFunds) {
+            require(tokenOnHandInfo.canBeUsedForAccountActivation, "tokenOnHand cannot be used for activation");
```
tbwebb22, line 918: why revert here rather than returning `true` to initiate the fallback flow?

contracts/libraries/HyperCoreLib.sol
grasphoper: Oh interesting, you're right!
I switched to `>` cause it was one less symbol haha

contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
grasphoper: Yeah good point. This PR is unpolished. I've since understood that this way of doing it doesn't work :(

Because for example we can create an account and pay for it and then fall back to evm flow. This is not ideal.

I'm working on an alternative approach right now

PR #1223, opened 2025-12-16
feat(HypercoreDepositHandler): Allow different types of messages passed to handleAcrossV3Message

nicholaspai:
Read first byte to determine how to decode `message`:
- type 1: message = abi.encode(user)
- type 2: message = abi.encode(user, signature)

type 2 messages can be used to activate new accounts, type 1 cannot

### DIFF
--- a/contracts/handlers/HyperliquidDepositHandler.sol
+++ b/contracts/handlers/HyperliquidDepositHandler.sol
@@ -124,7 +124,7 @@ contract HyperliquidDepositHandler is AcrossMessageHandler, ReentrancyGuard, Own
         if (message[0] == 0) {
             address user = abi.decode(message[1:], (address));
             _depositToHypercore(token, amount, user, false);
-        } else if (message[0] == bytes1("1")) {
+        } else if (message[0] == bytes1(0x01)) {
             (address user, bytes memory signature) = abi.decode(message[1:], (address, bytes));
             _verifySignature(user, signature);
             _depositToHypercore(token, amount, user, true);


grasphoper:
@nicholaspai are we moving forward with sig-based account creation? As we discussed in Slack, this doesn't fully protect us from griefing, but makes griefing a bit harder

contracts/handlers/HyperliquidDepositHandler.sol
```diff
@@ -108,18 +105,30 @@ contract HyperliquidDepositHandler is AcrossMessageHandler, ReentrancyGuard, Own
...
-        _verifySignature(user, signature);
-        _depositToHypercore(token, amount, user);
+        _decodeMessageAndDepositToHypercore(token, amount, message);
+    }
+
+    function _decodeMessageAndDepositToHypercore(address token, uint256 amount, bytes calldata message) internal {
+        if (message[0] == 0) {
+            address user = abi.decode(message[1:], (address));
+            _depositToHypercore(token, amount, user, false);
+        } else if (message[0] == bytes1("1")) {
```
fusmanii:
```suggestion
        } else if (message[0] == bytes1(0x01)) {
```

grasphoper:
> > @nicholaspai are we moving forward with sig-based account creation? As we discussed in Slack, this doesn't fully protect us from griefing, but makes griefing a bit harder
> 
> Yes, because we want the griefing protection to be possible at the API level. With signatures we can control do this, whereas if we don't have signatures at all, then even the API can't protect against this.

Oh, you mean that the API can block by IP and such?

PR #1226, opened 2025-12-18
refactor: migrate tasks

tbwebb22:
Migrates `TestChainAdapter` and `UpdateSpokePool` tasks to Foundry - these seem to be the only tasks that are currently used. Deletes old hardhat tasks and updates relevant documentation


script/tasks/TestChainAdapter.s.sol
```diff
@@ -0,0 +1,111 @@
...
+        address l2Token
+    ) external {
+        uint256 hubChainId = block.chainid;
+
+        // Try to get adapter from deployed addresses
+        address adapterAddress = getAddress(hubChainId, adapterName);
+        require(adapterAddress != address(0), string.concat("Adapter not found: ", adapterName));
+
+        this.run(spokeChainId, adapterAddress, l1Token, amount, l2Token);
+    }
```
grasphoper: I suggest that we either add a usage example for this method, or remove it completely. I think at this point it's easier(less of a mental load heh) for a dev to manually insert "wanted adapter address" than to figure out how to call this other method (considering the example call is unavailable in the comments)

script/tasks/TestChainAdapter.s.sol
```diff
@@ -0,0 +1,111 @@
...
+ *   forge script script/tasks/TestChainAdapter.s.sol:TestChainAdapter \
+ *     --sig "run(uint256,address,address,uint256,address)" \
+ *     <spokeChainId> <adapterAddress> <l1Token> <amount> <l2Token> \
+ *     --rpc-url mainnet --broadcast
+ *
+ * Example (bridge 1 USDC to Optimism):
+ *   forge script script/tasks/TestChainAdapter.s.sol:TestChainAdapter \
+ *     --sig "run(uint256,address,address,uint256,address)" \
+ *     10 0x... 0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48 1000000 0x... \
+ *     --rpc-url mainnet --broadcast
```
grasphoper: Mb include an example with real addrs? Or we can perhaps remove the second example as it's not materially different from `Usage`. Wdyt?

script/tasks/UpgradeSpokePool.s.sol
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * Usage:
+ *   forge script script/tasks/UpgradeSpokePool.s.sol:UpgradeSpokePool \
+ *     --sig "run(address)" <NEW_IMPLEMENTATION_ADDRESS> \
+ *     -vvvv
+ *
+ * Example:
+ *   forge script script/tasks/UpgradeSpokePool.s.sol:UpgradeSpokePool \
+ *     --sig "run(address)" 0x1234567890123456789012345678901234567890 \
+ *     -vvvv
```
grasphoper: Are you sure we need both of these examples?

script/tasks/UpgradeSpokePool.s.sol
grasphoper, line 1: This impl looks good to me. Have you checked against the old script to confirm that the outputs match? As we use this to generate multisig data (so it's sensitive)

grasphoper (commented):
Left a few minor comments!

script/tasks/UpgradeSpokePool.s.sol
tbwebb22: Yep! Outputs do match

script/tasks/TestChainAdapter.s.sol
tbwebb22: Agreed - removed it here https://github.com/across-protocol/contracts/pull/1226/commits/ef8475435218f388a547151b19df3bd97fe6bb29

script/tasks/TestChainAdapter.s.sol
tbwebb22: Agree its redundant - https://github.com/across-protocol/contracts/pull/1226/commits/60142c807478cc9dce690a681db91949fb1b1523

script/tasks/UpgradeSpokePool.s.sol
tbwebb22: Nope haha definitely redundant https://github.com/across-protocol/contracts/pull/1226/commits/7d11e7822d4afbcfb8085e1bffd4530b2879b4e3

PR #1228, opened 2025-12-23
improve: Iterate on OP deployment tooling

pxrl:
This PR adds support for CCTP to the OP Adapter implementation.

### DIFF
--- a/script/utils/Constants.sol
+++ b/script/utils/Constants.sol
@@ -199,7 +199,7 @@ contract Constants is Script {
         } else if (cctpDomain < 0 || cctpDomain > 2 ** 32 - 1) {
             revert("Invalid Circle CCTP domain ID");
         }
-        return uint32(cctpDomain);
+        return uint32(uint256(cctpDomain));
     }
 
     function hasCctpDomain(uint256 chainId) public view returns (bool) {


broadcast/deployed-addresses.json
```diff
@@ -172,6 +172,11 @@
           "address": "0x89415a82d909a7238d69094C3Dd1dCC1aCbDa85C",
           "block_number": 22789839
         },
+        "OP_Adapter": {
```
pxrl: nb. Naming needs to be chain-specific (i.e. `OP_Adapter_57073`).

broadcast/deployed-addresses.md
```diff
@@ -34,6 +34,7 @@ This file contains the latest deployed smart contract addresses from the broadca
 | LpTokenFactory             | [0x7dB69eb9F52eD773E9b03f5068A1ea0275b2fD9d](https://etherscan.io/address/0x7dB69eb9F52eD773E9b03f5068A1ea0275b2fD9d) |
 | Mode_Adapter               | [0xf1B59868697f3925b72889ede818B9E7ba0316d0](https://etherscan.io/address/0xf1B59868697f3925b72889ede818B9E7ba0316d0) |
 | MulticallHandler           | [0x0F7Ae28dE1C8532170AD4ee566B5801485c13a0E](https://etherscan.io/address/0x0F7Ae28dE1C8532170AD4ee566B5801485c13a0E) |
+| OP_Adapter                 | [0x91769164bb5B820818ce816426D10A725f785330](https://etherscan.io/address/0x91769164bb5B820818ce816426D10A725f785330) |
```
pxrl: Should also have chainId appended (i.e. `OP_Adapter_57073`).

broadcast/deployed-addresses.md
```diff
@@ -34,6 +34,7 @@ This file contains the latest deployed smart contract addresses from the broadca
...
 | Optimism_Adapter           | [0xE1e74B3D6A8E2A479B62958D4E4E6eEaea5B612b](https://etherscan.io/address/0xE1e74B3D6A8E2A479B62958D4E4E6eEaea5B612b) |
 | Optimism_Adapter           | [0x3562e309C6C79626E5F0Cf746FB5Bf4f6b8EebE5](https://etherscan.io/address/0x3562e309C6C79626E5F0Cf746FB5Bf4f6b8EebE5) |
```
pxrl: Driveby comment - we shouldn't have two `Optimism_Adapter` instances - that's pre-existing but can be tidied up here.

script/utils/Constants.sol
```diff
@@ -222,12 +222,14 @@ contract Constants is Script {
...
+    function getCircleDomainId(uint256 chainId) public view returns (uint32) {
         int256 cctpDomain = _getCctpDomain(chainId);
         if (cctpDomain == -1) {
-            revert("Circle domain ID not found");
+            revert("Circle CCTP domain ID not found");
+        } else if (cctpDomain < 0 || cctpDomain > 2 ** 32 - 1) {
+            revert("Invalid Circle CCTP domain ID");
         }
-        return uint256(cctpDomain);
+        return uint32(uint256(cctpDomain));
```
pxrl: The CCTP Domain ID is max 32 bits, so truncate it here after bounds checking.

script/utils/Constants.sol
```diff
@@ -85,7 +85,6 @@ contract Constants is Script {
     struct OpStackAddresses {
         address L1CrossDomainMessenger;
         address L1StandardBridge;
-        address L1BlastBridge;
```
pxrl, line 66: The Blast bridge is a total snowflake and I don't anticipate any further deployments there because the chain seems a bit stale these days. If we do need to support additional Blast depoyments in future then I'd propose we handle it as a special case, rather than baking it into the common OP deployment.

script/056DeployOPAdapter.s.sol
pxrl, line 1: nb. the bones of this file were taken from script/024DeployBaseAdapter.s.sol.

script/056DeployOPAdapter.s.sol
```diff
@@ -0,0 +1,68 @@
...
+import { OP_Adapter } from "../contracts/chain-adapters/OP_Adapter.sol";
+import { Constants } from "./utils/Constants.sol";
+
+// How to run:
+// 1. `source .env` where `.env` has MNEMONIC="x x x ... x" and ETHERSCAN_API_KEY="x" entries
+// 2. forge script script/056DeployOPAdapter.s.sol:DeployOPAdapter --rpc-url $NODE_URL_1 -vvvv
+// 3. Verify the above works in simulation mode.
+// 4. Deploy on mainnet by adding --broadcast --verify flags.
+// 5. forge script script/056DeployOPAdapter.s.sol:DeployOPAdapter --rpc-url $NODE_URL_1 --broadcast --verify -vvvv
+
```
pxrl:
```suggestion
```

broadcast/056DeployOPAdapter.s.sol/1/run-latest.json
pxrl, line 1: Not sure if it's inteded to add this; can easily drop it if not.

broadcast/deployed-addresses.md
```diff
@@ -34,6 +34,7 @@ This file contains the latest deployed smart contract addresses from the broadca
 | LpTokenFactory             | [0x7dB69eb9F52eD773E9b03f5068A1ea0275b2fD9d](https://etherscan.io/address/0x7dB69eb9F52eD773E9b03f5068A1ea0275b2fD9d) |
 | Mode_Adapter               | [0xf1B59868697f3925b72889ede818B9E7ba0316d0](https://etherscan.io/address/0xf1B59868697f3925b72889ede818B9E7ba0316d0) |
 | MulticallHandler           | [0x0F7Ae28dE1C8532170AD4ee566B5801485c13a0E](https://etherscan.io/address/0x0F7Ae28dE1C8532170AD4ee566B5801485c13a0E) |
+| OP_Adapter                 | [0x83057C549C6489899651012F84A36fA58FE2079e](https://etherscan.io/address/0x83057C549C6489899651012F84A36fA58FE2079e) |
```
pxrl: Reviving old comment - this should ideally be `OP_Adapter_<chainId>` (`OP_Adapter_57073` in this case).

script/056DeployOPAdapter.s.sol
```diff
@@ -0,0 +1,67 @@
...
+// 5. forge script script/056DeployOPAdapter.s.sol:DeployOPAdapter --rpc-url $NODE_URL_1 --broadcast --verify -vvvv
+contract DeployOPAdapter is Script, Test, Constants {
+    function run() external {
+        string memory deployerMnemonic = vm.envString("MNEMONIC");
+        uint256 spokeChainId = vm.envUint("SPOKE_CHAIN_ID");
+        uint256 deployerPrivateKey = vm.deriveKey(deployerMnemonic, 0);
+
+        // Get the current chain ID
+        uint256 chainId = block.chainid;
+        uint32 cctpDomain = getCircleDomainId(spokeChainId);
```
pxrl: This has to be conditional on `hasCctpDomain()` because otherwise it'll revert.

broadcast/deployed-addresses.md
pxrl: Implemented: https://github.com/across-protocol/contracts/pull/1228/commits/5917380c21dd1109c72ca1e32c381ae06b687ddf

script/utils/ExtractDeployedFoundryAddresses.ts
```diff
@@ -178,29 +178,33 @@ function extractContractAddresses(broadcastFile: BroadcastFile): Contract[] {
...
+            // nb. This is fragile. @todo: Improve.
+            switch (contractName) {
+              case "Universal_Adapter":
+                cctpDomainId = tx.arguments.at(3);
+                oftDstEid = tx.arguments.at(5);
+                break;
+              case "OP_Adapter":
+                cctpDomainId = tx.arguments.at(6);
+                break;
+            }
```
pxrl, line 194: @fusmanii I extended this to support the OP adapter contract, but we're relying on a hard coupling of constructor argument ordering here. I've just extended what we already have but wondered if you had any thoughts about how it could be improved. wdyt?

broadcast/deployed-addresses.json
pxrl: Resolved in this commit: https://github.com/across-protocol/contracts/pull/1228/commits/5917380c21dd1109c72ca1e32c381ae06b687ddf

script/056DeployOPAdapter.s.sol
pxrl: Addresed here: https://github.com/across-protocol/contracts/pull/1228/commits/b3a9f9f8534381e2bd37ac0e7d75c7f5e4c439d1

broadcast/056DeployOPAdapter.s.sol/1/run-latest.json
fusmanii: This is fine since we want to track the latest deployments and they are needs to generate `broadcast/deployed-addresses.json`

broadcast/deployed-addresses.md
fusmanii: Nice catch, this is known problem, it only happens in the md file and the script needs to be updated to only take the latest one

script/utils/ExtractDeployedFoundryAddresses.ts
fusmanii: Yes I agree its a bit fragile. I think a better approach would be to have the deploy script save the chain id and then have this script read that so that its not relying argument ordering

bmzig (approved):

PR #1229, opened 2025-12-23
[WIP] feat: Support OFT in OP_Adapter & OP_SpokePool

pxrl:


contracts/OP_SpokePool.sol
pxrl, line 1: nb. OP_SpokePool deployment script still needs to be updated.

PR #1231, opened 2026-01-06
feat: Migrate SP1Helios

tbwebb22:
This PR adds the SP1Helios contract, tests, and a Foundry deployment script. The deployment script downloads the appropriate genesis binary from our sp1-helios GitHub releases (detecting macOS vs Linux automatically), verifies the calculated binary checksum against checksums.json, and runs it to generate the initial light client configuration, and then deploys the SP1Helios contract with those parameters. 

A new script/universal folder is added here that includes a readme on the full deployment process of SP1Helios & UniversalSpokePool to a new chain. 

@bmzig @pxrl when you get a chance, please give a test run through the deployment process outlined in the new readme and lmk if there are any gaps in the described process.

### DIFF
--- a/script/universal/README.md
+++ b/script/universal/README.md
@@ -1,260 +1,30 @@
-# Universal Deployment Scripts
+# Universal SpokePool & SP1Helios Deployment
 
-This directory contains deployment scripts for the **Universal SpokePool** infrastructure—a chain-agnostic spoke pool system that uses SP1 zero-knowledge proofs for cross-chain message verification.
-
-## Scripts Overview
-
-| Script                           | Purpose                                     |
-| -------------------------------- | ------------------------------------------- |
-| `DeploySP1Helios.s.sol`          | Deploys the SP1Helios light client contract |
-| `DeployUniversalSpokePool.s.sol` | Deploys the Universal_SpokePool contract    |
-
-## Deployment Order
-
-For a new chain, deploy in this order:
-
-1. **SP1Helios** - The light client that verifies Ethereum L1 state
-2. **Universal_SpokePool** - The spoke pool that uses Helios for admin verification
-
----
-
-# SP1Helios Deployment
-
-The `DeploySP1Helios.s.sol` script deploys the SP1Helios contract—a light client that verifies Ethereum consensus state using SP1 zero-knowledge proofs.
-
-## What It Does
-
-1. Downloads a pre-built genesis binary from GitHub releases
-2. Verifies the binary's SHA256 checksum for security
-3. Runs the binary to generate genesis parameters from the Ethereum consensus layer
-4. Deploys the SP1Helios contract with those parameters
+This guide covers deploying the **Universal SpokePool** infrastructure to a new chain. The system uses SP1 zero-knowledge proofs via the SP1Helios light client for cross-chain message verification.
 
 ## Prerequisites
 
 - **Foundry** installed with FFI enabled
-- **curl** available in PATH (for downloading binaries)
-- **sha256sum** available in PATH (for checksum verification)
+- **curl** and **sha256sum** available in PATH
 - Access to Ethereum consensus layer RPC endpoints (beacon nodes)
 - A funded deployer wallet (via mnemonic)
+- `generated/constants.json` with chain configuration (run `yarn generate-constants` if missing)
 
 ## Environment Variables
 
-| Variable                  | Required | Description                                                                     |
-| ------------------------- | -------- | ------------------------------------------------------------------------------- |
-| `MNEMONIC`                | Yes      | BIP-39 mnemonic to derive the deployer's private key (uses index 0)             |
-| `SP1_RELEASE`             | Yes      | Genesis binary version (e.g., `0.1.0-alpha.20`)                                 |
-| `SP1_PROVER_MODE`         | Yes      | SP1 prover type: `mock`, `cpu`, `cuda`, or `network`                            |
-| `SP1_VERIFIER_ADDRESS`    | Yes      | Address of the SP1 verifier contract (use `0x0` to auto-deploy a mock verifier) |
-| `SP1_STATE_UPDATERS`      | Yes      | Comma-separated list of addresses authorized to submit state updates            |
-| `SP1_VKEY_UPDATER`        | Yes      | Address authorized to update the verification key                               |
-| `SP1_CONSENSUS_RPCS_LIST` | Yes      | Comma-separated list of Ethereum consensus (beacon) RPC URLs                    |
+Create a `.env` file with the following variables:
 
-### Example `.env` for SP1Helios
-
-```bash
-MNEMONIC="your twelve word mnemonic phrase goes here with spaces"
-SP1_RELEASE="0.1.0-alpha.20"
-SP1_PROVER_MODE="network"
-SP1_VERIFIER_ADDRESS="0x397A5f7f3dBd538f23DE225B51f532c34448dA9B"
-SP1_STATE_UPDATERS="0xf7bac63fc7ceacf0589f25454ecf5c2ce904997c"
-SP1_VKEY_UPDATER="0x9a8f92a830a5cb89a3816e3d267cb7791c16b04d"
-SP1_CONSENSUS_RPCS_LIST="https://ethereum-mainnet.core.chainstack.com/beacon/YOUR_KEY,https://lodestar-mainnet.chainsafe.io"
-```
-
-## Running SP1Helios Deployment
-
-```bash
-forge script script/universal/DeploySP1Helios.s.sol \
-  --rpc-url <RPC_URL> \
-  --broadcast \
-  --verify \
-  --etherscan-api-key <API_KEY> \
-  --ffi \
-  -vvvv
-```
-
-> **Note**: The `--ffi` flag is **required** as the script uses Foundry's FFI to execute shell commands (curl, chmod, sha256sum, and the genesis binary).
-
-## SP1Helios Behind the Scenes
-
-### Step 1: Platform Detection
-
-The script detects your operating system using `uname -s` and selects the appropriate binary:
-
-- **macOS**: `genesis_{version}_arm64_darwin`
-- **Linux**: `genesis_{version}_amd64_linux`
-
-### Step 2: Binary Download
-
-The genesis binary is downloaded from:
-
-```
-https://github.com/across-protocol/sp1-helios/releases/download/v{version}/{binary_name}
-```
-
-The binary is saved to `script/universal/genesis-binary`.
-
-### Step 3: Checksum Verification
-
-Before execution, the script verifies the downloaded binary against known-good checksums stored in `checksums.json`:
-
-```json
-{
-  "genesis_0.1.0-alpha.20_amd64_linux": "312d253ce...",
-  "genesis_0.1.0-alpha.20_arm64_darwin": "38607203e..."
-}
-```
-
-If the checksum doesn't match, the script aborts.
-
-### Step 4: Genesis Binary Execution
-
-The genesis binary connects to the Ethereum consensus layer RPCs and:
-
-1. Fetches the current sync committee information
-2. Gets the latest finalized header
-3. Retrieves execution state roots
-4. Generates the SP1 verification key
-
-The output is written to `script/universal/genesis.json` containing:
-
-- `executionStateRoot` - Current execution layer state root
-- `genesisTime` - Beacon chain genesis timestamp
-- `head` - Current head slot number
-- `header` - Beacon block header hash
-- `heliosProgramVkey` - SP1 program verification key
-- `secondsPerSlot` - Slot duration (12 seconds on mainnet)
-- `slotsPerEpoch` - Slots per epoch (32 on mainnet)
-- `slotsPerPeriod` - Slots per sync committee period (8192 on mainnet)
-- `syncCommitteeHash` - Current sync committee commitment
-- `verifier` - SP1 verifier contract address
-- `vkeyUpdater` - Address that can update the vkey
-- `updaters` - Array of state updater addresses
-
-### Step 5: Contract Deployment
-
-The script reads `genesis.json` and deploys the SP1Helios contract with the genesis parameters. If `SP1_VERIFIER_ADDRESS` was set to `0x0`, it automatically deploys a `SP1MockVerifier` for testing purposes.
+| Variable                  | Description                                                                     |
+| ------------------------- | ------------------------------------------------------------------------------- |
+| `MNEMONIC`                | BIP-39 mnemonic to derive the deployer's private key (uses index 0)             |
+| `SP1_RELEASE`             | Genesis binary version (e.g., `0.1.0-alpha.20`)                                 |
+| `SP1_PROVER_MODE`         | SP1 prover type: `mock`, `cpu`, `cuda`, or `network`                            |
+| `SP1_VERIFIER_ADDRESS`    | Address of the SP1 verifier contract (use `0x0` to auto-deploy a mock verifier) |
+| `SP1_STATE_UPDATERS`      | Comma-separated list of addresses authorized to submit state updates            |
+| `SP1_VKEY_UPDATER`        | Address authorized to update the verification key                               |
+| `SP1_CONSENSUS_RPCS_LIST` | Comma-separated list of Ethereum consensus (beacon) RPC URLs                    |
 
 ---
-
-# Universal SpokePool Deployment
-
-The `DeployUniversalSpokePool.s.sol` script deploys the Universal_SpokePool contract—a chain-agnostic spoke pool that verifies admin messages from L1 using the SP1Helios light client.
-
-## What It Does
-
-1. Reads deployment configuration from `constants.json`
-2. Automatically detects chain-specific addresses (USDC, CCTP, wrapped native token, etc.)
-3. Deploys the Universal_SpokePool implementation and proxy
-4. Initializes the proxy with the HubPool as cross-domain admin
-
-## Prerequisites
-
-- **Foundry** installed
-- `generated/constants.json` file with chain configuration
-- SP1Helios already deployed on the target chain
-- A funded deployer wallet (via mnemonic)
-
-## Environment Variables
-
-| Variable   | Required | Description                                                         |
-| ---------- | -------- | ------------------------------------------------------------------- |
-| `MNEMONIC` | Yes      | BIP-39 mnemonic to derive the deployer's private key (uses index 0) |
-
-## Running Universal SpokePool Deployment
-
-The script requires an `oftFeeCap` parameter that sets the maximum fee for OFT (Omnichain Fungible Token) transfers.
-
-### Dry Run (Simulation)
-
-```bash
-source .env
-forge script script/universal/DeployUniversalSpokePool.s.sol:DeployUniversalSpokePool \
-  --sig "run(uint256)" <OFT_FEE_CAP> \
-  --rpc-url <RPC_URL> \
-  -vvvv
-```
-
-### Broadcast to Network
-
-```bash
-source .env
-forge script script/universal/DeployUniversalSpokePool.s.sol:DeployUniversalSpokePool \
-  --sig "run(uint256)" <OFT_FEE_CAP> \
-  --rpc-url <RPC_URL> \
-  --broadcast \
-  -vvvv
-```
-
-### Example with Specific Values
-
-```bash
-# OFT_FEE_CAP of 78000 (example value)
-forge script script/universal/DeployUniversalSpokePool.s.sol:DeployUniversalSpokePool \
-  --sig "run(uint256)" 78000 \
-  --rpc-url $NODE_URL \
-  --broadcast \
-  -vvvv
-```
-
-## Universal SpokePool Behind the Scenes
-
-### Step 1: Load Deployment Configuration
-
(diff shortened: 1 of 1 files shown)


hardhat.config.ts
```diff
@@ -4,10 +4,13 @@ const { TASK_COMPILE_SOLIDITY_GET_SOURCE_PATHS } = require("hardhat/builtin-task
 subtask(TASK_COMPILE_SOLIDITY_GET_SOURCE_PATHS).setAction(async (_: any, __: any, runSuper: any) => {
   const paths = await runSuper();
 
+  // Filter out sp1-helios contracts (uses Foundry-only git submodule imports)
+  const filteredPaths = paths.filter((p: any) => !p.includes("contracts/sp1-helios"));
```
tbwebb22, line 8: filters out SP1Helios from getting compiled by hardhat since it uses the sp1helios-contracts submodule rather than an npm package. Maybe there is a better solution for this?

contracts/sp1-helios/SP1Helios.sol
```diff
@@ -0,0 +1,319 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.28;
```
tbwebb22: Changed from `pragma solidity 0.8.28;`

contracts/sp1-helios/SP1Helios.sol
```diff
@@ -0,0 +1,319 @@
...
+    /// @dev Verifies an SP1 proof and updates the light client state based on the proof outputs
+    /// @param proof The proof bytes for the SP1 proof
+    /// @param publicValues The public commitments from the SP1 proof
+    function update(bytes calldata proof, bytes calldata publicValues) external onlyRole(STATE_UPDATER_ROLE) {
+        // Parse the outputs from the committed public values associated with the proof.
+        ProofOutputs memory po = abi.decode(publicValues, (ProofOutputs));
+
+        uint256 fromHead = po.prevHead;
+        // Ensure that po.newHead is strictly greater than po.prevHead
+        if (po.newHead <= fromHead) revert NonIncreasingHead(po.newHead);
```
tbwebb22: Changed from `require(po.newHead > fromHead, NonIncreasingHead(po.newHead));`, along with other reverts throughout the contract. 

In this codebase we are stripping revert strings - this was causing tests checking for reverts with custom error + data to fail

contracts/sp1-helios/SP1Helios.sol
```diff
@@ -0,0 +1,319 @@
...
+
+        bytes32 storedPrevHeader = headers[fromHead];
+        if (storedPrevHeader == bytes32(0)) revert PreviousHeaderNotSet(fromHead);
```
tbwebb22: changed from
`require(storedPrevHeader != bytes32(0), PreviousHeaderNotSet(fromHead));`

contracts/sp1-helios/SP1Helios.sol
```diff
@@ -0,0 +1,319 @@
...
+        if (storedPrevHeader != po.prevHeader) revert PreviousHeaderMismatch(po.prevHeader, storedPrevHeader);
```
tbwebb22: changed from
`
        require(
            storedPrevHeader == po.prevHeader,
            PreviousHeaderMismatch(po.prevHeader, storedPrevHeader)
        );
`

contracts/sp1-helios/SP1Helios.sol
```diff
@@ -0,0 +1,319 @@
...
+
+        // Check if the head being proved against is older than allowed.
+        if (block.timestamp - slotTimestamp(fromHead) > MAX_SLOT_AGE) revert PreviousHeadTooOld(fromHead);
```
tbwebb22: changed from
`require(
            block.timestamp - slotTimestamp(fromHead) <= MAX_SLOT_AGE, PreviousHeadTooOld(fromHead)
        );`

contracts/sp1-helios/SP1Helios.sol
```diff
@@ -0,0 +1,319 @@
...
+
+        uint256 currentPeriod = getSyncCommitteePeriod(fromHead);
+
+        // Note: We should always have a sync committee for the current head.
+        // The "start" sync committee hash is the hash of the sync committee that should sign the next update.
+        bytes32 currentSyncCommitteeHash = syncCommittees[currentPeriod];
+        if (currentSyncCommitteeHash != po.startSyncCommitteeHash)
+            revert SyncCommitteeStartMismatch(po.startSyncCommitteeHash, currentSyncCommitteeHash);
```
tbwebb22: changed from
`require(currentSyncCommitteeHash == po.startSyncCommitteeHash,
SyncCommitteeStartMismatch(po.startSyncCommitteeHash, currentSyncCommitteeHash)
);`

contracts/sp1-helios/SP1Helios.sol
```diff
@@ -0,0 +1,319 @@
...
+            syncCommittees[newPeriod] = po.syncCommitteeHash;
+            emit SyncCommitteeUpdate(newPeriod, po.syncCommitteeHash);
+        }
+        // Set next period's sync committee hash if value exists.
+        if (po.nextSyncCommitteeHash != bytes32(0)) {
+            uint256 nextPeriod = newPeriod + 1;
+
+            // If the next sync committee is already correct, we don't need to update it.
+            if (syncCommittees[nextPeriod] != po.nextSyncCommitteeHash) {
+                if (syncCommittees[nextPeriod] != bytes32(0)) revert SyncCommitteeAlreadySet(nextPeriod);
```
tbwebb22: changed from
`require(syncCommittees[nextPeriod] == bytes32(0), SyncCommitteeAlreadySet(nextPeriod));`

contracts/sp1-helios/SP1Helios.sol
```diff
@@ -0,0 +1,319 @@
...
+    }
+
+    /// @notice Updates the helios program vkey
+    /// @dev Only callable by the VKEY_UPDATER_ROLE.
+    /// @param newHeliosProgramVkey The new helios program vkey
+    function updateHeliosProgramVkey(bytes32 newHeliosProgramVkey) external onlyRole(VKEY_UPDATER_ROLE) {
+        bytes32 oldHeliosProgramVkey = heliosProgramVkey;
+        heliosProgramVkey = newHeliosProgramVkey;
+
+        if (oldHeliosProgramVkey == newHeliosProgramVkey) revert VkeyNotChanged(newHeliosProgramVkey);
```
tbwebb22: changed from
`require(oldHeliosProgramVkey != newHeliosProgramVkey, VkeyNotChanged(newHeliosProgramVkey));`

hardhat.config.ts
grasphoper: Let's add it to where `!p.includes("contracts/periphery/mintburn")` and `!p.includes("contracts/external/libraries/BytesLib.sol")` conditions are

contracts/sp1-helios/SP1Helios.sol
grasphoper: We shouldn't be changing audited contracts, even with something minor as this. We should rather change the tests to accomodate. In `hardhat`, we were using different strip settings for debug vs prod builds. Alternatively, we could use something similar here.

Although that wasn't perfect because it was causing constant rebuilds of the whole repo.

hardhat.config.ts
tbwebb22: but then it would only get filtered out when `IS_TEST` is true, if `IS_TEST` is false hardhat compilation would fail for it right?

contracts/sp1-helios/SP1Helios.sol
tbwebb22: Yeah totally makes sense. I've reverted all the require statement changes. So now the only change from the SP1Helios repo is the `pragma solidity ^0.8.28;`. Will look into solutions for the tests

contracts/sp1-helios/SP1Helios.sol
fusmanii: we might still need to get it audited since we would be deploying it under 0.8.30 and it can have unexpected behaviour

grasphoper: Is there a way we can use some path-conditional compiler settings in foundry to still just compile with 0.8.28?

I.e., if path-to-sp1-helios, compile with 0.8.28, else compile with default repo version

hardhat.config.ts
grasphoper: Ah I see, you're right. And the remapping doesn't fix the hardhat compilation?
We had `forge-std/=lib/forge-std/src/`: was forge std being used somehow? We didn't need to filter that out before, so we might be able to figure this out too?

Or is this one different?

contracts/sp1-helios/SP1Helios.sol
grasphoper: @tbwebb22 take a look at https://getfoundry.sh/config/reference/solidity-compiler/ : `compilation_restrictions`

hardhat.config.ts
tbwebb22: yeah hardhat doesn't look at the remappings afaik. 
`forge-std` is just imported in Foundry tests, so the hardhat compilation should never encounter it

going to try this out to see if it enables hardhat to use the foundry remappings:
https://v2.hardhat.org/hardhat-runner/plugins/nomicfoundation-hardhat-foundry

contracts/sp1-helios/SP1Helios.sol
tbwebb22: @grasphoper @fusmanii lmk if you guys are good with this solution for being able to still test revert strings. This creates a separate cache and contract artifact folder for contracts that are compiled with the local test profile, and doesn't strip revert strings for these.

So these tests are run with `FOUNDRY_PROFILE=local forge test`
The downside is if you run `forge tests`, any tests that are checking revert strings or custom errors returning data will fail.

https://github.com/across-protocol/contracts/pull/1231/commits/36ebd2e1cad7e5b7a46110b4c652a08dfab541ea

contracts/sp1-helios/SP1Helios.sol
tbwebb22: Haven't been able to get the compilation restrictions to work with two different solc versions - so went back to `SP1Helios` at `pragma solidity ^0.8.28;`. So we should get it audited for this change

hardhat.config.ts
tbwebb22: I tested this plugin and it did do what we're looking for. But it also required bumping the hardhat version, and that for some reason broke a lot of tests. So I reverted back - I think since we're eventually getting rid of hardhat, the current solution of skipping hardhat compilation SP1Helios is fine.

cache-foundry-local/solidity-files-cache.json
grasphoper, line 1: This file seems out of place

grasphoper: Or maybe should just be ignored

tbwebb22: Yep! Its ignored now

hardhat.config.ts
grasphoper: Agree

contracts/sp1-helios/SP1Helios.sol
grasphoper: This solution looks good to me. We're supposed to be running local foundry tests with a `yarn test-evm-foundry` command, or a FOUNDRY_PROFILE=local flag for that matter too.

For reference, were there a lot of tests that are failing because of the strings being stripped? I see there are code changes to change a single test in the commit above, but I think you just didn't start fixing the rest right?

Eventually (or now) we might want to rename this profile to something else, e.g. `local-test`?

script/universal/DeploySP1Helios.s.sol
```diff
@@ -0,0 +1,202 @@
...
+        // Run the genesis binary with all required env vars passed directly
+        string[] memory runCmd = new string[](9);
+        runCmd[0] = "env";
+        runCmd[1] = "SOURCE_CHAIN_ID=1";
+        runCmd[2] = string.concat("SP1_PROVER=", sp1Prover);
+        runCmd[3] = string.concat("PRIVATE_KEY=", privateKeyHex);
+        runCmd[4] = string.concat("SP1_VERIFIER_ADDRESS=", sp1VerifierAddress);
+        runCmd[5] = string.concat("UPDATERS=", stateUpdaters);
+        runCmd[6] = string.concat("VKEY_UPDATER=", vkeyUpdater);
+        runCmd[7] = string.concat("CONSENSUS_RPCS_LIST=", consensusRpcsList);
```
grasphoper, line 180: The variables we're assigning here are read via `vm.envString` above. Why do we need to reassign here? Is the environment not inherited by this process that we're launching?

script/universal/DeploySP1Helios.s.sol
```diff
@@ -0,0 +1,202 @@
...
+        downloadCmd[5] = downloadUrl;
+        vm.ffi(downloadCmd);
+        console.log("Download complete");
+
+        // Make executable
+        string[] memory chmodCmd = new string[](3);
+        chmodCmd[0] = "chmod";
+        chmodCmd[1] = "+x";
+        chmodCmd[2] = binaryPath;
+        vm.ffi(chmodCmd);
```
grasphoper, line 165: We're downloading a binary and executing it immediately. Feels a bit scary. I suggest we do some kind of a checksum check

2 options I see are:
- developer provides a checksum as a script arg (that they get, say, from a release page Web UI) _or_
- we donwload a checksum file simultaneously (from the same resource, or from where?) and check against that
- or any other ideas to make this a bit more safe

contracts/sp1-helios/SP1Helios.sol
tbwebb22: So initially, no tests were failing - all the foundry tests expected strings to be stripped. Then when I migrated the SP1Helios tests over - 4 Sp1Helios tests started failing as they expected revert strings to be there. When turning off revert string stripping, the 4 SP1Helios tests passed, but that one MulticallUpgradeable test modified in this PR began to fail. So I modified that test to expect revert strings and all are passing now. Hopefully that makes sense.

script/universal/DeploySP1Helios.s.sol
tbwebb22: They're reassigned here because the binary expects different env var names than what I wanted to use in this repo.  I wanted the additional env vars in this repo to all be prefixed with "SP1_". Also the binary expects a private key, while the contracts repo env has mnemomic, so pk needs to be derived from that and passed to the binary.

contracts/sp1-helios/SP1Helios.sol
tbwebb22: @grasphoper @fusmanii I updated this contract to `pragma solidity 0.8.30` to lock it in at that version since that's what we're getting it audited at.

script/universal/DeploySP1Helios.s.sol
tbwebb22: Yeah this makes sense. What about if we maintain in this repo somewhere a .json file of the latest known good checksums per toolchain binary? That way to change a checksum here requires a PR and multiple people looking at those changes

contracts/sp1-helios/SP1Helios.sol
tbwebb22: Updated to `pragma solidity ^0.8.30` based on conversation with Faisal

script/universal/DeploySP1Helios.s.sol
grasphoper: Ok, sounds good to me

script/universal/DeploySP1Helios.s.sol
grasphoper: Hm, that's actually even safer that just downloading it from the same exact location that we just downloaded the binary from. I like it. How are we getting the hash sums to put in there though? Do we need to release them from the sp1-helios repo anyway? Alternatively, we could sign a binary with a key and use pgp to check the sig. But that's more involved, we'd have to embed the pgp key in the sp1-helios dev env probably?

Anyway, I think your solution is good

contracts/sp1-helios/SP1Helios.sol
grasphoper: Yep, sounds good!

script/universal/DeploySP1Helios.s.sol
grasphoper: Btw, the protection here is more so for the developer machine that's running the binary, so a developer will be looking at updating the checksum all on their own if they need to update before running the deploy script. That still should protect the dev in 99.9% cases

tbwebb22: Yep so my plan going forward is to update the SP1Helios workflow to generate a checksums.txt file on each release - this will have the checksums for each binary. Then I'll add the the checksums.json file to this repo and update the deploy script to verify checksums of downloaded binaries against the json file

grasphoper: `checksums.txt`: curious about this `s` part at the end. Is it not just the genesis binary checksum? Also, I wonder if there's some standard for doing something like this

tbwebb22: So I think there needs to be multiple checksums - assuming the checksum is of the final generated binary file. So ie genesis_1.0.0_amd64_linux would have a different checksum than genesis_1.0.0_arm64_darwin since those files would have different formats and contents

grasphoper: Yeah 100% you're right

script/universal/checksums.json
```diff
@@ -0,0 +1,4 @@
+{
+  "genesis_0.1.0-alpha.20_amd64_linux": "312d253ce634131d5eb5faa89d8843259c87ba585ac7eac4e777aa325fec54f6",
```
grasphoper: This is what I see on the release in the `sp1-helios` repo, am I looking in the wrong place?

```
"genesis_0.1.0-alpha.20_amd64_linux": "996741dfdb79dce9b13fe1e01336ca0fedf93faaafc379f1ca16e1d6360b4e91",
```

script/universal/DeployUniversalSpokePool.s.sol
```diff
@@ -36,7 +35,7 @@ contract DeployUniversalSpokePool is Script, Test, DeploymentUtils {
         vm.startBroadcast(deployerPrivateKey);
 
         uint256 heliosAdminBufferUpdateSeconds = 1 days;
-        address helios = getL2Address(info.spokeChainId, "helios");
+        address helios = getDeployedAddress("SP1Helios", info.spokeChainId, true);
```
grasphoper, line 38: We should move existing `Sp1Helios` deployment information from `sp1-helios` repo to here right?

script/universal/README.md
grasphoper, line 1: 2 comments on this:

1. This `README` seems very long and goes into a lot of details about the internals. Do you think we can make this shorter? Think about the dev. who's going to be deploying the system. They want a succinct README they can follow. Also, about the checklist after deployment. Unless that is exhaustive and 100% up-to-date, we shouldn't keep it. "Trust" in a file like this is very important for it to be used properly. We need to make sure it doesn't include details that are redundant / extra / 80% correct
2. Now that we have the SP1Helios contract and deployment script in the `contracts` repo, do you think it's possible to create a single script that will deploy the full Universal stack for a chain via a single command? Totally fine to do it in a follow-up. We might even complete that script with additional helper commands, like the `update-vkey` command that would generate a multisig multicall TX to send all of the admin acitons to the target universal spokes(just an idea, the single-deploy script is more important)

grasphoper (commented):
Left a few comments

script/universal/checksums.json
tbwebb22: oh nope you're right - checksum there was from alpha.19 release
fixed here https://github.com/across-protocol/contracts/pull/1231/commits/4c9b0888d4a4cd0f26a035829d745051bec2c4d8

script/universal/DeployUniversalSpokePool.s.sol
tbwebb22: All `SP1Helios` deployments from the other repo are already in this repo in `deployments.json`. So whenever `yarn extract-addresses` is run, these deployments are added to `deployed-addresses.json`. However, these deployments were named "Helios" rather than "SP1Helios", so I renamed them here: https://github.com/across-protocol/contracts/pull/1231/commits/0ca30fc6d57d6730a1394f37c18c2d44ddebe0a9

tbwebb22: or do you mean moving the actual broadcast files?

script/universal/README.md
tbwebb22: 1. Yep totally agree, signficantly simplified the readme here: https://github.com/across-protocol/contracts/pull/1231/commits/b5282dee8b08376a62b53f02b07e8fb4bb45cda9
2. For sure, definitely should be possible. I'll build a script for this in a followup PR

script/universal/DeployUniversalSpokePool.s.sol
grasphoper: > All `SP1Helios` deployments from the other repo are already in this repo in `deployments.json`. So whenever `yarn extract-addresses` is run, these deployments are added to `deployed-addresses.json`. However, these deployments were named "Helios" rather than "SP1Helios", so I renamed them here: [0ca30fc](https://github.com/across-protocol/contracts/commit/0ca30fc6d57d6730a1394f37c18c2d44ddebe0a9)

Sounds good, this should be enough :)

PR #1232, opened 2026-01-08
chore: Remove Aleph Zero

fusmanii:
We not longer support Aleph Zero


deployments/deployments.json
grasphoper, line 1: Interesting. We didn't have a SpokePool in deployments for `41455`?. Maybe we removed it in the past already?

grasphoper (approved):

deployments/deployments.json
fusmanii: must have been removed when we first disabled the chain

PR #1237, opened 2026-01-12
chore: HubPool tests migration

tbwebb22:
Migrates HubPool tests from Hardhat to Foundry
Closes ACP-20

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -33,7 +33,7 @@ yarn build-evm                        # Hardhat
 
 # Run tests
 yarn test-evm-foundry                 # Foundry local tests (recommended)
-FOUNDRY_PROFILE=local forge test      # Same as above
+FOUNDRY_PROFILE=local-test forge test      # Same as above
 yarn test-evm-hardhat                 # Hardhat tests (legacy)
 
 # Run specific Foundry tests


test/evm/foundry/local/chain-adapters/Arbitrum_SendTokensAdapter.t.sol
```diff
@@ -34,10 +34,6 @@ contract Arbitrum_SendTokensAdapterTest is HubPoolTestBase {
 
     uint256 constant ARBITRUM_CHAIN_ID = 42161;
 
-    // ============ Test Amounts ============
-
-    uint256 constant AMOUNT_TO_LP = 1000 ether;
```
tbwebb22, line 39: moved to HubPoolTestBase

test/evm/foundry/local/MultiCallerUpgradeable.t.sol
```diff
@@ -102,10 +102,8 @@ contract MultiCallerUpgradeableTest is Test {
                 assert(results[i].success);
             } else {
                 assert(!results[i].success);
-                assertEq(
-                    "", // Error messages are stripped.
-                    results[i].returnData
-                );
+                // Revert data is preserved by tryMulticall
+                assert(results[i].returnData.length > 0);
```
tbwebb22: Updated this as revert strings are no longer stripped in the local foundry test profile

test/evm/hardhat/BondToken.e2e.ts
tbwebb22, line 1: This test had a dependency on `constructSimpleTree` from the deleted `HubPool.ExecuteRootBundle` test, so moved that function into here for now

broadcast/deployed-addresses.md
grasphoper, line 1: I'm a bit lost RE whether this is needed or if a master merge will fix this

tbwebb22: yeah this change certainly isn't intentional from my end

I think if we get this one merged https://github.com/across-protocol/contracts/pull/1262
and then merge master into here, should fix this right?

grasphoper: Yea I think so

test/evm/foundry/local/HubPool_Admin.t.sol
```diff
@@ -0,0 +1,345 @@
...
+        );
+
+        assertTrue(lpToken != address(0), "lpToken should not be zero after enabling");
+        assertTrue(isEnabled, "isEnabled should be true");
+        assertEq(lastLpFeeUpdate, block.timestamp, "lastLpFeeUpdate should be current time");
+
+        // Verify LP token was created (using mock factory, so just verify it's a valid ERC20)
+        MintableERC20 lpTokenContract = MintableERC20(lpToken);
+        assertTrue(bytes(lpTokenContract.symbol()).length > 0, "LP token should have a symbol");
+        assertTrue(bytes(lpTokenContract.name()).length > 0, "LP token should have a name");
```
grasphoper: Should we also verify the exact name here like the hardhat test was doing?

broadcast/deployed-addresses.md
fusmanii: merged

test/evm/foundry/local/HubPool_Admin.t.sol
tbwebb22: https://github.com/across-protocol/contracts/pull/1237/commits/ab70069cb2c7d35c0f0566ec4ac2ede9a240a60d

tbwebb22:
> Nice work! One minor feedback from claude
> 
> ```
> 1. CLAUDE.md Documentation Mismatch - The doc says FOUNDRY_PROFILE=local but the actual profile name is local-test. This will cause all      
>   revert-message tests to fail if someone follows the docs literally.
> ```

Good catch! Fixed https://github.com/across-protocol/contracts/pull/1237/commits/bd1bda0ab9dd555e52bae05407ba60d66181457d

PR #1243, opened 2026-01-15
chore: Bump @solana/kit 2.3 -> 5.4

pxrl:
Kit has moved from 2.2 to 5.4 in the past few months, so we're playing catch-up. I worked progressively through the major releases and found there weren't many changes required, so the jump from 2.3 -> 5.4 has been squashed into a single upgrade.

Change mostly facilitated by Claude.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@across-protocol/contracts",
-  "version": "4.1.23-alpha.0",
+  "version": "4.1.23",
   "author": "UMA Team",
   "license": "AGPL-3.0-only",
   "repository": {


.github/actions/cache-svm-artifacts/action.yml
```diff
@@ -26,7 +26,7 @@ runs:
         # The job that generates the artifacts is responsible for archiving them to the cache tarball. This avoids any
         # conflicts with other caching actions that might have cleaned some of cached contents.
         path: svm-${{ inputs.type }}.tar
-        key: svm-${{ inputs.type }}-${{ runner.os }}-node-${{ steps.resolved-node.outputs.version }}-${{ hashFiles('Cargo.lock', 'programs/**/Cargo.toml', 'programs/**/Xargo.toml', 'programs/**/*.rs') }}
+        key: svm-${{ inputs.type }}-${{ runner.os }}-node-${{ steps.resolved-node.outputs.version }}-${{ hashFiles('yarn.lock', 'Cargo.lock', 'programs/**/Cargo.toml', 'programs/**/Xargo.toml', 'programs/**/*.rs') }}
```
pxrl, line 29: `yarn.lock` added here to ensure that changes in codama versions will trigger a new cache key. Without this change there are type mismatches due to changes in kit between 2.3 and 3.0.

test/svm/SvmSpoke.Fill.AcrossPlus.ts
pxrl, line 1: I preferred to drop types unless they're required. Types are resolving OK dynamically on my end.

src/svm/web3-v2/transactionUtils.ts
```diff
@@ -32,10 +34,11 @@ import { RpcClient } from "./types";
  */
 export const signAndSendTransaction = async (
   rpcClient: RpcClient,
-  transactionMessage: CompilableTransactionMessage & TransactionMessageWithBlockhashLifetime,
+  transactionMessage: BaseTransactionMessage & TransactionMessageWithFeePayer & TransactionMessageWithBlockhashLifetime,
   commitment: Commitment = "confirmed"
 ) => {
   const signedTransaction = await signTransactionMessageWithSigners(transactionMessage);
+  assertIsTransactionWithinSizeLimit(signedTransaction);
```
grasphoper, line 43: before, the TX would fail. Now, we will panic here?
Also, I forget, do downstream repos rely on `contracts` repo to keep production-ready utils, or are these just for testing / prototyping?

Just for the future, it's the same thing that happened with hardhat / foundry.
Hardhat used to generate artifacts, and those became ingrained into all the downstream repos. Changing that became painful.

@Reinis-FRP @md0x do you guys know if there's any way to "export some minimal artifacts" from the contracts repo that e.g. the `sdk` will use to define some production utils? Then the `contracts` repo can worry only about the contracts, and the `sdk` can worry about maintaining the utils

grasphoper (approved):
LGTM, interested to discuss this https://github.com/across-protocol/contracts/pull/1243#discussion_r2696329562

I.e., what repo should be maintaining the TS types and utils for contract interactions? Ideally we keep the `contracts` repo _pure_. But maybe that's impossible / impractical

src/svm/web3-v2/transactionUtils.ts
Reinis-FRP: As far as I know `sdk` mainly depends on codama clients and anchor IDL/types that are generated and exported in `contracts` repo. Everything else is expected to be used for `contracts` testing and prototyping scripts here. I've seen some exceptions to this in `sdk` where we import some utils from `contracts` in `sdk` tests as well as some cctp decoding and some other conversion utils, but those should be agnostic to the solana/kit versioning.

pxrl: > before, the TX would fail. Now, we will panic here?

We have a similar situation in the v3 -> v4 migration; I left some comments here: https://github.com/across-protocol/contracts/pull/1245#discussion_r2698073356

md0x: Yeah, I think what happened is that, in a few places (sdk and possibly others down the line) we ended up reusing functions from the contracts repo simply because it was the fastest way to ship instead of reimplementing them in the sdk.

I agree with the concern: ideally, any utils living in the contracts repo should be internal and for test only, not something downstream repos should use.

Related to that, one change that could make sense is moving all codama client generation logic into the sdk, we didn't intend to use them in the contracts. In the contracts repo those clients have mainly been useful in tests to validate functional equivalence between codama clients and anchor but they don’t necessarily need to live ther. 

So I agree and align with your suggestions @pxrl @grasphoper : contracts should focus on contracts + correctness, and the sdk should have the utils and client for prod with a minimal interface (if any) from the contracts repo.

PR #1245, opened 2026-01-15
chore: Bump @solana/kit 3.0 -> 4.0

pxrl:
Transition kit to 4.0. This turns out to be a pretty small change locally, so it will probably get rolled directly into a single 2.3 -> 4.0 commit.


src/svm/web3-v2/transactionUtils.ts
```diff
@@ -38,6 +39,7 @@ export const signAndSendTransaction = async (
   commitment: Commitment = "confirmed"
 ) => {
   const signedTransaction = await signTransactionMessageWithSigners(transactionMessage);
+  assertIsTransactionWithBlockhashLifetime(signedTransaction);
```
grasphoper, line 42: Why is this required now? Do the types not work out without it?

pxrl: The return type of some functions has been modified in the v4 release, so the asserts are used to narrow types to satisfy tsc. The docs give some guidance:

```
The modifyAndSignTransactions function for a TransactionModifyingSigner must now return a Transaction & TransactionWithLifetime & TransactionWithinSizeLimit. Previously it technically needed to return a type derived from the input TransactionMessage, but this wasn't checked.
```

The use of the assert pattern is also hinted at in another context in the same release:

```
For example, if you want to pass it to a function returned by sendAndConfirmTransactionFactory then you must use isTransactionWithBlockhashLifetime or assertIsTransactionWithBlockhashLifetime to check its lifetime first.
```

grasphoper: Yeah, makes sense!

PR #1246, opened 2026-01-15
chore: Remove leading number from foundry scripts

fusmanii:


pxrl:
@fusmanii meta, but I noticed that the commit hook is really persistent at adding perceived updated deployments. Your PR here seems to be impacted by the same. There seems to be some persistent state and I haven't figured out where it is; this can cause issues when swapping branches (etc) and generally seems like it could be a source of nasty merge conflicts. Do you know how we can make it a bit more graceful?

fusmanii:
@pxrl I think the issue with the commit hook is that it also extracts local/unstaged deployed addresses. Created [this](https://github.com/across-protocol/contracts/pull/1257) to ignore local unstaged files

broadcast/deployed-addresses.json
```diff
@@ -173,10 +173,15 @@
           "block_number": 24220587,
           "transaction_hash": "0x211a6f322257b67f4e1283844af54d26bfccb76557474165de6490673b38b436"
         },
-        "SponsoredCCTPSrcPeriphery": {
-          "address": "0x89004ea51bac007fec55976967135b2aa6e838d4",
-          "block_number": 23977771,
-          "transaction_hash": "0x0b05b7bc28e6d8ea9d9c135bbada908dd2b971e8428054a62598c2cbf9e24baa"
+        "PermissionSplitterProxy": {
```
fusmanii, line 176: These are just being re-ordered

broadcast/deployed-addresses.json
```diff
@@ -514,14 +514,9 @@
           "block_number": 15142204
         },
         "DonationBox": {
-          "address": "0x3d589d40312bf2d20f13cd0af26a11144a9ca844",
-          "block_number": 21341868,
-          "transaction_hash": "0xe617eb954e31dc4305f9ddcd9badd2e0e707e4b9c266e16fddaf8e7c23180fde"
-        },
-        "SponsoredCCTPDstPeriphery": {
-          "address": "0x1c709fd0db6a6b877ddb19ae3d485b7b4add879f",
+          "address": "0x039d62c549f27ead0eb9b567d8776289e5020583",
```
fusmanii, line 517: The address here is changing because there is a collision on `DonationBox` name and after the script renaming the other donation box address is being picked up. I have follow up work to fix this issue

grasphoper (approved):

PR #1251, opened 2026-01-16
fix: revert on invalid signature in receiveMessage to prevent griefing

fusmanii:
This fix addresses ACP-27 where an attacker could grief users by front-running receiveMessage with valid CCTP message/attestation but an invalid signature. Previously, this would cause the transaction to succeed with fallback behavior, consuming the CCTP nonce and causing the user to lose sponsorship and custom EVM actions.

Changes:
- Replace _isQuoteValid() with _validateQuoteOrRevert() that reverts with specific errors (InvalidSignature, InvalidNonce, InvalidDeadline)
- Remove conditional fallback logic since validation now reverts
- Update tests to expect reverts instead of fallback behavior
- Add griefing attack prevention tests

Now when an attacker submits an invalid signature, the entire transaction reverts, preserving the CCTP nonce so the legitimate caller can still submit with the correct signature.

### DIFF
--- a/contracts/periphery/mintburn/sponsored-cctp/SponsoredCCTPDstPeriphery.sol
+++ b/contracts/periphery/mintburn/sponsored-cctp/SponsoredCCTPDstPeriphery.sol
@@ -188,11 +188,10 @@ contract SponsoredCCTPDstPeriphery is BaseModuleHandler, SponsoredCCTPInterface,
         (SponsoredCCTPInterface.SponsoredCCTPQuote memory quote, uint256 feeExecuted) = SponsoredCCTPQuoteLib
             .getSponsoredCCTPQuoteData(message);
 
-        // Validate the quote and the signature.
-        bool isQuoteValid = _isQuoteValid(quote, signature);
-        if (isQuoteValid) {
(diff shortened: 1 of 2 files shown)


contracts/periphery/mintburn/sponsored-cctp/SponsoredCCTPDstPeriphery.sol
```diff
@@ -188,30 +188,26 @@ contract SponsoredCCTPDstPeriphery is BaseModuleHandler, SponsoredCCTPInterface,
...
 
-        // Validate the quote and the signature.
-        bool isQuoteValid = _isQuoteValid(quote, signature);
-        if (isQuoteValid) {
-            _getMainStorage().usedNonces[quote.nonce] = true;
-        }
+        // Validate the quote and the signature. Revert on invalid to prevent griefing attacks
+        // where an attacker provides correct message/attestation but invalid signature.
+        _validateQuoteOrRevert(quote, signature);
+        _getMainStorage().usedNonces[quote.nonce] = true;
```
grasphoper, line 194: Should we avoid an extra `SLOAD` here and just do this before `_validateQuoteOrRevert` and pass `MainStorage` down as an arg?

PR #1252, opened 2026-01-16
feat: Update ZkSync_SpokePool to support new L2 ZkStack L2AssetRouter

grasphoper:
This contract is designed to be used with ZkStack chains that migrate to the new [ZkGateway](https://forum.zknation.io/t/zip-6-prepare-zksync-for-zk-gateway/490#p-1247-changes-to-bridging-architecture-9) system. For example, this will be required for ZkSync (chain ID 324) and Lens.

~I don't know yet whether currently supported  ZkStack chains like Lens will use this new architecture or the current one. Therefore, to deploy hot fixes to Lens we would need to checkout an older commit and deploy that version of the ZkSync_SpokePool.~

~The alternative is introducing a new contract, but I'd like to consolidate changes.~

Lens will be upgraded with ZkSync, according to the ZkSync team.

P.S.: a cherry-pick of https://github.com/across-protocol/contracts/pull/1190 on top of `audit-jan19` branch

### DIFF
--- a/foundry.toml
+++ b/foundry.toml
@@ -25,7 +25,7 @@ src = "contracts/Lens_SpokePool.sol"
 compile = true
 fallback_oz = true
 mode = '3'
-zksolc = "1.5.7"
+zksolc = "1.5.15"
 
 [rpc_endpoints]
 arbitrum = "${NODE_URL_42161}"


contracts/ZkSync_SpokePool.sol
```diff
@@ -170,14 +185,25 @@ contract ZkSync_SpokePool is SpokePool, CircleCCTPAdapter {
...
 
-    function _setZkBridge(ZkBridgeLike _zkErc20Bridge) internal {
-        address oldErc20Bridge = address(zkErc20Bridge);
-        zkErc20Bridge = _zkErc20Bridge;
-        emit SetZkBridge(address(_zkErc20Bridge), oldErc20Bridge);
+    // Implementation from https://github.com/matter-labs/era-contracts/blob/48e189814aabb43964ed29817a7f05aa36f09fd6/l1-contracts/contracts/common/libraries/DataEncoding.sol#L117C14-L117C62
+    function _getAssetId(address _l2TokenAddress) internal view returns (bytes32) {
+        address l1TokenAddress = l2AssetRouter.l1TokenAddress(_l2TokenAddress);
+        return keccak256(abi.encode(l1ChainId, L2_NATIVE_TOKEN_VAULT_ADDR, l1TokenAddress));
+    }
```
grasphoper, line 198: @nicholaspai I'm a bit confused: would `L2_NATIVE_TOKEN_VAULT_ADDR` -style encoding only work for native assets, e.g. ETH? In `_bridgeTokensToHubPool`, we're using it as a catch-all bridge function. Is this correct?

grasphoper: Or I guess a "native asset" is something that was bridged from L1 using a native bridge?

PR #1257, opened 2026-01-19
chore: Only extract staged/tracked deployed addresses

fusmanii:


script/utils/ExtractDeployedFoundryAddresses.ts
```diff
@@ -59,9 +60,53 @@ interface JsonOutput {
...
+ * Get the git repository root directory.
+ */
+function getGitRoot(): string {
+  try {
+    const output = execSync("git rev-parse --show-toplevel", { encoding: "utf8" });
+    return output.trim();
+  } catch (error) {
+    // Fallback to current working directory if not in a git repo
+    return process.cwd();
+  }
```
grasphoper: I don't think we need this fallback here

script/utils/ExtractDeployedFoundryAddresses.ts
```diff
@@ -59,9 +60,53 @@ interface JsonOutput {
...
+      output
+        .trim()
+        .split("\n")
+        .filter(Boolean)
+        .map((f) => path.resolve(gitRoot, f))
+    );
+  } catch (error) {
+    console.error("Error getting tracked files from git:", error);
+    return new Set();
+  }
```
grasphoper: Wdyt if we default to failing (forcing the dev to fix the issue) instead of catchin here?

script/utils/ExtractDeployedFoundryAddresses.ts
fusmanii: done https://github.com/across-protocol/contracts/pull/1257/changes/4f9a06310f668261384ff63c64ea18af7ca22785

script/utils/ExtractDeployedFoundryAddresses.ts
fusmanii: good point, removed both try/catches https://github.com/across-protocol/contracts/pull/1257/changes/4f9a06310f668261384ff63c64ea18af7ca22785

grasphoper (approved):

PR #1258, opened 2026-01-19
chore: Add .claude folder

fusmanii:


.claude/hooks/prevent-env-read.py
```diff
@@ -0,0 +1,37 @@
...
+    if ".env" in file_path or ".env" in path or ".env" in glob:
+        output = {
+            "hookSpecificOutput": {
+                "hookEventName": "PreToolUse",
+                "permissionDecision": "deny",
+                "permissionDecisionReason": "Reading .env files is not allowed for security reasons. Environment variables contain sensitive credentials.",
+            }
+        }
+        print(json.dumps(output))
+        sys.exit(0)
```
grasphoper: We `sys.exit(0)` here too, would that allow the tool call? Should we exit with a non-zero code instead?

fusmanii: nice catch, changed to 2

grasphoper (approved):

PR #1260, opened 2026-01-19
improve: Add AcccessControl to DonationBox

fusmanii:
DonationBox.sol:                                                                                                                             
  - Now uses AccessControl instead of Ownable                                                                                                  
  - Deployer receives only DEFAULT_ADMIN_ROLE (can manage roles)
  - withdraw() requires WITHDRAWER_ROLE instead of onlyOwner                                                                                   
  - Multiple addresses can now be granted WITHDRAWER_ROLE 
  - Use OZ v5 contracts

### DIFF
--- a/script/DeployHyperliquidDepositHandler.s.sol
+++ b/script/DeployHyperliquidDepositHandler.s.sol
@@ -23,6 +23,8 @@ contract DeployHyperliquidDepositHandler is Script, Test {
         // Get the current chain ID
         uint256 chainId = block.chainid;
 
+        address donationBox = 0x1648fC159a5c13c060EFdF44f3CEE9bD184fa168;
+
         // Set up USDH as a supported token for this handler.
         IERC20 usdh = IERC20(0x111111a1a0667d36bD57c0A9f569b98057111111);
         uint32 usdhTokenIndex = 360;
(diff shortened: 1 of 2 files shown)


contracts/chain-adapters/DonationBox.sol
```diff
@@ -1,30 +1,36 @@
...
-contract DonationBox is Ownable {
+contract DonationBox is AccessControl {
     using SafeERC20 for IERC20;
 
+    /// @notice Role for addresses authorized to withdraw funds
+    bytes32 public constant WITHDRAWER_ROLE = keccak256("WITHDRAWER_ROLE");
+
+    constructor() {
+        _grantRole(DEFAULT_ADMIN_ROLE, msg.sender);
+        _grantRole(WITHDRAWER_ROLE, msg.sender);
```
grasphoper: Do you think we need this by default?

```
_grantRole(WITHDRAWER_ROLE, msg.sender);
```

It seems like the deployed should explicitly give themselves a role if they desire.

Although you probably did it to keep this backwards-compatible right? I'm just thinking about situations where we deploy from a dev wallet and will have to renounce that role from dev wallet always.

Wdyt?

grasphoper: I think we can change this if not a lot of our contracts rely on `new DonationBox() -> withdraw from it later` logic. I know there are a few for sure

It's almost like we want two modes available (set self as withdrawer and admin or set noone). Maybe a constructor param ..?

fusmanii: yeah added it to be similar to the current implementation. I think it makes sense to remove it and have the deployer set up the withdrawer as desired

fusmanii: done https://github.com/across-protocol/contracts/pull/1260/changes/85ec50b5b8f3f26a22429930bd55484b20a1e211

script/DeployHyperliquidDepositHandler.s.sol
grasphoper, line 1: https://github.com/across-protocol/contracts/pull/1313 will change this script, so hardcoding here is fine for now. This version has some errors

grasphoper (approved):

PR #1264, opened 2026-01-21
chore: Remove redundant OP stack implementations

pxrl:
These are now handled via the OP_Adapter and OP_SpokePool implementations.


contracts/Cher_SpokePool.sol
pxrl, line 1: This needs to be restored because the OP_SpokePool does not support Circle Bridged USDC.

script/DeployCherSpokePool.s.sol
pxrl, line 1: Restore this (per above).

PR #1266, opened 2026-01-22
chore: migrate more tests

tbwebb22:
Migrates tests from Hardhat to Foundry:
- BondToken tests
- ConfigStore test
- MerkleLib tests
- PermissionSplitterProxy test

Added murky library for building Merkle tree and generating proofs to use in the MerkleTree tests
https://github.com/dmfxyz/murky/

Closes ACP-40 & ACP-41


grasphoper:
It seems like using murky here is good. We have to be careful not to use it as a dependency in real contracts. Basically this is our "dev dependency"

grasphoper:
Yep, all you said makes sense to me

grasphoper:
Should be good to merge

PR #1267, opened 2026-01-23
chore: redeploy spoke pools to deprecate 0x0 output token

bmzig:
This redeploys all spoke pools which are not in this PR: https://github.com/across-protocol/contracts/pull/1276. It also does not include a redeployment of the SvmSpoke.

### DIFF
--- a/contracts/external/interfaces/ICoreDepositWallet.sol
+++ b/contracts/external/interfaces/ICoreDepositWallet.sol
@@ -30,7 +30,11 @@ interface IForwardDepositReceiver {
      * @param destinationId Forwarding-address-specific id used in conjunction with
      * recipient to route the deposit to a specific location.
      */
-    function depositFor(address recipient, uint256 amount, uint32 destinationId) external;
+    function depositFor(
+        address recipient,
+        uint256 amount,
+        uint32 destinationId
+    ) external;
 }
 
 /**
--- a/contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
+++ b/contracts/periphery/mintburn/HyperCoreFlowExecutor.sol
@@ -1184,12 +1184,7 @@ contract HyperCoreFlowExecutor is AccessControlUpgradeable, AuthorizedFundedFlow
         IERC20(token).safeTransfer(msg.sender, amount);
     }
 
-    function sweepOnCore(
-        address token,
-        uint64 amount,
-        uint32 sourceDex,
-        uint32 destinationDex
-    ) external onlyRole(FUNDS_SWEEPER_ROLE) {
+    function sweepOnCore(address token, uint64 amount, uint32 sourceDex, uint32 destinationDex) external onlyRole(FUNDS_SWEEPER_ROLE) {
         HyperCoreLib.transferERC20CoreToCore(
             _getMainStorage().coreTokenInfos[token].coreIndex,
             msg.sender,


contracts/external/interfaces/ICoreDepositWallet.sol
```diff
@@ -30,11 +30,7 @@ interface IForwardDepositReceiver {
      * @param destinationId Forwarding-address-specific id used in conjunction with
      * recipient to route the deposit to a specific location.
      */
-    function depositFor(
-        address recipient,
-        uint256 amount,
-        uint32 destinationId
-    ) external;
+    function depositFor(address recipient, uint256 amount, uint32 destinationId) external;
```
bmzig: @fusmanii It looks like the formatter is acting differently per person. Do you guys normally just live with these changes or is there a way to back out of the pre commit script without changing the script itself?

script/DeployLineaSpokePool.s.sol
```diff
@@ -0,0 +1,83 @@
...
+            usdcAddress, // _l2Usdc
+            cctpTokenMessenger // _cctpTokenMessenger
+        );
+
+        // Initialize deposit counter to very high number of deposits to avoid duplicate deposit ID's
+        // with deprecated spoke pool.
+        // Set hub pool as cross domain admin since it delegatecalls the Adapter logic.
+        bytes memory initArgs = abi.encodeWithSelector(
+            Linea_SpokePool.initialize.selector,
+            1_000_000, // _initialDepositId
```
grasphoper, line 51: Do we always need to be bumping this on new deployments? Or would that only matter if we redeploy proxy + impl together?

script/DeployOPSpokePool.s.sol
```diff
@@ -60,7 +63,7 @@ contract DeployOPSpokePool is Script, Test, DeploymentUtils {
             "OP_SpokePool",
             constructorArgs,
             initArgs,
-            false // implementationOnly
+            true // implementationOnly
```
grasphoper, line 66: Do we have to manually configure it during deployment?

script/DeployZkSyncSpokePool.s.sol
grasphoper, line 1: I'm surprised that you had to create this. I couldn't find a zksync .ts deployment script either in `master`. @fusmanii does this look right to you?

script/DeployOPSpokePool.s.sol
grasphoper: Seems like this flag is to force deploying impl only. If it finds existing proxy, should auto-deploy impl only. Did it not work?

script/DeployZkSyncSpokePool.s.sol
```diff
@@ -0,0 +1,92 @@
...
+            info.hubPool // _withdrawalRecipient
+        );
+
+        // Deploy the proxy
+        DeploymentResult memory result = deployNewProxy(
+            "ZkSync_SpokePool",
+            constructorArgs,
+            initArgs,
+            true // implementationOnly
+        );
```
grasphoper, line 73: @fusmanii out of scope of this PR, but related. `deployNewProxy` uses `Core.deploy`. I get this message when inspecting the code:

Image

Do you think we should switch to using those libs instead?

script/DeployLineaSpokePool.s.sol
bmzig: It should only matter for new proxy deployments since it's not an immutable var. I suppose we could have a smart way to determine what the initial deposit ID is based on what the deployed proxy's current ID is, but I think new proxy deployments are so rare that the marginal benefit of doing that is pretty small.

script/DeployOPSpokePool.s.sol
bmzig: I didn't try... It might have worked if it find a proxy, but I just saw we had `true` in other deploy scripts and I also only needed an implementation so I changed it.

script/DeployZkSyncSpokePool.s.sol
fusmanii: might have been missed when migrating the deploy scripts, this looks good

script/DeployZkSyncSpokePool.s.sol
fusmanii: Looks like we can just use `deployUUPSProxy/deployImplementation` from `Upgrades.sol`

PR #1270, opened 2026-01-24
improve: N-03

grasphoper:


contracts/libraries/HyperCoreLib.sol
```diff
@@ -219,12 +219,12 @@ library HyperCoreLib {
...
-        if (destinationDex != CORE_SPOT_DEX_ID) {
-            action = abi.encode(to, address(0), sourceDex, destinationDex, erc20CoreIndex, amountCore);
-            payload = abi.encodePacked(SEND_ASSET_TO_DEX_HEADER, action);
-        } else {
+        if (destinationDex == CORE_SPOT_DEX_ID && sourceDex == CORE_SPOT_DEX_ID) {
             action = abi.encode(to, erc20CoreIndex, amountCore);
             payload = abi.encodePacked(SPOT_SEND_HEADER, action);
+        } else {
+            action = abi.encode(to, address(0), sourceDex, destinationDex, erc20CoreIndex, amountCore);
+            payload = abi.encodePacked(SEND_ASSET_TO_DEX_HEADER, action);
```
grasphoper, line 227: @fusmanii @tbwebb22 

Before, this function assumed that we're always transferring from SPOT. I changed it such that we're transferring from source to destination dex, as the function signature implies. Will the `SEND_ASSET_TO_DEX_HEADER` action be able to handle this?

fusmanii: In their [docs](https://hyperliquid.gitbook.io/hyperliquid-docs/for-developers/hyperevm/interacting-with-hypercore#action-encoding-details) they don't mention any restrictions so I am assuming as long as the token supports the dex it should work

tbwebb22: Agree it should work

PR #1273, opened 2026-01-26
fix: `HyperliquidDepositHandler`

grasphoper:
This PR addresses all outstanding issues wrt `HyperliquidDepositHandler`:

L-04: fixed in [bcebeeb](https://github.com/across-protocol/contracts/pull/1273/commits/bcebeeb5d00423a858be4a20a9d85f81dd79e6a9)
L-05: didn't fix
L-07: no longer relevant (fixed as a part of #1254 )
N-04: fixed in [96ec161](https://github.com/across-protocol/contracts/pull/1273/commits/96ec161c0e3e1af28bb7e20cdc5be949d6073dec)
(extra issue) CR-01: some tokens cannot be used for account activation. `CoreTokenInfo` provides us with the relevant configuration param and we revert if account activation is requested when the deposit is not in token usable for account activation.

### DIFF
--- a/contracts/handlers/HyperliquidDepositHandler.sol
+++ b/contracts/handlers/HyperliquidDepositHandler.sol
@@ -6,10 +6,9 @@ import "@openzeppelin/contracts-v4/token/ERC20/IERC20.sol";
 import "@openzeppelin/contracts-v4/token/ERC20/utils/SafeERC20.sol";
 import "@openzeppelin/contracts-v4/security/ReentrancyGuard.sol";
 import { ECDSA } from "@openzeppelin/contracts-v4/utils/cryptography/ECDSA.sol";
-import { HyperCoreLib } from "../libraries/HyperCoreLib.sol";
+import { HyperCoreLib, CoreTokenInfo } from "../libraries/HyperCoreLib.sol";
 import { Ownable } from "@openzeppelin/contracts-v4/access/Ownable.sol";
 import { DonationBox } from "../chain-adapters/DonationBox.sol";
-import { CoreTokenInfo, CoreTokenInfoLib } from "../periphery/mintburn/Structs.sol";
 
 /**
  * @title Allows caller to bridge tokens from HyperEVM to Hypercore and send them to an end user's account
@@ -185,7 +184,7 @@ contract HyperliquidDepositHandler is AcrossMessageHandler, ReentrancyGuard, Own
         uint64 accountActivationFeeCore,
         uint64 bridgeSafetyBufferCore
     ) external onlyOwner {
-        CoreTokenInfo memory coreTokenInfo = CoreTokenInfoLib.build(
+        CoreTokenInfo memory coreTokenInfo = HyperCoreLib.buildCoreTokenInfo(
             coreIndex,
             canBeUsedForAccountActivation,
             accountActivationFeeCore,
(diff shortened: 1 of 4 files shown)


contracts/handlers/HyperliquidDepositHandler.sol
```diff
@@ -282,15 +282,15 @@ contract HyperliquidDepositHandler is AcrossMessageHandler, ReentrancyGuard, Own
...
         /// @dev There is no nonce in this signature because an account on Hypercore can only be activated once
         /// by this contract, so reusing a signature cannot be used to grief the DonationBox.
         bytes32 expectedHash = keccak256(abi.encode(expectedUser));
         if (ECDSA.recover(expectedHash, signature) != signer) revert InvalidSignature();
     }
 
-    function _getTokenInfo(address evmAddress) internal view returns (TokenInfo memory) {
-        if (supportedTokens[evmAddress].evmAddress == address(0)) {
+    function _getTokenInfo(address evmAddress) internal view returns (CoreTokenInfo memory) {
+        if (supportedTokens[evmAddress].tokenInfo.evmContract == address(0)) {
```
grasphoper, line 317: Hm, there's been some weirdness around `evmContract` before. Is this check sufficient? What else can we do?

contracts/handlers/HyperliquidDepositHandler.sol
```diff
@@ -244,20 +245,37 @@ contract HyperliquidDepositHandler is AcrossMessageHandler, ReentrancyGuard, Own
...
+        if (
+            !HyperCoreLib.isCoreAmountSafeToBridge(
+                coreTokenInfo.coreIndex,
+                quotedCoreAmount,
+                coreTokenInfo.bridgeSafetyBufferCore
+            )
+        ) {
+            // Bridge doesn't have enough liquidity - fall back to HyperEVM transfer
+            IERC20(token).safeTransfer(user, evmAmount);
+            emit FallbackToHyperEVM(user, token, evmAmount);
```
grasphoper, line 278: Do we fall back to EVM or do we revert and refund the user instead? If feels like this is better because there's no capital lockup

contracts/handlers/HyperliquidDepositHandler.sol
fusmanii: this check is good, we just can't compare `evmAddress` with `tokenInfo.evmContract` since they are not always the same

contracts/periphery/mintburn/Structs.sol
```diff
@@ -20,6 +20,38 @@ struct CoreTokenInfo {
     uint64 bridgeSafetyBufferCore;
 }
 
+library CoreTokenInfoLib {
```
fusmanii: Should this be placed in this file as the name of this file is `Structs.sol`? What do you think about adding this `build` function to `HyperCoreLib`?

tbwebb22: Agree with @fusmanii

grasphoper: Fixed in [95ef73c](https://github.com/across-protocol/contracts/pull/1273/commits/95ef73cfe59a62e2876ddff3852bc4d74162ba70)

Had to also move the `CoreTokenInfo` struct into the lib. This created some moving-code-around diff but wcyd

PR #1275, opened 2026-01-27
feat: move to single signature ERC-3009 and use deterministic depositIds for gasless deposits

mrice32:
This PR makes two primary changes that have some ripple effects on the `SpokePoolPeriphery` contract:
1. We hash the witness data and use it as the ERC-3009 nonce to allow for single-signature auth.
2. We move all gasless deposit methods to using deterministic depositIds, so prefills/simultaneous deposits and fills are possible. This is safe because the gasless mechanisms all guarantee that the nonce is not reused, so this means that we will never have a repeat depositId.

In addition to this, we now do not ignore the nonce passed in the structs. For user-submitted deposits, we require a 0 nonce (which signals an incrementing depositId using the normal deposit method). For Permit2, we now require that the permit2 nonce match the nonce in the witness data, so they can be interchangeably used to compute depositId (reduce confusion).

Relatedly, we also add some convenience methods for computing depositId (and the witness value) to enable easy prefills.


contracts/SpokePoolPeriphery.sol
```diff
@@ -341,9 +345,12 @@ contract SpokePoolPeriphery is SpokePoolPeripheryInterface, ReentrancyGuard, Mul
         SwapAndDepositData calldata swapAndDepositData,
         uint256 validAfter,
         uint256 validBefore,
-        bytes calldata receiveWithAuthSignature,
-        bytes calldata swapAndDepositDataSignature
+        bytes calldata receiveWithAuthSignature
     ) external override nonReentrant {
+        bytes32 witness = keccak256(
+            abi.encodePacked(BRIDGE_AND_SWAP_WITNESS_IDENTIFIER, abi.encode(swapAndDepositData))
+        );
```
fusmanii: how do we ensure `witness` uniqueness given that its now obtained from deposit data?

tbwebb22: My understanding is that the token implementing ERC-3009 will revert on the `receiveWithAuthorization` call if the `witness` has already been used

Also there's a `nonce` in `SwapAndDepositData` thats part of what gets hashed to generate the `witness` - so this can create uniqueness even if all the other intent data is identical

grasphoper: Yeah, my undertsanding was the same as Taylor's, `witness` here is essentially the `nonce` for the ERC-3009 token's permit with auth

From https://eips.ethereum.org/EIPS/eip-3009 reference implementation:

```
        require(
            !_authorizationStates[from][nonce],
            "EIP3009: authorization is used"
        );
```

mrice32: Agreed with the above. There is nonce reuse prevention that is per-user. The nonce within the hashed data means that a user should never have an identical witness hash unless they are accidentally reusing the same API response twice.

contracts/SpokePoolPeriphery.sol
```diff
@@ -470,6 +473,7 @@ contract SpokePoolPeriphery is SpokePoolPeripheryInterface, ReentrancyGuard, Mul
             depositData.baseDepositData.outputAmount,
             depositData.baseDepositData.destinationChainId,
             depositData.baseDepositData.exclusiveRelayer,
+            depositData.nonce, // TODO: should we hash this nonce (if nonzero) with the permit2 nonce to protect the user against nonce reuse?
```
mrice32: @grasphoper @fusmanii @tbwebb22 thoughts?

contracts/SpokePoolPeriphery.sol
```diff
@@ -423,6 +424,7 @@ contract SpokePoolPeriphery is SpokePoolPeripheryInterface, ReentrancyGuard, Mul
             depositData.baseDepositData.outputAmount,
             depositData.baseDepositData.destinationChainId,
             depositData.baseDepositData.exclusiveRelayer,
+            depositData.nonce + 1, // +1 to avoid 0 (which triggers regular deposit) and ensure uniqueness
```
grasphoper: Not sure why this is needed. If the relayer is doing a pre-fill and a deposit after, wouldn't they just submit a non-zero nonce? Feels like it should be their responsibility

grasphoper: Wait, I take it back, the user signed over depositData. Wouldn't we, then, trigger a lot of deposits to have nonce = 1?

I like the solution proposed for the Permit2: hashing this with the permit nonce seems like it'd ensure uniqueness

grasphoper: Also, why are we forcing the user to have a non-zero nonce here, but not in Permit2?

contracts/SpokePoolPeriphery.sol
grasphoper: Hmm, hashing it together sounds right to me. Then the relayer would know to pre-fill with the hash that they can derive easily right

contracts/SpokePoolPeriphery.sol
mrice32: Fair point -- let me see if we can allow 0 nonces on all of them.

contracts/SpokePoolPeriphery.sol
```diff
@@ -564,23 +631,25 @@ contract SpokePoolPeriphery is SpokePoolPeripheryInterface, ReentrancyGuard, Mul
      * @param providedNonce The provided nonce value.
      */
     function _validateAndIncrementNonce(address user, uint256 providedNonce) private {
-        if (permitNonces[user] != providedNonce) {
+        // Pre-increment nonce so it can never be 0.
+        if (++permitNonces[user] != providedNonce) {
```
fusmanii: in future if we re-deploy the periphery contract these nonces will reset and there could be collisions right? might make sense to version the identifiers or have a global version number that gets hashed in `_deposit()`

contracts/SpokePoolPeriphery.sol
```diff
@@ -413,6 +420,7 @@ contract SpokePoolPeriphery is SpokePoolPeripheryInterface, ReentrancyGuard, Mul
         _validateAndIncrementNonce(signatureOwner, depositData.nonce);
         // Verify that the signatureOwner signed the input depositData.
         _validateSignature(signatureOwner, PeripherySigningLib.hashDepositData(depositData), depositDataSignature);
+        // Use nonce + 1 to avoid 0 (which triggers regular deposit) and ensure uniqueness
```
tbwebb22: I think this comment is no longer accurate

contracts/SpokePoolPeriphery.sol
mrice32: But they will be from different msg.senders, so I think we'd be okay.

contracts/SpokePoolPeriphery.sol
mrice32: Fixed.

PR #1283, opened 2026-01-30
feat: USDFree interfaces

grasphoper:
### Project overview

**Why:** 

Improve sponsored system: the current iteration of the sponsored bridging system is limited to a specific use case: sponsored stable bridging into HyperLiquid using mint/burn bridges.

Unify Across product: mint/burn, Across, sponsored Phase0 are currently different parts of the stack that could benefit from unification in terms of both operational and UX sides.

**Goals:**

bridge tokens to other chain in a mechanism-agnostic way: CCTP, OFT, Across etc.

use a single upgradeable entry point for order submission so that the users don't ever have to re-approve and integrators don't have to maintain complicated mappings

support gasless

support best-path swapping on either source or destination or both

support user action after the desired amount of tokens is received on destination (like AAVE deposits or deposits to Hyperliquid Core etc.)

support sponsoring transactions to cover the bridge and / or swap fee partially or in full

support operation chaining: allow performing multiple cross-chain actions in a row

mitigate abuse of destination-side actions (e.g. by using obfuscation)

**Outcomes:** 

Architecture for contracts, API and relayer, showing how the system would function together

Implementation of the above system

Closes ACP-10

### DIFF
--- a/contracts/usdfree/Interfaces.sol
+++ b/contracts/usdfree/Interfaces.sol
@@ -107,24 +107,24 @@ enum OrderFundingType {
     TransferWithAuthorization
 }
 
-abstract contract OrderGateway {
+interface IOrderGateway {
     function submit(
         MerkleOrder calldata order,
         MerkleRoute calldata route,
         // Funding by the party authorizing the order. Funding has the amount of a single ERC20 token
         TypedData calldata orderFunding,
         SubmitterData calldata submitterData
-    ) external payable virtual;
+    ) external payable;
 }
 
-abstract contract Executor {
+interface IExecutor {
     function execute(
         bytes32 orderId,
         TokenAmount calldata userTokenAmount,
         StepAndNext calldata stepAndNext,
         address submitter,
         bytes[] calldata submitterParts
-    ) external payable virtual; // onlyOrderGateway / onlyOrderStore
+    ) external payable; // onlyOrderGateway / onlyOrderStore
 }
 
 interface IUserActionExecutor {
@@ -136,12 +136,8 @@ interface IUserActionExecutor {
     ) external payable;
 }
 
-abstract contract OrderStore {
-    function handle(
-        bytes32 orderId,
-        TokenAmount calldata tokenAmount,
-        TypedData calldata steps
-    ) external payable virtual;
+interface IOrderStore {
+    function handle(bytes32 orderId, TokenAmount calldata tokenAmount, TypedData calldata steps) external payable;
 
     // handle + fill immediately
     function handleAtomic(
@@ -155,7 +151,7 @@ abstract contract OrderStore {
         // submitter address passed
         address submitter,
         SubmitterData calldata submitterData
-    ) external payable virtual;
+    ) external payable;
 
-    function fill(uint256 localOrderIndex, SubmitterData calldata submitterData) external payable virtual;
+    function fill(uint256 localOrderIndex, SubmitterData calldata submitterData) external payable;
 }


contracts/usdfree/Interfaces.sol
```diff
@@ -0,0 +1,127 @@
...
+
+// Entrypoint for all order submissions
+contract OrderGateway {
+    function submit(Order calldata order, SubmitterData calldata submitterData) external payable;
+    function submitGasless(Order calldata order, SubmitterData calldata submitterData, bytes calldata gaslessSig) external payable;
+    function submitWithAuction(
+        OrderWithAuction memory order,
+        SubmitterData calldata submitterData,
+        bytes calldata gaslessSig,
+        AuctionChanges calldata auctionChanges
```
tbwebb22: are there smart contract enforced restrictions on what the auctionChanges can be?

tbwebb22: actually from this comment, seems like the answer is "yes"

> // Auction changes can do things like change the submitter, improve the token requirement or shorten the deadline etc.

contracts/usdfree/Interfaces.sol
```diff
@@ -0,0 +1,127 @@
...
+    bytes tokenReq; // Mandatory: (token, amount). Token has to be correct. Amount can be 0 to mean "no enforcement"
+    // NOTE: if an order is sponsored, this is forced by the API to be one of the trusted relayers, to prevent self-
+    // submission and self-sandwiching on the sponsored orders
+    bytes submitterReq; // None OR address (can be bytes32 for non-EVM chains)
+    bytes deadlineReq; // None OR deadline
+    // A list of other requiremens that can be checked in a static way after the submitter has already executed
+    bytes[] otherStaticReqs;
+    // NOTE: if we want obfuscation to encompass both the requirements and the final action, we can record auction `Changes` 
+    // and apply them after the deobfuscation
+    bytes hashOrFinalAction;
```
tbwebb22: `dstChainId` will be represented in this right?

contracts/usdfree/Interfaces.sol
```diff
@@ -0,0 +1,127 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.30;
+
+struct ExecutionStep {
```
tbwebb22: is `srcChainId` represented anywhere in here?

contracts/usdfree/Interfaces.sol
```diff
@@ -0,0 +1,127 @@
...
+// Contract responsible for executing submitter actions, checking user requirements and executing the final user action
+interface Executor {
+    function execute(
+        // `submitter` is propagated by the caller: either Gateway or `IntentStore`
+        address submitter,
+        // Provided by the submitter to try to meet user's requirements
+        bytes calldata submitterActions, // weiroll
+        // Current and the remaining execution steps, initially defined by the user in `Order.steps`
+        ExecutionStep calldata currentStep,
+        ExecutionStep[] calldata nextSteps
```
tbwebb22: Do these need to be separate args? or could `currentStep` just always be the step at index 0

contracts/usdfree/Interfaces.sol
```diff
@@ -0,0 +1,127 @@
...
+struct ExecutionStep {
+    bytes tokenReq; // Mandatory: (token, amount). Token has to be correct. Amount can be 0 to mean "no enforcement"
+    // NOTE: if an order is sponsored, this is forced by the API to be one of the trusted relayers, to prevent self-
+    // submission and self-sandwiching on the sponsored orders
+    bytes submitterReq; // None OR address (can be bytes32 for non-EVM chains)
+    bytes deadlineReq; // None OR deadline
+    // A list of other requiremens that can be checked in a static way after the submitter has already executed
+    bytes[] otherStaticReqs;
+    // NOTE: if we want obfuscation to encompass both the requirements and the final action, we can record auction `Changes` 
+    // and apply them after the deobfuscation
```
tbwebb22: I understand how auctionChanges are applied on source - `submitWithAuction` on `OrderGateway`. But I'm a little fuzzy on how this happens on dst.

contracts/usdfree/Interfaces.sol
tbwebb22: All my questions were answered in our call, can be disregarded

contracts/usdfree/Interfaces.sol
```diff
@@ -0,0 +1,145 @@
...
+pragma solidity ^0.8.30;
+
+struct ExecutionStep {
+    /*
+    NOTE: `tokenReq` is a mandatory param:
+    - amount == 0 means no enforccement
+    - token is used to inform the final user action. The action is made with (tokenReq.token, balanceOf(tokenReq.token, address(this))) on the executor
+    */
+    bytes tokenReq; // (token, amount)
+    // NOTE: if an order is sponsored, this is forced by the API to be one of the trusted relayers, to prevent self-
```
fusmanii: how does the API enforce this?

contracts/usdfree/Interfaces.sol
```diff
@@ -0,0 +1,145 @@
...
+    bytes actions; // MulticallHandler.Instructions or weiroll (support can be expanded to any format)
+    bytes deobfuscation; // if user's finalAction is obfuscated
+}
+
+// Entrypoint contract for all order submissions
+abstract contract OrderGateway {
+    // NOTE: `funding` here is (token, amount) for approval flows. For gasless flows something like (transfer data object + sig)
+    function submit(
+        Order calldata order,
+        bytes calldata funding,
```
tbwebb22: so the input token and amount aren't part of whats signed by the user right? Thinking through if there are any security implications with this

contracts/usdfree/Interfaces.sol
```diff
@@ -0,0 +1,145 @@
...
+    // and apply them after the deobfuscation (send them to the next steps similarly to how we send `nextSteps`)
+    bytes hashOrUserAction;
+    // NOTE: If the current step hasn't been executed by `deadline`, this address can break the execution chain and withdraw
+    // the assets associated with the order
+    address refundRecipient;
+}
+
+// The struct that the user signs/submits. Contains all of the user's preferences
+struct Order {
+    // NOTE: there's no (token, amount) here that the user submits. That's provided separately at the time of the submit.. call
```
nicholaspai: I think your comment is incomplete here?

contracts/usdfree/Interfaces.sol
```diff
@@ -0,0 +1,145 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.30;
+
+struct ExecutionStep {
+    /*
+    NOTE: `tokenReq` is a mandatory param:
+    - amount == 0 means no enforccement
+    - token is used to inform the final user action. The action is made with (tokenReq.token, balanceOf(tokenReq.token, address(this))) on the executor
+    */
+    bytes tokenReq; // (token, amount)
```
nicholaspai: What's the thinking here that this is type `bytes` if this is always going to be abi.encode(address, uint256)? I guess i'm just wondering why not just explicitly require amount and token instead of squashing them into bytes

contracts/usdfree/Interfaces.sol
```diff
@@ -0,0 +1,145 @@
...
+struct OrderWithAuction {
+    Order order;
+    // `auctionAuthority` has the power to change the execution steps (see `struct Change` below)
+    address auctionAuthority;
+}
+
+// A struct that represents some change in user requirement as a result of the auction
+struct Change {
+    // enum for: token, deadline, submitter or custom
+    uint8 typ;
```
nicholaspai: type?
```suggestion
    uint8 type;
```

contracts/usdfree/Interfaces.sol
```diff
@@ -0,0 +1,145 @@
...
+    // NOTE: if a type is custom, data can inlcude an index of the otherStaticReqs that the auction wants to change. The
```
nicholaspai:
```suggestion
    // NOTE: if a type is custom, data can include an index of the otherStaticReqs that the auction wants to change. The
```

contracts/usdfree/Interfaces.sol
```diff
@@ -0,0 +1,145 @@
...
+    // funtion to apply the change has to be implemented on the OrderGateway for this to work
```
nicholaspai:
```suggestion
    // function to apply the change has to be implemented on the OrderGateway for this to work
```

contracts/usdfree/Interfaces.sol
```diff
@@ -0,0 +1,145 @@
...
+    uint256 amount;
+}
+
+// Provided to support the execution of a single step
+struct SubmitterData {
+    // ? TODO: hard to create relevant good `witness`, as `orderId` is only present on initial submission. Maybe there's a way?
+    // NOTE: only approval-based token pulls are supported from submitters
+    TokenAmount[] extraFunding;
+    bytes actions; // MulticallHandler.Instructions or weiroll (support can be expanded to any format)
+    bytes deobfuscation; // if user's finalAction is obfuscated
```
nicholaspai: how would deobfuscation bytes work?

contracts/usdfree/Interfaces.sol
```diff
@@ -0,0 +1,145 @@
...
+        ExecutionStep[] calldata nextSteps
+    ) external payable virtual;
+}
+
+// NOTE: it is a responsibility of `IUserActionExecutor` to propagate `nextSteps` to the future execution layer
+interface IUserActionExecutor {
+    function execute(
+        // `ExecutionStep.tokenReq.token`
+        address token,
+        // `balanceOf(address(IExecutor), ExecutionStep.tokenReq.token)`
```
nicholaspai: isn't the balanceOf step vulnerable to someone dropping tokens on to the contract?

contracts/usdfree/README.md
```diff
@@ -0,0 +1,140 @@
...
+- `submit()` - Submit order with funding
+- `submitWithAuction()` - Submit with auction-based price improvement
+
+**Executor** - Executes a single step
+
+- Runs submitter actions (swaps, etc.)
+- Checks requirements after submitter actions complete
+- Calls IUserActionExecutor with token, amount, params, and remaining steps
+
+**IUserActionExecutor** - Final action interface
```
nicholaspai: Why is this called an I-contract, and what's the reasoning behind the naming? The executor/userActionExecutor names aren't the most intuitive to me

contracts/usdfree/README.md
```diff
@@ -0,0 +1,140 @@
...
+}
+```
+
+## Execution Flow
+
+1. **OrderGateway** computes orderId from Order + domain separation
+2. **OrderGateway** applies auction changes if present (signed by auctionAuthority)
+3. **Executor** runs submitter actions
+4. **Executor** checks all requirements are met (token balance, submitter, deadline, etc.)
+5. **Executor** calls **IUserActionExecutor** with current step's action
```
nicholaspai: So you keep mentioning in the docs that the Executor executes a single step. This implies that an action might have multiple steps. How would the execution flow look like for a multi step action? And what would a multi step action even do, in practice?

contracts/usdfree/Interfaces.sol
grasphoper: Can't use `type`, it's a reserved keyword

contracts/usdfree/Interfaces.sol
grasphoper: To make this extensible without changing the format of the `ExecutionStep`. I think the idea is more so make it

```
(tokenReqTypeEnum, someOtherData)
```

Which, to start can be something like (0, (token, amount))

But also could be something like an onchain dutch auction: (1, (startingPrice, startingTime, endPrice))

This is interpreted and checked by the internal function on the `Executor` contract, so as we expand it's functionality, we can add more types to the token requirement

contracts/usdfree/Interfaces.sol
grasphoper: The `OrderStore` handles the deobfuscation. If length `hashOrUserAction` is 32 bytes, safe to assume it's a hash. Or we could just add a single byte at the start to specify whether this is obfuscated to start with.

Then the `OrderStore` receives deobfuscated data as a part of submitter data, checks against the hash and propagates to `Executor`. So the `Executor` never deals with deobfusaction

contracts/usdfree/Interfaces.sol
grasphoper: Is it a problem? `Executor` contract is not at all suitable for escrowing any funds (it allows submitter to do anything with the funds on the contract as long as the `tokenReq` is met). Seems fine

contracts/usdfree/Interfaces.sol
grasphoper: For sponsored orders, API is fully responsible for keeping it's own database of `orderId => sponsorshipAmount`, so when `submitterReq` is not one of the trusted entities, the API just doesn't put the `orderId` into the DB

contracts/usdfree/Interfaces.sol
grasphoper: It is complete. `submit..` here just means `OrderGateway.submit` or `OrderGateway.submitWithAuction`

When submission is direct by the user, it's an approval-based flow and relies on approvals and `(token, amount)` decoded from `funding` argument.

If submission is gasless, `funding` still holds the relevant information, e.g. in `IPermit2.PermitTransferFrom`:

```
    struct TokenPermissions {
        address token;
        uint256 amount;
    }

    struct PermitTransferFrom {
        TokenPermissions permitted;
        uint256 nonce;
        uint256 deadline;
    }
```

contracts/usdfree/README.md
grasphoper: I-contract to show that it's an interface and there are many implementations possible, in contrast to something like an `Executor`, which is a single contract.

As for name intuitiveness, these are open to change, open to suggestions!

contracts/usdfree/README.md
grasphoper: Single step is executed on one chain. A step is:

- submitter actions
- check user reqs
- user action

What could 2 steps do?

For example, mint-burn + deposit into Hyperliquid:

[srcStep, dstStep]

- srcStep: tokenReq == input amount (so no submitter actions needed). user action = deposit into CCTP
- dstStep: check tokenReq (e.g. input from source - a few bps) + call `HyperliquidDepositHandler.depositToHypercore`

Alternatively, a single-step similar thing could be:

[srcStep]

tokenReq == input amount, user action == deposit into spokepool such that message = HyperliquidDepositHandler interaction

In the case where there are 2 actions, mint-burn dst contracts are interacting with `OrderStore`. `OrderStore` on dst is analogous to `OrderGateway` on src, expect it doesn't apply auction changes and doesn't calculate `orderId`

PR #1287, opened 2026-01-31
fix: fix depositId reuse bug in SpokePoolPeriphery

mrice32:
There is a subtle bug that makes it possible to reuse `depositIds` in the SpokePoolPeriphery. The code currently relies on whatever fund acquisition system is being used (Permit/Permit2/ERC-3009) to ensure that depositIds are unique. This works _as long as_ the signer == depositor (specified in the input struct). However, an authorizer is free to specify a different depositor. This depositor could, in theory, be the same depositor specified by some other authorizer. If both authorizers use the same nonce as well.

If that happens, the SpokePool will hash the msg.sender (this contract) with the same depositor and nonce, creating a collision.

To prevent this outcome, we include the authorizer in the derivation of the nonce passed to the SpokePool.

### DIFF
--- a/contracts/SpokePoolPeriphery.sol
+++ b/contracts/SpokePoolPeriphery.sol
@@ -554,7 +554,7 @@ contract SpokePoolPeriphery is SpokePoolPeripheryInterface, ReentrancyGuard, Mul
     /**
      * @notice Returns the deposit ID for a given nonce and depositor.
      * @param depositor The depositor to use for the deposit.
-     * @param authorizer The authorizer for the movement of funds..
+     * @param authorizer The authorizer for the movement of funds.
      * @param nonceIdentifier The nonce identifier to use for the deposit.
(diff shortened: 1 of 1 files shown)


contracts/SpokePoolPeriphery.sol
```diff
@@ -544,13 +546,15 @@ contract SpokePoolPeriphery is SpokePoolPeripheryInterface, ReentrancyGuard, Mul
...
-            AUTHORIZATION_NONCE_IDENTIFIER
+            AUTHORIZATION_NONCE_IDENTIFIER,
+            signatureOwner
         );
     }
 
     /**
      * @notice Returns the deposit ID for a given nonce and depositor.
      * @param depositor The depositor to use for the deposit.
+     * @param authorizer The authorizer for the movement of funds..
```
grasphoper: extra .

PR #1297, opened 2026-02-04
chore: remove old Succinct contracts

grasphoper:
Closes ACP-12


contracts/interfaces/IHelios.sol
```diff
@@ -1,16 +1,11 @@
 // SPDX-License-Identifier: BUSL-1.1
```
fusmanii, line 1: how to we distinguish what goes in external/interfaces vs what goes in interfaces?

broadcast/deployed-addresses.json
```diff
@@ -554,11 +554,6 @@
         "SpokePoolPeriphery": {
           "address": "0x89415a82d909a7238d69094C3Dd1dCC1aCbDa85C",
           "block_number": 18113085
-        },
-        "Lisk_SpokePool": {
```
fusmanii, line 558: these are just implementation contracts, I dont think we should be adding these to, wdyt?

grasphoper: This was done by a pre-commit hook. Yeah, agreed, we don't need these and it's potentially confusing

contracts/interfaces/IHelios.sol
grasphoper: My understanding is that if it's our own contract, it goes into `interfaces/` , if it's an external dependency, to `external/interfaces`

grasphoper: Technically since we have `Sp1Helios` contract here now, we can throw this `IHelios` interface away all together and just use the `Sp1Helios` directly with the same result, but this change is smaller diff

PR #1299, opened 2026-02-05
fix: L-05

fusmanii:


contracts/libraries/HyperCoreLib.sol
```diff
@@ -430,6 +431,9 @@ library HyperCoreLib {
         uint256 maximumEVMSendAmount,
         int8 decimalDiff
     ) internal pure returns (uint256 amountEVMToSend, uint64 amountCoreToReceive) {
+        // If the maximum EVM send amount is greater than the maximum uint64 value, revert
+        if (maximumEVMSendAmount > type(uint64).max) revert MaximumEVMSendAmountTooLarge();
```
grasphoper: This does not cover the latest branch: `amountCoreToReceive = uint64(amountEVMToSend * scale);`

I think we could do something like:
- define some `amountCoreToReceive256` at the top of the function
- each branch does the logic and assigns `amountCoreToReceive256`
- at the bottom of the function, the max check is performed and conversion is done

fusmanii: nice catch, done https://github.com/across-protocol/contracts/pull/1299/changes/f527cb924609e01b991d7ee641bc075f64ab662c

tbwebb22 (commented):
LGTM, wondering if we should just use OZ's safeCast here though?

tbwebb22 (approved):

grasphoper (approved):

PR #1300, opened 2026-02-05
fix: L-04 add EIP-712 to HyperLiquidDepositHandler

tbwebb22:
> _verifySignature [authorizes](https://github.com/across-protocol/contracts/blob/d3bc16ecaecb84b71c80b826be1b93d0cd973ddc/contracts/handlers/HyperliquidDepositHandler.sol#L285-L289) donation-box-funded activation using ECDSA.recover(keccak256(abi.encode(user)), signature) without domain separation or an expiry. The signed message omits contract-specific data (e.g., contract address, chain ID) and time-bounding metadata, so the signature is a raw hash of user that can be replayed, in theory, wherever an identical hash is accepted. Lack of nonce or deadline further allows indefinite reuse.
> 
> In this current specific case of account activation sponsorship authorization, the scope is limited to one-time use and no other immediate cross-context replay is possible. Nonetheless, it is advised to future-proof signature authorization and adopt industry best practices.
> 
> Consider adopting EIP-712-style domain separation (including contract address, chain ID, and intent identifier), expanding the signed payload to cover the action being authorized, and adding an explicit expiry to limit signature lifetime.

### DIFF
--- a/broadcast/deployed-addresses.json
+++ b/broadcast/deployed-addresses.json
@@ -554,6 +554,11 @@
         "SpokePoolPeriphery": {
           "address": "0x89415a82d909a7238d69094C3Dd1dCC1aCbDa85C",
           "block_number": 18113085
+        },
+        "Lisk_SpokePool": {
+          "address": "0xe58480ca74f1a819fafd777beded4e2d5629943d",
+          "block_number": 27152120,
+          "transaction_hash": "0x0129be4b65f57157e7cbd7a8b8688afed1293a3442620df4d1c85736d451b0fc"
         }
       }
     },
@@ -590,9 +595,9 @@
           "block_number": 8910059
         },
         "Cher_SpokePool": {
-          "address": "0x48e687205d3962c43891b8cde5a4fe75fa6c8d7a",
-          "block_number": 11504355,
-          "transaction_hash": "0x5e126b5845bf4425d053febf06089e7c02411ee3edf7df6ffa888bc530d85c7f"
+          "address": "0x1c8243198570658f818fc56538f2c837c2a32958",
+          "block_number": 17949038,
+          "transaction_hash": "0x7a3ca40368652c84c2ca47f5722df58e098efcb8af08427192d55ae0baf92e91"
         }
       }
     },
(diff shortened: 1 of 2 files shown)


contracts/handlers/HyperliquidDepositHandler.sol
```diff
@@ -72,7 +76,7 @@ contract HyperliquidDepositHandler is AcrossMessageHandler, ReentrancyGuard, Own
      * should be one controlled by the Across API to prevent griefing attacks that attempt to drain the Donation Box.
      * @param _spokePool Address of the SpokePool contract that can call handleV3AcrossMessage.
      */
-    constructor(address _signer, address _spokePool) {
+    constructor(address _signer, address _spokePool) EIP712("HyperliquidDepositHandler", "1") {
```
grasphoper: Is using `"1"` canonical, or should it be `"v1.0"` or something?

broadcast/deployed-addresses.json
```diff
@@ -554,11 +554,6 @@
         "SpokePoolPeriphery": {
           "address": "0x89415a82d909a7238d69094C3Dd1dCC1aCbDa85C",
           "block_number": 18113085
-        },
```
fusmanii, line 557: should we undo these to not pollute the audit diff?

contracts/handlers/HyperliquidDepositHandler.sol
tbwebb22: There's not really a standard AFAIK. I checked within our codebase and `SpokePoolPeriphery` has "1.0.0" so I updated to that
https://github.com/across-protocol/contracts/pull/1300/changes/e9303702344db06b0c39db4a6be7f2daf3a1589f

broadcast/deployed-addresses.json
tbwebb22: done!

contracts/handlers/HyperliquidDepositHandler.sol
grasphoper: Sounds good!

PR #1302, opened 2026-02-05
chore: Use Upgrade.sol in deploymentsUtil

fusmanii:


script/utils/DeploymentUtils.sol
```diff
@@ -99,7 +99,8 @@ contract DeploymentUtils is Script, Test, Constants, DeployedAddresses, Config {
         Options memory opts;
 
         opts.constructorData = constructorArgs;
-        // opts.referenceBuildInfoDir = "artifacts";
+        opts.unsafeAllow = "delegatecall";
```
fusmanii, line 102: this is because we use `delegatecall` in `SpokePool` which OZ doesn't allow by default

script/utils/DeploymentUtils.sol
```diff
@@ -99,7 +99,8 @@ contract DeploymentUtils is Script, Test, Constants, DeployedAddresses, Config {
...
+        opts.unsafeSkipAllChecks = true;
```
fusmanii: If the checks are not skipped then OZ doesn't allow partial compilation, so any file changes requires removing all artifacts and building again. The positive is that OZ will do some validation before deploying so might be worth adding but it will/can cause dev friction. WDYT @tbwebb22 @grasphoper?

script/utils/DeploymentUtils.sol
```diff
@@ -110,20 +111,20 @@ contract DeploymentUtils is Script, Test, Constants, DeployedAddresses, Config {
...
 
             result = DeploymentResult({ proxy: existingProxy, implementation: implementation, isNewProxy: false });
 
             console.log("New", contractName, "implementation deployed @", implementation);
         } else {
-            address implementation = Core.deploy(contractName, constructorArgs, opts);
+            // Deploy implementation first, then proxy using the OZ v4 ERC1967Proxy directly
+            // to avoid vm.getCode ambiguity with multiple ERC1967Proxy artifacts.
+            address implementation = Upgrades.deployImplementation(contractName, opts);
+            address proxy = address(new ERC1967Proxy(implementation, initArgs));
```
grasphoper, line 128: Do we want to use native `Upgrades.deployUUPSProxy` or something? Or is the comment above explaining why we're not using that?

script/utils/DeploymentUtils.sol
grasphoper: This is only when you're deploying though right? I kind of lean towards just adding a bit of friction when deploying normally. Normally the dev shouldn't be changing any code when deploying right? It might cause a new recompile if a dev had saved artifacts previously right?

So I would leave
```
opts.unsafeSkipAllChecks = false;
```
Wdyt?

script/utils/DeploymentUtils.sol
fusmanii: yeah its because we are using V4 and V5 OZ, there is a name collision for `ERC1967Proxy` so need to deploy it explicitly

script/utils/DeploymentUtils.sol
fusmanii: it mainly around requiring a full recompile before deployment vs partial recompile (e.g. recompiling only changed files). I agree, I like adding some friction when deploying. I can also add some notes so that devs know to do full recompile when deployment fails

fusmanii: done https://github.com/across-protocol/contracts/pull/1302/changes/ec07f3ab3ea35b3340cfb5260c3a6fd7ffedc4c9

script/utils/DeploymentUtils.sol
grasphoper: Ah, makes sense

grasphoper (approved):

PR #1303, opened 2026-02-06
chore: Add prettier check to CI

tbwebb22:
Add prettier format check for Solidity files (contracts/ and script/) to CI, and add yarn format-contracts convenience command. Also fixes the broken glob in lint-fix.

VSCode / Cursor setup - add this to your workspace or user settings.json:

```
    "[solidity]": {
        "editor.defaultFormatter": "esbenp.prettier-vscode",
        "editor.formatOnSave": true
    },
```

This ensures VS Code uses the project's node_modules/prettier and .prettierrc to format Solidity files on save, matching what CI enforces. If you don't want format-on-save, you can omit that line and just run `yarn format-contracts` before committing.

Closes ACP-47


grasphoper (commented):
A few Qs:
- Is `"editor.defaultFormatter": "esbenp.prettier-vscode",` referring to the prettier -- the node module, or prettier -- the vs code extension? If the extension, are there any extra settings to change on the extension for it to use `node_modules/prettier` and .prettierrc, .prettierignore and something like `prettier-plugin-solidity`? E.g. can it happen that the extension is overriding the prettierrc config?
- `yarn pretty-quick --staged`: what prettier path + version is that using?
- should we perhaps upgrade versions of `prettier-plugin-solidity`, `pretty-quick` as a part of this change?

grasphoper:
@tbwebb22 got it to work only after also adding this to `.vscode/settings.json`:

```
...
  "prettier.prettierPath": "./node_modules/prettier",
...
```

grasphoper (commented):
Do you think we should just disable the trailing comma option to reduce this diff? The change otherwise looks good. I think we wanna hold off on merging this though and wait for the latest audit to land on master first (better fix conflicts here than in the audit PR)

tbwebb22:
> Do you think we should just disable the trailing comma option to reduce this diff? The change otherwise looks good. I think we wanna hold off on merging this though and wait for the latest audit to land on master first (better fix conflicts here than in the audit PR)

Sounds good, we can always re-enable the trailing comma option later on if we want to.

tbwebb22:
> Do you think we should just disable the trailing comma option to reduce this diff? The change otherwise looks good. I think we wanna hold off on merging this though and wait for the latest audit to land on master first (better fix conflicts here than in the audit PR)

https://github.com/across-protocol/contracts/pull/1303/changes/a09c7b20edfd79f8befac17d6dcf1afafe7eb236

Apparently there's no way to configure line-wrapping length in prettier-plugin-solidity v2, so we'll have to live with those changes in the diff if we want to stick with the newer version

PR #1304, opened 2026-02-09
chore: deploy periphery contracts to chosen testnets

dohaki:
Closes https://linear.app/uma/issue/ACB-164/deploy-testnet-periphery-contracts

### DIFF
--- a/broadcast/deployed-addresses.json
+++ b/broadcast/deployed-addresses.json
@@ -16,7 +16,7 @@
           "block_number": 23086526
         },
         "Arbitrum_Adapter": {
-          "address": "0xc0b6d2f794cc787c71f2ca5cecd57102c32379b3",
+          "address": "0xc0b6d2f794cc787C71f2cA5ceCD57102C32379B3",
           "block_number": 23498485,
           "transaction_hash": "0xedc397bdc758e28f61e8b35ddab220911ccfe605ddd3649d13223dc12bc137ef"
         },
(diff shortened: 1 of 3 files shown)


deployments/deployments.json
```diff
@@ -51,7 +51,8 @@
   },
```
fusmanii, line 51: If you are deploying with foundry you don't need to update `deployments.json`, the pre-commit hook will automatically read from the broadcast folder and update `broadcast/deployed-addresses.json` (which is the new source of truth)

foundry.toml
```diff
@@ -51,11 +51,15 @@ megaeth = "${NODE_URL_4326}"
...
 arbitrum_sepolia = "${NODE_URL_421614}"
+base_sepolia = "${NODE_URL_84532}"
+optimism_sepolia = "${NODE_URL_11155420}"
+sepolia = "${NODE_URL_11155111}"
 hyperevm_testnet = "${NODE_URL_998}"
 
 
 [etherscan]
 ethereum = { key = "${ETHERSCAN_API_KEY}" }
+sepolia = { key = "${ETHERSCAN_API_KEY}" }
```
fusmanii: is this needed? I though with etherscan V2 works on testnets as well (based on this list https://docs.etherscan.io/supported-chains). was it not working for you?

dohaki: removed !

deployments/deployments.json
dohaki: Hmm, somehow it doesn't seemt to work for testnets 🤔

dohaki: Added a fix here https://github.com/across-protocol/contracts/pull/1304/commits/360d22e65ea46fc84fac9b551357842ea7c7024e

fusmanii: nice catch!

PR #1305, opened 2026-02-09
feat: script to deploy both SP1Helios & UniversalSpokePool

tbwebb22:
Add a combined deployment shell script (DeploySP1HeliosAndUniversalSpokePool.sh) that automates the full 
SP1Helios + Universal SpokePool deployment in a single command. The script deploys SP1Helios, deploys the       Universal SpokePool proxy, transfers the admin role, and handles contract verification.                         
DeployUniversalSpokePool.s.sol now takes the SP1Helios address as a parameter instead of reading it from 
deployed-addresses.json, removing the need to run extract-addresses between deployments. README updated to      
reflect the new workflow.                                                    

Why a shell script instead of a Foundry script:

The deployment requires orchestrating two separate Forge scripts (SP1Helios needs --ffi, the SpokePool uses
DeploymentUtils), plus cast send calls for the admin role transfer. A shell script can invoke each Forge script
independently — preserving their separate broadcast directories — and pipe the SP1Helios address from one
script's broadcast output into the next. This would be difficult to do in a single .s.sol file without
duplicating the SP1Helios deployment logic and merging broadcast artifacts.

Closes ACP-49

### DIFF
--- a/script/universal/DeploySP1HeliosAndUniversalSpokePool.s.sol
+++ b/script/universal/DeploySP1HeliosAndUniversalSpokePool.s.sol
@@ -3,6 +3,7 @@ pragma solidity ^0.8.0;
 
 import { Script } from "forge-std/Script.sol";
 import { console } from "forge-std/console.sol";
+import { DeployedAddresses } from "../utils/DeployedAddresses.sol";
 
 /// @title DeploySP1HeliosAndUniversalSpokePool
 /// @notice Combined Foundry script that orchestrates the full deployment of SP1Helios + Universal_SpokePool.
@@ -32,7 +33,7 @@ import { console } from "forge-std/console.sol";
 /// Required env vars (loaded from .env, passed through to sub-scripts):
 ///   MNEMONIC, SP1_RELEASE, SP1_PROVER_MODE, SP1_VERIFIER_ADDRESS,
 ///   SP1_STATE_UPDATERS, SP1_VKEY_UPDATER, SP1_CONSENSUS_RPCS_LIST
-contract DeploySP1HeliosAndUniversalSpokePool is Script {
+contract DeploySP1HeliosAndUniversalSpokePool is Script, DeployedAddresses {
     bytes32 internal constant DEFAULT_ADMIN_ROLE = 0x00;
     bytes32 internal constant VKEY_UPDATER_ROLE = keccak256("VKEY_UPDATER_ROLE");
 
@@ -54,22 +55,13 @@ contract DeploySP1HeliosAndUniversalSpokePool is Script {
         // =====================================================================
         // Safety check: abort if a SpokePool already exists on this chain
         // =====================================================================
-        string memory deployedAddressesPath = string.concat(vm.projectRoot(), "/broadcast/deployed-addresses.json");
-        if (vm.isFile(deployedAddressesPath)) {
-            string memory existing = _ffiText(
-                string.concat(
-                    "jq -r '.chains[\"",
-                    chainId,
-                    '"].contracts["SpokePool"].address // empty\' ',
-                    deployedAddressesPath
-                )
-            );
-            if (bytes(existing).length > 0) {
-                console.log("Error: SpokePool already deployed on chain", chainId, "at", existing);
-                console.log("This script is intended for fresh deployments only.");
-                console.log("Remove the chain entry from", deployedAddressesPath, "if you want to redeploy.");
-                revert("SpokePool already exists on target chain");
-            }
+        uint256 targetChainId = vm.parseUint(chainId);
+        address existingSpokePool = getAddress(targetChainId, "SpokePool");
+        if (existingSpokePool != address(0)) {
+            console.log("Error: SpokePool already deployed on chain", chainId, "at", existingSpokePool);
+            console.log("This script is intended for fresh deployments only.");
+            console.log("Remove the chain entry from broadcast/deployed-addresses.json if you want to redeploy.");
+            revert("SpokePool already exists on target chain");
         }
 
         // =====================================================================
--- a/script/universal/README.md
+++ b/script/universal/README.md
@@ -5,6 +5,7 @@ This guide covers deploying the **Universal SpokePool** infrastructure to a new 
 ## Prerequisites
 
 - **Foundry** installed with FFI enabled
+- **jq** installed (used by the combined deployment script to parse broadcast JSON)
 - **curl** and **sha256sum** available in PATH
 - Access to Ethereum consensus layer RPC endpoints (beacon nodes)
 - A funded deployer wallet (via mnemonic)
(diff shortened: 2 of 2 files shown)


script/universal/DeploySP1HeliosAndUniversalSpokePool.sh
```diff
@@ -0,0 +1,237 @@
...
+# Step 1: Deploy SP1Helios
+# Runs DeploySP1Helios.s.sol which downloads a genesis binary via FFI,
+# generates genesis parameters, and deploys the SP1Helios light client.
+# --slow sends transactions one at a time (safer for new/unstable RPCs).
+# ===========================================================================
+echo ""
+echo "=== Step 1: Deploying SP1Helios ==="
+forge script script/universal/DeploySP1Helios.s.sol \
+  --rpc-url "$RPC_URL" \
+  $BROADCAST --slow --ffi -vvvv $EXTRA_ARGS
```
grasphoper: Do we need this `--slow` flag always?

script/universal/DeploySP1HeliosAndUniversalSpokePool.sh
```diff
@@ -0,0 +1,237 @@
...
+ensure_run_latest "broadcast/DeploySP1Helios.s.sol/$CHAIN_ID"
+
+# Wait for in-flight transactions to confirm before the next deployment,
+# so the RPC node has an up-to-date nonce for the deployer.
+if [[ -n "$BROADCAST" ]]; then
+  echo "Waiting for transactions to confirm..."
+  sleep 10
+fi
```
grasphoper: Should we loop here?

grasphoper: I'm thinking about chains similar to HyperEVM. Their block times can be 60s for large blocks, so this will not be enough. Does `cast` have some functionality like "waitForTX"?

script/universal/DeploySP1HeliosAndUniversalSpokePool.sh
```diff
@@ -0,0 +1,237 @@
...
+
+# ===========================================================================
+# Step 2: Deploy Universal_SpokePool
+# Parse the SP1Helios address from the Step 1 broadcast output, then pass it
+# as a constructor argument to the SpokePool deployment script.
+# The SpokePool is deployed behind an ERC1967 proxy via DeploymentUtils.
+# ===========================================================================
+SP1_HELIOS=$(jq -r '.transactions[] | select(.contractName == "SP1Helios") | .contractAddress' \
+  "broadcast/DeploySP1Helios.s.sol/$CHAIN_ID/run-latest.json")
```
grasphoper: The synchronization here is happening via this `run-latest` file. Could we produce some output from the deployment scripts instead perhaps so that we wouldn't need this logic?

IMO that would be less fragile

script/universal/DeploySP1HeliosAndUniversalSpokePool.sh
```diff
@@ -0,0 +1,237 @@
...
+fi
+echo ""
+echo "SP1Helios deployed at: $SP1_HELIOS"
+
+echo ""
+echo "=== Step 2: Deploying Universal_SpokePool ==="
+forge script script/universal/DeployUniversalSpokePool.s.sol:DeployUniversalSpokePool \
+  --sig "run(address,uint256)" "$SP1_HELIOS" "$OFT_FEE_CAP" \
+  --rpc-url "$RPC_URL" \
+  $BROADCAST --slow -vvvv $EXTRA_ARGS
```
grasphoper: same about `--slow`

script/universal/DeploySP1HeliosAndUniversalSpokePool.sh
```diff
@@ -0,0 +1,237 @@
...
+
+if [[ -n "$BROADCAST" ]]; then
+  # Grant DEFAULT_ADMIN_ROLE to the SpokePool
+  cast send "$SP1_HELIOS" \
+    "grantRole(bytes32,address)" "$DEFAULT_ADMIN_ROLE" "$SPOKE_POOL" \
+    --rpc-url "$RPC_URL" \
+    --private-key "$DEPLOYER_PRIVATE_KEY"
+
+  # Brief pause so the RPC node's nonce tracking catches up after the previous tx.
+  sleep 5
```
grasphoper: Can we:
- read nonce
- send tx 1 with read nonce
- send tx 2 with read nonce + 1

Seems like this could be less fragile

script/universal/DeploySP1HeliosAndUniversalSpokePool.sh
```diff
@@ -0,0 +1,237 @@
...
+
+  # Renounce DEFAULT_ADMIN_ROLE from the deployer
+  cast send "$SP1_HELIOS" \
+    "renounceRole(bytes32,address)" "$DEFAULT_ADMIN_ROLE" "$DEPLOYER_ADDRESS" \
+    --rpc-url "$RPC_URL" \
+    --private-key "$DEPLOYER_PRIVATE_KEY"
+
+  echo "Admin role transferred successfully."
```
grasphoper: I'd honestly wait for a TX here and `cast` a view function before reporting this

script/universal/DeploySP1HeliosAndUniversalSpokePool.sh
grasphoper: We could then also get rid of `ensure_run_latest`

script/universal/DeploySP1HeliosAndUniversalSpokePool.sh
grasphoper: Hmm, not sure waitForTx will 100% work here, since the deployment script already waits for a TX receipt by default I think. One mitigation I can think about is getting the nonce at the start of the script (via cast e.g.) and then using deterministic nonces in some places

cast "waitForTx" can be useful in other places tho

contracts/Universal_SpokePool.sol
```diff
@@ -18,18 +18,21 @@ import "./SpokePool.sol";
 contract Universal_SpokePool is OwnableUpgradeable, SpokePool, CircleCCTPAdapter {
     /// @notice The data store contract that only the HubPool can write to. This spoke pool can only act on
     /// data that has been written to this store.
+    /// @custom:oz-upgrades-unsafe-allow state-variable-immutable
```
tbwebb22, line 21: needed to add these annotations because of this change: https://github.com/across-protocol/contracts/pull/1302/changes

script/universal/DeploySP1HeliosAndUniversalSpokePool.sh
tbwebb22: Yeah turns out this sleep isn't necessary - the RPC tracks nonce even for pending transactions, and Foundry reads the current account nonce from the RPC before each send. Removed both sleeps
https://github.com/across-protocol/contracts/pull/1305/changes/6a2a731accea8a933635d307d7e988abab264cd3

script/universal/DeploySP1HeliosAndUniversalSpokePool.sh
tbwebb22: Yeah makes sense. Made that change here https://github.com/across-protocol/contracts/pull/1305/changes/10e7cf61f5169599287ec4081a34596428d847c1

and removed ensure_run_latest
https://github.com/across-protocol/contracts/pull/1305/changes/86b6af12dd669f1c5844ecc0832fd4da7c3e2328
https://github.com/across-protocol/contracts/pull/1305/changes/776110ce38430823515d8c8275423b4c0c3e0d50

script/universal/DeploySP1HeliosAndUniversalSpokePool.sh
tbwebb22: Turns out it isn't needed, removed here https://github.com/across-protocol/contracts/pull/1305/changes/83e0bc7bf6aa01c8be08ac77f4c7939cc1869746

script/universal/DeploySP1HeliosAndUniversalSpokePool.sh
tbwebb22: Turns out it isn't needed, removed here https://github.com/across-protocol/contracts/pull/1305/changes/83e0bc7bf6aa01c8be08ac77f4c7939cc1869746

script/universal/DeploySP1HeliosAndUniversalSpokePool.sh
tbwebb22: https://github.com/across-protocol/contracts/pull/1305/changes/230b91844def5802bcc98d327680643c22316db2

script/universal/DeploySP1HeliosAndUniversalSpokePool.s.sol
```diff
@@ -0,0 +1,376 @@
...
+/// Example (broadcast + verify):
+///   forge script script/universal/DeploySP1HeliosAndUniversalSpokePool.s.sol \
+///     --sig "run(uint256,string,string,bool)" 78000000000000000000000 "https://rpc.hyperliquid.xyz/evm" "YOUR_API_KEY" true --ffi
+///
+/// Required env vars (loaded from .env, passed through to sub-scripts):
+///   MNEMONIC, SP1_RELEASE, SP1_PROVER_MODE, SP1_VERIFIER_ADDRESS,
+///   SP1_STATE_UPDATERS, SP1_VKEY_UPDATER, SP1_CONSENSUS_RPCS_LIST
+contract DeploySP1HeliosAndUniversalSpokePool is Script {
+    bytes32 internal constant DEFAULT_ADMIN_ROLE = 0x00;
+    bytes32 internal constant VKEY_UPDATER_ROLE = keccak256("VKEY_UPDATER_ROLE");
```
grasphoper, line 38: Can we use `SP1Helios.VKEY_UPDATER_ROLE` directly here? Hm, same with admin role?

script/universal/DeploySP1HeliosAndUniversalSpokePool.s.sol
```diff
@@ -0,0 +1,376 @@
...
+
+    function run() external pure {
+        revert(
+            "Usage: forge script ... --sig 'run(uint256,string,string,bool)' <OFT_FEE_CAP> <RPC_URL> <ETHERSCAN_API_KEY> <BROADCAST> --ffi"
+        );
+    }
```
grasphoper: Is having this entrypoint a requirement? If not, I'd lean remove it since usage is documented above and I don;'t think anyone realistically will go try to run this script blindly without arguments (it will revert either way?)

script/universal/DeploySP1HeliosAndUniversalSpokePool.s.sol
```diff
@@ -0,0 +1,376 @@
...
+                    deployedAddressesPath
+                )
+            );
+            if (bytes(existing).length > 0) {
+                console.log("Error: SpokePool already deployed on chain", chainId, "at", existing);
+                console.log("This script is intended for fresh deployments only.");
+                console.log("Remove the chain entry from", deployedAddressesPath, "if you want to redeploy.");
+                revert("SpokePool already exists on target chain");
+            }
+        }
```
grasphoper: Can we use `DeploymentUtils` for this check?

script/universal/DeploySP1HeliosAndUniversalSpokePool.s.sol
```diff
@@ -0,0 +1,376 @@
...
+        );
+        address sp1Helios = vm.parseAddress(
+            _ffiText(
+                string.concat(
+                    "jq -r '.transactions[] | select(.contractName == \"SP1Helios\") | .contractAddress' ",
+                    heliosRunDir,
+                    "/run-latest.json"
+                )
+            )
+        );
```
grasphoper, line 114: Might use `DeploymentUtils` here too

grasphoper: Ah, not directly. We'd need to run an address extraction script first to use it. Might just run it here `extract_foundry_addresses.sh` , and then use DeploymentUtils. Wdyt?

script/universal/DeploySP1HeliosAndUniversalSpokePool.s.sol
grasphoper: Actually doesn't seem like `Sp1Helios.VKEY_UPDATER_ROLE` works

script/universal/DeploySP1HeliosAndUniversalSpokePool.s.sol
grasphoper: Actually I'm wondering if `run` in a foundry script can return anything? Could we then parse a returned address instead of using these broadcast/ files as a medium of communication?

E.g. make `DeploySP1Helios` return `(address)` perhaps? (I checked it, it works). Up to you if this is better:

```
Script ran successfully.

Gas used: 24188

== Return ==                                                               
0: address 0x000000000000000000000000000000000000bEEF

== Logs ==                                                           
Deployed address: 0x000000000000000000000000000000000000bEEF
```

script/universal/DeploySP1HeliosAndUniversalSpokePool.s.sol
```diff
@@ -0,0 +1,376 @@
...
+                string.concat(
+                    "jq -r '.transactions[] | select(.contractName == \"ERC1967Proxy\") | .contractAddress' ",
+                    spokeRunDir,
+                    "/run-latest.json"
+                )
+            )
+        );
+        require(spokePool != address(0), "Could not find SpokePool address in broadcast output");
+        console.log("");
+        console.log("SpokePool deployed at:", spokePool);
```
grasphoper, line 157: Same w/ DeploymentUtils / or returning the address

script/universal/DeploySP1HeliosAndUniversalSpokePool.s.sol
```diff
@@ -0,0 +1,376 @@
...
+        console.log("SpokePool:", spokePool);
+
+        string memory vkeyRole = vm.toString(VKEY_UPDATER_ROLE);
+        string memory adminRole = vm.toString(DEFAULT_ADMIN_ROLE);
+
+        if (doBroadcast) {
+            _castSend(rpcUrl, pkHex, sp1Helios, "grantRole(bytes32,address)", vkeyRole, spokePool);
+            _castSend(rpcUrl, pkHex, sp1Helios, "renounceRole(bytes32,address)", vkeyRole, deployer);
+            _castSend(rpcUrl, pkHex, sp1Helios, "grantRole(bytes32,address)", adminRole, spokePool);
+            _castSend(rpcUrl, pkHex, sp1Helios, "renounceRole(bytes32,address)", adminRole, deployer);
```
grasphoper, line 175: I'm wondering why we need `cast` here. Can't we do it in a more direct way from a script?

script/universal/DeploySP1HeliosAndUniversalSpokePool.s.sol
```diff
@@ -0,0 +1,376 @@
...
+            console.log("SpokePool has DEFAULT_ADMIN_ROLE:", spokeHasAdmin);
+            console.log("Deployer has DEFAULT_ADMIN_ROLE: ", deployerHasAdmin);
+
+            require(
+                keccak256(bytes(spokeHasVkey)) == keccak256("true") &&
+                    keccak256(bytes(deployerHasVkey)) == keccak256("false") &&
+                    keccak256(bytes(spokeHasAdmin)) == keccak256("true") &&
+                    keccak256(bytes(deployerHasAdmin)) == keccak256("false"),
+                "Role verification failed!"
+            );
```
grasphoper, line 221: Heh, a funny require statement. But the same about `cast`: couldn't we use more native script tools here?

script/universal/DeploySP1HeliosAndUniversalSpokePool.s.sol
```diff
@@ -0,0 +1,376 @@
...
+                    vm.toString(oftFeeCap),
+                    " --rpc-url ",
+                    rpcUrl,
+                    " --verify ",
+                    verifyArgs,
+                    " -vvvv --resume",
+                    " --private-key ",
+                    pkHex,
+                    " 1>&2 || echo 'Warning: Universal_SpokePool verification failed (may already be verified)' >&2"
+                )
```
grasphoper, line 266: Interesting!

script/universal/DeploySP1HeliosAndUniversalSpokePool.s.sol
tbwebb22: yeah agreed its not necessary
https://github.com/across-protocol/contracts/pull/1305/changes/f1bf351523da8f59bd25c314422cf08cdfa130ad

script/universal/DeploySP1HeliosAndUniversalSpokePool.s.sol
tbwebb22: Yep, they definitely can return things. The challenge is that we're invoking them through ffi rather than just calling the run function. But looking into if its possible

script/universal/DeploySP1HeliosAndUniversalSpokePool.s.sol
tbwebb22: Yeah I tried quite a bit to do it this way, but ultimately wasn't able to get it to work. Some strange things start happening when you have two levels of broadcasts happening

script/universal/DeploySP1HeliosAndUniversalSpokePool.s.sol
tbwebb22: Yep, was able to use the `DeployedAddresses` util here https://github.com/across-protocol/contracts/pull/1305/changes/b5b12bc2e73ab8e1cdce8f72f76f1fd8924f8e32

script/universal/DeploySP1HeliosAndUniversalSpokePool.s.sol
tbwebb22: Yeah there doesn't seem to be a clean way to do it. Leaning towards leaving this as is

grasphoper: Sounds good, let's leave it

PR #1320, opened 2026-02-17
chore: zkSync and Lens SpokePool deploy

fusmanii:


foundry.toml
```diff
@@ -14,6 +14,7 @@ solc_version = "0.8.30"
 revert_strings = "strip"
 fs_permissions = [{ access = "read-write", path = "./" }]
 use_literal_content = true
+metadata = "none"
```
grasphoper: I think settings are different? https://www.getfoundry.sh/guides/deterministic-deployments-using-create2#configuring-for-determinism

Also, I'm hesitant if we _really_ want to add this as a global default. Let's discuss in the meet tomorrow. Now that we know that whitespace affects the things and that we have a CI check for formatting, our deployments should become more deterministic overtime even without this setting

fusmanii: > Disable metadata hash for deterministic bytecode

looks like they recommend to disable for deterministic bytecode

fusmanii: only disabling metadata in eraVM https://github.com/across-protocol/contracts/pull/1320/changes/60a6f789d412ec4969b4943228d058d2a9ca3274

grasphoper (dismissed):
LGTM, let's merge after we execute the multisig right

grasphoper (approved):

PR #1410, opened 2026-04-13
feat: Operations multisig deployments

tbwebb22:
Adds script to deploy Safe multisigs using the Safe protocol kit.

Deploys to the following chains

Chain ID | Chain Name | Safe UI Supports?
-- | -- | --
56 | BSC | Yes
143 | Monad | Yes
232 | Lens | Yes
324 | zkSync | Yes
999 | HyperEVM | Yes
1135 | Lisk | No
1868 | Soneium | No
4217 | Tempo | Yes
9745 | Plasma | Yes
81457 | Blast | No
7777777 | Zora | No

Note that Safes deployed to Lens and zkSync have different addresses than the other deployments - the protocol kit resolves to different Safe infra contracts for zkSync flavored chains.


script/safe-multisig/deploySafe.ts
```diff
@@ -0,0 +1,170 @@
...
+  chainId: number;
+  provider: ethers.providers.JsonRpcProvider;
+  nodeUrl: string;
+}> {
+  const explicitChainId = getArg("--chain-id");
+  if (!explicitChainId) throw new Error("--chain-id is required");
+  const chainId = Number(explicitChainId);
+  if (!Number.isInteger(chainId) || chainId <= 0) throw new Error(`Invalid --chain-id value: ${explicitChainId}`);
+  const nodeUrl = getNodeUrl(chainId);
+  return { chainId, provider: getProvider(nodeUrl), nodeUrl };
```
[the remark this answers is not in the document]
tbwebb22: https://github.com/across-protocol/contracts/pull/1410/changes/c0d0c710d064a06e1a35102e40614a4947fc8912

bmzig:
Any chance we can deploy these multisigs to Monad/Tempo/TRON as well? It's convenient to deploy them to all universal networks.

tbwebb22:
> Any chance we can deploy these multisigs to Monad/Tempo/TRON as well? It's convenient to deploy them to all universal networks.

Tron isn't supported by the Safe protocol kit, I'm assuming the Safe contract has some incompatibility with Tron. Tron does have some native multisig solution that I'm looking into - we'll probably have to end up using it.

Went ahead and deployed to the other Universal networks here https://github.com/across-protocol/contracts/pull/1410/changes/690be72e2d7aa57acf5d3922bd6c184cb84d5021

script/safe-multisig/deploySafe.ts
```diff
@@ -0,0 +1,175 @@
...
+  if (process.argv.includes("--help")) {
+    printUsage();
+    return;
+  }
+
+  const configPath = DEFAULT_CONFIG_PATH;
+  const { chainId, provider, nodeUrl } = await resolveChainContext();
+  const config = loadMultisigConfig(configPath);
+  const wallet = getSigner(provider);
+  const signer = wallet.connect(new ethers.providers.JsonRpcProvider(nodeUrl));
```
grasphoper: nit: can we just use `provider` here?

script/safe-multisig/broadcast.ts
```diff
@@ -0,0 +1,112 @@
...
+import { execSync } from "child_process";
+import { BigNumber, ethers } from "ethers";
+import { getAddress } from "ethers/lib/utils";
+
+const BROADCAST_SCRIPT_NAME = "DeploySafe.s.sol";
+
+function toHex(value: BigNumber | number | string | null | undefined): string | null {
+  if (value === null || value === undefined) return null;
+  return BigNumber.from(value).toHexString();
+}
```
grasphoper, line 12: nit: might use `ethers.utils.hexlify` here? We're due to upgrade ethers to at least v6 and it will prob stay on as a dev dependency or get replaced by view, so may not be relevant. But prob. a bit cleaner

script/safe-multisig/broadcast.ts
```diff
@@ -0,0 +1,112 @@
...
+        contractAddress: getAddress(opts.contractAddress),
+        function: null,
+        arguments: opts.args,
+        transaction: {
+          type: toHex(tx.type ?? 2),
+          from: getAddress(tx.from),
+          to: tx.to ? getAddress(tx.to) : getAddress(opts.deploymentTransaction.to),
+          gas: toHex(tx.gasLimit),
+          value: toHex(tx.value ?? opts.deploymentTransaction.value),
+          data: tx.data ?? opts.deploymentTransaction.data,
```
grasphoper: I think `input` is correct here for a field name

script/safe-multisig/broadcast.ts
```diff
@@ -0,0 +1,112 @@
...
+        type: toHex(receipt.type ?? tx.type ?? 2),
+        effectiveGasPrice: toHex(receipt.effectiveGasPrice),
+      },
+    ],
+    libraries: [],
+    pending: [],
+    returns: {},
+    timestamp,
+    chain: opts.chainId,
+    multi: false,
```
grasphoper: Seems extra

script/safe-multisig/config.json
```diff
@@ -0,0 +1,11 @@
+{
+  "threshold": 3,
+  "saltNonce": "0x0",
```
grasphoper, line 3: a Q: why 0x0, a change from 0x1234 we used before?

script/safe-multisig/broadcast.ts
```diff
@@ -0,0 +1,112 @@
...
+  args: unknown[];
+}): string {
+  const timestamp = Math.floor(Date.now() / 1000);
+  const tx = opts.sentTransaction;
+  const receipt = opts.receipt;
+  const broadcast = {
+    transactions: [
+      {
+        hash: receipt.transactionHash,
+        transactionType: "CREATE",
```
grasphoper, line 55: Is this correct? `CREATE2`? It's more like a factory-based deployment right? Might be fine to keep this so that `ExtractDeployedFoundryAddresses.ts` handles this well

grasphoper (dismissed):
Looks good, a few minor comments

script/safe-multisig/deploySafe.ts
tbwebb22: cleaned this up here https://github.com/across-protocol/contracts/pull/1410/changes/9851a8f4ab704ade8934bcd6b5d0602f4616f7d4

script/safe-multisig/broadcast.ts
tbwebb22: looks like hexlify doesn't work as a drop in here since this helper accepts strings, which hexlify doesn't

script/safe-multisig/broadcast.ts
tbwebb22: good catch https://github.com/across-protocol/contracts/commit/77e3aca44885f65bc14ec1e7262653c07944e0e4

script/safe-multisig/broadcast.ts
tbwebb22: for sure https://github.com/across-protocol/contracts/commit/55ceebb2db35ff250700d32876fc6a20465b3b78

script/safe-multisig/config.json
tbwebb22: no real reason, aside from 0x0 being sort of a default salt, it is arbitrary. Safes will have different addresses anyways since the signers have changed

script/safe-multisig/broadcast.ts
tbwebb22: Yeah it is technically CREATE2 under the hood. These broadcast files are actually a different format than they would be if they were deployed with an actual foundry script - the deployed Safe would appear in the "additionalContracts" section. But agree its probably fine to keep as is for the extract addresses script

grasphoper (dismissed):
Looks good. We should probably updatae the run-latest artifacts to reflect the data -> input change + multi: false removal

tbwebb22:
> Looks good. We should probably updatae the run-latest artifacts to reflect the data -> input change + multi: false removal

done https://github.com/across-protocol/contracts/commit/0e0ec994115e4b4a9e250ad9364ccce0888da59b

grasphoper (approved):

PR #1420, opened 2026-05-01
feat(SpokePoolPeriphery): support extended ERC-3009 bytes-signature

mrice32:
## Summary

- Adds two new entry points on `SpokePoolPeriphery` that pull tokens via the extended EIP-3009 `receiveWithAuthorization(...,bytes signature)` overload (e.g. USDC v2.2):
  - `depositWithAuthorizationBytes`
  - `swapAndBridgeWithAuthorizationBytes`
- Adds the matching `IERC20AuthBytes` interface alongside `IERC20Auth`. Selectors differ from the canonical v/r/s `receiveWithAuthorization`, so a token can implement both — and our periphery now supports either.
- This unlocks contract-wallet (EIP-1271) signers for ERC-3009 deposit/swap-and-bridge, which the existing v/r/s flows cannot support.
- Tail logic (witness binding, submission fee payment, dispatch) is shared with the v/r/s variants via small private helpers; only the token-pull call differs.
- `MockERC20` gains the bytes overload (sharing verification logic with the v/r/s overload) so we can exercise both code paths.

## Test plan

- [x] `yarn build-evm-foundry`
- [x] `yarn test-evm-foundry -- --match-contract SpokePoolPeripheryTest` — 33 passed (existing 29 + 4 new)
- [x] New tests cover:
  - `testTransferWithAuthBytesDepositValidWitness` — happy path, EOA signer
  - `testTransferWithAuthBytesSwapAndBridgeValidWitness` — happy path, swap + bridge
  - `testTransferWithAuthBytesSwapAndBridgeInvalidWitness` — wrong witness reverts (fuzz)
  - `testTransferWithAuthBytesDepositEIP1271ContractSigner` — proves a contract wallet can deposit, which is impossible via the v/r/s entry points
- [x] `yarn lint-fix`


contracts/interfaces/SpokePoolPeripheryInterface.sol
```diff
@@ -276,6 +297,27 @@ interface SpokePoolPeripheryInterface {
...
+     * signers.
+     * @dev If `acrossInputToken` does not implement the bytes-signature `receiveWithAuthorization`
+     * overload, this call will revert.
+     * @param signatureOwner The owner of the EIP3009 signature and depositData signature. Assumed to be the depositor for the Across spoke pool.
+     * @param depositData Specifies the Across deposit params to send.
+     * @param validAfter The unix time after which the `receiveWithAuthorization` signature is valid.
+     * @param validBefore The unix time before which the `receiveWithAuthorization` signature is valid.
+     * @param receiveWithAuthSignature EIP3009 signature, unstructured bytes (EOA or EIP-1271).
+     */
+    function depositWithAuthorizationBytes(
```
fusmanii, line 325: should `depositWithAuthorization` be changed to accept `r, s,` and `v` and this can also be called `depositWithAuthorization`? similar to how token contracts overload `receiveWithAuthorization`

fusmanii (commented):
just one Q

contracts/periphery/SpokePoolPeriphery.sol
```diff
@@ -380,10 +380,50 @@ contract SpokePoolPeriphery is SpokePoolPeripheryInterface, ReentrancyGuard, Mul
...
+        IERC20AuthBytes(address(swapAndDepositData.swapToken)).receiveWithAuthorization(
+            signatureOwner,
+            address(this),
+            swapAndDepositData.swapTokenAmount + _submissionFeeAmount,
+            validAfter,
+            validBefore,
+            witness,
+            receiveWithAuthSignature
+        );
+        _finishSwapAndBridgeWithAuthorization(swapAndDepositData, witness, signatureOwner);
```
grasphoper, line 430: nit: we could either pass in `_submissionFeeAmount` into `_finishSwapAndBridgeWithAuthorization` (save an `MLOAD`?) or just read from memory in this line directly:
```
swapAndDepositData.swapTokenAmount + _submissionFeeAmount,
```
->
```
swapAndDepositData.swapTokenAmount + swapAndDepositData.submissionFees.amount,
```

grasphoper (dismissed):

contracts/periphery/SpokePoolPeriphery.sol
grasphoper: Ok fixed in [6e2a16b](https://github.com/across-protocol/contracts/pull/1420/commits/6e2a16bc126199a58411f591f0f2f75c0a7c66d1) and [3d43733](https://github.com/across-protocol/contracts/pull/1420/commits/3d437331c21793ae6d05ac67b5f66371db2b216b)

grasphoper (dismissed):

fusmanii (approved):

PR #1423, opened 2026-05-05
chore: OFT & CCTP direct deposit deploy

fusmanii:


tbwebb22:
Looks like dev wallet is still default admin on the DonationBox

broadcast/DeploySponsoredCCTPSrcPeriphery.s.sol/999/run-latest.json
```diff
@@ -1,18 +1,18 @@
 {
   "transactions": [
     {
-      "hash": "0x13b2a5eee6960ab615fe83eaac7203a16878d039fa4a91cb9e1dfafc1bc56298",
+      "hash": "0xef3bcc6d1bb0adc5a2b88c6dbe64a2492f72ac7d27a4aa6e42e5ace34df1f531",
       "transactionType": "CREATE",
       "contractName": "SponsoredCCTPSrcPeriphery",
-      "contractAddress": "0xc611d7c33d1c7915f13ef35f488eaee9d9dc3555",
+      "contractAddress": "0xf4e32c4ac479f0b007bc005ec0f481a2c78ba1b4",
```
tbwebb22, line 7: looks like this one is still owned by dev wallet

broadcast/DeploySrcPeriphery.s.sol/999/run-latest.json
```diff
@@ -0,0 +1,68 @@
+{
+  "transactions": [
+    {
+      "hash": "0xc92f27a2736d47bed6ea737d590a27a2498fc73829ffd0b828775d0628aa14cf",
+      "transactionType": "CREATE",
+      "contractName": "SponsoredOFTSrcPeriphery",
+      "contractAddress": "0x93e1e5e9d840e758cc016f8b13cdf5dbaeca4b06",
```
tbwebb22, line 7: This one also still owned by the dev wallet
Signer here is 0x8d6e1729FeCe1AC62c05aA4F57819A90B4Cd22aF - is that correct?

broadcast/DeploySponsoredCCTPSrcPeriphery.s.sol/999/run-latest.json
fusmanii: nice catch, just updated this and 0x93e1E5E9D840e758CC016F8b13cDf5dbAeca4b06 owner to multisig

broadcast/DeploySrcPeriphery.s.sol/999/run-latest.json
fusmanii: both owner and signer should be correct now

tbwebb22 (dismissed):

script/mintburn/oft/usdt0.toml
```diff
@@ -45,7 +45,8 @@ endpoint_url = "${NODE_URL_999}"
 token = "0xB8CE59FC3717ada4C02eaDF9682A9e934F625ebb"
 oft_messenger = "0x904861a24F30EC96ea7CFC3bE9EA4B476d237e98"
 multicall_handler = "0xfD0876712DD9003D014CDCd8e5140B4EFAC9BFCC"
```
grasphoper: This is the current on-chain address `0xeaa13e675Fb95821B99D34B9c94c3D60fecAdd4A`. I see the deploy script was changed to not record this into the toml. I think that's fine, but we should remove this entry then

script/mintburn/cctp/config.toml
```diff
@@ -7,8 +7,8 @@ cctpTokenMessenger = "0x28b5a0e9C621a5BadaA536219b3a228C8168cf5d"
 usdc = "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48"
 cctpMessageTransmitter = "0x81D40F21F12A8F0E3252Bccb954D722d4c464B64"
 baseToken = "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48"
-sponsoredCCTPDstPeriphery = "0xde589398a3fe1dcca80c12053a14811f3e4033f4"
-multicallHandler = "0x64a43393866DBA0044879979fAa7AD3d000622e9"
+sponsoredCCTPDstPeriphery_usdc = "0x214C19FBcdFb683F2c726B4BBAf24AB483bFaDf3"
+multicallHandler_usdc = "0xDd52f8134f85f3979fbA24387Ce0CEC05937259E"
```
grasphoper: Were these updated manually?

Base doesn't have the `_usdc` suffix. IMO, we should be consistent with the naming scheme here. Also, could e.g. just add `_usdh` to the one "unusual" deployment we have, since other wise `sponsoredCCTPDstPeriphery` should imply USDC. We could have: `sponsoredCCTPDstPeriphery` and `sponsoredCCTPDstPeriphery_usdh` just for the odd case of cctp dst handler + usdh on HEVM

broadcast/deployed-addresses.json
```diff
@@ -849,15 +849,25 @@
           "block_number": 31071129,
           "transaction_hash": "0xd78f71ca11de6d28fe02006adf91b55f05efe0ffb1e0b2f51cb450cdf5ade5c0"
         },
-        "DonationBox_OFT": {
-          "address": "0xf9a0c1C775f1B6e156ad3f1DB784520461DDb19e",
-          "block_number": 27226495,
-          "transaction_hash": "0x79d57ee428509b8a739a619580ef0154d3618f782d5cab1bfad01c0d542d2e3e"
+        "DonationBox_CCTP_USDC_EVM": {
```
grasphoper, line 852: Why do we have an extra `_EVM` here?

script/mintburn/cctp/config.toml
fusmanii: yeah I like that, just having `_usdh`

grasphoper (approved):

PR #1430, opened 2026-05-13
fix: Tron USDT fix

tbwebb22:
Adds Tron-specific variants of SpokePool and the counterfactual contracts (CounterfactualDepositSpokePool, WithdrawImplementation) to handle Tron USDT's non-standard transfer, which moves balances correctly but always returns false — breaking SafeERC20.safeTransfer and the existing _noRevertTransfer return-value check.

Introduces a shared SafeTransferERC20 mixin in contracts/libraries/ exposing a virtual _safeTransfer hook (default: OZ safeTransfer); both the base SpokePool and the counterfactual contracts inherit from it. The Tron variants override the hook with a balance-delta success check (via TronTransferLib._safeTransferBalanceCheck for the counterfactual variants, inline in Tron_SpokePool), reverting with TronTransferFailed on genuine failure. transferFrom is correct on Tron USDT, so safeTransferFrom paths are unchanged, as are CCTP (USDC-only) and OFT (bridges USDT0, a compliant ERC20).

Tron_SpokePool extends Universal_SpokePool and overrides only _safeTransfer and _noRevertTransfer; CounterfactualDepositSpokePoolTron and WithdrawImplementationTron follow the same single-hook-override pattern. Includes deploy scripts and Foundry tests covering both the success path and genuine-failure paths (e.g. blacklisted recipient).

Re-deploys the changed contracts to Tron


contracts/periphery/counterfactual/CounterfactualDepositSpokePoolTr.sol
```diff
@@ -0,0 +1,36 @@
...
+ *      returns true correctly on Tron USDT.
+ *
+ *      The EIP-712 domain name is inherited from the parent (`CounterfactualDepositSpokePool`).
+ *      Cross-implementation signature replay is already prevented by the `verifyingContract`
+ *      field of the EIP-712 domain: each clone's address is derived via CREATE2 from its
+ *      implementation address, so a signature for a mainline clone does not verify against
+ *      a Tron-variant clone.
+ * @custom:security-contact bugs@across.to
+ */
+contract CounterfactualDepositSpokePoolTr is CounterfactualDepositSpokePool {
```
tbwebb22, line 23: Tron network limits contract names to 32 characters. Contracts with names longer than that aren't deployable on the network - so had to truncate this one

tbwebb22:
@codex review

contracts/periphery/counterfactual/CounterfactualDepositSpokePoolTr.sol
fusmanii: another Tron limitation :(

fusmanii (dismissed):

grasphoper (approved):

PR #1431, opened 2026-05-13
fix: `ArbitraryEVMFlowExecutor`

grasphoper:


grasphoper:
@codex please review. Note: this solution is meant to protect a user from receiving dust in all cases. Dust living on MulticallHandler and small random fee subractions should NOT fool ArbitraryEVMFlowExecutor into choosing an incorrect finalToken , finalAmount. One more thing to consider is that user should not be able to inflate sponsorship amount arbitrarily. The latter is less of a concern, since user actions are never picked by untrusted users, only by a trusted actor given user's intent.

contracts/periphery/mintburn/ArbitraryEVMFlowExecutor.sol
```diff
@@ -83,19 +84,13 @@ abstract contract ArbitraryEVMFlowExecutor {
...
-            uint256 finalBalance = IERC20(params.commonParams.finalToken).balanceOf(address(this));
-            if (finalBalance >= finalAmountSnapshot) {
-                // This means the swap did happen, so we check the balance of the output token and send it.
-                finalAmount = finalBalance - finalAmountSnapshot;
-            } else {
-                // If we somehow lost final tokens (e.g. by depositing into some contract), just set the finalAmount to 0.
-                finalAmount = 0;
-            }
+            params.commonParams.finalToken = params.initialToken;
+            finalAmount = params.commonParams.amountInEVM + eBI - sBI;
```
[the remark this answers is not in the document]
grasphoper: @codex addressed by draining dust upfront. Please review

test/evm/foundry/local/ArbitraryEVMFlowExecutor.t.sol
```diff
@@ -0,0 +1,264 @@
...
+            200e6,
+            300e6
+        );
+
+        EVMFlowParams memory p = _params(address(tokenIn), address(tokenIn), 500e6, abi.encode(calls));
+        CommonFlowParams memory out = harness.executeFlow(p);
+
+        assertEq(out.finalToken, address(tokenIn));
+        // Branch 1 triggers (initialBalance > initialAmountSnapshot via `<=`); reports the original input.
+        assertEq(out.amountInEVM, 500e6, "branch 1 reports amountInEVM, ignoring surplus");
```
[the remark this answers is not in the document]
grasphoper: @codex fixed, re-review please

tbwebb22 (dismissed):

contracts/periphery/mintburn/ArbitraryEVMFlowExecutor.sol
```diff
@@ -155,6 +158,51 @@ abstract contract ArbitraryEVMFlowExecutor {
         return abi.encode(instructions);
     }
 
+    /**
+     * @notice Drains any pre-existing balances of initialToken and finalToken from MulticallHandler
+     * @dev No-op when MulticallHandler holds no dust in either token. Otherwise routes a minimal set of
+     *      direct IERC20.transfer calls through handleV3AcrossMessage with fallbackRecipient = 0
+     *      (skipping the handler's own post-drain), then asserts both balances are zero.
+     */
+    function _drainMulticallHandlerDust(address initialToken, address finalToken) internal {
```
mrice32: Personally, I would just make two calls instead of having this complexity. It's rare that dust is left, so trying to optimize for a single call in the rare case that there's dust in two tokens seems like overkill (not sure how it got there in the first place).

Something like:
```solidity
  // Calling function
  ...
  _drainMulticallHandlerDust(initialToken);
  if (finalToken != initialToken) {
    _drainMulticallHandler(finalToken);
  }
  ...

function _drainMulticallHandlerDust(address token) internal {
  if (token.balanceOf(multicallHandler) != 0) {
    // drain
  }
}
```

contracts/periphery/mintburn/ArbitraryEVMFlowExecutor.sol
```diff
@@ -155,6 +158,51 @@ abstract contract ArbitraryEVMFlowExecutor {
...
+            });
+        }
+
+        bytes memory instructions = abi.encode(
+            MulticallHandler.Instructions({ calls: calls, fallbackRecipient: address(0) })
+        );
+
+        MulticallHandler(payable(multicallHandler)).handleV3AcrossMessage(initialToken, 0, address(this), instructions);
+
+        if (dustInitial > 0) {
```
mrice32: What are these checks?

contracts/periphery/mintburn/ArbitraryEVMFlowExecutor.sol
grasphoper: Agree, will change to this

contracts/periphery/mintburn/ArbitraryEVMFlowExecutor.sol
grasphoper: I guess it's me trying to be cautious

I.e. we're not checking the `bool` output from the `IERC20.transfer` call on the MulticallHandler, so we wouldn't know if the call failed. Wanted to give the rest of the code (that comes after) a guarantee that dust will not be present. Makes reasoning about the rest of the code easier. Wdyt?

grasphoper: > wouldn't know if the call failed

I mean, we wouldn't know if IERC20.transfer returned success = false

grasphoper:
@codex review one last time please

contracts/periphery/mintburn/ArbitraryEVMFlowExecutor.sol
```diff
@@ -82,19 +90,17 @@ abstract contract ArbitraryEVMFlowExecutor {
...
+        uint256 finalAmount = params.commonParams.amountInEVM + eBI - sBI;
+        if (differentTokens) {
+            // Ending balance final token
+            uint256 eBF = IERC20(params.commonParams.finalToken).balanceOf(address(this));
+            if (eBF > sBF) {
+                finalAmount = eBF - sBF;
             } else {
-                // If we somehow lost final tokens (e.g. by depositing into some contract), just set the finalAmount to 0.
-                finalAmount = 0;
+                params.commonParams.finalToken = params.initialToken;
```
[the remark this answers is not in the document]
grasphoper: This is fine.

- no dust + `eBF > sBF` means that calls succeded on MulticallHandler
- finalToken was declared as a desired token and calls succeeded => we don't have to worry about initialToken here anymore

fusmanii (approved):

contracts/periphery/mintburn/ArbitraryEVMFlowExecutor.sol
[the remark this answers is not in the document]
grasphoper: added comment to document this behavior [44f724c](https://github.com/across-protocol/contracts/pull/1431/commits/44f724ca3cee59aec4f81fd514f11c6d282acec3)

PR #1433, opened 2026-05-14
feat: Multisig migration script

tbwebb22:
Open `MULTISIGS.md` in preview mode to see the formatted table 

Closes ACP-142


package.json
```diff
@@ -51,6 +51,7 @@
     "extract-addresses": "./script/utils/extract_foundry_addresses.sh",
     "generate-constants-json": "ts-node ./script/utils/GenerateConstantsJson.ts",
     "deploy-safe": "ts-node ./script/safe-multisig/deploySafe.ts",
+    "list-multisigs": "ts-node ./script/safe-multisig/generateMultisigList.ts",
```
fusmanii, line 54:
```suggestion
    "list-multisigs": "ts-node ./script/safe-multisig/generateMultisigList.ts && yarn prettier --write script/safe-multisig/MULTISIGS.md",
```
so that it gets formatted

script/safe-multisig/MULTISIGS.md
```diff
@@ -0,0 +1,68 @@
+# Safe Multisig Migration Status
+
+**Migration progress: 9.1%** — 6 of 66 checks pass (60 outstanding).
+
+| Chain ID  | Chain           | Ops Multisig Deployed                                                                                        | Universal SpokePool Owner                                                                               | Counterfactual WithdrawManager Owner                                                                                     | Counterfactual WithdrawManager directWithdrawer                                                                          | Sponsored CCTP Periphery Owner                                                                                           | Sponsored OFT Periphery Owner                                                                            | DonationBox Admin |
```
fusmanii: I noticed that the donation boxes are missing on all chains

fusmanii (commented):
the md file looks nice, easy to spot check at a glance.

script/safe-multisig/MULTISIGS.md
tbwebb22: good catch! Updated the script and re-ran it - Donation boxes populated now. Since the DonationBox uses AccessControl (non-enumerable) - the table will just show "No" if the admin isn't one of the multisigs or fallback EOA

tbwebb22: also just realized we probably need two columns for DonationBox - some chains have CCTP and OFT deployments of it

tbwebb22: ok made this update - if there are multiple DonationBoxes for a chain, it will show each of them in the DonationBox column

PR #1434, opened 2026-05-19
feat: Counterfactual route policies

tbwebb22:
Closes ACP-106


grasphoper:
Closes ACP-106

contracts/periphery/counterfactual/CounterfactualDeposit.sol
```diff
@@ -5,52 +5,89 @@ import { Clones } from "@openzeppelin/contracts/proxy/Clones.sol";
...
+        // always present — new impls must preserve this property.
+        if (params.length < 0x60) revert ParamsTooShort();
+        uint256 leafDestinationChainId;
+        bytes32 leafOutputToken;
+        assembly {
+            leafDestinationChainId := calldataload(add(params.offset, 0x20))
+            leafOutputToken := calldataload(add(params.offset, 0x40))
+        }
+        if (leafDestinationChainId != cloneArgs.destinationChainId || leafOutputToken != cloneArgs.outputToken)
+            revert InvalidIdentity();
```
fusmanii: submitter passed in both `params` and `cloneArgs` kind of strange that we are checking them against each other, is there a way you think we can just have the submitter pass them once?

contracts/periphery/counterfactual/CounterfactualDepositCCTP.sol
```diff
@@ -27,90 +34,130 @@ struct CCTPDepositParams {
...
  */
 struct CCTPSubmitterData {
     uint256 amount;
     address executionFeeRecipient;
     bytes32 nonce;
     uint256 cctpDeadline;
+    uint256 executionFee;
+    uint32 signatureDeadline;
+    bytes peripherySignature;
     bytes signature;
```
fusmanii:
```suggestion
    bytes authSignature;
```
or something similar to distinguish it form the other signature

contracts/periphery/counterfactual/CounterfactualDepositCCTP.sol
```diff
@@ -27,90 +34,130 @@ struct CCTPDepositParams {
...
+
+        if (sd.executionFee > dp.maxExecutionFee) revert MaxExecutionFee();
+
         address inputToken = address(uint160(uint256(dp.burnToken)));
 
-        if (dp.executionFee > 0) IERC20(inputToken).safeTransfer(sd.executionFeeRecipient, dp.executionFee);
+        if (sd.executionFee > 0) IERC20(inputToken).safeTransfer(sd.executionFeeRecipient, sd.executionFee);
 
-        uint256 depositAmount = sd.amount - dp.executionFee;
+        uint256 depositAmount = sd.amount - sd.executionFee;
```
fusmanii: what happens if `executionFee > amount`? I guess the funds will just get stuck until more funds are sent

contracts/periphery/counterfactual/CounterfactualDepositSpokePool.sol
```diff
@@ -168,34 +168,39 @@ contract CounterfactualDepositSpokePool is ICounterfactualImplementation, EIP712
...
         uint256 outputAmount,
-        uint256 depositAmount
+        uint256 depositAmount,
+        uint256 executionFee
     ) private pure {
         uint256 outputInInputToken = (outputAmount * dp.stableExchangeRate) / EXCHANGE_RATE_SCALAR;
         uint256 relayerFee = depositAmount > outputInInputToken ? depositAmount - outputInInputToken : 0;
-        uint256 totalFee = relayerFee + dp.executionFee;
+        uint256 totalFee = relayerFee + executionFee;
         uint256 maxFee = dp.maxFeeFixed + (dp.maxFeeBps * inputAmount) / BPS_SCALAR;
```
fusmanii: why does only SpokePool have a fixed and dynamic fee but the rest CCTP/OFT have only fixed?

contracts/periphery/counterfactual/RoutePolicy.sol
```diff
@@ -0,0 +1,30 @@
...
+ * @title RoutePolicy
+ * @notice Minimal `Ownable` contract holding one merkle root that enumerates the routes a set of
+ *         counterfactual clones may execute on this chain. The owner (typically a multisig) calls
+ *         `updateRoot(newRoot)` to upgrade the route set globally for every clone pointing at this policy.
+ * @dev The contract address is intended to be identical across every EVM chain (deployed via the
+ *      deterministic-deployment proxy with constant constructor args); each chain maintains its own
+ *      independent `activeRoot` storage.
+ * @custom:security-contact bugs@across.to
+ */
+contract RoutePolicy is IRoutePolicy, Ownable {
```
fusmanii: Should all the leaf verification happen here? like even the leaf construction? that way when a RouterPolicy is replaced the leaf computation also gets updated

fusmanii (commented):
first pass

contracts/periphery/counterfactual/CounterfactualDeposit.sol
tbwebb22: Well the leaves on the policy (params) can point to different destination chains and tokens. We have to ensure that the selected leaf targets the same destination as this clone is. If we don't the submitter could select a leaf that sends to the wrong destination chain or token.

contracts/periphery/counterfactual/CounterfactualDepositCCTP.sol
tbwebb22: yeah good call, maybe rename it to `counterfactualSignature`?

contracts/periphery/counterfactual/CounterfactualDepositCCTP.sol
tbwebb22: Thats right. And I think thats fine. Basically `maxExecutionFee` is a lower bound for how much the user should be sending through the counterfactual.

contracts/periphery/counterfactual/CounterfactualDepositSpokePool.sol
tbwebb22: So the CCTP / OFT implementations take `maxFee` as a param, so we're kind of delegating that dynamic fee part to the sponsored bridging contracts

contracts/periphery/counterfactual/RoutePolicy.sol
tbwebb22: Can you elaborate on what you mean here? I'm not sure I follow

contracts/periphery/counterfactual/CounterfactualDepositCCTP.sol
tbwebb22: renamed to `counterfactualSignature` here https://github.com/across-protocol/contracts/pull/1434/changes/0c5f57d9c33e7eb191dd44e70f98a5de08a906f7

contracts/periphery/counterfactual/CounterfactualDeposit.sol
fusmanii: right, I was just thinking more along the lines submitter provides `destinationChainId` and `outputToken` and the function then constructs params/cloneArgs, but that might be uglier

contracts/periphery/counterfactual/RoutePolicy.sol
fusmanii: mainly thinking if these lines from `CounterfactualDeposit.sol` can be moved here?

```
        bytes32 leaf = keccak256(bytes.concat(keccak256(abi.encode(implementation, keccak256(params)))));
        bytes32 root = IRoutePolicy(cloneArgs.routePolicyAddress).activeRoot();
        if (!MerkleProof.verify(proof, root, leaf)) revert InvalidProof();
```

tbwebb22: yeah that definitely could work. Do you think its cleaner?
It would be a little more gas (Claude is telling me +1k gas) since we're passing more calldata in the external call

contracts/periphery/counterfactual/CounterfactualDeposit.sol
grasphoper: I think we can remove duplication here, see slack

tbwebb22: done https://github.com/across-protocol/contracts/commit/62597e55cc38d87a93834b947126ba099e1a0ebd

contracts/periphery/counterfactual/CounterfactualDepositCCTP.sol
```diff
@@ -27,90 +32,131 @@ struct CCTPDepositParams {
...
 
-    constructor(address _srcPeriphery, uint32 _sourceDomain) {
+    /// @notice Local signer that authorizes the runtime `executionFee`.
+    address public immutable signer;
+
+    constructor(
+        address _srcPeriphery,
+        uint32 _sourceDomain,
+        address _signer
+    ) EIP712("CounterfactualDepositCCTP", "v1.1.0") {
```
grasphoper: nit: this one is a breaking change, so we should have `v2.0.0`, rather than a v1.1.0

contracts/periphery/counterfactual/CounterfactualDepositCCTP.sol
```diff
@@ -27,90 +32,131 @@ struct CCTPDepositParams {
...
-     * @dev Bridges tokens via SponsoredCCTP. `params` is ABI-encoded as `CCTPDepositParams`;
-     *      `submitterData` as `CCTPSubmitterData` (includes a signature forwarded to the CCTP periphery).
-     *      ERC-20 only (no native tokens). No local signature verification — delegated to `srcPeriphery`.
+     * @dev ERC-20 only (no native tokens). `finalRecipient` and `finalToken` for the CCTP quote
+     *      come from `cloneArgs.recipient` / `cloneArgs.outputToken`.
      */
-    function execute(bytes calldata params, bytes calldata submitterData) external payable {
+    function execute(
+        CloneArgs calldata cloneArgs,
+        bytes calldata params,
```
grasphoper: `params` is a bit confusing now. What about `routeParams` or `routePolicyParams`? `params` used to mean "something that counterfactual committed to. Now it means, something that we confirmed against a route policy Merkle tree

contracts/periphery/counterfactual/CounterfactualDepositCCTP.sol
```diff
@@ -27,90 +32,131 @@ struct CCTPDepositParams {
...
-     * @param sd Submitter-provided execution data.
-     * @param depositAmount Amount to deposit after deducting the execution fee.
-     */
-    function _depositForBurn(CCTPDepositParams memory dp, CCTPSubmitterData memory sd, uint256 depositAmount) private {
+    function _verifySignature(bytes32 paramsHash, CCTPSubmitterData memory sd) private view {
+        if (block.timestamp > sd.signatureDeadline) revert SignatureExpired();
+        bytes32 structHash = keccak256(
+            abi.encode(
+                EXECUTE_CCTP_TYPEHASH,
+                address(this),
```
grasphoper: `address(this)` is already part of `_hashTypedDataV4` derivation below, so this one is redundant

contracts/interfaces/ICounterfactualImplementation.sol
```diff
@@ -1,17 +1,25 @@
...
+     * @param cloneArgs Dispatcher-verified identity fields (output token, destination, recipient, etc.).
      * @param params ABI-encoded route parameters committed to in the merkle leaf.
      * @param submitterData ABI-encoded data supplied by the caller at execution time.
      */
-    function execute(bytes calldata params, bytes calldata submitterData) external payable;
+    function execute(
+        CloneArgs calldata cloneArgs,
+        bytes calldata params,
+        bytes calldata submitterData
+    ) external payable;
```
grasphoper, line 35: I think we should change this interface to something like:

```
recipient,
tokenOut,
destinationChainId, // sometimes unused, but has to be on every leaf IMO
routeParams,
submitterData
```

Reasoning: 
1. cloneArgs is something that's responsibility or CounterfactualDeposit. 
2. Not all of cloneArgs params were checked against the route root
3. By passing cloneArgs here, we're essentially passing half-vetted params into the implementation (essentially an adapter)

Instead, by changing to what I proposed above, this adapter will know to have received only vetted params. Or rather, this impl doesn't have to think about the params at all, it's just executing with all of the params that it's been given

contracts/periphery/counterfactual/RoutePolicy.sol
fusmanii: actually now that I think about this, its not going to work since `RoutePolicy` isn't upgradable so doesn't give us any benefit with moving the logic there

contracts/periphery/counterfactual/CounterfactualDepositCCTP.sol
tbwebb22: removed `address(this)` and `amount` from both the CCTP and OFT sigs https://github.com/across-protocol/contracts/commit/59e71b6d9801b09860210011b23a4abdf475c657

contracts/interfaces/ICounterfactualImplementation.sol
tbwebb22: done here https://github.com/across-protocol/contracts/commit/178fbaecb76f0dc5268a739ec31a9de3056be4dc
in a later commit changed `WithdrawImplementation` back to being `ICounterfactualImplementation`

contracts/periphery/counterfactual/CounterfactualDepositCCTP.sol
tbwebb22: https://github.com/across-protocol/contracts/commit/c0c4e2246e000d57c988a8bccba0e35747dcb0f5

contracts/periphery/counterfactual/CounterfactualDepositCCTP.sol
tbwebb22: renamed to `routeParams` https://github.com/across-protocol/contracts/commit/ede145a6b24155a746be5a6471df352802356ffc

droplet-rl (commented):
Solid PR — the spec doc reads cleanly and the contract changes track it 1:1. I went through the dispatcher, factory, RoutePolicy, all three bridge impls, WithdrawImplementation, AdminWithdrawManager, and the tests. No blocking bugs found. The two open questions you flagged are the right ones to focus on; opinions inline.

## On the open questions

**1. Withdraw architecture — recommend (a) structural + separate (current), with a tweak.**
The fund-recovery guarantee is the load-bearing property here, and (a) is the cheapest way to preserve it without coupling the dispatcher to transfer logic. The vestigial `address(this), recipient, outputToken, destinationChainId` args on `WithdrawImplementation.execute` are a minor cost relative to keeping the dispatcher's "every execute goes through an impl" invariant clean.

That said, the `withdrawImpl` immutable in `AdminWithdrawManager` is mostly cosmetic: since `admin` can call any impl through the dispatcher's escape, the immutable doesn't actually pin anything. It does make the manager's two paths slightly less flexible (can't migrate to a new withdraw impl without redeploying the manager — and since `cloneArgs.admin = manager` is itself immutable in the clone's argsHash, that effectively pins withdrawImpl per-clone for life). Consider whether you'd rather make it mutable (owner-set) or drop the field entirely and have callers pass the impl. (c) policy-committed is correctly ruled out for the reason you noted.

**2. Signature replay — strong recommend the one-time signature marks (option c) before mainnet.**
The replay surface is real and asymmetric across the four paths:
- `SignedWithdraw` is the worst — committed `(token, to, amount)` triple → a single refund of `amount` lets the same sig drain again to the same `to`, no further coordination.
- CCTP/OFT — replayable if the clone is refunded to `≥ sd.amount` within the deadline.
- SpokePool — same shape; refund-then-redeposit is the attack.

The "rely on signer-side bookkeeping + short deadlines" path puts the entire safety property in a single off-chain process. A leaked or buffered signature inside its deadline is exploitable with zero on-chain protection. ~1 SSTORE per execute is cheap insurance, and the ERC-7201 mixin for the three bridge impls is a clean ~30 LOC. I'd push to land this in this PR rather than ship the design with the known footgun.

## Other observations

A few smaller things worth considering — none blocking.

- **`RoutePolicy` uses single-step `Ownable`.** The deployment story has a deployer EOA calling `transferOwnership(chainLocalMultisig)` per chain. A typo there permanently bricks the policy on that chain (no way back — and the address is supposed to be uniform cross-chain, so recovering means deploying a new policy at a new address and re-rooting every clone, breaking address consistency for affected chains). `Ownable2Step` shifts that risk meaningfully without adding much complexity.

- **Inconsistent EIP-712 binding of `clone`.** `CounterfactualDepositSpokePool.EXECUTE_DEPOSIT_TYPEHASH` includes `address clone` as a struct field; CCTP / OFT / SignedWithdraw rely on the EIP-712 domain separator's `verifyingContract` alone. Both are safe (domain separator already binds `address(this)`), but the inconsistency is a future-confusion vector. Either pick one and apply it everywhere, or note the asymmetry in the impl comments.

- **Test for cross-chain address consistency is tautological.** `RoutePolicy.t.sol` lines 108–126 compute `computeCreate2Address(...)` twice with identical inputs and assert equality — that's a no-op. The interesting property is "same `(deployer, salt, codeHash)` across chains where only `block.chainid` differs gives the same address" or "different deployer addresses give different addresses." Worth either fixing or dropping.

- **`updateRoot(bytes32(0))` is unguarded.** Open Question #1 covers the broader compromise mitigation, but worth noting the specific footgun: an owner can brick the policy in one tx. Admin escape covers fund recovery, so this is a liveness issue rather than a safety issue. A min-non-zero check would catch fat-fingers cheaply.

- **`AdminWithdrawManager.setSigner` / `setDirectWithdrawer` accept `address(0)`.** OZ ECDSA v5 reverts on signatures recovering to `address(0)`, so no forgery risk — just a liveness footgun.

- **`RootUpdated(bytes32 newRoot)` is not indexed.** Indexers can filter on it regardless, but `indexed` makes per-root subscriptions cheaper. Drive-by suggestion.

- **`CounterfactualDeposit.execute` is `payable` and `msg.value` carries through the delegatecall.** Fine for OFT (LayerZero fees) and native SpokePool deposits; just confirming this is intentional — the dispatcher itself does no msg.value accounting and trusts each impl to handle it. (It does.)

contracts/periphery/counterfactual/RoutePolicy.sol
droplet-rl: Consider `Ownable2Step` instead of `Ownable`. The deployment flow has the deployer EOA calling `transferOwnership(chainLocalMultisig)` separately per chain (per the spec and `Deployment` section of `RoutePolicies.md`); single-step transfer means a typo or wrong-address transfer permanently bricks the policy on that chain. Recovery would require deploying a new `RoutePolicy` at a new address, which breaks the cross-chain address-consistency invariant for the affected chain. `Ownable2Step` requires the multisig to `acceptOwnership()` and eliminates this class of fat-finger.

contracts/periphery/counterfactual/RoutePolicy.sol
```diff
@@ -0,0 +1,30 @@
...
+    /// @inheritdoc IRoutePolicy
+    bytes32 public activeRoot;
+
+    constructor(address initialOwner, bytes32 initialRoot) Ownable(initialOwner) {
+        activeRoot = initialRoot;
+    }
+
+    /// @inheritdoc IRoutePolicy
+    function updateRoot(bytes32 newRoot) external onlyOwner {
```
droplet-rl: `updateRoot(bytes32(0))` is allowed and bricks the policy in one tx (admin escape still recovers funds, so this is liveness, not safety). The PR's Open Question #1 already covers the broader compromise mitigation; if you want the cheapest possible fat-finger guard without committing to a timelock, an explicit `if (newRoot == bytes32(0)) revert` here would catch it. Note that this would also block the legitimate "freeze the policy" use case — your call.

contracts/periphery/counterfactual/AdminWithdrawManager.sol
```diff
@@ -5,108 +5,110 @@ import { Ownable } from "@openzeppelin/contracts/access/Ownable.sol";
...
     error SignatureExpired();
 
-    /// @notice EIP-712 typehash for signed withdraw messages.
+    /// @notice EIP-712 typehash for signed withdraw messages. The signer fixes `to`.
     bytes32 public constant SIGNED_WITHDRAW_TYPEHASH =
-        keccak256("SignedWithdraw(address depositAddress,address token,uint256 amount,uint256 deadline)");
+        keccak256("SignedWithdraw(address depositAddress,address token,address to,uint256 amount,uint256 deadline)");
+
+    /// @notice Canonical `WithdrawImplementation` address (passed to the dispatcher's escape path).
+    address public immutable withdrawImpl;
```
droplet-rl: `withdrawImpl` is immutable and `cloneArgs.admin = address(this)` is committed in the clone's argsHash for life. Together that effectively pins the withdraw implementation per clone — to migrate to a fixed-up withdraw impl, you'd have to deploy a new manager *and* redeploy every clone (different argsHash → different CREATE2 address). Worth either (a) making this an owner-settable field, or (b) dropping it and having `directWithdraw` / `signedWithdraw` take the impl as a parameter. Option (b) costs ~1 calldata word per call and removes the only deployment-ordering invariant for this contract.

contracts/periphery/counterfactual/AdminWithdrawManager.sol
```diff
@@ -5,108 +5,110 @@ import { Ownable } from "@openzeppelin/contracts/access/Ownable.sol";
...
-     * @param token Token to withdraw.
-     * @param amount Amount to withdraw.
-     * @param proof Merkle proof for the withdrawal leaf.
-     * @param deadline Timestamp after which the signature is no longer valid.
-     * @param signature EIP-712 signature from `signer`.
+     * @notice Signed withdraw — anyone can trigger with a valid `signer` signature.
+     * @dev The signer fixes the recipient (`to`) in the EIP-712 message; the caller cannot redirect.
      */
-    function signedWithdrawToUser(
+    function signedWithdraw(
```
droplet-rl, line 106: Option (c) one-time signature marks (per the PR description) is the right fix here — the EIP-712 message commits `(depositAddress, token, to, amount, deadline)` with no nonce or used-flag, so a refund of `amount` during the deadline window enables the same `(to, amount)` drain a second time. This path is the most replay-prone of the four. For the manager specifically, a regular `mapping(bytes32 sigHash => bool used)` in normal storage (no ERC-7201 needed since the manager isn't delegatecalled) is the minimal fix — set the flag after `ECDSA.recover` succeeds.

contracts/periphery/counterfactual/CounterfactualDepositSpokePool.sol
```diff
@@ -79,24 +69,25 @@ contract CounterfactualDepositSpokePool is ICounterfactualImplementation, EIP712
...
 
     error MaxFee();
     error InvalidSignature();
     error SignatureExpired();
     error NativeTransferFailed();
 
-    /// @notice EIP-712 typehash for execute deposit signature verification.
+    /// @notice EIP-712 typehash binding the signature to (clone, leaf, runtime fields).
     bytes32 public constant EXECUTE_DEPOSIT_TYPEHASH =
         keccak256(
```
droplet-rl, line 100: The typehash includes `address clone` as a struct field; CCTP / OFT typehashes rely on the EIP-712 domain separator (`verifyingContract = address(this) = clone`) alone. Both are equivalent for replay safety, but the inconsistency between the three bridge impls is a future-confusion vector. Either drop `clone` here (domain separator already binds it) or add the field to CCTP/OFT for symmetry. Mild preference for dropping — the typehash gets shorter and the binding lives in exactly one place (the domain separator).

contracts/periphery/counterfactual/CounterfactualDepositCCTP.sol
```diff
@@ -27,90 +31,129 @@ struct CCTPDepositParams {
...
     }
 
-    /**
-     * @notice Calls depositForBurn on the SponsoredCCTPSrcPeriphery with the constructed quote.
-     * @param dp Route parameters from the merkle leaf.
-     * @param sd Submitter-provided execution data.
-     * @param depositAmount Amount to deposit after deducting the execution fee.
-     */
-    function _depositForBurn(CCTPDepositParams memory dp, CCTPSubmitterData memory sd, uint256 depositAmount) private {
+    function _verifySignature(bytes32 routeParamsHash, CCTPSubmitterData memory sd) private view {
```
droplet-rl: Same replay concern as `AdminWithdrawManager.signedWithdraw` and the other bridge impls: the EIP-712 sig is reusable until `signatureDeadline` if the clone is re-funded with `≥ sd.amount` of the burn token during that window. The shared ERC-7201 `SignatureMarks` mixin you described in Open Question #2 is the cleanest fix — adding `if (used[sigHash]) revert; used[sigHash] = true;` here after `recover` would shut this path. Worth landing in the same PR rather than shipping with the known footgun.

contracts/periphery/counterfactual/CounterfactualDepositOFT.sol
```diff
@@ -26,85 +30,128 @@ struct OFTDepositParams {
...
     }
 
-    /**
-     * @notice Calls deposit on the SponsoredOFTSrcPeriphery with the constructed quote.
-     * @param dp Route parameters from the merkle leaf.
-     * @param sd Submitter-provided execution data.
-     * @param depositAmount Amount to deposit after deducting the execution fee.
-     */
-    function _deposit(OFTDepositParams memory dp, OFTSubmitterData memory sd, uint256 depositAmount) private {
+    function _verifySignature(bytes32 routeParamsHash, OFTSubmitterData memory sd) private view {
```
droplet-rl: Same replay concern as CCTP — `(paramsHash, executionFee, signatureDeadline)` plus the clone-bound domain separator is sufficient to bind *identity*, but nothing prevents a second execute with the same sig during the deadline window if the clone is re-funded. Recommend the shared `SignatureMarks` ERC-7201 mixin per Open Question #2.

test/evm/foundry/local/RoutePolicy.t.sol
```diff
@@ -0,0 +1,127 @@
...
+
+    function testNonOwnerCannotTransferOwnership() public {
+        vm.expectRevert(abi.encodeWithSelector(Ownable.OwnableUnauthorizedAccount.selector, other));
+        vm.prank(other);
+        policy.transferOwnership(other);
+    }
+
+    // --- Cross-chain address consistency: identical constructor args → identical address ---
+
+    function testIdenticalArgsProduceIdenticalAddress() public {
```
droplet-rl: This test is tautological — `computeCreate2Address(salt, codeHash, address(this))` returns the same value when called twice with identical inputs by construction. The interesting cross-chain consistency property (constructor args identical across chains → same address regardless of `block.chainid`) is implicit in how CREATE2 works (it doesn't use chainid) but isn't actually exercised here. Either rewrite to vary an input that *would* matter on different chains, or drop this test and rely on the production-side `CheckCounterfactualDeployments.s.sol` for cross-chain verification.

contracts/interfaces/IRoutePolicy.sol
```diff
@@ -0,0 +1,20 @@
...
+/**
+ * @title IRoutePolicy
+ * @notice Interface for `RoutePolicy` — a per-chain merkle root that authorizes which routes
+ *         a clone may execute. Clones reference a policy via their `routePolicyAddress` immutable
+ *         arg and the dispatcher queries `activeRoot()` on every non-withdraw execute.
+ * @custom:security-contact bugs@across.to
+ */
+interface IRoutePolicy {
+    /// @notice Emitted on every successful root update.
+    event RootUpdated(bytes32 newRoot);
```
droplet-rl: Drive-by: `event RootUpdated(bytes32 indexed newRoot)` makes per-root subscriptions cheaper for indexers. Non-blocking.

tbwebb22:
@codex review

grasphoper (dismissed):

contracts/periphery/counterfactual/CounterfactualDepositCCTP.sol
```diff
@@ -27,90 +31,131 @@ struct CCTPDepositParams {
...
+        if (block.timestamp > sd.signatureDeadline) revert SignatureExpired();
+        bytes32 structHash = keccak256(
+            abi.encode(EXECUTE_CCTP_TYPEHASH, routeParamsHash, sd.executionFee, sd.signatureDeadline)
+        );
```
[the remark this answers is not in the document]
grasphoper: I actually think that `structHash = abi.encode(routeParams.nonce, sd.executionFee, sd.signatureDeadline` can be all we need here.

Will tie the two signatures together effectively, since API commits to once nonce per unique quote

tbwebb22:
@codex review - note the changes to the IRoutePolicy interface, and the new RoutePolicyImmutableRoot implementation

contracts/periphery/counterfactual/CounterfactualCloneArgs.sol
```diff
@@ -0,0 +1,42 @@
...
+ * @param recipient Destination-chain address that receives `outputToken`.
+ * @param admin EVM address with full execution authority over this clone — can call any implementation
+ *              with any routeParams, bypassing the policy's merkle-proof requirement.
+ * @param routePolicyAddress `RoutePolicy` contract whose root authorizes this clone's routes.
+ */
+struct CloneArgs {
+    bytes32 outputToken;
+    uint256 destinationChainId;
+    bytes32 recipient;
+    address admin;
```
mrice32: Who do we intend the admin to be set to? Some admin multisig or the user?

contracts/periphery/counterfactual/RoutePolicyImmutableRoot.sol
```diff
@@ -0,0 +1,65 @@
...
+ *      implementation's constructor disables initializers on the implementation itself; the proxy
+ *      is initialized exactly once via `initialize(initialOwner)`. Both the implementation and the
+ *      proxy can be deployed deterministically (e.g. via the deterministic-deployment proxy) — on
+ *      day 0 every chain deploys with the same `initialRoot` (typically `bytes32(0)`), so impl and
+ *      proxy land at the same address everywhere. Subsequent per-chain rotations diverge: each
+ *      chain deploys its own implementation carrying that chain's root, and upgrades its proxy.
+ *      The proxy address stays constant on each chain throughout all rotations.
+ * @custom:security-contact bugs@across.to
+ */
+contract RoutePolicyImmutableRoot is IRoutePolicy, Initializable, OwnableUpgradeable, UUPSUpgradeable {
```
droplet-rl, line 31: Same recommendation as my prior review, now applied to the upgradeable variant: `Ownable2StepUpgradeable` instead of `OwnableUpgradeable`. The deployment story (`RoutePolicies.md` step 4) has the deployer EOA calling `transferOwnership(chainLocalMultisig)` separately per chain — single-step transfer leaves a fat-finger trap that bricks the policy on the affected chain, and recovery requires a fresh policy at a new address, which breaks cross-chain address consistency. Two-step transfer eliminates this class of footgun cheaply.

contracts/periphery/counterfactual/CounterfactualDepositCCTP.sol
```diff
@@ -27,112 +31,169 @@ struct CCTPDepositParams {
...
+    /// @notice EIP-712 typehash binding the local fee signature to (nonce, runtime fee, deadline).
+    /// @dev The clone is bound implicitly via the EIP-712 domain separator's `verifyingContract`
+    ///      field (= `address(this)` = the clone during delegatecall). `amount` is bound implicitly
+    ///      via the periphery signature, which covers `depositAmount = sd.amount - sd.executionFee`.
+    ///      The route is bound transitively: the periphery signature commits `(route, nonce)`
+    ///      together, so binding the local sig to `nonce` pins the route via the periphery's quote
+    ///      (and cleanly gives single-use replay protection — once the periphery consumes the
+    ///      nonce, the local sig can never be replayed).
+    bytes32 public constant EXECUTE_CCTP_TYPEHASH =
+        keccak256("ExecuteCCTP(bytes32 nonce,uint256 executionFee,uint32 signatureDeadline)");
```
droplet-rl, line 96: Confirmed this is now safe against in-deadline replay: `SponsoredCCTPSrcPeriphery.sol:85,89` rejects duplicate nonces, so binding the local sig to `nonce` transitively closes the replay window. Nice idea — no extra storage in the impl. (The comment block already explains this; flagging for reviewers' confidence.)

contracts/periphery/counterfactual/CounterfactualDepositOFT.sol
```diff
@@ -26,109 +30,171 @@ struct OFTDepositParams {
...
+    /// @notice EIP-712 typehash binding the local fee signature to (nonce, runtime fee, deadline).
+    /// @dev The clone is bound implicitly via the EIP-712 domain separator's `verifyingContract`
+    ///      field (= `address(this)` = the clone during delegatecall). `amount` is bound implicitly
+    ///      via the periphery signature, which covers `depositAmount = sd.amount - sd.executionFee`.
+    ///      The route is bound transitively: the periphery signature commits `(route, nonce)`
+    ///      together, so binding the local sig to `nonce` pins the route via the periphery's quote
+    ///      (and cleanly gives single-use replay protection — once the periphery consumes the
+    ///      nonce, the local sig can never be replayed).
+    bytes32 public constant EXECUTE_OFT_TYPEHASH =
+        keccak256("ExecuteOFT(bytes32 nonce,uint256 executionFee,uint32 signatureDeadline)");
```
droplet-rl, line 95: Same as CCTP — `SponsoredOFTSrcPeriphery.sol:221,98` consumes nonces single-use, so the nonce-binding here closes the local-sig replay too. Both peripheries verified.

contracts/periphery/counterfactual/CounterfactualDepositSpokePool.sol
```diff
@@ -79,24 +69,25 @@ contract CounterfactualDepositSpokePool is ICounterfactualImplementation, EIP712
...
     );
 
     error MaxFee();
     error InvalidSignature();
     error SignatureExpired();
     error NativeTransferFailed();
 
-    /// @notice EIP-712 typehash for execute deposit signature verification.
+    /// @notice EIP-712 typehash binding the signature to (clone, leaf, runtime fields).
     bytes32 public constant EXECUTE_DEPOSIT_TYPEHASH =
```
droplet-rl, line 99: Still replayable in deadline (OQ #2). Unlike CCTP/OFT, the SpokePool path has no downstream nonce consumption to lean on, so the `routeParamsHash`-bound sig remains valid against the same clone until `signatureDeadline` — a refund of `inputAmount` within that window enables a second deposit. Strong recommend either (a) adding a `bytes32 nonce` to the typehash + a `mapping(bytes32 => bool)` in clone storage (ERC-7201 to avoid colliding with other impls), or (b) committing `block.timestamp` lower-bound the signer must move forward for each subsequent sig. (a) is the cleaner fit for the broadcast-and-pick-up executor model.

contracts/periphery/counterfactual/AdminWithdrawManager.sol
droplet-rl, line 106: Still replayable in deadline (OQ #2). This path is the more dangerous of the two open replays — the signature commits a fixed `(token, to, amount)` triple, so any refund of `amount` during the deadline window enables a second drain to the same `to` with no further coordination. A regular `mapping(bytes32 sigHash => bool used)` in the manager's storage (no ERC-7201 since this contract isn't delegatecalled) checked + set after `ECDSA.recover` is the minimal fix.

contracts/periphery/counterfactual/AdminWithdrawManager.sol
droplet-rl: Same observation as my prior review: this immutable plus the fact that `cloneArgs.admin = address(this)` is itself in the clone's argsHash for life effectively pins the withdraw implementation per-clone. To rotate the withdraw impl you'd need a new manager *and* a redeploy of every clone (which changes their addresses and defeats the address-persistence goal). Making this owner-settable, or moving it to a per-call argument, removes that constraint. Not blocking, but it costs nothing structurally and unlocks future flexibility.

contracts/periphery/counterfactual/CounterfactualDepositSpokePool.sol
```diff
@@ -79,24 +69,25 @@ contract CounterfactualDepositSpokePool is ICounterfactualImplementation, EIP712
...
         keccak256(
-            "ExecuteDeposit(uint256 inputAmount,uint256 outputAmount,bytes32 exclusiveRelayer,uint32 exclusivityDeadline,uint32 quoteTimestamp,uint32 fillDeadline,uint32 signatureDeadline)"
+            "ExecuteDeposit(address clone,bytes32 routeParamsHash,uint256 inputAmount,uint256 outputAmount,bytes32 exclusiveRelayer,uint32 exclusivityDeadline,uint32 quoteTimestamp,uint32 fillDeadline,uint32 signatureDeadline,uint256 executionFee)"
```
droplet-rl, line 101: Mild inconsistency with CCTP/OFT typehashes — those bind `clone` implicitly via the EIP-712 domain separator alone, while this one redundantly includes `address clone` in the struct. Both are safe; consider dropping `clone` here and relying on the domain separator (matches the CCTP/OFT pattern and shortens the typehash).

contracts/periphery/counterfactual/RoutePolicyImmutableRoot.sol
```diff
@@ -0,0 +1,65 @@
...
+     *      clone bound to this policy. Returns the implementation's immutable `_root`.
+     */
+    function activeRoot(address /* clone */) external view returns (bytes32) {
+        return _root;
+    }
+
+    /// @dev Required by UUPS — only the owner can upgrade the implementation. Upgrading is the
+    ///      mechanism by which the active root changes: the owner deploys a new implementation
+    ///      carrying the new root and points the proxy at it.
+    function _authorizeUpgrade(address) internal override onlyOwner {}
```
droplet-rl, line 64: Worth verifying as part of the deployment runbook: a chain whose policy proxy points at an impl whose code is somehow lost (selfdestruct of the impl, etc.) would be unable to `upgradeToAndCall` because the proxy's `_dispatchUpgradeToAndCall` selfdelegate calls into the current impl's `upgradeTo` via UUPS. The impls here have no selfdestruct, so this is theoretical, but the `CheckCounterfactualDeployments.s.sol` (still pending) should at least assert each chain's proxy has a non-zero current impl with the expected code hash for the deployed impl revision.

droplet-rl (commented):
Re-reviewed against the head of `taylor/counterfactual-route-policy` (`9d6aee9`). I focused on what changed since my prior review (`59e71b6`): the CCTP/OFT replay mitigation, the new `RoutePolicyImmutableRoot`, the `admin` arg added to `ICounterfactualImplementation`, and the `WithdrawImplementation` defense-in-depth check.

## What the new diff resolves cleanly

- **CCTP/OFT replay (OQ #2 partial close).** Binding the local `counterfactualSignature` to `nonce` instead of `routeParamsHash` is the right shape. I verified both peripheries (`SponsoredCCTPSrcPeriphery.sol:85,89` and `SponsoredOFTSrcPeriphery.sol:221,98`) reject duplicate nonces and mark them used, so once the periphery consumes the nonce, the local sig is also burned — no on-chain bookkeeping needed in the bridge impl itself. The replay window is genuinely closed for these two paths.
- **`WithdrawImplementation` defense-in-depth.** The explicit `if (msg.sender != admin) revert Unauthorized()` covers the "policy owner accidentally adds a withdraw leaf" footgun. `admin` flows through the dispatcher's verified `cloneArgs`, so this is safe to trust here.
- **`RoutePolicyImmutableRoot`.** Bytecode-immutable root + UUPS upgrade-as-rotation is a reasonable trade — reads become bytecode constant loads instead of SLOADs, at the cost of a small impl deploy per rotation. The interface adds the `clone` arg for forward compatibility with per-clone roots without a future breaking change. Good move.
- **Cross-chain consistency test.** `testIdenticalDay0ArgsProduceIdenticalProxyAddress` now actually exercises the property (impl → proxy CREATE2 chain, then asserts an arg change cascades through to a different proxy address). My prior comment about the tautological test is closed.

## Still open

- **OQ #2 for `SpokePool` + `AdminWithdrawManager.signedWithdraw`.** Neither has a nonce-like field that something downstream consumes single-use, so both remain replayable within `signatureDeadline` if the clone is re-funded. The PR description acknowledges this. My recommendation hasn't changed: land a one-time signature-mark mapping for these two before mainnet. `SignedWithdraw` is the more dangerous of the pair (committed `(token, to, amount)` triple → re-drain to the same `to` on refund).
- **Single-step `Ownable` → `Ownable2Step`.** Still applies, now on `OwnableUpgradeable`. The deployment story calls a single-step `transferOwnership(chainLocalMultisig)` per chain (`RoutePolicies.md` step 4); a typo permanently bricks the policy on that chain, which then can't be recovered without breaking cross-chain address consistency. `Ownable2StepUpgradeable` removes that class of fat-finger.
- **`AdminWithdrawManager.withdrawImpl` immutable.** Still vestigial in the sense I flagged — since `cloneArgs.admin = manager` is in the clone's argsHash for life and the manager pins its `withdrawImpl`, the withdraw impl is effectively immutable per-clone. Either making it owner-settable or dropping it (and having the manager pass it in per call) would eliminate the only deployment-ordering invariant left on this contract.
- **Inconsistent typehash binding.** SpokePool typehash now includes `address clone` while CCTP/OFT bind clone only via the EIP-712 domain separator. Both are safe; the inconsistency is the only thing worth tidying up.

## Drive-by on the new code

- `RoutePolicyImmutableRoot._authorizeUpgrade(address)` doesn't constrain what the new impl looks like beyond `onlyOwner`. Documented in OQ #1. The owner can in principle deploy an impl whose `activeRoot` does anything; the admin escape covers fund safety regardless. Fine.
- `_disableInitializers()` in the impl constructor is the right move (line 39). Test at line 52 covers it.
- The genesis-root-must-be-identical-across-chains invariant is now load-bearing for cross-chain proxy address consistency. Worth a sanity check in `CheckCounterfactualDeployments.s.sol` when that's wired up: assert the proxy's current impl matches a chain-specific impl after first rotation, and that the proxy address itself matches a reference chain.

Overall: the design has tightened substantially since my first pass. With OQ #2 closed for SpokePool + the manager, and `Ownable2Step` on the policy, this is solidly ship-ready.

contracts/periphery/counterfactual/CounterfactualDeposit.sol
```diff
@@ -5,52 +5,93 @@ import { Clones } from "@openzeppelin/contracts/proxy/Clones.sol";
...
+            // Verify merkle proof against the policy's active root. The leaf preimage binds the
+            // clone's identity (outputToken, destinationChainId) so a leaf can only be proven against
+            // the clone it was authored for — no separate identity check needed.
+            bytes32 leaf = keccak256(
+                bytes.concat(
+                    keccak256(
+                        abi.encode(
+                            implementation,
+                            cloneArgs.outputToken,
+                            cloneArgs.destinationChainId,
```
mrice32: Can we drop these and let the leaf validate that it knows how to handle this particular outputToken and destinationChainId (via routeParams or otherwise)?

contracts/periphery/counterfactual/CounterfactualDepositCCTP.sol
```diff
@@ -27,112 +31,169 @@ struct CCTPDepositParams {
...
- * @dev Called via delegatecall from the CounterfactualDeposit dispatcher.
+ * @notice Bridges tokens from a counterfactual clone via SponsoredCCTP.
+ * @dev Called via delegatecall from the dispatcher. Two signatures are checked per execute:
+ *      (1) the periphery quote signature (forwarded to the CCTP src periphery unchanged), and
+ *      (2) the local EIP-712 signature authorizing the runtime `executionFee` (bounded by
+ *      the leaf's `maxExecutionFee`).
  * @custom:security-contact bugs@across.to
  */
-contract CounterfactualDepositCCTP is ICounterfactualImplementation {
+contract CounterfactualDepositCCTP is ICounterfactualImplementation, EIP712 {
```
mrice32, line 71: Does this only work with chains where destination SponsoredBridging contracts are deployed?

I'm assuming we don't have a way to do a vanilla CCTP leaf, right? I could see that being nice for destination Solana, for instance.

contracts/periphery/counterfactual/CounterfactualDepositOFT.sol
```diff
@@ -26,109 +30,171 @@ struct OFTDepositParams {
...
- * @dev Called via delegatecall from the CounterfactualDeposit dispatcher.
- *      msg.value covers LayerZero native messaging fees.
+ * @notice Bridges tokens from a counterfactual clone via SponsoredOFT (LayerZero).
+ * @dev Called via delegatecall from the dispatcher. `msg.value` is forwarded to the OFT periphery
+ *      to cover LayerZero native messaging fees. Two signatures are checked per execute: the
+ *      periphery quote signature (forwarded unchanged) and the local EIP-712 fee signature.
  * @custom:security-contact bugs@across.to
  */
-contract CounterfactualDepositOFT is ICounterfactualImplementation {
+contract CounterfactualDepositOFT is ICounterfactualImplementation, EIP712 {
```
mrice32, line 70: Similar to CCTP, we don't have a way to talk to non sponsored bridging OFT contracts, right?

I'm assuming there's nothing stopping us from doing that in the future, right?

contracts/periphery/counterfactual/WithdrawImplementation.sol
```diff
@@ -5,39 +5,38 @@ import { ICounterfactualImplementation } from "../../interfaces/ICounterfactualI
...
+    /// @inheritdoc ICounterfactualImplementation
+    function execute(
+        bytes32 /* recipient */,
+        bytes32 /* outputToken */,
+        uint256 /* destinationChainId */,
+        address admin,
+        bytes calldata /* routeParams */,
+        bytes calldata submitterData
+    ) external payable {
+        if (msg.sender != admin) revert Unauthorized();
```
mrice32: This would be the AdminWithdrawManager, right? Does that mean that we would need to have a distinct AdminWithdrawManager contract deployed for each counterfactual address that has their EOA encoded? Or am I missing something?

mrice32 (commented):
Overall looks great, just a few comments!

contracts/periphery/counterfactual/WithdrawImplementation.sol
mrice32: And if it doesn't, then does that mean the signed withdrawer (i.e. the API) can authorize a withdrawal to any address (since the unique refund address isn't known/stored onchain)?

contracts/periphery/counterfactual/CounterfactualCloneArgs.sol
tbwebb22: so this is intended to be set to the `AdminWithdrawManager` contract - which in turn allows its`directWithdrawer` (should be a multsig) to call `directWithdraw`, or for an executor to call `signedWithdraw`

contracts/periphery/counterfactual/CounterfactualDepositCCTP.sol
tbwebb22: Yep thats right, this is specific to our SponsoredBridging contracts. And yeah for a vanilla CCTP leaf we would need a new implementation for that specifically

contracts/periphery/counterfactual/CounterfactualDepositOFT.sol
tbwebb22: Correct, and yeah no reason we couldn't add an implementation for this in the future

contracts/periphery/counterfactual/WithdrawImplementation.sol
tbwebb22: > This would be the AdminWithdrawManager, right? Does that mean that we would need to have a distinct AdminWithdrawManager contract deployed for each counterfactual address that has their EOA encoded? Or am I missing something?

so I'm envisioning that there would generally be a global AdminWithdrawManager instance per chain that all counterfactual clones point to as their `admin`

tbwebb22: > does that mean the signed withdrawer (i.e. the API) can authorize a withdrawal to any address (since the unique refund address isn't known/stored onchain)?

thats correct. Do you think this is placing to much trust in the API? And that the user's refund address should be store onchain?

tbwebb22:
@codex review the last commit that restructured the merkle tree and moved outputToken and destinationChain into routeParams

contracts/periphery/counterfactual/CounterfactualDeposit.sol
tbwebb22: https://github.com/across-protocol/contracts/commit/8897c8171ba798e9d89c1a0a182a770ea0b39117

tbwebb22:
@codex review the last commit that updated the withdraw flow

contracts/periphery/counterfactual/WithdrawImplementation.sol
```diff
@@ -5,47 +5,57 @@ import { ICounterfactualImplementation } from "../../interfaces/ICounterfactualI
...
+    /// @inheritdoc ICounterfactualImplementation
+    function execute(
+        bytes32 /* recipient */,
+        bytes32 /* outputToken */,
+        uint256 /* destinationChainId */,
+        address userAddress,
+        bytes calldata /* routeParams */,
+        bytes calldata submitterData
+    ) external payable {
+        if (msg.sender != admin && msg.sender != userAddress) revert Unauthorized();
```
droplet-rl, line 52: This is the right shape. Forcing the destination to `userAddress` instead of a caller-supplied `to` collapses three separate trust concerns (signer compromise, directWithdrawer compromise, signature replay) into a single benign outcome: the user receives their own funds. Strong improvement over the prior `(token, to, amount)` triple.

contracts/periphery/counterfactual/AdminWithdrawManager.sol
```diff
@@ -5,108 +5,119 @@ import { Ownable } from "@openzeppelin/contracts/access/Ownable.sol";
...
 
-    /// @notice EIP-712 typehash for signed withdraw messages.
+    /// @notice EIP-712 typehash for signed withdraw messages. The signature commits to the target
+    ///         `withdrawImpl` so a submitter cannot redirect an authorized withdrawal to a
+    ///         different impl. Recipient is not signed — it is fixed to `cloneArgs.userAddress`
+    ///         by `WithdrawImplementation`.
     bytes32 public constant SIGNED_WITHDRAW_TYPEHASH =
-        keccak256("SignedWithdraw(address depositAddress,address token,uint256 amount,uint256 deadline)");
+        keccak256(
+            "SignedWithdraw(address depositAddress,address withdrawImpl,address token,uint256 amount,uint256 deadline)"
```
droplet-rl, line 53: Good that `withdrawImpl` is committed in the typehash — a compromised submitter can't substitute a malicious impl on a valid signature. Combined with the policy's merkle gate (only policy-authorized impls can be reached) and the impl's `msg.sender == admin` check (only the manager or user can trigger this impl), the signer's authority is genuinely "when + how much" only. The destination question is resolved structurally inside the impl.

contracts/periphery/counterfactual/CloneIdentity.sol
```diff
@@ -0,0 +1,41 @@
...
+    error WrongDestinationChain();
+
+    /**
+     * @notice Revert unless the route-bound identity matches the clone's identity.
+     * @param routeOutputToken         `outputToken` value committed inside `routeParams`.
+     * @param cloneOutputToken         `outputToken` value forwarded by the dispatcher from `cloneArgs`.
+     * @param routeDestinationChainId  `destinationChainId` committed inside `routeParams`.
+     * @param cloneDestinationChainId  `destinationChainId` forwarded by the dispatcher from `cloneArgs`.
+     */
+    function enforce(
```
droplet-rl, line 32: The opt-in model relies on impl authors remembering to call this. The current three impls that need binding (SpokePool, CCTP, OFT) all do, and the natspec spells out the contract. Worth adding to the security-review checklist: every new `ICounterfactualImplementation` either calls `CloneIdentity.enforce` or has an explicit comment justifying why identity binding is unnecessary (e.g. `WithdrawImplementation` because destination is forced to `userAddress`).

contracts/periphery/counterfactual/CounterfactualDeposit.sol
```diff
@@ -5,52 +5,87 @@ import { Clones } from "@openzeppelin/contracts/proxy/Clones.sol";
...
+        if (cloneArgs.hash() != storedHash) revert InvalidCloneArgs();
 
-        // Double-hash to prevent leaf/internal-node ambiguity (OpenZeppelin standard).
-        bytes32 leaf = keccak256(bytes.concat(keccak256(abi.encode(implementation, keccak256(params)))));
+        // User escape — the clone's user can execute any implementation, bypassing the policy.
+        // Works even if `activeRoot == bytes32(0)` or the policy contract is bricked.
+        if (msg.sender != cloneArgs.userAddress) {
+            // Verify merkle proof against the policy's active root. The leaf commits only
+            // `(implementation, keccak256(routeParams))`
+            bytes32 leaf = keccak256(bytes.concat(keccak256(abi.encode(implementation, keccak256(routeParams)))));
```
droplet-rl, line 59: The leaf format change is a real design pivot — worth a paragraph in the changelog (or PR description) calling out that off-chain leaf builders must be updated to match the new commitment `keccak256(impl, keccak256(routeParams))` (without `outputToken` / `destinationChainId` at the dispatcher level). Easy to miss when porting indexers / SDK encoders.

contracts/periphery/counterfactual/CounterfactualDepositSpokePool.sol
```diff
@@ -79,24 +86,25 @@ contract CounterfactualDepositSpokePool is ICounterfactualImplementation, EIP712
...
 
     error MaxFee();
     error InvalidSignature();
     error SignatureExpired();
     error NativeTransferFailed();
 
-    /// @notice EIP-712 typehash for execute deposit signature verification.
+    /// @notice EIP-712 typehash binding the signature to (clone, leaf, runtime fields).
     bytes32 public constant EXECUTE_DEPOSIT_TYPEHASH =
         keccak256(
```
droplet-rl, line 100: Pure cosmetic: typehash still binds `address clone` explicitly while CCTP/OFT typehashes bind clone implicitly via the EIP-712 domain separator. Both are safe — the inconsistency is the only thing left to tidy. Drop `clone` here for symmetry, or add it to CCTP/OFT, but pick one.

droplet-rl (dismissed):
Re-reviewed against `af3fe4d`. Two new commits since my last pass: the withdraw-flow rewrite and the dispatcher-leaf simplification (`outputToken`/`destinationChainId` moved out of the leaf, into per-impl `routeParams` with opt-in binding via `CloneIdentity.enforce`).

This iteration is a notable security improvement. The two design changes work together to materially shrink the trust footprint of every authority in the system:

## Why this is meaningfully better

**`WithdrawImplementation` now forces destination to `userAddress`.** The destination is no longer chosen at call time — it's the clone-bound `userAddress`, period. This collapses several of my prior concerns at once:
- `SignedWithdraw` replay (OQ #2) is now benign: a replayed signature still routes to the user, not to an attacker. The "blast radius bounded" note in the open-questions section is correct.
- A compromised `directWithdrawer` can force withdrawals but cannot redirect them.
- A compromised `signer` can force withdrawals but cannot redirect them.
- The trust model collapses to "when and how much," never "where." Clean.

**`AdminWithdrawManager` no longer pins `withdrawImpl`.** Passed per-call now, with the signed typehash committing `withdrawImpl` so a compromised submitter can't substitute. The construction-order constraint I flagged is replaced with a clean dependency: manager deploys first, impl deploys second pointing at the manager. My prior concern about the immutable being vestigial / pinning impl per-clone is closed.

**Identity binding moved from dispatcher to per-impl `CloneIdentity.enforce`.** I went back and forth on this trade. The dispatcher's leaf is now simpler (`(impl, keccak256(routeParams))`), and each impl declares whether it needs identity binding. The footgun is real — a future impl author who needs binding and forgets to call `CloneIdentity.enforce` ships a vulnerability — but the convention is explicit in the lib's natspec and in the per-impl natspec, and the three current impls that need binding all call it. Auditors can grep for `CloneIdentity.enforce` to verify coverage on future additions. Acceptable trade.

**The `SpokePoolRouteParams` natspec now spells out *why* binding is required there** (the `stableExchangeRate` per-pair assumption). Important context that was implicit before; making it explicit reduces the chance a future maintainer drops the enforcement.

## What's still open (all flagged, none blocking)

- **`RoutePolicyImmutableRoot` still uses single-step `OwnableUpgradeable`.** I flagged this twice. The deployment story still has the deployer EOA single-step-transferring to a per-chain multisig; a typo bricks the policy on that chain, and recovery breaks cross-chain address consistency. `Ownable2StepUpgradeable` removes that class of fat-finger for free. Not blocking — it's an operational risk you can offset with paranoid deployment scripting + reviews — but it's the cheapest hardening left on the table. `AdminWithdrawManager`'s single-step `Ownable` is less load-bearing (per-chain manager, no cross-chain invariant) but worth flipping in the same pass for consistency.
- **OQ #2 SpokePool replay** is the last unresolved replay window, and it too is now mostly benign (the replay funds go to `cloneArgs.recipient`, which is the legitimate user destination). Worth still doing the one-time-marks fix before mainnet — "user receives their own money again" is still a UX/liveness footgun (someone can involuntarily move funds out of a clone the user wanted to keep funded) — but no longer security-load-bearing.
- **OQ #1 owner-compromise / timelock split** is the genuine remaining design question. Not for this PR.
- **`SpokePool` typehash redundantly includes `address clone`** while CCTP/OFT bind clone only via the domain separator. Pure cosmetic at this point.

## Process-level observation

The implementation plan's Scripts checklist (`RoutePolicies.md`) is still entirely unchecked. The deploy/check/rotate scripts (`DeployRoutePolicy.s.sol`, `RotateRoutePolicyRoot.s.sol`, `CheckCounterfactualDeployments.s.sol`, etc.) aren't in this PR. That's fine — splitting into a follow-up keeps this PR reviewable — but make sure the follow-up lands before any testnet deployment, and that `CheckCounterfactualDeployments.s.sol` asserts current-impl code hashes on every chain (the cross-chain proxy address consistency invariant depends on no chain having upgraded prematurely).

A defensive item for the security-review checklist: every new `ICounterfactualImplementation` either calls `CloneIdentity.enforce` or has an explicit comment justifying why identity binding is unnecessary (e.g. `WithdrawImplementation` because destination is forced to `userAddress`). Catches the future-maintainer footgun cheaply.

Approving. Nice work tightening this through three review cycles.

mrice32:
@codex review

mrice32 (approved):
LGTM!

PR #1435, opened 2026-05-19
fix: Check for token balance diff & localToken

fusmanii:


grasphoper (approved):
LGTM, but the tests are failing for some reason. Intermittent or should we fix?

PR #1446, opened 2026-05-29
feat: Tron MulticallHandler

tbwebb22:
Adds TronMulticallHandler, a Tron-specific MulticallHandler variant that keeps the existing multicall behavior but drains leftover ERC20/TRC20 balances through TronTransferLib so Tron USDT’s false-returning transfer does not break recovery. 

The base handler now exposes a virtual transfer hook, and the Tron build/deploy path, docs, and focused tests were updated to cover the new contract.

### DIFF
--- a/contracts/handlers/MulticallHandler.sol
+++ b/contracts/handlers/MulticallHandler.sol
@@ -109,7 +109,7 @@ contract MulticallHandler is AcrossMessageHandler, ReentrancyGuard {
             // ERC20 token.
             uint256 amount = IERC20(token).balanceOf(address(this));
             if (amount > 0) {
-                IERC20(token).safeTransfer(destination, amount);
+                _safeTransfer(token, destination, amount);
                 emit DrainedTokens(destination, token, amount);
             }
         } else {
@@ -177,6 +177,10 @@ contract MulticallHandler is AcrossMessageHandler, ReentrancyGuard {
 
         (bool success, ) = target.call{ value: value }(callData);
         if (!success) revert ReplacementCallFailed(callData);
+    }
+
+    function _safeTransfer(address token, address to, uint256 amount) internal virtual {
+        IERC20(token).safeTransfer(to, amount);
     }
 
     function _requireSelf() internal view {
--- a/test/evm/foundry/local/TronMulticallHandler.t.sol
+++ b/test/evm/foundry/local/TronMulticallHandler.t.sol
@@ -1,0 +1,50 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.0;
+
+import { Test } from "forge-std/Test.sol";
+
+import { MulticallHandler } from "../../../../contracts/handlers/MulticallHandler.sol";
+import { TronMulticallHandler } from "../../../../contracts/handlers/TronMulticallHandler.sol";
+import { TronTransferLib } from "../../../../contracts/libraries/TronTransferLib.sol";
+import { MockTronUSDT } from "../../../../contracts/test/MockTronUSDT.sol";
+
+contract TronMulticallHandlerTest is Test {
+    TronMulticallHandler handler;
+    MockTronUSDT usdt;
+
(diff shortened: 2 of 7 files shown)


tbwebb22:
@droplet-rl review

contracts/handlers/MulticallHandler.sol
```diff
@@ -179,6 +179,10 @@ contract MulticallHandler is AcrossMessageHandler, ReentrancyGuard {
         if (!success) revert ReplacementCallFailed(callData);
     }
 
+    function _safeTransfer(address token, address to, uint256 amount) internal virtual {
+        IERC20(token).safeTransfer(to, amount);
+    }
+
```
droplet-rl, line 185: Optional: a short comment here noting that this hook exists for chain-specific variants (e.g. Tron) to swap in non-standard ERC20 success checks would save future readers a trip to `TronMulticallHandler`. The `virtual` keyword alone doesn't communicate intent.

test/evm/foundry/local/TronMulticallHandler.t.sol
```diff
@@ -0,0 +1,50 @@
...
+
+        MulticallHandler.Instructions memory instructions = MulticallHandler.Instructions({
+            calls: new MulticallHandler.Call[](0),
+            fallbackRecipient: fallbackRecipient
+        });
+
+        vm.expectRevert(TronTransferLib.TronTransferCallReverted.selector);
+        handler.handleV3AcrossMessage(address(usdt), amount, address(0), abi.encode(instructions));
+    }
+}
```
droplet-rl, line 50: Consider adding a direct test for `drainLeftoverTokens()` (the external `onlySelf`-callable path) and one for `TronTransferBalanceMismatch` (e.g. a mock that returns ok but doesn't move balance). Both share `_drainRemainingTokens`, so coverage is transitive today, but explicit cases would lock in the contract surface against future refactors. Non-blocking.

droplet-rl (approved):
## Summary

LGTM. Small, focused change that mirrors the established Tron-variant pattern (`CounterfactualDepositSpokePoolTr`, `WithdrawImplementationTron`): expose a virtual `_safeTransfer` hook on the mainline contract and override only that hook in the Tron child to route through `TronTransferLib._safeTransferBalanceCheck`.

### What I verified

- **Surface area is minimal.** Parent change is one method extraction + `virtual` — keeps existing semantics (still calls `IERC20(token).safeTransfer`) and existing `MulticallHandler.t.sol` `expectCall` assertions on `transfer(...)` calldata remain valid.
- **Native ETH path unaffected** — `_drainRemainingTokens` only goes through `_safeTransfer` on the ERC20 branch; the native branch uses `destination.sendValue` and is untouched.
- **`makeCallWithBalance` correctly left alone** — it only reads `balanceOf` to inject calldata, no transfer, so no Tron-USDT exposure.
- **Tron build wiring is correct** — `TronPeripheryImports.sol` already aggregates OZ-v4 entry points; `TronMulticallHandler` (transitively OZ-v4 via parent) fits cleanly. `TronTransferLib`'s OZ-v5 `IERC20` import is fine: it's selector-compatible and only used inside the library.
- **Deploy script + README + `package.json` script** match the form of the sibling Tron periphery scripts.

### Tests

- Happy path (transfer returns `false`, balance moves) and revert path (blacklisted recipient → `TronTransferCallReverted`) are both covered against the existing `MockTronUSDT`.
- No direct test of `drainLeftoverTokens()` (the external `onlySelf` path) or of the `TronTransferBalanceMismatch` branch at the handler level, but both share `_drainRemainingTokens` / are exercised at the library level — non-blocking.

### Nits (non-blocking)

- A one-line note in `MulticallHandler.sol` next to the new `_safeTransfer` hook saying "override in chain-specific variants for non-standard ERC20 return values" would help future readers; the rationale otherwise lives only in `TronMulticallHandler` and the counterfactual README.
- `using SafeERC20 for IERC20;` in the parent remains active but unused on the override path — harmless.

Approving.

grasphoper (approved):

PR #1452, opened 2026-06-05
feat: Upgradeable Counterfactuals

tbwebb22:
Closes ACP-106


tbwebb22:
@codex review

droplet-rl (commented):
## Summary

Nice, well-scoped refactor. The architecture (BeaconProxy + dispatcher + per-bridge impls + signed runtime fees) is internally consistent and matches the very detailed `DESIGN.md`. All 48 counterfactual-related tests pass locally under `FOUNDRY_PROFILE=local-test`, including the new beacon-retargeting, `updateRootAndExecute`, `checkStableExchangeRate` flag, and Tron leaf-impl variants.

I have no blocking concerns. Verdict: **COMMENT** with a handful of small items, mostly stale docs/comments that don't match the post-D27/D28 (beacon + no-versioning) world, plus one minor consistency nit on the beacon initializer and one merge-hygiene note.

### What I looked at
- Core: `CounterfactualBeacon`, `CounterfactualDeposit`, `CounterfactualDepositFactory(Tron)`.
- Per-bridge impls: SpokePool / CCTP / OFT (route params, EIP-712 fee sig, `_checkFee`, `checkStableExchangeRate` gate, source-chain binding) and the Tron variants.
- Withdraw: `WithdrawImplementation(Tron)` + `AdminWithdrawManager` (direct + signed paths).
- Tests: all suites listed in Phase 4 of `DESIGN.md`.

### Architectural notes (positive)
- Storage-slot hygiene: beacon (`across.counterfactual.beacon.storage`) vs proxy state (`across.counterfactual.upgradeable.storage`) are distinct ERC-7201 namespaces; no risk of collision when the beacon-target is delegatecalled into a proxy frame.
- OZ `BeaconProxy`'s constructor calls `IBeacon(beacon).implementation()` and then delegatecalls into it, which means the factory's first `deploy()` will revert if the beacon's `implementation` is still `address(0)` — operationally enforces the deploy ordering (beacon → impl → `setImplementation` → factory).
- `updateRootAndExecute`'s skip-when-current pattern cleanly avoids the `RootUnchanged` revert in races where two relayers each try to bump+execute.
- SpokePool's typehash binding everything explicitly vs CCTP/OFT relying transitively on the periphery quote signature is well-justified given there's no SpokePool periphery in this path.

### Merge hygiene — please rebase

The PR is forked from `8a45d6eb`, one commit before master's #1446 (`init Tron MulticallHandler`). As a side effect, this diff currently shows `contracts/handlers/MulticallHandler.sol` reverted to its pre-#1446 state and `contracts/handlers/TronMulticallHandler.sol` (plus its test, its deploy script, and the `package.json` / `script/tron/README.md` references) deleted. Almost certainly accidental — please rebase against current master so #1446 is preserved on merge.

### Stale docs/comments that didn't have a clean anchor inside the diff hunks
A couple that I'd flag separately because the affected file isn't part of this PR's diff or the line is in an unchanged hunk:

- `contracts/periphery/counterfactual/CounterfactualDepositSpokePoolTr.sol` (unchanged on this branch, follow-up suggestion): the rationale comment "each clone's address is derived via CREATE2 from its implementation address" was true under the old EIP-1167 clone model, but under `BeaconProxy` the proxy's address depends on `(factory, salt, beacon, initialRoot)` — the *implementation* is not in the proxy preimage at all. The stated protection (cross-impl signature replay) still holds, but for a different reason: the route impl address is committed in the merkle leaf, and SpokePool's typehash binds `routeParamsHash`. The Tron variant also deliberately inherits the same EIP-712 domain name, so the domain-name argument doesn't add anything either.
- `contracts/periphery/counterfactual/CounterfactualDepositSpokePool.sol` around line 50: a couple of unchanged docstrings still say "the clone address" (leftover EIP-1167 wording — under the new model it's the `BeaconProxy` address). Cosmetic but inconsistent with the rest of the (well-written) DESIGN.md.

### Optional hardening
`CounterfactualDeposit._execute` does `implementation.delegatecall(...)` without validating the target is a contract. A delegatecall to an EOA succeeds as a no-op with empty returndata. The merkle-tree authoring is the trust boundary here (the leaf commits the address), so this is fine in principle — but a one-line `if (implementation.code.length == 0) revert NotAContract();` would catch tree-builder mistakes loudly. Up to you; if rejected, worth noting why in DESIGN.md.

### One DESIGN.md sanity check
Phase 0 still flags pinning a fixed `solc` / optimizer combo and adding a `forge inspect <c> bytecode` cross-chain parity check for the factory and `CounterfactualBeacon` as remaining work. With `salt` now caller-supplied (D30), automatic cross-chain parity only holds when the caller reuses `salt = 0`; the determinism guarantee on the *substrate* (factory + beacon address) is what makes that caller obligation meaningful. Worth confirming this Phase-0 bytecode-parity check is on the deploy checklist before any production deployment.

contracts/interfaces/ICounterfactualBeacon.sol
```diff
@@ -0,0 +1,29 @@
...
+import { IBeacon } from "@openzeppelin/contracts/proxy/beacon/IBeacon.sol";
+
+/**
+ * @title ICounterfactualBeacon
+ * @notice Global, per-chain registry that governs how upgradeable counterfactual proxies behave. It is
+ *         the **beacon** for every counterfactual `BeaconProxy`: `implementation()` (from `IBeacon`)
+ *         returns the single canonical implementation all proxies run, so changing it upgrades every
+ *         proxy at once (no per-proxy action). It also holds the `(proxy, latestRoot)` upgrade tree
+ *         `root` and the `version` / `minRequiredVersion` that gate per-proxy root freshness.
+ * @dev `implementation()` here is the **counterfactual** implementation (the beacon target) — distinct
```
droplet-rl: **Stale NatSpec.** Per D28 (versioning removed) and D29 (rename), this interface no longer holds `version` / `minRequiredVersion`. The docstring still claims it does. Suggest replacing the last sentence with something like:

> It also holds the `(proxy, latestRoot)` upgrade tree `root` that authorizes per-proxy root updates. Root updates are best-effort — there is no on-chain version/freshness gate.

Keeping the stale text will mislead an auditor reading the interface in isolation.

contracts/periphery/counterfactual/CounterfactualBeacon.sol
```diff
@@ -0,0 +1,94 @@
...
+     * @notice Initialize the registry.
+     * @param owner_ The admin (use a multisig).
+     * @param implementation_ Initial global implementation / beacon target (may be address(0) and set
+     *        later via `setImplementation` — the deploy flow is registry → impl → `setImplementation`).
+     * @param upgradeRoot_ Initial upgrade-tree root (may be 0 and set later).
+     */
+    function initialize(address owner_, address implementation_, bytes32 upgradeRoot_) external initializer {
+        __Ownable_init(owner_);
+        __Ownable2Step_init();
+        _setImplementation(implementation_);
```
droplet-rl: **Consistency gap.** `setImplementation` enforces `newImplementation.code.length != 0`, but `initialize` calls `_setImplementation(implementation_)` directly with no such check. The docstring above explicitly permits `address(0)` for lazy init, which is fine — but it would *also* silently accept an arbitrary non-zero EOA (a typo'd address), which the public setter would have rejected.

In practice, the first `factory.deploy()` would revert because OZ's `BeaconProxy` constructor calls `IBeacon(beacon).implementation()` and then delegatecalls into it, so a bad address surfaces quickly — but it's still a UX gap. Suggest:

```solidity
if (implementation_ != address(0) && implementation_.code.length == 0) revert NotAContract();
_setImplementation(implementation_);
```

Keeps the "may be address(0) and set later" affordance but rejects EOAs.

contracts/periphery/counterfactual/CounterfactualDeposit.sol
```diff
@@ -1,53 +1,117 @@
...
- *      that cover execution-time parameters (amounts, deadlines, etc.) but do not commit to the leaf's
- *      route-specific `params` (destination chain, tokens, recipient, etc.). If two leaves share the same
- *      implementation address, a caller could prove leaf A's route params while submitting an authorization
- *      signature intended for leaf B's route, since the signature is valid for either leaf. The system is
- *      intended to be used such that a clone's merkle tree never contains multiple leaves with the same
+ *      Note: some leaf implementations use authorization signatures that cover execution-time
+ *      parameters (amounts, deadlines) but not the leaf's route `params`. If two leaves share the same
+ *      implementation address, a caller could prove leaf A's params while submitting a signature meant
+ *      for leaf B. A clone's tree must therefore never contain multiple leaves with the same
  *      implementation address.
```
droplet-rl: Small precision nit on this caveat: the SpokePool impl's typehash *does* bind `routeParamsHash`, so the cross-leaf scenario you describe (prove leaf A, submit B's signature) is *only* a concern for CCTP/OFT — where the typehash is `(nonce, executionFee, signatureDeadline)` and the route/amount are bound only transitively via the periphery quote signature.

For CCTP/OFT it then collapses to a nonce-uniqueness invariant on the periphery: a single nonce can't authorize two different (route, amount) tuples. The tree-builder invariant (“never two leaves with the same implementation”) is still a fine belt-and-braces rule, but the failure mode is bridge-specific. Tightening the language here would make this easier to verify in audit.

contracts/periphery/counterfactual/CounterfactualDeposit.sol
```diff
@@ -1,53 +1,117 @@
...
+    function updateRootAndExecute(
+        bytes32 newRoot,
+        bytes32[] calldata updateProof,
+        address implementation,
+        bytes calldata params,
+        bytes calldata submitterData,
+        bytes32[] calldata executeProof
+    ) external payable {
+        // Skip the update (and its `RootUnchanged` revert) when the proxy is already current.
+        if (newRoot != activeRoot()) _updateRoot(newRoot, updateProof);
```
droplet-rl, line 98: Subtle but worth documenting: when `newRoot == activeRoot()`, the `updateProof` arg is **never validated**. The test `testUpdateRootAndExecuteSkipsUpdateWhenCurrent` deliberately passes a bogus proof and confirms it's ignored. That's the intended D32 behavior, but a caller reading the diff might think they're getting a fresh upgrade-tree authorization check on every call. A one-line `@dev` note like *“`updateProof` is ignored when the proxy is already at `newRoot`”* would make this less surprising.

contracts/periphery/counterfactual/CounterfactualDepositCCTP.sol
```diff
@@ -62,77 +68,125 @@ contract CounterfactualDepositCCTP is ICounterfactualImplementation {
...
+        if (block.chainid != routeParams.sourceChainId) revert SourceChainMismatch();
+        _verifySignature(submitterData);
+        if (submitterData.executionFee > routeParams.maxExecutionFee) revert MaxExecutionFee();
 
-        if (dp.executionFee > 0) IERC20(inputToken).safeTransfer(sd.executionFeeRecipient, dp.executionFee);
+        address inputToken = address(uint160(uint256(routeParams.burnToken)));
 
-        uint256 depositAmount = sd.amount - dp.executionFee;
+        if (submitterData.executionFee > 0)
+            IERC20(inputToken).safeTransfer(submitterData.executionFeeRecipient, submitterData.executionFee);
```
droplet-rl, line 114: Worth confirming explicitly in DESIGN.md (Open Question #2 / pre-deployment ordering) what happens if a relayer reuses a fee signature within `signatureDeadline` and the user re-funds the proxy. The CCTP periphery's nonce-uniqueness check is what blocks the *bridge* call, but the *local* `_verifySignature` here only binds `(nonce, executionFee, signatureDeadline)` — so a replay attempt would pay this execution fee, then fail at the periphery, then revert the whole tx (atomic). Good — but only because the fee payout precedes the periphery call in the same tx. Worth a one-line comment near this `safeTransfer` making that atomicity-ordering load-bearing claim explicit. (Same applies to OFT.)

contracts/periphery/counterfactual/CounterfactualDeposit.sol
```diff
@@ -1,53 +1,117 @@
...
+
+    /// @notice Emitted when `activeRoot` is updated via the upgrade tree.
+    event RootUpdated(bytes32 newRoot);
+
+    /// @dev Merkle proof against the registry's `(proxy, latestRoot)` tree failed.
+    error InvalidUpgradeProof();
+    /// @dev New root equals the current `activeRoot` (no-op).
+    error RootUnchanged();
+
+    constructor(ICounterfactualBeacon beacon) {
```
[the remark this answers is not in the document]
tbwebb22: Yes, deploy scripts have not been updated yet

contracts/periphery/counterfactual/CounterfactualBeacon.sol
tbwebb22: https://github.com/across-protocol/contracts/commit/98c903201f0ab25be3180131b3896aedb6b0d9e2

contracts/periphery/counterfactual/CounterfactualDeposit.sol
tbwebb22: https://github.com/across-protocol/contracts/commit/46c5d9955997356b3c7e4fd3e213d26718abb098

contracts/periphery/counterfactual/CounterfactualDeposit.sol
tbwebb22: https://github.com/across-protocol/contracts/commit/46c5d9955997356b3c7e4fd3e213d26718abb098

contracts/periphery/counterfactual/CounterfactualDepositCCTP.sol
tbwebb22: https://github.com/across-protocol/contracts/commit/8a55bbc8f7615ed0f3e54ca2a9d627b6f6d68daf

contracts/interfaces/ICounterfactualBeacon.sol
tbwebb22: https://github.com/across-protocol/contracts/commit/37f1aeceda90fac1544054c3c47940118d5b5c85

contracts/periphery/counterfactual/DESIGN.md
```diff
@@ -0,0 +1,414 @@
...
+### Implementation — global, via the beacon (instant)
+
+Implementation is shared logic, so it is administered **once, globally**: the admin calls
+`setImplementation(impl)` on the registry (the beacon). Because every counterfactual is a `BeaconProxy`
+that resolves `beacon.implementation()` **live on each call**, all proxies use the new implementation
+**immediately** — there is **no per-proxy upgrade, no `syncImplementation`, and no bootstrap**. The
+registry validates the target is a contract (`NotAContract`) **and that its immutable `BEACON()` points
+back at this beacon** (`WrongBeacon`) — catching the catastrophic error of retargeting every proxy to logic
+bound to a different beacon (which would silently brick `updateRoot`). (This guards the wrong-beacon
+footgun, **not** ERC-7201 storage-layout drift — see Open Question #5.) The admin is otherwise trusted (and
```
droplet-rl: **Dangling cross-reference.** The `## Open Questions` section was removed in this round, so "see Open Question #5" (and the similar refs on lines 238 and 383) point at nothing now. Either inline the relevant one-sentence answer or drop the parenthetical. Non-blocking.

droplet-rl (dismissed):
## Re-review — Approve

Branch was force-pushed (`e3164476 → f2c8dd0d`) with a clean rebase onto current master and 9 follow-up commits addressing my prior review. Re-verified the diff and reran the full counterfactual test suite under `FOUNDRY_PROFILE=local-test`: **74/74 passing** (was 70 — the new beacon-binding checks come with new tests).

### Prior review items — all addressed

| Item | Where | Resolution |
|---|---|---|
| Stale `version`/`minRequiredVersion` natspec | `ICounterfactualBeacon.sol:11-13` | Replaced with "Root updates are best-effort — there is no on-chain version/freshness gate." ✅ |
| `initialize` skipped the `code.length` check that `setImplementation` enforced | `CounterfactualBeacon.sol:55-63, 92-99` | Resolved with *more* rigor than requested: introduces `_validateImplementation` that (a) rejects EOAs (`NotAContract`) AND (b) checks the impl's immutable `BEACON()` points back at this beacon (`WrongBeacon`) — catching the catastrophic admin error of retargeting every proxy to logic bound to a different beacon. Applied in both `initialize` (when non-zero) and `setImplementation`. Three new tests: `testInitializeRejectsEoaImplementation`, `testInitializeAllowsZeroImplementation`, `testSetImplementationRejectsWrongBeacon`, `testSetImplementationRejectsNonConformingContract`. ✅ |
| "Cross-leaf same-impl" caveat was too broad (didn't apply to SpokePool) | `CounterfactualDeposit.sol:28-33` | Now correctly scopes to CCTP/OFT and explicitly notes SpokePool is unaffected because its typehash binds `routeParamsHash`. ✅ |
| `updateRootAndExecute` silently ignores `updateProof` when already at `newRoot` | `CounterfactualDeposit.sol:96-97` | Inline comment now states *"NOTE: `updateProof` is therefore NOT validated in that case — there is no root change to authorize."* ✅ |
| CCTP/OFT atomicity-ordering of fee transfer + periphery call | `CounterfactualDepositCCTP.sol:125-128`, `CounterfactualDepositOFT.sol:124-127` | Added matching multi-line comment in both impls making the load-bearing ordering explicit, with the nonce-uniqueness rationale. ✅ |
| PR forked from stale base, accidental Tron-multicall deletions | branch state | Resolved by the rebase — `334d063b init Tron MulticallHandler (#1446)` is now in the branch's history, and the prior accidental deletions are gone from the diff. ✅ |

The optional hardening I'd flagged (delegatecall-to-EOA) is unaddressed but called out as "up to you"; the tree-builder responsibility model is reasonable.

### Bonus improvements I noticed
- DESIGN.md gained a useful **Terminology** callout disambiguating the **Route Tree** (per-counterfactual, `activeRoot`) from the **Upgrade Tree** (per-chain, `beacon.upgradeRoot`).
- New "Upgrade Tree construction (two rules)" section explaining why per-chain Upgrade Trees are correct-by-construction (every leaf's `latestRoot` targets that chain's Route Tree only) and why the no-downgrade invariant collapses to one-leaf-per-proxy-per-chain (D25).
- New "Trust split (CCTP/OFT)" callout explaining the two-signature model (`counterfactualSignature` = fee only; `peripherySignature` = route/amount/recipient/action).
- Updated `setImplementation` paragraph documents both `NotAContract` and `WrongBeacon` checks.

### One small lingering nit
DESIGN.md still has three references to **"Open Question #N"** (lines 238, 249, 383) but the **Open Questions** section was removed in this round, so those refs are now dangling for a reader. Worth either re-inlining the relevant short answers or dropping the cross-references. Non-blocking — happy to land as-is and have it cleaned up in a follow-up doc pass.

Nice work. LGTM.

contracts/periphery/counterfactual/CounterfactualDeposit.sol
```diff
@@ -58,4 +124,19 @@ contract CounterfactualDeposit is ICounterfactualDeposit {
...
         }
     }
+
+    function _updateRoot(bytes32 newRoot, bytes32[] calldata proof) private {
+        CounterfactualStorage storage $ = _getStorage();
+        if (newRoot == $.activeRoot) revert RootUnchanged();
+        bytes32 leaf = keccak256(bytes.concat(keccak256(abi.encode(address(this), newRoot))));
+        if (!MerkleProof.verify(proof, BEACON.upgradeRoot(), leaf)) revert InvalidUpgradeProof();
+        $.activeRoot = newRoot;
+        emit RootUpdated(newRoot);
```
mrice32, line 134: Should this be done in the beacon itself? That way, if we decide to change the upgrade procedure in some way, we just upgrade the beacon?

tbwebb22: You're referring to just the proof verfication moving to the beacon right? But activeRoot is still stored in the counterfactual, not the beacon right?

I don't see a huge difference with the proof verification living here or on the beacon. In this current system we can still update this verification procedure by updating the implementation address that the beacon returns to the proxy

contracts/periphery/counterfactual/CounterfactualDeposit.sol
```diff
@@ -58,4 +124,19 @@ contract CounterfactualDeposit is ICounterfactualDeposit {
             }
         }
     }
+
+    function _updateRoot(bytes32 newRoot, bytes32[] calldata proof) private {
+        CounterfactualStorage storage $ = _getStorage();
+        if (newRoot == $.activeRoot) revert RootUnchanged();
+        bytes32 leaf = keccak256(bytes.concat(keccak256(abi.encode(address(this), newRoot))));
+        if (!MerkleProof.verify(proof, BEACON.upgradeRoot(), leaf)) revert InvalidUpgradeProof();
```
fusmanii, line 132: Given that both `CounterfactualBeacon` and `CounterfactualDeposit` are upgradable, does is make sense to just make active roots a mapping in `CounterfactualBeacon`? don't think we are gaining much by having it be a merkle tree

tbwebb22: Yeah a mapping just wouldn't really scale though. Imagine if we have thousands of counterfactuals on a chain, then to do an update we'd have to update each of those indexes in the mapping

fusmanii: right, yeah didn't realize that the upgrade root leaf includes the `CounterfactualDeposit` address

contracts/periphery/counterfactual/CounterfactualDeposit.sol
mrice32: Wiat, I'm still a bit confused. Do all counterfactualdeposit addresses point to the same implementation? Or is each one individually upgradable?

grasphoper (approved):

PR #1467, opened 2026-06-12
feat(SpokePoolPeriphery): support ERC-6492 counterfactual contract-wallet signers

mrice32:
> Stacked on top of #1420 (`feat/erc3009-bytes-signature`). Review/merge that first; this PR's diff is against that branch.

## Summary

PR #1420 added the bytes-signature ERC-3009 entry points (`depositWithAuthorizationBytes`, `swapAndBridgeWithAuthorizationBytes`) which pull tokens via the extended `receiveWithAuthorization(...,bytes signature)` overload (e.g. USDC v2.2). Those delegate signature verification **to the token**, which supports EIP-1271 contract wallets — but the token has no concept of ERC-6492. A **counterfactual** (not-yet-deployed) smart-contract wallet therefore can't authorize a transfer: there's no code at the signer's address to run `isValidSignature` on.

We can't change the token's verification. So, following the approach in [Coinbase's `ERC6492SignatureHandler`](https://github.com/base/commerce-payments/blob/main/src/collectors/ERC6492SignatureHandler.sol), this PR implements **only the deploy half** of ERC-6492 in the periphery — it is *not* the signature verifier:

- New abstract `ERC6492SignatureHandler` base contract. `_handleERC6492Signature` detects the ERC-6492 magic suffix (`0x6492…6492`) and, if present, runs the wrapped `(factory, factoryCalldata)` to materialize the wallet, then returns the unwrapped inner signature.
- `SpokePoolPeriphery` inherits it and runs `_handleERC6492Signature` on the signature in both `*WithAuthorizationBytes` entry points before forwarding it to the token. Once the wallet exists, the token's own EIP-1271 check against the signer succeeds.
- The wrapped `factory`/`factoryCalldata` are attacker-controlled, so the prepare call is routed through the canonical **Multicall3** singleton (supplied as an **immutable** constructor arg) rather than made directly — the periphery is never the `msg.sender` of that call and so can't be tricked into abusing its own allowances. The call uses `requireSuccess=false`, so an already-deployed wallet is tolerated; the token remains the source of truth on validity.
- New minimal `IMulticall3` interface.
- Constructor now takes the Multicall3 address; the deploy script + new `Constants.getMulticall3()` supply the canonical `0xcA11bde05977b3631167028862bE2a173976CA11`.

Non-ERC-6492 signatures (EOA / already-deployed EIP-1271) fall straight through `_handleERC6492Signature` unchanged, so existing behavior is preserved.

## Scope

Only the two `*WithAuthorizationBytes` flows need this — they're the only entry points where the **token** performs EIP-1271 verification. The v/r/s, permit, and permit2 flows are unaffected.

## Test plan

- [x] `yarn build-evm-foundry`
- [x] `yarn test-evm-foundry -- --match-contract "SpokePoolPeripheryTest|TransferProxy"` — 53 passed (49 existing + 4 new)
- [x] New tests:
  - `testTransferWithAuthBytesDepositERC6492CounterfactualWallet` — wallet deployed by the prepare step, deposit succeeds
  - `testTransferWithAuthBytesSwapAndBridgeERC6492CounterfactualWallet` — same for swap-and-bridge
  - `testTransferWithAuthBytesDepositERC6492AlreadyDeployedWallet` — wrapped sig + already-deployed wallet (idempotent prepare noop)
  - `testTransferWithAuthBytesDepositERC6492FailedPrepareTolerated` — failing prepare target is swallowed (`requireSuccess=false`), deposit still succeeds via the already-deployed signer
  - Existing EOA + EIP-1271 bytes tests continue to exercise the passthrough path
- [x] `yarn lint-fix`

### DIFF
--- a/contracts/periphery/ERC6492SignatureHandler.sol
+++ b/contracts/periphery/ERC6492SignatureHandler.sol
@@ -1,0 +1,86 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity ^0.8.0;
+
+import { IMulticall3 } from "../external/interfaces/IMulticall3.sol";
+
+/**
+ * @title ERC6492SignatureHandler
+ * @notice Base contract that performs the "prepare" (deploy) step of an ERC-6492 signature without
+ * being the signature verifier itself.
+ * @dev Some downstream verifiers — most importantly the ERC-3009 `receiveWithAuthorization(...,bytes)`
+ * overload on tokens such as USDC v2.2 — validate contract-wallet signatures via EIP-1271 but have no
+ * concept of ERC-6492. A signature from a *counterfactual* (not-yet-deployed) smart-contract wallet
+ * therefore fails, because there is no code at the signer's address to call `isValidSignature` on.
+ *
+ * ERC-6492 (https://eips.ethereum.org/EIPS/eip-6492) solves this by wrapping the real signature with
+ * the factory call needed to deploy the wallet. Since we cannot change the token's verification, we
+ * replicate only the *deploy* half of the flow here: if a signature carries the ERC-6492 magic
+ * suffix, we run its embedded factory call to materialize the wallet, then hand the unwrapped inner
+ * signature back to the caller to pass on to the real verifier (the token). Once the wallet exists,
+ * the token's own EIP-1271 check against the signer succeeds.
+ *
+ * @dev This contract does NOT validate the signature; it only ensures the signer is deployed. The
+ * embedded factory `target`/`callData` is fully attacker-controlled, so the call is routed through a
+ * neutral public Multicall3 singleton rather than made directly. This guarantees the prepare call can
+ * never execute with this contract as `msg.sender` and thus cannot abuse any of this contract's
+ * allowances or privileges. The call is made with `requireSuccess: false`, so an already-deployed
+ * wallet (whose factory call reverts on redeploy) is tolerated; correctness ultimately rests on the
+ * downstream verifier.
+ *
+ * Inspired by Coinbase's ERC6492SignatureHandler
+ * (https://github.com/base/commerce-payments/blob/main/src/collectors/ERC6492SignatureHandler.sol).
+ * @custom:security-contact bugs@across.to
+ */
+abstract contract ERC6492SignatureHandler {
+    // Magic suffix that marks a signature as ERC-6492 wrapped.
+    bytes32 internal constant _ERC6492_MAGIC_VALUE = 0x6492649264926492649264926492649264926492649264926492649264926492;
+
+    // Canonical Multicall3 singleton used to make the neutral ERC-6492 prepare/deploy call.
+    IMulticall3 public immutable multicall3;
+
+    /**
+     * @notice Constructs the handler.
+     * @param _multicall3 Address of the canonical Multicall3 singleton used to route prepare calls.
+     */
+    constructor(address _multicall3) {
+        multicall3 = IMulticall3(_multicall3);
+    }
+
+    /**
+     * @notice Detects an ERC-6492 wrapped signature and, if present, runs its embedded factory call
+     * to deploy the (counterfactual) signer before returning the unwrapped inner signature.
+     * @param signature The user-provided signature, possibly ERC-6492 wrapped.
+     * @return The inner signature to forward to the real verifier. If `signature` is not ERC-6492
+     * wrapped it is returned unchanged.
+     */
+    function _handleERC6492Signature(bytes memory signature) internal returns (bytes memory) {
+        // An ERC-6492 signature is at least the 32-byte magic suffix; anything shorter passes through.
+        uint256 signatureLength = signature.length;
+        if (signatureLength < 32) return signature;
+
+        // Pass through unless the trailing 32 bytes are the ERC-6492 magic value.
+        bytes32 suffix;
+        assembly {
+            suffix := mload(add(add(signature, 32), sub(signatureLength, 32)))
+        }
+        if (suffix != _ERC6492_MAGIC_VALUE) return signature;
+
+        // Strip the magic suffix and decode the wrapped (factory, factoryCalldata, innerSignature).
+        bytes memory erc6492Data = new bytes(signatureLength - 32);
+        for (uint256 i; i < signatureLength - 32; ++i) {
+            erc6492Data[i] = signature[i];
+        }
+        address prepareTarget;
+        bytes memory prepareData;
+        (prepareTarget, prepareData, signature) = abi.decode(erc6492Data, (address, bytes, bytes));
+
+        // Route the deploy call through the neutral Multicall3 so this contract is never the sender of
+        // attacker-controlled calldata. Failure is tolerated (e.g. the wallet is already deployed);
+        // the downstream verifier is the source of truth on signature validity.
+        IMulticall3.Call[] memory calls = new IMulticall3.Call[](1);
+        calls[0] = IMulticall3.Call(prepareTarget, prepareData);
+        multicall3.tryAggregate(false, calls);
+
+        return signature;
+    }
+}
--- a/contracts/periphery/ERC6492SignatureHandler.sol
+++ b/contracts/periphery/ERC6492SignatureHandler.sol
@@ -53,26 +53,20 @@ abstract contract ERC6492SignatureHandler {
      * @return The inner signature to forward to the real verifier. If `signature` is not ERC-6492
      * wrapped it is returned unchanged.
      */
-    function _handleERC6492Signature(bytes memory signature) internal returns (bytes memory) {
+    function _handleERC6492Signature(bytes calldata signature) internal returns (bytes memory) {
         // An ERC-6492 signature is at least the 32-byte magic suffix; anything shorter passes through.
         uint256 signatureLength = signature.length;
         if (signatureLength < 32) return signature;
 
         // Pass through unless the trailing 32 bytes are the ERC-6492 magic value.
-        bytes32 suffix;
-        assembly {
-            suffix := mload(add(add(signature, 32), sub(signatureLength, 32)))
-        }
-        if (suffix != _ERC6492_MAGIC_VALUE) return signature;
+        if (bytes32(signature[signatureLength - 32:]) != _ERC6492_MAGIC_VALUE) return signature;
 
-        // Strip the magic suffix and decode the wrapped (factory, factoryCalldata, innerSignature).
-        bytes memory erc6492Data = new bytes(signatureLength - 32);
-        for (uint256 i; i < signatureLength - 32; ++i) {
-            erc6492Data[i] = signature[i];
-        }
-        address prepareTarget;
-        bytes memory prepareData;
-        (prepareTarget, prepareData, signature) = abi.decode(erc6492Data, (address, bytes, bytes));
+        // Decode the wrapped (factory, factoryCalldata, innerSignature) directly from the calldata
+        // slice that strips the magic suffix.
+        (address prepareTarget, bytes memory prepareData, bytes memory innerSignature) = abi.decode(
+            signature[:signatureLength - 32],
+            (address, bytes, bytes)
+        );
 
         // Route the deploy call through the neutral Multicall3 so this contract is never the sender of
         // attacker-controlled calldata. Failure is tolerated (e.g. the wallet is already deployed);
@@ -81,6 +75,6 @@ abstract contract ERC6492SignatureHandler {
         calls[0] = IMulticall3.Call(prepareTarget, prepareData);
         multicall3.tryAggregate(false, calls);
 
-        return signature;
+        return innerSignature;
     }
 }
--- a/contracts/periphery/ERC6492SignatureHandler.sol
+++ b/contracts/periphery/ERC6492SignatureHandler.sol
@@ -77,7 +77,7 @@ abstract contract ERC6492SignatureHandler {
         // the downstream verifier is the source of truth on signature validity.
         IMulticall3.Call[] memory calls = new IMulticall3.Call[](1);
         calls[0] = IMulticall3.Call(prepareTarget, prepareData);
-        multicall3.tryAggregate(false, calls);
+        multicall3.tryAggregate({ requireSuccess: false, calls: calls });
 
         return innerSignature;
     }
--- a/contracts/periphery/ERC6492SignatureHandler.sol
+++ b/contracts/periphery/ERC6492SignatureHandler.sol
@@ -7,10 +7,11 @@ import { IMulticall3 } from "../external/interfaces/IMulticall3.sol";
  * @title ERC6492SignatureHandler
  * @notice Base contract that performs the "prepare" (deploy) step of an ERC-6492 signature without
  * being the signature verifier itself.
- * @dev Some downstream verifiers — most importantly the ERC-3009 `receiveWithAuthorization(...,bytes)`
- * overload on tokens such as USDC v2.2 — validate contract-wallet signatures via EIP-1271 but have no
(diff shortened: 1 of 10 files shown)


droplet-rl:
:robot: Babysitting this PR — picked up the assignment. Working from `feat/erc6492-signature-handling`; I'll post a summary when this pass is done.

contracts/periphery/ERC6492SignatureHandler.sol
```diff
@@ -0,0 +1,86 @@
...
+        if (suffix != _ERC6492_MAGIC_VALUE) return signature;
+
+        // Strip the magic suffix and decode the wrapped (factory, factoryCalldata, innerSignature).
+        bytes memory erc6492Data = new bytes(signatureLength - 32);
+        for (uint256 i; i < signatureLength - 32; ++i) {
+            erc6492Data[i] = signature[i];
+        }
+        address prepareTarget;
+        bytes memory prepareData;
+        (prepareTarget, prepareData, signature) = abi.decode(erc6492Data, (address, bytes, bytes));
```
droplet-rl: Nit: the byte-by-byte copy is O(n) words of overhead vs. a word-aligned memory copy. For a typical 65-byte ECDSA inner signature wrapped in a few hundred bytes of factory data, the loop is ~tens of thousands of gas heavier than necessary. Two cheaper options:

1. Use the calldata slice and `bytes.concat` / a direct memory copy (the entry points pass `bytes calldata`, so you could overload `_handleERC6492Signature(bytes calldata)` and just `signature[:signatureLength - 32]`).
2. Use a 32-byte-at-a-time assembly `mload`/`mstore` loop, or `mcopy` (the project compiles with a Solidity that supports it on recent EVMs).

Fine to leave as-is for a first cut since this path is opt-in and gas is paid by the caller, but worth picking up before final.

droplet-rl (commented):
Self-review on the babysit pass.

- Addressed codex's P2 thread by updating the Tron deployment script (and the Tron README) to accept and forward the new `multicall3` constructor argument. Without it any deployment via `yarn tron-deploy-spoke-pool-periphery <permit2>` would have reverted in the constructor's zero-address check.
- Otherwise the diff looks solid: the design routes the attacker-controlled prepare call through the canonical Multicall3 so the periphery is never `msg.sender`, the `requireSuccess=false` tolerates already-deployed wallets, the `nonReentrant` guard on the entry points prevents prepare-driven reentry, and the `abstract` handler doesn't pretend to be a verifier.
- One small nit on `_handleERC6492Signature` flagged inline — the manual byte-by-byte copy loop is correct but gas-wasteful and easy to replace with a memory copy. Non-blocking.

contracts/periphery/SpokePoolPeriphery.sol
```diff
@@ -188,10 +195,17 @@ contract SpokePoolPeriphery is SpokePoolPeripheryInterface, ReentrancyGuard, Mul
...
      */
-    constructor(IPermit2 _permit2) EIP712("ACROSS-PERIPHERY", "1.0.0") {
+    constructor(
+        IPermit2 _permit2,
+        address _multicall3
+    ) EIP712("ACROSS-PERIPHERY", "1.0.0") ERC6492SignatureHandler(_multicall3) {
         require(address(_permit2) != address(0), "Permit2 cannot be zero address");
         require(_isContract(address(_permit2)), "Permit2 must be a contract");
+        require(_multicall3 != address(0), "Multicall3 cannot be zero address");
+        require(_isContract(_multicall3), "Multicall3 must be a contract");
```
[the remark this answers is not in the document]
droplet-rl: Good catch. Updated `script/tron/periphery/tron-deploy-spoke-pool-periphery.ts` to accept and ABI-encode the `multicall3` Base58Check address as the second constructor argument, validate it via `validateTronAddresses`, and log it. Also bumped the usage string in `script/tron/README.md` to `yarn tron-deploy-spoke-pool-periphery <permit2> <multicall3> [--testnet]`.

droplet-rl:
Babysit pass complete.

**Reviewer threads (1 unresolved → 0)**
- codex P2 on `SpokePoolPeriphery.sol:208` (Tron deploy script left out the new `_multicall3` arg). Updated `script/tron/periphery/tron-deploy-spoke-pool-periphery.ts` to take and ABI-encode a second `<multicall3>` Tron Base58Check address (validated via `validateTronAddresses`) and bumped the usage string in `script/tron/README.md` to match. Replied + queued the thread for resolution.

**Self-review**
- Submitted one inline nit on `contracts/periphery/ERC6492SignatureHandler.sol:75`: the byte-by-byte copy loop is correct but ~tens of thousands of gas heavier than necessary — a calldata-overloaded slice or a 32-byte assembly/mcopy loop would be cheaper. Non-blocking.
- No correctness issues otherwise: prepare call is routed through Multicall3 so the periphery is never `msg.sender` of attacker-controlled calldata, `requireSuccess=false` tolerates already-deployed wallets, `nonReentrant` blocks prepare-driven reentry, and the `abstract` handler stays out of the verifier role.

**Commits this pass**
- `07de37dd` chore(tron): forward multicall3 arg to SpokePoolPeriphery deploy

Push will be handled by the checkpoint; the codex resolve is queued and will fire on push success.

droplet-rl:
@codex review

droplet-rl:
Acknowledged codex's clean re-review of `07de37dd`. No unresolved threads on the PR — both prior ones (codex's P2 on the Tron deploy script, my own nit on the byte-copy loop in `ERC6492SignatureHandler`) are resolved. Working tree clean, no further commits this pass.

mrice32:
@droplet-rl make the optimization you called out.

droplet-rl:
Done — applied the optimization in `64bb534d`. `_handleERC6492Signature` now takes `bytes calldata` and decodes `(factory, factoryCalldata, innerSignature)` directly from a calldata slice; the magic-suffix check is `bytes32(signature[signatureLength - 32:])`. That drops both the implicit calldata→memory copy at the call boundary and the byte-by-byte loop into a scratch memory buffer. Call sites in `SpokePoolPeriphery` already passed `bytes calldata`, so no logical changes there. Verified: `yarn build-evm-foundry` clean, 8/8 ERC-6492 + AuthBytes tests pass.

droplet-rl:
:warning: Couldn't finish 1 background task(s) before the 30m budget ran out — their results weren't gathered:
  • forge build completion or errors

droplet-rl:
@codex review

droplet-rl:
Codex came back clean on the optimization commit (`64bb534d`). No unresolved threads on the PR; the byte-copy nit is now marked outdated. No further commits this pass.

mrice32:
@droplet-rl please make the optimization you mentioned in your review.

droplet-rl:
Already in — pushed as `64bb534d` ("perf(ERC6492SignatureHandler): decode wrapped signature from calldata") on the previous trigger. It's the current PR head.

The handler now takes `bytes calldata` and decodes `(factory, factoryCalldata, innerSignature)` directly from a calldata slice (`signature[:signatureLength - 32]`); the magic-suffix check is `bytes32(signature[signatureLength - 32:])`. That drops both the implicit calldata→memory copy of the full signature at the call boundary and the byte-by-byte loop into a scratch `bytes memory` buffer before `abi.decode`. Codex re-ran on `64bb534d40` and came back clean. The original review thread is now marked `outdated`.

Nothing more to do this pass; let me know if you had a different optimization in mind.

contracts/periphery/ERC6492SignatureHandler.sol
```diff
@@ -0,0 +1,80 @@
...
+     * @notice Detects an ERC-6492 wrapped signature and, if present, runs its embedded factory call
+     * to deploy the (counterfactual) signer before returning the unwrapped inner signature.
+     * @param signature The user-provided signature, possibly ERC-6492 wrapped.
+     * @return The inner signature to forward to the real verifier. If `signature` is not ERC-6492
+     * wrapped it is returned unchanged.
+     */
+    function _handleERC6492Signature(bytes calldata signature) internal returns (bytes memory) {
+        // An ERC-6492 signature is at least the 32-byte magic suffix; anything shorter passes through.
+        uint256 signatureLength = signature.length;
+        if (signatureLength < 32) return signature;
```
fusmanii:
```suggestion
        if (signatureLength < _ERC6492_MAGIC_VALUE.length) return signature;
```
nit

contracts/periphery/SpokePoolPeriphery.sol
```diff
@@ -188,10 +195,17 @@ contract SpokePoolPeriphery is SpokePoolPeripheryInterface, ReentrancyGuard, Mul
...
      * @notice Construct a new Periphery contract.
      * @param _permit2 Address of the canonical permit2 contract.
+     * @param _multicall3 Address of the canonical Multicall3 singleton used to route ERC-6492
+     * prepare/deploy calls (see ERC6492SignatureHandler).
      */
-    constructor(IPermit2 _permit2) EIP712("ACROSS-PERIPHERY", "1.0.0") {
+    constructor(
+        IPermit2 _permit2,
+        address _multicall3
+    ) EIP712("ACROSS-PERIPHERY", "1.0.0") ERC6492SignatureHandler(_multicall3) {
```
fusmanii, line 204: should we change the version here?

fusmanii (approved):
Looks good to me, just two (optional) comments

contracts/periphery/ERC6492SignatureHandler.sol
droplet-rl: Applied in `13f38536`. Used `_ERC6492_MAGIC_VALUE.length` (a compile-time constant for the `bytes32`) for the length check and also for the two calldata-slice bounds, so all three sites stay self-documenting and in sync with the constant. Built + tests green (8/8 ERC-6492 + AuthBytes).

contracts/periphery/SpokePoolPeriphery.sol
droplet-rl: I don't think a bump is required here, but it's your call. The `"1.0.0"` version feeds the periphery's *own* EIP-712 domain (`_domainSeparatorV4`), which is used to verify the `DepositData`/`SwapAndDepositData` typed-data signatures the periphery itself checks. This PR doesn't change any of those typed-data structs — the ERC-6492 work only deploys the counterfactual signer before handing the unwrapped signature to the token, and the token's `receiveWithAuthorization` verifies against the *token's* EIP-712 domain (e.g. USDC's), not ours. So nothing the periphery signs over changed.

There's also no replay concern motivating a bump: this is a fresh deployment (new bytecode + new `multicall3` constructor arg → new address), so the domain separator already differs from the existing periphery by `verifyingContract`. A version bump would only be warranted if we'd changed the typed-data schema while keeping the same address.

Happy to bump to `"1.1.0"` purely as a human-facing signal of the redeploy if you'd prefer — just let me know and I'll add it. Leaving this open for your call.

droplet-rl:
Babysit pass done — handled fusmanii's two (optional) review comments.

**1. `ERC6492SignatureHandler.sol` — use `_ERC6492_MAGIC_VALUE.length` instead of literal `32`** → applied in `13f38536`. Since the constant is a `bytes32`, `.length` is a compile-time `32`; I used it for the length guard and both calldata-slice bounds so all three stay in sync with the constant. Replied + queued the thread for resolution.

**2. `SpokePoolPeriphery.sol:204` — "should we change the [EIP712] version?"** → replied, left open for the author's call (no change). Reasoning: `"1.0.0"` feeds the periphery's *own* EIP-712 domain (`_domainSeparatorV4`), used to verify the `DepositData`/`SwapAndDepositData` typed-data sigs the periphery checks itself. This PR changes none of those structs — ERC-6492 only deploys the counterfactual signer, and the token's `receiveWithAuthorization` verifies against the *token's* domain (e.g. USDC's), not ours. No replay motive either: it's a fresh deployment (new bytecode + `multicall3` arg → new address), so the domain separator already differs by `verifyingContract`. Offered to bump to `"1.1.0"` purely as a redeploy signal if preferred.

**Self-review:** no issues. Walked the handler edge cases (exact-32-byte input → `[0:]` slice valid; malformed wrapper just reverts the tx with no security impact; magic-suffix collision on a real ERC-3009 sig is negligible by EIP-6492 design). This pass's only delta is the cosmetic refactor.

**Verification:** `yarn build-evm-foundry` clean; `forge test --match-path test/.../SpokePoolPeriphery.t.sol --match-test "ERC6492|AuthBytes"` → 8/8 pass.

**Commits this pass:**
- `13f38536` refactor(ERC6492SignatureHandler): use magic-value length over literal 32

droplet-rl:
@codex review

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=f845e22e90488154a4a50d994dcdc2f2&trace_start_ts=1782841351&trace_end_ts=1782928051)

script/utils/Constants.sol
```diff
@@ -244,6 +244,13 @@ contract Constants is Script {
         return vm.parseJsonAddress(file, string.concat(".L2_ADDRESS_MAP.", vm.toString(chainId), ".permit2"));
     }
 
+    // Get the canonical Multicall3 singleton address. Multicall3 is deployed deterministically at the
+    // same address on virtually all EVM chains (see https://github.com/mds1/multicall). Used by the
+    // SpokePoolPeriphery to route ERC-6492 prepare/deploy calls through a neutral contract.
+    function getMulticall3() public pure returns (address) {
+        return 0xcA11bde05977b3631167028862bE2a173976CA11;
```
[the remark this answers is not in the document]
droplet-rl: Good catch. Made `getMulticall3` chain-aware: it now takes a `chainId` and returns the zkSync-stack Multicall3 deployment (`0xF9cda624FBC7e059355ce98a31693d299FACd963`) for `ZK_SYNC` (324) and `ZK_SYNC_SEPOLIA` (300), falling back to the canonical `0xcA11…CA11` everywhere else. `DeploySpokePoolPeriphery.s.sol` now passes `block.chainid` through.

I verified those two addresses against the canonical mds1/multicall `deployments.json` (both zkSync networks share `0xF9cda624…`). `LENS`/`LENS_SEPOLIA` aren't in that registry, so I left them on the canonical fallback rather than guess — and since the periphery constructor rejects a non-contract `_multicall3`, an unmapped chain whose Multicall3 turns out to be non-canonical fails loudly at deploy time instead of deploying with a dead reference. Easy to add a mapping there once an address is confirmed.

droplet-rl:
:warning: Couldn't finish 4 background task(s) before the 30m budget ran out — their results weren't gathered:
  • Build contracts + scripts
  • forge build completion or errors (bkx9cyu9r)
  • Reinstall deps and rebuild capturing exit code
  • reinstall+build result (bzg02jzt3)

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=fe95ff6c8f2318c7f723d10dd919ff42&trace_start_ts=1782842156&trace_end_ts=1782928856)

grasphoper (commented):
Looks good, 2 Qs:
- move `isContract` check to the `ERC6492SignatureHandler` constructor  instead of SpokePoolPeriphery?
- should we add the same logic to `Permit2` signatures?

grasphoper:
Added Permit2 support here https://github.com/across-protocol/contracts/pull/1488 if we want it

droplet-rl:
  • Build with exit-code capture
  • build exit for bh7mpcjkw

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=b9ff7ebd586dfd3f65394d09227d2e3e&trace_start_ts=1783227292&trace_end_ts=1783313992)

PR #1475, opened 2026-06-24
feat: SpokePool as deposit adapter

grasphoper:


contracts/spoke-pools/SpokePool.sol
```diff
@@ -1728,6 +1736,146 @@ abstract contract SpokePool is
...
+        uint256 flags = rules >> 160;
+        address authority = address(uint160(rules));
+
+        // If an authority is configured, the proposed modification set must be signed by it.
+        if (authority != address(0)) {
+            bytes32 digest = keccak256(
+                abi.encodePacked(
+                    VERSIONED_AUCTION_NAMEHASH,
+                    address(gateway),
+                    stepId,
```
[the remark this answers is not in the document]
grasphoper: `depositNonce` is going to be a random number for any v5 deposit we'll be trying to do in an adapter. So two different `Path`s will never have the same `depositNonce` anyway. This is a part of a broader push to rely on (stepId + random-number) for replay prevention rather than pathId, since for multiple entries of the same thing you still want (pathId + some number) and using stepId everywhere just makes sense.

grasphoper:
@codex the issues fixed. Please re-review

contracts/spoke-pools/SpokePool.sol
```diff
@@ -1728,6 +1752,100 @@ abstract contract SpokePool is
...
+        ParamsFromV5Input memory inputParams,
+        ParamsFromV5Jit memory jitParams,
+        bytes32 pathId
+    ) internal view {
+        uint256 rules = uint256(inputParams.paramModificationRules);
+        uint256 flags = rules >> 160;
+        address authority = address(uint160(rules));
+
+        // If an authority is configured, the proposed modification set must be signed by it.
+        if (authority != address(0)) {
```
fusmanii, line 1973: so if the `authority` is not set the submitter can still set overrides below?

contracts/spoke-pools/SpokePool.sol
```diff
@@ -1728,6 +1752,100 @@ abstract contract SpokePool is
...
+
+        bytes32 curPathId = gateway.currentPathId();
+        address curSubmitter = gateway.currentSubmitter();
+        if (curSubmitter == address(0)) revert InactiveV5Flow();
+        uint256 depositId = getUnsafeDepositId(
+            // gateway's submitter serves as a differentiating factor instead of msg.sender, which is going to be Executor for most deposits here
+            curSubmitter,
+            inputParams.depositor,
+            // depositNonce here serves as a differentiator of different deposits that might come in a single Path execution
+            uint256(keccak256(abi.encodePacked(curPathId, inputParams.depositNonce)))
```
fusmanii, line 1930: would be nice to do what [SpokePoolPeriphery](https://github.com/across-protocol/contracts/blob/master/contracts/periphery/SpokePoolPeriphery.sol#L686) does and allow for normal deposit here if `depositNonce` is zero

contracts/spoke-pools/SpokePool.sol
grasphoper: Yep!

PR #1476, opened 2026-06-24
feat: SpokePool V5 fill

fusmanii:


fusmanii:
@codex review

contracts/spoke-pools/SpokePool.sol
```diff
@@ -1003,6 +1023,52 @@ abstract contract SpokePool is
         fillRelay(convertedRelayData, repaymentChainId, msg.sender.toBytes32());
     }
 
+    function executeAcrossV5(bytes calldata input, bytes calldata jitData) external payable unpausedFills {
```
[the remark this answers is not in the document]
fusmanii: good point, https://github.com/across-protocol/contracts/pull/1476/commits/2532593212b1790dd1f935420a31438f5de16feb

contracts/spoke-pools/SpokePool.sol
```diff
@@ -1003,6 +1023,52 @@ abstract contract SpokePool is
...
+            fillDeadline: inputParamsV5.fillDeadline,
+            exclusivityDeadline: jitInputParamsV5.exclusivityDeadline,
+            message: new bytes(0)
+        });
+
+        // Exclusivity deadline is inclusive and is the latest timestamp that the exclusive relayer has sole right
+        // to fill the relay.
+        if (
+            _fillIsExclusive(relayData.exclusivityDeadline, uint32(getCurrentTime())) &&
+            relayData.exclusiveRelayer.toAddress() != msg.sender
```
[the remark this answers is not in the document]
fusmanii: nice catch, https://github.com/across-protocol/contracts/pull/1476/commits/2532593212b1790dd1f935420a31438f5de16feb

contracts/spoke-pools/SpokePool.sol
```diff
@@ -229,18 +246,21 @@ abstract contract SpokePool is
...
      */
     /// @custom:oz-upgrades-unsafe-allow constructor
     constructor(
         address _wrappedNativeTokenAddress,
         uint32 _depositQuoteTimeBuffer,
         uint32 _fillDeadlineBuffer,
         uint32 _oftDstEid,
-        uint256 _oftFeeCap
+        uint256 _oftFeeCap,
+        address _gateway
```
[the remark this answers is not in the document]
fusmanii: done https://github.com/across-protocol/contracts/pull/1476/commits/688d2284af3e9b96bc40463a185546c2acbddfe4

fusmanii:
@codex review

contracts/spoke-pools/SpokePool.sol
```diff
@@ -1003,6 +1019,54 @@ abstract contract SpokePool is
         fillRelay(convertedRelayData, repaymentChainId, msg.sender.toBytes32());
     }
 
+    function executeAcrossV5(bytes calldata input, bytes calldata jitData) external payable unpausedFills {
```
droplet-rl: **Add `nonReentrant`** (blocking). Every other fill entrypoint — `fillRelay`, `fillRelayWithUpdatedDeposit`, `requestSlowFill`, `executeSlowRelayLeaf` — carries `nonReentrant`; this one omits it. Since the V5 branch in `_transferTokensToRecipient` does `safeTransferFrom(submitter, …)`, a hooked output token can re-enter `fillRelay` with a message starting with `V5_DEPOSIT_HEADER` while no lock is held and `currentSubmitter()` is still non-zero, pulling the submitter's approved tokens to an attacker. Suggest: `external payable nonReentrant unpausedFills`.

Separately, this is marked `payable` but `msg.value` is never consumed or refunded — stranded ETH. Drop `payable` unless native-token fills are planned.

contracts/spoke-pools/SpokePool.sol
```diff
@@ -1003,6 +1019,54 @@ abstract contract SpokePool is
...
+        // Exclusivity deadline is inclusive and is the latest timestamp that the exclusive relayer has sole right
+        // to fill the relay.
+        if (
+            _fillIsExclusive(relayData.exclusivityDeadline, uint32(getCurrentTime())) &&
+            relayData.exclusiveRelayer.toAddress() != gateway.currentSubmitter()
+        ) {
+            revert NotExclusiveRelayer();
+        }
+
+        relayData.message = abi.encodePacked(V5_DEPOSIT_HEADER, gateway.currentStepId());
```
droplet-rl: Minor: the message is overwritten with `V5_DEPOSIT_HEADER || currentStepId` (64 bytes), so any depositor-supplied message is discarded for V5 fills and the recipient `handleV3AcrossMessage` callback is never invoked (the V5 branch returns early). Confirm V5 fills are not meant to carry recipient-executable messages in this iteration.

contracts/spoke-pools/SpokePool.sol
```diff
@@ -1654,6 +1719,22 @@ abstract contract SpokePool is
         uint256 amountToSend = relayExecution.updatedOutputAmount;
         address recipientToSend = relayExecution.updatedRecipient.toAddress();
 
+        if (_isV5Fill(relayData.message)) {
```
droplet-rl: This branch keys the "pull from submitter" decision on whether `relayData.message` starts with `V5_DEPOSIT_HEADER`. In the `fillRelay` / `executeSlowRelayLeaf` paths that message is externally supplied, so this is the entry point for the reentrancy concern above and for self-griefing (a normal deposit whose message happens to start with the header becomes unfillable). Prefer threading an explicit `bool isV5` from `executeAcrossV5` through `_fillRelayV3` → `_transferTokensToRecipient` so this branch is unreachable from message-controlled paths.

Nit: `outputToken`, `recipientToSend`, and `amountToSend` are already computed three lines above — reuse them instead of re-deriving `relayData.outputToken.toAddress()` / `relayExecution.updatedRecipient.toAddress()` / `relayExecution.updatedOutputAmount`.

contracts/spoke-pools/SpokePool.sol
```diff
@@ -1654,6 +1719,22 @@ abstract contract SpokePool is
...
+            if (isSlowFill) {
+                revert V5SlowFillNotAllowed();
+            }
+
+            address submitter = gateway.currentSubmitter();
+            if (submitter == address(0)) revert V5RequiresGateway();
+
+            IERC20Upgradeable(relayData.outputToken.toAddress()).safeTransferFrom(
```
droplet-rl: This always uses `safeTransferFrom` for the output token, so when `outputToken == wrappedNativeToken` the recipient receives WETH rather than unwrapped native (unlike the normal path at L1739). Intentional? If so, a one-line comment would help; if not, it diverges from V3 fill semantics for WETH recipients.

droplet-rl (changes requested):
## Review: feat: SpokePool V5 fill

Thanks @fusmanii. I cloned, checked out `faisal/spoke-pool-v5-fill`, read the diff, and confirmed it **compiles cleanly** (`yarn build-evm-foundry` → `Compiler run successful`). The chain-specific constructor wiring (adding `_gateway` to every `*_SpokePool` + deploy scripts + `getGateway()` env helper + README) is mechanical and consistent, and immutables don't disturb the UUPS storage layout — that part is good.

The core change adds a Gateway-only `executeAcrossV5` entrypoint that synthesizes a `V3RelayData`, stamps the message with `V5_DEPOSIT_HEADER || stepId`, and routes through `_fillRelayV3`. `_transferTokensToRecipient` then detects that header (`_isV5Fill`) and pulls the output token from `gateway.currentSubmitter()` instead of `msg.sender`. The mechanism is clever, but I have a few concerns I'd like resolved before approving — one security-relevant, plus a missing-tests gap on a fund-moving path.

### Blocking

**1. `executeAcrossV5` is missing `nonReentrant` (reentrancy via the shared, message-sniffing fill path).**
Every other fill entrypoint — `fillRelay`, `fillRelayWithUpdatedDeposit`, `requestSlowFill`, `executeSlowRelayLeaf` — is `nonReentrant`; this one isn't. During a V5 fill, `_transferTokensToRecipient` does `outputToken.safeTransferFrom(submitter, recipient, amount)`. If the output token has a transfer hook (ERC-777-style or a custom callback), that hook can re-enter the SpokePool *without any lock held* and call `fillRelay` with a crafted `relayData.message` that begins with `V5_DEPOSIT_HEADER`. Because `gateway.currentSubmitter()` is still non-zero mid-execution, the re-entrant fill is treated as a V5 fill and pulls the **submitter's** tokens to an attacker-chosen recipient/amount, bounded only by the submitter's outstanding approval. Note the Gateway's own `Reentrancy` guard doesn't help here — the re-entrant call hits the SpokePool directly, bypassing the Gateway. Please add `nonReentrant`.

**2. No test coverage.** There are no references to `executeAcrossV5`, `_isV5Fill`, `InputParamsV5`, or `V5_DEPOSIT_HEADER` anywhere under `test/`. This is a new external entrypoint that moves third-party (submitter) funds. At minimum I'd want: gateway-only access control, the exclusivity check against `currentSubmitter()`, the `submitter == address(0)` → `V5RequiresGateway` revert, slow-fill rejection (`V5SlowFillNotAllowed`), double-fill (`RelayFilled`), and a happy-path fill that pulls from the submitter.

### Non-blocking (worth addressing)

**3. Detecting V5 by sniffing `relayData.message` couples a fund-routing decision to externally-influenceable data.** Threading an explicit `bool isV5` through `_fillRelayV3`/`_transferTokensToRecipient` (so the submitter-pull branch is reachable *only* from `executeAcrossV5`) would remove the header-collision surface entirely and is the cleaner root-cause fix for #1.

**4. `payable` but `msg.value` is never used or refunded** — any ETH sent is stranded. Drop `payable` unless native fills are coming.

**5.** V5 path always `safeTransferFrom`s the output token, so a `wrappedNativeToken` output is delivered as WETH (not unwrapped to native like the normal path), and the recipient `handleV3AcrossMessage` callback is skipped. Both are likely intentional for this iteration — please confirm.

Happy to re-review once the reentrancy guard + tests land.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=28177cf84798acd7e857050f6628cd93&trace_start_ts=1782417507&trace_end_ts=1782504207)

contracts/spoke-pools/SpokePool.sol
```diff
@@ -1003,6 +1019,54 @@ abstract contract SpokePool is
...
+        });
+
+        // Exclusivity deadline is inclusive and is the latest timestamp that the exclusive relayer has sole right
+        // to fill the relay.
+        if (
+            _fillIsExclusive(relayData.exclusivityDeadline, uint32(getCurrentTime())) &&
+            relayData.exclusiveRelayer.toAddress() != gateway.currentSubmitter()
+        ) {
+            revert NotExclusiveRelayer();
+        }
```
grasphoper: I think we should reuse the logic from `fillRelay` here.

Something like:
- add `function fillRelayFrom`
- `fillRelay` calls `fillRelayFrom(..args, msg.sender)`
- `executeAcrossV5` calls `fillRelayFrom(..args, gateway.submitter())`

contracts/spoke-pools/SpokePool.sol
```diff
@@ -1003,6 +1019,54 @@ abstract contract SpokePool is
...
+            exclusiveRelayer: jitInputParamsV5.exclusiveRelayer,
+            inputToken: inputParamsV5.inputToken,
+            outputToken: inputParamsV5.outputToken,
+            inputAmount: inputParamsV5.inputAmount,
+            outputAmount: jitInputParamsV5.outputAmount,
+            originChainId: inputParamsV5.originChainId,
+            depositId: inputParamsV5.depositId,
+            fillDeadline: inputParamsV5.fillDeadline,
+            exclusivityDeadline: jitInputParamsV5.exclusivityDeadline,
+            message: new bytes(0)
```
grasphoper: I think we should just make it `abi.encodePacked(V5_DEPOSIT_HEADER, gateway.currentStepId());` here. Why make it bytes(0)?

contracts/spoke-pools/SpokePool.sol
```diff
@@ -1654,6 +1719,22 @@ abstract contract SpokePool is
...
+            address submitter = gateway.currentSubmitter();
+            if (submitter == address(0)) revert V5RequiresGateway();
+
+            IERC20Upgradeable(relayData.outputToken.toAddress()).safeTransferFrom(
+                submitter,
+                relayExecution.updatedRecipient.toAddress(),
+                relayExecution.updatedOutputAmount
+            );
+            return;
+        }
```
grasphoper: I think this logic should not live in this low-level function

I think `_transferTokensToRecipient` should also take a `from` for *who to pull funds from*

As for the v5 checking, it should happen up the stack, in all the public fill functions.

I think most public fill functions should just have a `notV5Fill() modifier` at the top. E.g. `requestSlowFill` should have that

fusmanii: good idea, I like `notV5Fill` modifier, added

contracts/spoke-pools/SpokePool.sol
fusmanii: done

contracts/spoke-pools/SpokePool.sol
fusmanii: I went with creating `_fillRelayV5` instead

contracts/spoke-pools/SpokePool.sol
```diff
@@ -289,6 +305,14 @@ abstract contract SpokePool is
         _;
     }
 
+    // Reverts if `message` belongs to a V5 deposit. V5 deposits are Gateway-fill-only and must never be settled
+    // out of this contract's reserves, so this guards functions (such as the slow-fill path) that must not
+    // process them.
+    modifier nonV5Fill(bytes memory message) {
+        if (_isV5Fill(message)) revert V5SlowFillNotAllowed();
```
grasphoper: The error looks off here

contracts/spoke-pools/SpokePool.sol
grasphoper: I just meant that we're duplicating exclusivity logic here, and IMO we should have it in one place somewhere

contracts/spoke-pools/SpokePool.sol
```diff
@@ -1003,6 +1027,56 @@ abstract contract SpokePool is
         fillRelay(convertedRelayData, repaymentChainId, msg.sender.toBytes32());
     }
 
+    function executeAcrossV5(bytes calldata input, bytes calldata jitData) external payable unpausedFills {
+        if (msg.sender != address(gateway)) revert V5RequiresGateway();
+
+        // all V3RelayData fields except for exclusiveRelayer, inputAmount, outputAmount, exclusivityDeadline from input
```
droplet-rl: This comment (and the one two lines down) is now stale after the struct reshuffle. `InputParamsV5` only carries `{recipient, outputToken, minOutputAmount}` — it does **not** contain depositor/originChainId/depositId/fillDeadline/inputToken/inputAmount, and `jitData` now carries almost everything. The current text ("all V3RelayData fields except for exclusiveRelayer, inputAmount, outputAmount, exclusivityDeadline from input") describes the previous layout and will mislead a Gateway integrator building this calldata. Please update both comments to match `InputParamsV5` / `JitInputParamsV5`.

contracts/spoke-pools/SpokePool.sol
```diff
@@ -1003,6 +1027,56 @@ abstract contract SpokePool is
         fillRelay(convertedRelayData, repaymentChainId, msg.sender.toBytes32());
     }
 
+    function executeAcrossV5(bytes calldata input, bytes calldata jitData) external payable unpausedFills {
```
droplet-rl: Two carry-over nits: (1) still `payable`, but `msg.value` is never used or refunded — strand-able ETH; drop `payable` unless native fills are planned. (2) Still no `nonReentrant`. The concrete drain from last round is closed now that the normal path no longer pulls from the submitter, so this is just defense-in-depth + consistency with every other fill entrypoint — cheap to add.

contracts/spoke-pools/SpokePool.sol
```diff
@@ -1589,17 +1657,52 @@ abstract contract SpokePool is
...
+
+    /**
+     * @notice Fills a V5 relay. V5 deposits are Gateway-fill-only and are always fast fills: the output tokens
+     * are pulled from the Gateway's current submitter (never from this contract's reserves or msg.sender) and
+     * sent to the recipient via the shared transfer logic, except the recipient message handler is never invoked.
+     * Slow fills are not supported for V5 deposits.
+     * @param relayExecution The relay execution parameters.
+     * @param relayer Address credited as the relayer (the repayment address) in the FilledRelay event.
+     */
+    function _fillRelayV5(V3RelayExecutionParams memory relayExecution, bytes32 relayer) internal {
```
droplet-rl: Clean separation here — settling from `submitter` with `callMessageHandler=false` reads much better than the old message-sniff branch. One edge note: this hardcodes `FillType.FastFill`. Because `requestSlowFill` lacks the `nonV5Fill` guard that `executeSlowRelayLeaf` now has, a matching V5 relay hash could be parked in `RequestedSlowFill` first, and this fill would then emit `FastFill` rather than `ReplacedSlowFill`. No fund impact, but adding `nonV5Fill(relayData.message)` to `requestSlowFill` would keep the invariant tight.

contracts/interfaces/V5SpokePoolInterface.sol
```diff
@@ -0,0 +1,31 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+interface V5SpokePoolInterface {
+    /// @notice Thrown when a V5 deposit or fill is attempted outside of a V5 entrypoint / live Gateway execution.
+    error V5RequiresGateway();
+    /// @notice Thrown when a slow fill is attempted against a V5 deposit (V5 deposits are Gateway-fill-only).
+    error V5SlowFillNotAllowed();
+    /// @notice
```
droplet-rl: Empty `/// @notice` — please document when `V5OutputAmountTooLow` is thrown (submitter's `outputAmount < minOutputAmount`), consistent with the other two errors.

droplet-rl (changes requested):
## Re-review (commits 9ba3686 → a2f8db8)

Thanks for the refactor, @fusmanii — this addresses the substance of my earlier review well. Pulled the latest, re-read the diff, and confirmed it still **compiles** (`yarn build-evm-foundry` → `Compiler run successful`). All refactored call sites line up (`_fillRelayV3` / `_fillRelayV5` / `_recordFill` / the new 5-arg `_transferTokensToRecipient`).

**Resolved from last round:**
- ✅ **The reentrancy/drain concern is gone.** V5 settlement now flows through a dedicated `_fillRelayV5` that explicitly passes `from = submitter`, and the normal `fillRelay` path no longer routes funds based on message bytes. `_isV5Fill` is now used only by the `nonV5Fill` guard, not to decide who funds a fill — exactly the explicit-flag separation I was hoping for.
- ✅ WETH output is now unwrapped (V5 goes through the shared transfer path).
- ✅ Message-handler skip is now explicit (`callMessageHandler=false`) and documented.
- ✅ Nice addition: the `minOutputAmount` floor (`V5OutputAmountTooLow`) protects the recipient against a submitter under-filling.

CEI looks right too — `_recordFill` marks the hash `Filled` before any token transfer.

**Still blocking:**

1. **No test coverage.** Same as last round — there are still zero references to `executeAcrossV5` / `_fillRelayV5` / `V5_DEPOSIT_HEADER` under `test/`. This is a new external entrypoint that moves a third party's (the submitter's) funds; I don't think it should merge without tests for at least: gateway-only access, the `minOutputAmount` floor, exclusivity vs `currentSubmitter()`, the `submitter == address(0)` revert, double-fill (`RelayFilled`), and `nonV5Fill` rejecting a V5 slow fill.

2. **Stale calldata comments in `executeAcrossV5`.** The two comments describing what comes from `input` vs `jitData` now describe the *old* struct split and are misleading (see inline). Worth fixing since they're the first thing a Gateway integrator reads.

**Non-blocking (worth a look):**
- `requestSlowFill` doesn't carry the new `nonV5Fill` guard that `executeSlowRelayLeaf` got, so a V5 relay hash can still be parked in `RequestedSlowFill`; a later V5 fill then records `FillType.FastFill` instead of `ReplacedSlowFill`. No fund impact, but it's inconsistent with the "V5 never enters the slow-fill flow" invariant. Mirroring the guard on `requestSlowFill` would tighten it.
- `executeAcrossV5` is still `payable` with `msg.value` unused/unrefunded, and still lacks `nonReentrant` (now only a defense-in-depth/consistency nit since the concrete vector is closed).
- Empty `/// @notice` on `V5OutputAmountTooLow`.

Close — mainly tests + the stale comments. Happy to re-review.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=d0e7704b1bc7c050ca29d80243e08524&trace_start_ts=1782426091&trace_end_ts=1782512791)

contracts/spoke-pools/SpokePool.sol
```diff
@@ -1589,17 +1657,52 @@ abstract contract SpokePool is
...
+        _recordFill(relayExecution, relayer, FillType.FastFill);
+
+        // V5 fills are settled by pulling the output tokens from the Gateway's current submitter, and never invoke
+        // the recipient's message handler (the V5 message carries the V5 header, not a recipient payload).
+        address submitter = gateway.currentSubmitter();
+        if (submitter == address(0)) revert V5RequiresGateway();
```
grasphoper: The call came from Gateway already and is using the IExecutor interface. I don't think we should be doing this zero check

contracts/spoke-pools/SpokePool.sol
```diff
@@ -1003,6 +1027,56 @@ abstract contract SpokePool is
...
     }
 
+    function executeAcrossV5(bytes calldata input, bytes calldata jitData) external payable unpausedFills {
+        if (msg.sender != address(gateway)) revert V5RequiresGateway();
+
+        // all V3RelayData fields except for exclusiveRelayer, inputAmount, outputAmount, exclusivityDeadline from input
+        InputParamsV5 memory inputParamsV5 = abi.decode(input, (InputParamsV5));
+
+        // get the rest of params plus repaymentChainId and repaymentAddress from jitData
+        JitInputParamsV5 memory jitInputParamsV5 = abi.decode(jitData, (JitInputParamsV5));
```
grasphoper: Both of these should include a second param `bytes` to pass through to the next executor (which could be deposit.recipient)

grasphoper (commented):
I'm a bit confused RE difference between `_fillRelayV5` and `_fillRelayV3`. It seems like they should be doing the same thing:
- check deadline
- emit event
- transfer tokens

Just the v5 flows will call `_fillRelayV3` with `false` as a `isSlowFill` argument.

`_fillRelayV3` might be a misnomer now, which we could rename to `_fillRelay`

contracts/spoke-pools/SpokePool.sol
fusmanii: updated

contracts/spoke-pools/SpokePool.sol
fusmanii: removed

contracts/spoke-pools/SpokePool.sol
fusmanii: done

contracts/spoke-pools/SpokePool.sol
fusmanii: factored out to `_buildRelayExecution`
https://github.com/across-protocol/contracts/pull/1476/changes/580971f22a06f0e6850bd1fb6b64f5b0bb65e117

contracts/spoke-pools/SpokePool.sol
```diff
@@ -1003,6 +1011,47 @@ abstract contract SpokePool is
         fillRelay(convertedRelayData, repaymentChainId, msg.sender.toBytes32());
     }
 
+    function executeAcrossV5(bytes calldata input, bytes calldata jitData) external payable unpausedFills {
```
droplet-rl: `payable` here is still unused, and it's now more pointed: `_fillRelayV5` invokes the recipient's **also-`payable`** `executeAcrossV5` callback without forwarding value (no `{value: …}`), so any `msg.value` the Gateway sends is stranded in the SpokePool and the executor sees `0`. Decide the intent: if executor commands can need native value, thread `msg.value` through to the callback; if not, drop `payable` from this entrypoint. Also still missing `nonReentrant` — the only fill entrypoint without it, and it now makes an unbounded external call to an arbitrary recipient.

contracts/spoke-pools/SpokePool.sol
```diff
@@ -1589,17 +1667,74 @@ abstract contract SpokePool is
...
+    function _fillRelayV5(
+        V3RelayExecutionParams memory relayExecution,
+        bytes32 relayer,
+        bytes memory message,
+        bytes memory executorMessage
+    ) internal {
+        _recordFill(relayExecution, relayer, FillType.FastFill);
+
+        // V5 fills are settled by pulling the output tokens from the Gateway's current submitter
+        address submitter = gateway.currentSubmitter();
```
droplet-rl, line 1730: The explicit `if (submitter == address(0)) revert V5RequiresGateway();` guard was removed. A zero submitter will now fail later as an opaque `transferFrom(address(0), …)` revert from the output token rather than the clear `V5RequiresGateway`. The entrypoint's `msg.sender == gateway` check means this shouldn't happen in practice, but restoring the explicit revert is cheap and keeps the failure mode legible.

contracts/spoke-pools/SpokePool.sol
```diff
@@ -1589,17 +1667,74 @@ abstract contract SpokePool is
...
+        _transferTokensToRecipient(relayExecution, relayExecution.relay, submitter, false);
+
+        address recipientToSend = relayExecution.updatedRecipient.toAddress();
+        if (message.length > 0 && AddressLibUpgradeable.isContract(recipientToSend)) {
+            IAcrossV5Executor(recipientToSend).executeAcrossV5(message, executorMessage);
```
droplet-rl: Good: routing the executor callback here (after `_recordFill` + transfer) keeps CEI intact and is clean. Note this is now an unbounded external call into arbitrary recipient code with no reentrancy lock held — combined with the missing `nonReentrant` on `executeAcrossV5`. I couldn't construct a drain (submitter-pull is Gateway-gated and the hash is already `Filled`), but it's worth the guard. Also: this is where `msg.value` would be forwarded if native-value executor commands are intended (see the entrypoint comment).

droplet-rl (changes requested):
## Re-review (commit 580971f `add executeAcrossV5 call`)

Pulled the latest, re-read the diff, and confirmed it still **compiles** (`yarn build-evm-foundry` → `Compiler run successful`).

**Resolved from last round — thanks:**
- ✅ Stale `executeAcrossV5` calldata comments removed.
- ✅ `requestSlowFill` now carries `nonV5Fill(relayData.message)`, so both slow-fill paths are guarded — the `V5SlowFillNotAllowed` → `V5MethodNotAllowed` rename reads accurately now.
- ✅ `V5OutputAmountTooLow` is documented.
- ✅ `_buildRelayExecution` is a nice DRY consolidation of the exclusivity check + execution-params build across `fillRelay` and the V5 path. I checked the V3 message-handler relocation out of `_transferTokensToRecipient` into `_fillRelayV3` — behavior is equivalent (same recipient/token/amount/`from`/ordering), and CEI still holds (`_recordFill` → transfer → callback).

**New in this commit:** V5 fills now invoke `IAcrossV5Executor(recipient).executeAcrossV5(message, executorMessage)` after settlement (executor-callback pattern). The CEI ordering is correct and the submitter-pull stays Gateway-gated, so I don't see a reentrancy drain. Two things to resolve, though:

**Blocking:**

1. **Still no tests.** Third round flagging this — there are still zero references to `executeAcrossV5` / `_fillRelayV5` / `V5_DEPOSIT_HEADER` under `test/`. The function now moves a third party's (submitter's) funds *and* makes an unbounded external call into an arbitrary recipient executor. That really shouldn't merge untested. Minimum coverage I'd want: gateway-only access, `minOutputAmount` floor, exclusivity vs `currentSubmitter()`, double-fill (`RelayFilled`), `nonV5Fill` rejecting a V5 slow fill, and the executor callback firing (and bubbling a revert).

2. **`payable` / `msg.value` is dropped on the floor.** `executeAcrossV5` is `payable`, and the recipient callback `IAcrossV5Executor.executeAcrossV5` is also `payable` — but `msg.value` is neither used by the SpokePool nor forwarded to the callback (it's invoked without `{value: …}`). So any native value sent by the Gateway is stranded in the SpokePool and the executor sees `msg.value == 0`. Either forward the value to the executor or drop `payable` from the entrypoint; as written the two `payable`s imply a native-value flow that doesn't actually thread through.

**Non-blocking:**
- `_fillRelayV5` dropped the explicit `if (submitter == address(0)) revert V5RequiresGateway()`. A zero submitter now surfaces as an opaque ERC20 `transferFrom(address(0))` revert instead of the clear error. It's cheap to restore and keeps the failure legible.
- `executeAcrossV5` still lacks `nonReentrant`. No concrete exploit (CEI is right, submitter-pull is Gateway-gated), but now that it calls into arbitrary recipient code it's the one fill entrypoint without the guard every sibling has — cheap defense-in-depth.
- Minor: the SpokePool both *implements* `IAcrossV5Executor.executeAcrossV5` (its Gateway entrypoint) and *calls* `executeAcrossV5` on recipients — same selector, two different roles. Harmless (a self-recipient fill would just revert the Gateway check) but the overloaded interface is a readability trap; a one-line comment would help.

Genuinely close now — it's mostly tests + the `payable`/value question. Happy to re-review.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=6b2213a742ff8ddc022e805749c54412&trace_start_ts=1782431015&trace_end_ts=1782517715)

contracts/spoke-pools/SpokePool.sol
fusmanii: done

contracts/spoke-pools/SpokePool.sol
```diff
@@ -1003,6 +1011,48 @@ abstract contract SpokePool is
         fillRelay(convertedRelayData, repaymentChainId, msg.sender.toBytes32());
     }
 
+    function executeAcrossV5(bytes calldata input, bytes calldata jitData) external payable unpausedFills {
```
droplet-rl: Non-blocking: this is the only fill entrypoint without `nonReentrant`. With `_fillRelayV5` now making an unbounded, value-bearing external call into arbitrary recipient code, adding the guard would be cheap defense-in-depth. I couldn't construct a drain (CEI is correct and the submitter-pull is Gateway-gated), so not a blocker.

contracts/spoke-pools/SpokePool.sol
```diff
@@ -1589,17 +1668,81 @@ abstract contract SpokePool is
...
+        V3RelayExecutionParams memory relayExecution,
+        bytes32 relayer,
+        bytes memory message,
+        bytes memory executorMessage,
+        uint256 msgValue
+    ) internal {
+        _recordFill(relayExecution, relayer, FillType.FastFill);
+
+        // V5 fills are settled by pulling the output tokens from the Gateway's current submitter
+        address submitter = gateway.currentSubmitter();
```
droplet-rl, line 1730: Non-blocking: the explicit `if (submitter == address(0)) revert V5RequiresGateway();` was dropped, so a zero submitter now fails as an opaque `transferFrom(address(0), …)` revert rather than the clear error. Unreachable in practice via the `msg.sender == gateway` gate, but restoring it keeps the failure legible — your call.

droplet-rl (approved):
## Approving ✅ (through commit 1eed547)

Pulled the latest four commits, re-read the diff, and **ran the new suite: `SpokePool_FillV5.t.sol` — 18/18 passing** (`yarn test-evm-foundry`), with the whole test tree compiling under the updated `MockSpokePool(_weth, _gateway)` constructor.

This round closes out the last standing items from my earlier reviews:

- ✅ **Tests added and thorough.** `SpokePool_FillV5.t.sol` covers gateway-only access, pause, `minOutputAmount` floor, exclusivity (active + expired), fill-deadline, happy-path submitter pull + fill status, the `FilledRelay` event (incl. `FastFill` type + message hash), double-fill (`RelayFilled`), the executor callback firing with the right `(message, executorMessage)` payloads, EOA-recipient no-callback, WETH→native unwrap, and all four `nonV5Fill` guard paths. `MockGateway` is a clean minimal stand-in.
- ✅ **`msg.value` no longer strandable.** It's forwarded to the executor callback via `executeAcrossV5{ value: msgValue }`, and rejected with `V5UnusedMsgValue` when there's no callback to consume it — directly tested both ways.
- ✅ **`nonV5Fill` extended to `fillRelay` and `fillRelayWithUpdatedDeposit`** (on top of the slow-fill paths), so a V5-tagged message can only ever be settled through `executeAcrossV5`.

Re-verified the security posture with the now-arbitrary, value-bearing executor callback: CEI holds (`_recordFill` marks the hash `Filled` before any transfer/callback), and the submitter-pull stays Gateway-gated, so a re-entrant recipient can't reach the submitter's approval or double-settle. The earlier reentrancy/drain concern stays closed.

Nice iteration across the rounds — the V3/V5 separation ended up clean.

### Non-blocking follow-ups (optional, fine to defer)
- `executeAcrossV5` is still the one fill entrypoint without `nonReentrant`. No demonstrable exploit (CEI + Gateway-gating cover it), but it now makes an unbounded, value-bearing call into arbitrary recipient code — adding the guard would be cheap belt-and-suspenders consistent with the other fill functions.
- `_fillRelayV5` no longer has the explicit `submitter == address(0)` → `V5RequiresGateway` check; a zero submitter now surfaces as an opaque `transferFrom(address(0))` revert. Practically unreachable given the `msg.sender == gateway` gate, but the explicit revert read better.
- Minor: SpokePool both implements `IAcrossV5Executor.executeAcrossV5` (its Gateway entrypoint) and calls `executeAcrossV5` on recipients — same selector, two roles. Harmless, but a one-line comment would save a future reader a double-take.

None of these block; approving as-is.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=8010dcc195237955c443e3a252c38c07&trace_start_ts=1782437266&trace_end_ts=1782523966)

contracts/spoke-pools/SpokePool.sol
grasphoper: `nonReentrant`?

contracts/spoke-pools/SpokePool.sol
```diff
@@ -1003,6 +1011,48 @@ abstract contract SpokePool is
...
+            exclusiveRelayer: jitInputParamsV5.exclusiveRelayer,
+            inputToken: jitInputParamsV5.inputToken,
+            outputToken: inputParamsV5.outputToken,
+            inputAmount: jitInputParamsV5.inputAmount,
+            outputAmount: jitInputParamsV5.outputAmount,
+            originChainId: jitInputParamsV5.originChainId,
+            depositId: jitInputParamsV5.depositId,
+            fillDeadline: jitInputParamsV5.fillDeadline,
+            exclusivityDeadline: jitInputParamsV5.exclusivityDeadline,
+            message: abi.encodePacked(V5_DEPOSIT_HEADER, gateway.currentStepId())
```
grasphoper: What I'm wondering is:

Should we have relayer specify all of the fill params, but just `require` that they hit a few conditions?

E.g. 

```
require(jitInputParamsV5.recipient == inputParamsV5.recipient);
require(jitInputParamsV5.outputToken == inputParamsV5.outputToken);
require(jitInputParamsV5.message == abi.encodePacked(V5_DEPOSIT_HEADER, gateway.currentStepId()));
```

This will retain the control of defining full fill params with the relayer, only putting a few extra conditions. In this sense, relayer will be as sure as they were before that their fill is not going to be incorrect

contracts/spoke-pools/SpokePool.sol
```diff
@@ -1643,11 +1786,14 @@ abstract contract SpokePool is
      * @notice Transfers tokens to the recipient based on the relay execution parameters.
      * @param relayExecution The relay execution parameters.
      * @param relayData The relay data.
+     * @param from Address the output tokens are pulled from for fast fills.
+     * Unused for slow fills, which are paid out of this contract's reserves.
      * @param isSlowFill Whether this is a slow fill execution.
      */
     function _transferTokensToRecipient(
```
grasphoper, line 1818: We can make this function return `recipientToSend` and `outputToken` to prevent duplicate bytes32 -> address conversion in some flows

contracts/spoke-pools/SpokePool.sol
```diff
@@ -1589,17 +1668,81 @@ abstract contract SpokePool is
...
+    ) internal {
+        _recordFill(relayExecution, relayer, FillType.FastFill);
+
+        // V5 fills are settled by pulling the output tokens from the Gateway's current submitter
+        address submitter = gateway.currentSubmitter();
+        _transferTokensToRecipient(relayExecution, relayExecution.relay, submitter, false);
+
+        address recipientToSend = relayExecution.updatedRecipient.toAddress();
+        if (message.length > 0 && AddressLibUpgradeable.isContract(recipientToSend)) {
+            IAcrossV5Executor(recipientToSend).executeAcrossV5{ value: msgValue }(message, executorMessage);
```
grasphoper: There's a nuance we need to carefully assess:

in traditional `fillRelay`, if outputtoken == wrapped native, we:
- send value to handler
- call handleV3AcrossMessage separately after with no `msg.value`

Should we do the same here or does it not matter which way to go?

grasphoper (commented):
Looking good, a few additional comments

contracts/spoke-pools/SpokePool.sol
fusmanii: good point, I don't think we need it because of `if (msg.sender != address(gateway)) ` check and that Gateway it self is non-reentrant (with the `if (currentSubmitter() != address(0))` check), do you agree?

contracts/spoke-pools/SpokePool.sol
fusmanii: when `outputToken == wrappedNative` we don't send value actually, in `_unwrapwrappedNativeTokenTo` it checks if the recipient is a contract then it still sends wrapped native token as ERC20

but in the V5 case `executeAcrossV5` is payable, so we need to pass in `msg.value` here if the recipient is a contract

contracts/spoke-pools/SpokePool.sol
fusmanii: good point, https://github.com/across-protocol/contracts/pull/1476/changes/5d71e259c01189a84b2408657783f96626677d6e

contracts/spoke-pools/SpokePool.sol
fusmanii: as discussed just adding `message` to `JitInputParamsV5`
https://github.com/across-protocol/contracts/pull/1476/commits/04b229005d0298b312eb936a6b43983557a8c882

contracts/spoke-pools/SpokePool.sol
```diff
@@ -1589,17 +1686,88 @@ abstract contract SpokePool is
...
+     * @param relayer Address credited as the relayer (the repayment address) in the FilledRelay event.
+     * @param message V5 path command sequence forwarded to the recipient's executeAcrossV5 callback.
+     * @param executorMessage Submitter-provided dynamic data forwarded to the recipient's executeAcrossV5 callback.
+     * @param msgValue Native value forwarded to the recipient's executeAcrossV5 callback. Must be zero unless a
+     * callback is invoked, otherwise it would be stranded in this contract.
+     */
+    function _fillRelayV5(
+        V3RelayExecutionParams memory relayExecution,
+        bytes32 relayer,
+        bytes memory message,
```
droplet-rl: Non-blocking naming nit: this param is still called `message` (and the next `executorMessage`), but they now carry `inputParamsV5.executorInput` / `jitInputParamsV5.executorJitInput`. Worse, `message` here collides with `relayData.message` (the V5 header||stepId tag) — the exact ambiguity the struct-field rename was meant to kill. Renaming these to `executorInput` / `executorJitInput` would finish the disambiguation.

contracts/spoke-pools/SpokePool.sol
```diff
@@ -1003,6 +1011,66 @@ abstract contract SpokePool is
...
 
+    /**
+     * @notice Gateway-only entrypoint that fast-fills a V5 deposit. Callable exclusively by the configured Gateway
+     * while it is executing a step. Output tokens are pulled from the Gateway's current submitter and sent to the
+     * recipient. If the recipient is a contract and an executor message is supplied, its IAcrossV5Executor.executeAcrossV5
+     * callback is invoked with any attached msg.value forwarded.
+     * @dev Submitter-provided message is compared against the current stepId from the Gateway's context and must
+     * be of the shape V5_DEPOSIT_HEADER || stepId
+     * @param input ABI-encoded InputParamsV5 — the user/deposit-committed constraints: recipient, outputToken,
+     * minOutputAmount, and the executor message.
```
droplet-rl: Stale wording after the rename: this field is `executorInput` now, not "the executor message".

droplet-rl (approved):
## Re-approving ✅ (through commit 8ef5437)

Reviewed the three follow-up commits and **re-ran the suite: `SpokePool_FillV5.t.sol` — 19/19 passing** (the 18 from before plus the new `testRevertsWhenRelayMessageDoesNotMatchStepId`), with the whole test tree compiling — which also confirms the new `require(..., CustomError())` syntax builds under the repo's solc.

The substantive change — having the submitter supply the relay `message` in `jitData` rather than the contract constructing it — is **behavior-preserving**: it's validated to be exactly `length == 64` with `header == V5_DEPOSIT_HEADER` and `stepId == gateway.currentStepId()`, which is byte-identical to the previous `abi.encodePacked(V5_DEPOSIT_HEADER, gateway.currentStepId())`. So the relay hash, `_isV5Fill` detection, and the emitted message hash are all unchanged, and the strict validation leaves the submitter no freedom to inject a different tag. Nicely covered by the new negative test.

The other changes are clean: struct-field renames (`message`→`executorInput`, `executorMessage`→`executorJitInput`, plus the new `message` tag) disambiguate the two notions of "message", returning `(outputToken, recipientToSend)` from `_transferTokensToRecipient` removes the recompute I'd noted earlier, and the `if/revert`→`require` conversions are equivalent.

Approving — no blocking issues. Two cosmetic follow-ups below, both optional.

### Non-blocking nits
- `_fillRelayV5`'s params are still named `message` / `executorMessage` even though they now carry `executorInput` / `executorJitInput` — and `message` there collides with `relayData.message` (the V5 tag), which is exactly the ambiguity the struct rename set out to remove. Renaming the params to match would finish the cleanup.
- The `executeAcrossV5` docstring still calls the `InputParamsV5` field "the executor message" (line ~1022); it's `executorInput` now.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=fcfd29412171cfd02a8adbeeb454457f&trace_start_ts=1782519532&trace_end_ts=1782606232)

contracts/spoke-pools/SpokePool.sol
fusmanii: done https://github.com/across-protocol/contracts/pull/1476/commits/3edbf8c1c25234bb198ced1757de8a18d78568fc

contracts/spoke-pools/SpokePool.sol
fusmanii: done https://github.com/across-protocol/contracts/pull/1476/commits/3edbf8c1c25234bb198ced1757de8a18d78568fc

contracts/spoke-pools/SpokePool.sol
fusmanii: also not having `nonReentrant` here will allow to chain `fillV5 -> deposit` I think

grasphoper: Hmm, agree with you on the Gateway reasoning.

This means though that fillV5 can reenter other Spoke's fns like deposit / fill (like you mention). Have to understand whether that's a problem or not

grasphoper (approved):

PR #1479, opened 2026-06-30
feat: SpokePool fill as adapter

fusmanii:


contracts/spoke-pools/SpokePool.sol
```diff
@@ -482,6 +480,23 @@ abstract contract SpokePool is
...
+     * action discriminator, so the Executor can compose V5 deposits and fills.
+     * @dev `input` is `abi.encodePacked(bytes1(uint8(V5AdapterAction)), abi.encode(actionInput))`: the leading byte is
+     * the action discriminator and the remainder is the action-specific payload (`V5DepositInput` for deposits,
+     * `V5FillInput` for fills); `jitData` is the matching JIT payload (`V5DepositJit` / `V5FillJit`).
+     * Callable by anyone, but every branch requires a live Gateway execution (non-zero current submitter).
+     */
+    function adapterExecuteAcrossV5(bytes calldata input, bytes calldata jitData) external payable nonReentrant {
+        V5AdapterAction action = V5AdapterAction(uint8(input[0]));
+        if (action == V5AdapterAction.Deposit) {
+            _adapterDepositAcrossV5(input[1:], jitData);
```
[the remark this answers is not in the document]
fusmanii: these contracts are not live yet, so this shouldn't be a problem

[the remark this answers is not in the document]
fusmanii: @codex review

fusmanii:
@claude review

contracts/spoke-pools/SpokePool.sol
droplet-rl, line 494: **Coverage gap (main ask):** the `V5AdapterAction.Deposit` branch is never exercised by any test. `SpokePool_FillAdapter.t.sol` only builds `Fill`-action payloads, and grepping the suite shows no test touches `_adapterDepositAcrossV5`, `V5DepositInput`, the action-byte deposit encoding, or `_resolveDynamicParams`. This path was previously a dedicated external function; the PR now reaches it through the dispatcher + `input[1:]` slice, and that new plumbing (plus the whole JIT param-modification flow) is untested. Please add at least one deposit-dispatch test — happy path that asserts the stamped `V5_MAGIC_PREFIX || dstStepId` message and the derived `depositId`, plus the `InactiveV5Flow` revert. The fill path proves the encoding pattern works, but the deposit branch is security-sensitive enough to warrant its own coverage.

contracts/spoke-pools/SpokePool.sol
```diff
@@ -482,6 +480,23 @@ abstract contract SpokePool is
...
+    /**
+     * @notice Single Executor adapter entrypoint for V5 flows. Dispatches to a deposit or a fill based on the leading
+     * action discriminator, so the Executor can compose V5 deposits and fills.
+     * @dev `input` is `abi.encodePacked(bytes1(uint8(V5AdapterAction)), abi.encode(actionInput))`: the leading byte is
+     * the action discriminator and the remainder is the action-specific payload (`V5DepositInput` for deposits,
+     * `V5FillInput` for fills); `jitData` is the matching JIT payload (`V5DepositJit` / `V5FillJit`).
+     * Callable by anyone, but every branch requires a live Gateway execution (non-zero current submitter).
+     */
+    function adapterExecuteAcrossV5(bytes calldata input, bytes calldata jitData) external payable nonReentrant {
+        V5AdapterAction action = V5AdapterAction(uint8(input[0]));
```
droplet-rl, line 492: Minor: `input[0]` on an empty `input` reverts with a low-level array out-of-bounds panic rather than a descriptive error. Harmless (it still reverts), but a `require(input.length > 0, ...)` would give a clearer failure for a malformed adapter call.

contracts/spoke-pools/SpokePool.sol
```diff
@@ -1466,6 +1445,49 @@ abstract contract SpokePool is
         );
     }
 
+    /**
+     * @notice Deposit branch of `adapterExecuteAcrossV5`: as an `ADAPTER_CALL` step (input tokens pulled from the
+     * Executor, live Gateway execution required), creates a V5-stamped Across deposit after applying any JIT auction params.
+     * @param input ABI-encoded `V5DepositInput` — committed deposit params + param-modification rules.
+     * @param jitData ABI-encoded `V5DepositJit` — the submitter's JIT modifications and authorizing signature.
+     */
+    function _adapterDepositAcrossV5(bytes calldata input, bytes calldata jitData) internal unpausedDeposits {
```
droplet-rl, line 1454: Minor footgun: the dispatcher is `payable` but this deposit branch ignores `msg.value`, so any native value attached to a `Deposit` action is silently stranded in the contract. It was technically possible before (the old deposit entrypoint was also payable), but now that deposit and fill share one payable entrypoint the mistake is easier to make. Consider `if (msg.value > 0) revert V5UnusedMsgValue();` on the deposit branch to mirror the fill path's stranded-value protection.

contracts/spoke-pools/SpokePool.sol
```diff
@@ -1719,37 +1741,81 @@ abstract contract SpokePool is
...
-        V3RelayExecutionParams memory relayExecution,
-        bytes32 relayer,
-        bytes memory executorInput,
-        bytes memory executorJitInput,
-        uint256 msgValue
-    ) internal {
-        _recordFill(relayExecution, relayer, FillType.FastFill);
+    function _adapterFillAcrossV5(bytes calldata input, bytes calldata jitData) internal unpausedFills {
+        // Require a live Gateway execution; the caller (Executor) is implicitly authorized by this context.
+        if (gateway.currentSubmitter() == address(0)) revert InactiveV5Flow();
```
droplet-rl, line 1751: Nit (gas): `gateway.currentSubmitter()` is read here for the liveness gate and then read again inside `_fillV5` → `_buildRelayExecution` for the exclusivity relayer. You could read it once and pass it down, saving an external call on the adapter fill path. Not worth much; flagging only for completeness.

droplet-rl (commented):
## Review: `feat: SpokePool fill as adapter`

Nice refactor. The change generalizes the single Gateway-driven `executeAcrossV5` fill into a shared `_fillV5` core reachable from two entrypoints, and folds the V5 deposit path behind a new dispatcher `adapterExecuteAcrossV5(input, jitData)` that selects deposit vs. fill via a leading 1-byte action discriminator. The one real behavioral difference between the two fill paths — Gateway-direct pulls output tokens from `gateway.currentSubmitter()`, adapter pulls from `msg.sender` (the Executor) — is threaded cleanly through a single `from` parameter. Structs are renamed to a clearer `V5{Deposit,Fill}{Input,Jit}` scheme.

**Verification:** installed deps, built with `FOUNDRY_PROFILE=local-test`, and ran the two affected suites — `SpokePoolFillAdapterTest` (15) and `SpokePoolFillV5Test` (19), all 34 pass. Full project compiles. I confirmed `keccak256("AcrossV5MessagePrefix.V1") == 0x89ae4bc75915265a3f10e926c3894a29534f1d6362ee8959cb0e5be00f3527fd`, so the hardcoded prefix is correct.

**Correctness spot-checks that hold up:**
- CEI preserved in `_fillV5`: checks (min-out, witness, deadline/exclusivity via `_buildRelayExecution`) → effect (`_recordFill`) → interactions (token transfer, executor callback), all under `nonReentrant`.
- Guard composition is intact after moving the pause checks onto the internal branches: `adapterExecuteAcrossV5` is `nonReentrant`, the deposit branch carries `unpausedDeposits`, the fill branch `unpausedFills` — equivalent to the pre-refactor modifiers.
- Invalid action bytes are safe: `V5AdapterAction(uint8(input[0]))` reverts with `Panic(0x21)` for any byte ≥ 2, so the `else` branch is only reachable for `Fill` (byte == 1) — no silent misdispatch.
- Cross-path double-fill is deduped because both paths key `_recordFill` on the same relay hash.

**Notable side-fix worth calling out explicitly in the PR description:** collapsing `V5_DEPOSIT_HEADER` into `V5_MAGIC_PREFIX` is more than a rename. On the base branch deposits were stamped with `V5_MAGIC_PREFIX` while both the fill-tag check *and* the `_isV5Fill` slow-fill guard compared against the *different* `V5_DEPOSIT_HEADER` — so the `nonV5Fill` guard never actually recognized a real V5-stamped deposit. Unifying to one constant closes that inconsistency. I grepped the repo (Solidity + TS/JS + JSON) and found no remaining references to the removed public constant, so the ABI-getter removal is safe internally; just confirm no off-chain tooling reads `V5_DEPOSIT_HEADER()`.

**Main ask:** the new deposit-dispatch branch has no test coverage (details inline). No blocking correctness issues found — the remaining inline notes are minor. Leaving as a comment rather than a block.

droplet-rl:
:warning: Couldn't finish 1 background task(s) before the 30m budget ran out — their results weren't gathered:
  • Confirm full project build

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=b3bf4ff764a0a4c95c16ed16fe15d18d&trace_start_ts=1782871382&trace_end_ts=1782958082)

PR #1481, opened 2026-07-02
feat: v5 + Spoke

grasphoper:
- spoke v5: 
  - fill when path.executor(see v5 repo) is SpokePool (spoke-as-executor)
  - fill through ADAPTER_CALL command (spoke-as-adapter)
- small addition to Hyperliquid contracts for easier reuse in v5 repo


contracts/spoke-pools/SpokePool.sol
```diff
@@ -1566,9 +1644,77 @@ abstract contract SpokePool is
...
+     * @param payer Address the output tokens are pulled from: the Gateway's current submitter (executor mode)
+     * or msg.sender, the executing tape's own balance (adapter mode).
+     */
+    function _fillV5(V5FillInput memory fillInput, V5FillJit memory fillJit, address payer) internal unpausedFills {
+        if (fillJit.outputAmount < fillInput.minOutputAmount) revert V5OutputAmountTooLow();
+        _requireExclusiveFiller(fillJit.exclusivityDeadline, fillJit.exclusiveRelayer, gateway.currentSubmitter());
+
+        V3RelayData memory relayData = V3RelayData({
+            depositor: fillJit.depositor,
+            recipient: fillInput.recipient,
```
mrice32: Why is the recipient and outputToken an input, while other fields are JIT? Not saying that's wrong, just trying to understand the reasoning.

contracts/spoke-pools/SpokePool.sol
```diff
@@ -1566,9 +1644,77 @@ abstract contract SpokePool is
...
+            exclusiveRelayer: fillJit.exclusiveRelayer,
+            inputToken: fillJit.inputToken,
+            outputToken: fillInput.outputToken,
+            inputAmount: fillJit.inputAmount,
+            outputAmount: fillJit.outputAmount,
```
mrice32: Any concerns around the case where someone sends an invalid fill (0 output amount), but the indexer thinks this is complete because the rest of the actions don't care about the output amount of the fill?

contracts/spoke-pools/SpokePool.sol
grasphoper: user's money is tied to the `witness`, which in turn gets constructed (waves hands) from `input`

- input.recipient makes sure that when a `pathId` is emitted => the funds went to the user (same as having `TRANSFER` cmd in the general tape)
- input.token + input.minAmount makes sure that no one can spam-create `pathId` from a user order without first spending at least `minAmount` (this acts as a sort of `BALANCE_REQ`)

So these two make spoke-as-executor similar to:
```
[BALANCE_REQ, TRANSFER(recipient), optional-handleV3]
```

Let's say we didn't commit to input.recipient:
- pathId could be emitted for an unrelated fill that spammer would just send to self

If we didn't commit to token, minAmount:
- pathId could be emitted with 1 wei

Also, take another tape:

```
[ADAPTER_CALL, TRANSFER-or-actions-from-input]
```
ADAPTER_CALL gets BALANCE_REQ included in it basically for free, giving us the same pathId guarantees as in the spoke-as-executor case.

If we didn't have these, we'd have to have a `BALANCE_REQ` in this flow. (just in the spirit of making simple flows more lean)

contracts/spoke-pools/SpokePool.sol
grasphoper: That's what `minAmount` protection gives us (see above). Even if minAmount is static (e.g. with 5% slippage), whereas fill.outputAmount is dynamic, noone can "just emit a pathId for free"

mrice32: Makes sense. I missed that.

contracts/spoke-pools/SpokePool.sol
mrice32: Makes sense. The minAmount ties it together. I missed that in my review.

mrice32 (dismissed):
LGTM!

fusmanii (approved):

tbwebb22 (approved):

PR #1495, opened 2026-07-17
feat: SpokePool SVM V5 modifier

fusmanii:
Similar to EVM V5 SpokePool [modifier](https://github.com/across-protocol/contracts/blob/if/v5-iter2/contracts/spoke-pools/SpokePool.sol#L315)

Closes ACP-190


programs/svm-spoke/src/instructions/fill.rs
```diff
@@ -92,6 +95,11 @@ pub fn fill_relay<'info>(
     let FillRelayParams { relay_data, repayment_chain_id, repayment_address } =
         unwrap_fill_relay_params(relay_data, repayment_chain_id, repayment_address, &ctx.accounts.instruction_params);
 
+    // V5 tagged deposits must be filled through new V5 entrypoints
+    if is_v5_message(&relay_data.message) {
+        return err!(CommonError::V5FillOnly);
+    }
```
droplet-rl, line 101: Non-blocking: the comment (and the error message) refer to "V5 fill entrypoints", but no such instructions exist in the SVM program yet. Until they land, a V5-tagged deposit destined to Solana is simply unfillable on every path (fast and slow) and settles via expiry + origin-chain refund. That's presumably the intent — quarantine first, entrypoints later — but a few words here saying so would save the next reader from hunting for entrypoints that don't exist.

test/svm/SvmSpoke.Fill.ts
```diff
@@ -327,6 +328,27 @@ describe("svm_spoke.fill", () => {
...
+    );
+    const dstStepId = Buffer.alloc(32, 7);
+    updateRelayData({ ...relayData, message: Buffer.concat([v5MagicPrefix, dstStepId]) });
+
+    const relayHash = Array.from(calculateRelayHashUint8Array(relayData, chainId));
+    try {
+      await approvedFillRelay([relayHash, relayData, new BN(1), relayer.publicKey]);
+      assert.fail("Fill should have failed due to V5-tagged message");
+    } catch (err: any) {
+      assert.include(err.toString(), "V5FillOnly", "Expected V5FillOnly error");
```
droplet-rl, line 348: Nice pattern deriving the prefix instead of hardcoding it — this is effectively the cross-VM byte-parity test. One optional addition while you're here: a companion test asserting that `deposit` still *accepts* a V5-tagged message would pin down the other half of the contract (quarantine is fill-side only; origin-side V5 deposits from Solana must keep working).

droplet-rl (approved):
Clean, well-scoped port of the EVM `nonV5Fill` quarantine to the SVM spoke. Verified the details that matter here:

**Constant parity** — I recomputed `keccak256("AcrossV5MessagePrefix.V1")` independently and it matches the hardcoded bytes exactly (`0x89ae4bc7...27fd`), so the SVM constant is byte-for-byte identical to the EVM `V5_MAGIC_PREFIX` in #1481.

**Predicate parity** — `is_v5_message` (`len >= 32 && message[..32] == PREFIX`) has identical semantics to the EVM `_isV5Message` (length `< 32` → false, first 32 bytes compared), including the edge case of a message that is exactly the 32-byte prefix.

**Coverage** — the guard lands on all three settlement entrypoints that consume `relay_data.message`: `fill_relay`, `request_slow_fill`, and `execute_slow_relay_leaf`. The EVM PR guards a fourth function (`fillRelayWithUpdatedDeposit`), which has no SVM counterpart, so this is the complete set. The deposit side is intentionally unguarded, matching EVM (V5 deposits are ordinary deposits with a stamped witness message).

**Placement** — in all three handlers the check runs after `unwrap_*_params`, so it covers both the direct-argument path and the `instruction_params` buffer path, and it runs before any state mutation or token transfer.

**Error enum** — `V5FillOnly` is appended at the end of `CommonError`, so no existing error codes shift, and the name matches the EVM `V5FillOnly()` error.

**Tests** — deriving the prefix via `keccak256` in the tests (rather than copying the program constant) makes them double as the cross-VM parity check, which is a nice touch. Test isolation is fine: both suites reset `relayData` in `beforeEach`, and `updateRelayData` in `SvmSpoke.Fill.ts` is synchronous, so the non-awaited call is correct. The untested `execute_slow_relay_leaf` guard is genuinely unreachable end-to-end (the fill-status PDA for a V5 relay hash can never exist since both `request_slow_fill` and `fill_relay` reject first), and the test comment says so — no action needed.

`cargo check` on the program passes with no errors. Two non-blocking notes inline.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=3820dd52fcc629820dd7199279ab9311&trace_start_ts=1784561668&trace_end_ts=1784648368)

Reinis-FRP (approved):

PR #1520, opened 2026-08-11
chore: counterfactual v3 deployments

tbwebb22:
Note that Across is removing support for the chains below, and will therefore not be supporting PDAs on them. So there are some deleted broadcast files for these chains:
232 Lens
1135 Lisk
1868 Soneium
34443 Mode
81457 Blast
7777777 Zora

For checking deployed configurations, I recommend running:
```
source .env
FOUNDRY_PROFILE=counterfactual forge script \
  script/counterfactual/CheckCounterfactualDeployments.s.sol:CheckCounterfactualDeployments \
  --rpc-url $NODE_URL_1 --ffi --gas-limit 100000000000 -vv
```

Note that since the sponsored bridging redeployments from [PR#1436](https://github.com/across-protocol/contracts/pull/1436) were removed from this branch, these checks in the checker script above will fail. The deployed counterfactual contracts should point to OFT/CCTP src peripheries in the #1436 branch.

Counterfactual roles should all be transferred to multisigs: `owner` on the beacon, `owner` on the `AdminWithdrawManager`, and `directWithdrawer` on the `AdminWithdrawManager`.

On Tron, these should be set to the fallback EOA


grasphoper (commented):
This one looks mostly fine to me, left a few Qs in Slack

grasphoper (approved):

PR #1533, opened 2026-08-19
fix: split beacon interface

tbwebb22:
Splits the beacon interface in two: ICounterfactualBeaconBase now holds just the registry surface (implementation(), upgradeRoot()), and ICounterfactualBeacon extends it with this repo's chain-config getters. 

The base contract, the dispatcher, and CounterfactualImplementationBase now depend only on the registry half, so using the beacon as a beacon no longer drags in the v4 config surface — which is what lets the V5 beacon in contracts-v5 extend the base and declare its own config interface. 

This is a pure refactor with no behavior change: all ten counterfactual contracts compile to bytecode byte-identical to master, so every deployed address is unaffected.

Claude report that bytecode is unchanged:
https://claude.ai/code/artifact/181aaa4e-aa33-4be8-b877-2c12eb72a8bd

### DIFF
--- a/contracts/interfaces/ICounterfactualDepositFactory.sol
+++ b/contracts/interfaces/ICounterfactualDepositFactory.sol
@@ -3,80 +3,59 @@ pragma solidity ^0.8.0;
 
 /**
  * @title ICounterfactualDepositFactory
- * @notice Interface for the generic counterfactual deposit factory
- * @dev This factory creates reusable deposit addresses via CREATE2. It is bridge-agnostic:
- *      each implementation defines its own immutables struct, and the factory stores only the hash.
+ * @notice The deterministic deployer for counterfactual `BeaconProxy` instances. A proxy's address is
+ *         `f(salt, initialRoot)`, because the `initialize(initialRoot)` call data sits in the proxy's init
+ *         code — so a depositor can be quoted an address before anything is deployed, and the proxy can be
+ *         created later (or never, if funds are swept by another route).
+ * @dev Implemented by `CounterfactualDepositFactory`. Cross-chain address parity requires the factory and
+ *      the beacon to sit at identical addresses on every chain, and the caller to reuse the same `salt`;
+ *      `salt = 0` is the canonical choice, giving one address per `initialRoot`. Chain-specific variants
+ *      (e.g. `CounterfactualDepositFactoryTron`, for Tron's 0x41 CREATE2 prefix) override prediction only.
  * @custom:security-contact bugs@across.to
  */
 interface ICounterfactualDepositFactory {
-    /// @notice Emitted when a new clone is deployed.
-    event DepositAddressCreated(
-        address indexed depositAddress,
-        address indexed counterfactualDepositImplementation,
-        bytes32 indexed paramsHash,
-        bytes32 salt
-    );
+    /// @notice Emitted when a counterfactual proxy is deployed.
+    event CounterfactualDeployed(address indexed counterfactual, bytes32 initialRoot);
+
+    /// @notice The beacon (the `CounterfactualBeacon`) every deployed proxy points at.
+    function BEACON() external view returns (address);
 
     /**
-     * @notice Predicts the deterministic address of a clone before deployment.
-     * @param counterfactualDepositImplementation Implementation contract address.
-     * @param paramsHash keccak256 hash of the ABI-encoded route parameters.
-     * @param salt Unique salt for address generation.
-     * @return Predicted address.
+     * @notice Predict the proxy address for a given `salt` and `initialRoot`, deployed or not.
+     * @param salt CREATE2 salt; `0` for the canonical address of this `initialRoot`.
+     * @param initialRoot The proxy's initial route tree, bound into its address via the init code.
+     * @return The counterfactual address.
      */
-    function predictDepositAddress(
-        address counterfactualDepositImplementation,
-        bytes32 paramsHash,
-        bytes32 salt
-    ) external view returns (address);
+    function predictAddress(bytes32 salt, bytes32 initialRoot) external view returns (address);
 
     /**
-     * @notice Deploys a counterfactual deposit clone via CREATE2.
-     * @param counterfactualDepositImplementation Implementation contract address.
-     * @param paramsHash keccak256 hash of the ABI-encoded route parameters.
-     * @param salt Unique salt for address generation.
-     * @return depositAddress Address of deployed clone.
+     * @notice Deploy the proxy for `salt` and `initialRoot`, already initialized and resolving its
+     *         implementation live from the beacon. Reverts if it is already deployed.
+     * @return counterfactual The deployed proxy address.
      */
-    function deploy(
-        address counterfactualDepositImplementation,
-        bytes32 paramsHash,
-        bytes32 salt
-    ) external returns (address depositAddress);
+    function deploy(bytes32 salt, bytes32 initialRoot) external returns (address counterfactual);
 
     /**
-     * @notice Forwards calldata to a deployed clone.
-     * @param depositAddress Address of the deployed clone.
-     * @param executeCalldata Calldata to forward (e.g. abi.encodeCall of executeDeposit).
-     */
-    function execute(address depositAddress, bytes calldata executeCalldata) external payable;
-
-    /**
-     * @notice Deploys and executes a deposit in one transaction.
-     * @param counterfactualDepositImplementation Implementation contract address.
-     * @param paramsHash keccak256 hash of the ABI-encoded route parameters.
-     * @param salt Unique salt for address generation.
-     * @param executeCalldata Calldata to forward to the clone.
-     * @return depositAddress Address of deployed clone.
+     * @notice Deploy the proxy, then forward `executeCalldata` to it. Reverts if already deployed.
+     * @dev Any `msg.value` is forwarded with the call; a revert in the proxy is bubbled up verbatim.
+     * @param executeCalldata Calldata for the proxy (e.g. `abi.encodeCall(ICounterfactualDeposit.execute, …)`).
+     * @return counterfactual The deployed proxy address.
      */
     function deployAndExecute(
-        address counterfactualDepositImplementation,
-        bytes32 paramsHash,
         bytes32 salt,
+        bytes32 initialRoot,
         bytes calldata executeCalldata
-    ) external payable returns (address depositAddress);
+    ) external payable returns (address counterfactual);
 
     /**
-     * @notice Deploys (if needed) and executes a deposit in one transaction.
-     * @param counterfactualDepositImplementation Implementation contract address.
-     * @param paramsHash keccak256 hash of the ABI-encoded route parameters.
-     * @param salt Unique salt for address generation.
-     * @param executeCalldata Calldata to forward to the clone.
-     * @return depositAddress Address of deployed clone.
+     * @notice Deploy the proxy if it does not exist yet, then forward `executeCalldata` to it. Idempotent
+     *         in the deployment step, so this is the safe entry point when a proxy may already be live.
+     * @param executeCalldata Calldata for the proxy (e.g. `abi.encodeCall(ICounterfactualDeposit.execute, …)`).
+     * @return counterfactual The proxy address, newly deployed or pre-existing.
      */
     function deployIfNeededAndExecute(
-        address counterfactualDepositImplementation,
-        bytes32 paramsHash,
         bytes32 salt,
+        bytes32 initialRoot,
         bytes calldata executeCalldata
-    ) external payable returns (address depositAddress);
+    ) external payable returns (address counterfactual);
 }
--- a/contracts/periphery/counterfactual/CounterfactualBeaconBase.sol
+++ b/contracts/periphery/counterfactual/CounterfactualBeaconBase.sol
@@ -5,7 +5,7 @@ import { Initializable } from "@openzeppelin/contracts-upgradeable/proxy/utils/I
 import { UUPSUpgradeable } from "@openzeppelin/contracts-upgradeable/proxy/utils/UUPSUpgradeable.sol";
 import { Ownable2StepUpgradeable } from "@openzeppelin/contracts-upgradeable/access/Ownable2StepUpgradeable.sol";
 import { IBeacon } from "@openzeppelin/contracts/proxy/beacon/IBeacon.sol";
-import { ICounterfactualBeacon } from "../../interfaces/ICounterfactualBeacon.sol";
+import { ICounterfactualBeaconBase } from "../../interfaces/ICounterfactualBeaconBase.sol";
 
 /// @dev Minimal view used to verify a candidate beacon target is bound to this beacon — every
 ///      counterfactual implementation embeds its beacon as the immutable `BEACON` (for `updateRoot`).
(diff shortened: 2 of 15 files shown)


script/counterfactual/DeployCounterfactualBeacon.s.sol
```diff
@@ -29,7 +29,7 @@ import { CounterfactualBeaconBootstrap } from "../../contracts/periphery/counter
 //      pointing it at the chain-specific impl. A proxy already on a real impl keeps that impl (no upgrade);
 //      the dispatcher wiring in steps 4-5 still runs so an interrupted deploy is completed.
 //      The bootstrap already consumed the initializer slot, so pass empty calldata (no re-init).
-//   4. The dispatcher `new CounterfactualDeposit(ICounterfactualBeacon(proxy))` via CREATE2 (proxy is
+//   4. The dispatcher `new CounterfactualDeposit(ICounterfactualBeaconBase(proxy))` via CREATE2 (proxy is
```
droplet-rl, line 32: This comment now describes code that doesn't exist. The CREATE2 init code is actually built in `script/counterfactual/CounterfactualConfig.sol:66`, which still reads:

```solidity
return abi.encodePacked(type(CounterfactualDeposit).creationCode, abi.encode(ICounterfactualBeacon(proxy)));
```

No functional impact — `abi.encode` of either contract type yields the same 32-byte address word, and I confirmed `CounterfactualDeposit`'s creation code (and therefore the dispatcher's CREATE2 address) is byte-identical to base. But a stale comment in the *deploy script* is a bad place for drift: this file is what someone reads when reconstructing an init-code preimage to check a deployed address.

Suggest updating `CounterfactualConfig.sol:66` to `ICounterfactualBeaconBase(proxy)` so the comment is true. `test/evm/foundry/local/CounterfactualTestBase.sol:79` has the same leftover (`new CounterfactualDeposit(ICounterfactualBeacon(address(beacon)))`) — it compiles via the implicit upcast to the base interface, but it's the same inconsistency.

contracts/periphery/counterfactual/CounterfactualBeaconBase.sol
```diff
@@ -18,18 +18,18 @@ interface IBeaconTarget {
...
- *         that touches no logic here (see `CounterfactualBeacon`).
- * @dev Abstract — the config getters declared in `ICounterfactualBeacon` are implemented by the derived
- *      contract's immutables. A UUPS proxy, so the registry address is permanent (anchoring every
- *      `BeaconProxy`) while logic/config evolve. `Ownable2Step` admin (no timelock) — it can retarget every
- *      proxy instantly, so use a trusted multisig. `implementation()` is the beacon target, not the
- *      registry's own UUPS implementation.
+ *         lives in a derived contract as immutables. Splitting it out keeps the audit boundary clean: this
+ *         base is the reviewable logic; a config-only change is a new derived contract that touches no logic
+ *         here (see `CounterfactualBeacon`).
+ * @dev Abstract — the config getters are declared by whichever config interface the derived contract
```
droplet-rl, line 24: This is the one real consequence of the split and I think the natspec undersells it.

Before, the base inherited `ICounterfactualBeacon`, so the compiler *forced* every concrete beacon to expose the full config getter set — you couldn't ship a derived beacon that silently lacked `usdc()` or `signer()`. That enforcement is gone. `CounterfactualBeacon` re-declares `ICounterfactualBeacon` explicitly, which restores it for this repo's beacon, but that's now convention rather than something the base can guarantee.

The current wording ("the config getters are declared by whichever config interface the derived contract implements") reads as description, not obligation. A future derived beacon that forgets to list a config interface compiles cleanly, and the failure only shows up at runtime — and asymmetrically: the hardcoded typed getters (`beacon.usdc()`, `beacon.signer()`) revert, but `_resolveBeaconAddress` swallows the failed staticcall and returns `address(0)`, surfacing as `RouteNotConfigured` rather than anything that points at a missing getter.

Suggest phrasing it as a requirement — something like "a derived beacon MUST declare a config interface extending `ICounterfactualBeaconBase`; the base cannot enforce this" — since that's the invariant the V5 beacon in contracts-v5 will also need to honor. Worth noting `CheckCounterfactualBeaconImpls.s.sol` already covers this operationally by probing every getter on a deployed impl, so the safety net exists; it's just no longer at compile time.

contracts/periphery/counterfactual/CounterfactualImplementationBase.sol
```diff
@@ -20,16 +21,16 @@ abstract contract CounterfactualImplementationBase is ICounterfactualImplementat
     error RouteNotConfigured();
 
     /// @dev The per-chain registry that anchors this proxy (resolved from the ERC-1967 beacon slot).
-    function _beacon() internal view returns (ICounterfactualBeacon) {
-        return ICounterfactualBeacon(ERC1967Utils.getBeacon());
+    function _beacon() internal view returns (address) {
```
droplet-rl, line 24: Non-blocking, and I can see you already went back and forth here (`795d39b2` then `2c5927f8`) — just recording the trade-off.

Returning `address` makes the base genuinely config-agnostic, which is the goal, but it pushes `ICounterfactualBeacon(_beacon())` out to 9 call sites across the four v4 leaves, including twice inside `CounterfactualDepositCCTP` in functions that sit right next to one that already holds a typed `beacon` local. Against the repo's "prioritize succinctness / reuse existing helpers" guidance, the dropped `_configBeacon()` read better at the call sites.

If it ever comes back, the fix for what I assume prompted dropping it — a private helper duplicated in four files — is to hang it off a shared v4-specific intermediate (`abstract contract CounterfactualV4ImplementationBase is CounterfactualImplementationBase`) rather than per-leaf, so the base stays config-agnostic and the cast lives in exactly one place.

The other option that preserves typing is `returns (ICounterfactualBeaconBase)` — still config-agnostic, keeps the registry surface typed — but it costs `address(_beacon()).staticcall(...)` here and `ICounterfactualBeacon(address(_beacon()))` at the leaves, which is worse. So `address` is defensible; fine as-is.

If you do change any of this, re-run the two-branch bytecode diff. An inlined `internal view` helper *should* be free, but the byte-identity property is the main asset of this PR and worth re-confirming rather than assuming.

droplet-rl (dismissed):
Clean split, and the "no behavior change" claim holds up — I reproduced it rather than taking it on faith.

## What I verified

**Bytecode identity.** Built `contracts/periphery/counterfactual/` at `taylor/beacon-base-split` and at `origin/epic/aug-26` (both `FOUNDRY_PROFILE=counterfactual forge build --skip test --skip script`, solc 0.8.30) and diffed artifacts. All **13** contracts in that directory — not just the ten — are identical on **both** creation and runtime bytecode:

```
CounterfactualBeacon              creation=SAME runtime=SAME
CounterfactualBeaconBootstrap     creation=SAME runtime=SAME
CounterfactualDeposit             creation=SAME runtime=SAME
CounterfactualDepositCCTP         creation=SAME runtime=SAME
CounterfactualDepositOFT          creation=SAME runtime=SAME
CounterfactualDepositSpokePool    creation=SAME runtime=SAME
CounterfactualDepositSpokePoolTr  creation=SAME runtime=SAME
CounterfactualDepositVanillaCCTP  creation=SAME runtime=SAME
CounterfactualDepositFactory      creation=SAME runtime=SAME
CounterfactualDepositFactoryTron  creation=SAME runtime=SAME
WithdrawImplementation            creation=SAME runtime=SAME
WithdrawImplementationTron        creation=SAME runtime=SAME
AdminWithdrawManager              creation=SAME runtime=SAME
```

Worth stating explicitly for whoever re-checks this later: identity holds **because deploys run under `[profile.counterfactual]`, which pins `bytecode_hash = "none"`**. Under the default profile the artifacts carry an IPFS metadata hash covering source content, and `ICounterfactualBeaconBase.sol` is a new source while `CounterfactualBeaconBase.sol`'s comments changed — so default-profile artifacts *will* differ in those trailing 82 hex chars. That's expected and irrelevant to CREATE2 addresses, but it would look alarming to someone verifying with the wrong profile.

I also checked the Tron angle, since `[profile.tron]` does *not* set `bytecode_hash = "none"`: no issue, because `tron-deploy-counterfactual-beacon.ts` documents that TronWeb uses plain CREATE and cross-chain address parity is impossible there anyway. No CREATE2 preimage to disturb.

**ABI.** `CounterfactualBeacon` — ABI and `methodIdentifiers` byte-identical. `CounterfactualDeposit` — selectors identical; the only ABI delta is the constructor arg's `internalType` string (`contract ICounterfactualBeacon` → `contract ICounterfactualBeaconBase`). Canonical `type` stays `address`, so ethers/viem/typechain consumers are unaffected. `ICounterfactualBeacon`'s own ABI surface is unchanged — `implementation()`, `upgradeRoot()`, and both events still resolve through the new base.

**Compilation and tests.** Full-repo `forge build` (default profile) succeeds — scripts and tests included. All 6 counterfactual suites pass: 78 tests, 0 failed.

## Reasoning check on the split itself

The inheritance change is sound. `CounterfactualBeacon is CounterfactualBeaconBase, ICounterfactualBeacon` linearizes fine, and re-declaring `ICounterfactualBeacon` on the concrete contract preserves the compile-time guarantee that the config getters exist — which matters, because the base no longer supplies it (see inline comment). `CounterfactualBeaconBootstrap` doesn't derive from `CounterfactualBeaconBase`, so the frozen-bytecode contract is untouched, as the byte comparison confirms. `CounterfactualDeposit` only ever calls `BEACON.upgradeRoot()`, so narrowing it to the registry surface is exactly right.

## Findings

All non-blocking; nothing here changes behavior or addresses. Three inline comments — the first is a real comment/code drift introduced by this PR, the other two are about making the new convention self-enforcing. Anything acted on should be re-checked against the byte-identity property, since that's the thing worth protecting here.

One thought for a follow-up rather than this PR: `DESIGN.md` describes the system purely at the contract level and never names the interfaces, so nothing in it is stale. But given the whole point of this split is to let an out-of-repo V5 beacon extend `CounterfactualBeaconBase`, a sentence in the registry section noting the base/config interface boundary would help the next reader find it.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=cd3c7189c59f144b06c8361cdd1a7c3c&trace_start_ts=1787173169&trace_end_ts=1787259869)

grasphoper (dismissed):

tbwebb22:
@droplet-rl I pushed some more changes, double check again that contract bytecodes are unchanged

droplet-rl:
Re-checked `a20cd0b5`. **Bytecode is still unchanged** — verified two independent ways.

## Bytecode

Clean builds on both sides, `FOUNDRY_PROFILE=counterfactual forge build --skip test --skip script`, solc 0.8.30.

**1. New head `a20cd0b5` vs base `origin/epic/aug-26` (`8aa73521`)** — all 13 contracts in `contracts/periphery/counterfactual/`, identical on **creation and runtime**:

```
CounterfactualBeacon              creation=SAME runtime=SAME
CounterfactualBeaconBootstrap     creation=SAME runtime=SAME
CounterfactualDeposit             creation=SAME runtime=SAME
CounterfactualDepositCCTP         creation=SAME runtime=SAME
CounterfactualDepositOFT          creation=SAME runtime=SAME
CounterfactualDepositSpokePool    creation=SAME runtime=SAME
CounterfactualDepositSpokePoolTr  creation=SAME runtime=SAME
CounterfactualDepositVanillaCCTP  creation=SAME runtime=SAME
CounterfactualDepositFactory      creation=SAME runtime=SAME
CounterfactualDepositFactoryTron  creation=SAME runtime=SAME
WithdrawImplementation            creation=SAME runtime=SAME
WithdrawImplementationTron        creation=SAME runtime=SAME
AdminWithdrawManager              creation=SAME runtime=SAME
```

**2. New head vs previous head `2c5927f8`** — also all SAME, i.e. the new commit moved nothing on its own.

I also added a check I'd skipped last round: `BeaconProxy` and `ERC1967Proxy` creation code are byte-identical too. That matters because a user's counterfactual deposit address is `f(factory, salt, keccak256(BeaconProxy initcode))` — so the *user-facing* addresses are confirmed unaffected, not just the system contracts.

## ABI

`CounterfactualDepositFactory` and `CounterfactualDepositFactoryTron` ABIs are **byte-identical to base** — which confirms `CounterfactualDeployed` survived the move into the interface and that inheriting `ICounterfactualDepositFactory` neither added nor dropped a member. `CounterfactualDeposit` selectors identical; still just the constructor `internalType` string delta from the earlier commits.

## Everything else

Full default-profile `forge build` clean (contracts + scripts + tests). 78 counterfactual tests pass, plus 21 in `AdminWithdrawManager.t.sol` / `WithdrawImplementation.t.sol` since those import `ICounterfactualDeposit`. Prettier clean. Solhint 0 errors — 2 new `func-name-mixedcase` warnings for the `BEACON()` interface declarations, matching the ones already on the contracts, so non-failing.

## On the commit itself

Worth saying this is more than a docs pass: `ICounterfactualDepositFactory` was genuinely dead and wrong. I grepped — before this commit **nothing in the repo referenced it**, and it still described a `predictDepositAddress` / `deploy(address,bytes32,bytes32)` / `execute` API with a `DepositAddressCreated` event that no longer existed anywhere. Binding it with `contract CounterfactualDepositFactory is ICounterfactualDepositFactory` means the compiler now enforces conformance, so it can't rot silently again. Good fix.

I spot-checked the new interface text against the implementation rather than just reading it:
- "Reverts if it is already deployed" — correct, `new BeaconProxy{salt: salt}` reverts on CREATE2 collision.
- "`CounterfactualDepositFactoryTron` … override prediction only" — correct, it overrides only `_computeProxyAddress`.
- `deployIfNeededAndExecute` idempotency — correct.

**One nit, non-blocking.** You made `CounterfactualDeposit.sol` fully `@inheritdoc`-driven, but `CounterfactualDepositFactory.sol` only uses it for `BEACON`. `predictAddress`, `deploy`, `deployAndExecute` and `deployIfNeededAndExecute` still carry duplicated `@notice` text — and it has *already* diverged from the interface wording:

> interface: "Deploy the proxy for `salt` and `initialRoot`, already initialized and resolving its implementation live from the beacon."
> contract: "Deploy the proxy for `salt` and `initialRoot` (already initialized + always-current via the beacon)."

Two copies of the same doc drifting apart is the exact failure mode this commit is fixing. Switching those four to `@inheritdoc` is doc-only and can't touch bytecode. Also a small asymmetry: `ICounterfactualDeposit.BEACON()` returns `ICounterfactualBeaconBase` while `ICounterfactualDepositFactory.BEACON()` returns `address` — each matches its own impl so nothing's wrong, just inconsistent between siblings.

Happy to push that `@inheritdoc` cleanup if you want it — say the word and I'll commit it to the branch and re-run the byte comparison to confirm it's still a no-op. Left the branch untouched otherwise; my approval from the earlier review still stands.

droplet-rl:
  • Full default-profile build
  • Wait for solc to finish then show build output

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=bb4743a4da32f6ae2131bc8060249b36&trace_start_ts=1787197073&trace_end_ts=1787283773)

contracts/interfaces/ICounterfactualDepositFactory.sol
```diff
@@ -3,80 +3,59 @@ pragma solidity ^0.8.0;
...
-     * @param salt Unique salt for address generation.
-     * @return Predicted address.
-     */
-    function predictDepositAddress(
-        address counterfactualDepositImplementation,
-        bytes32 paramsHash,
-        bytes32 salt
-    ) external view returns (address);
+    /// @notice The beacon (the `CounterfactualBeacon`) every deployed proxy points at.
+    function BEACON() external view returns (address);
```
fusmanii, line 21: Return type of this `BEACON()` is different than contracts/interfaces/ICounterfactualDeposit.sol

Is this intentional?

fusmanii (commented):
Looks good, just one Q

fusmanii (approved):

PR #1535, opened 2026-08-27
feat(svm): add V5 adapter foundations

Reinis-FRP:
## Summary

- freeze the versioned Borsh input and JIT wire boundary for Gateway-facing V5 deposits and fills
- define deposit identity, signed parameter modification, amount resolution, PDA, and account-validation foundations
- add a compatibility specification and shared golden vectors verified from Rust, TypeScript, and Solidity

Linear: ACP-184

Stacked on #1495.

## Testing

- `cargo test -p svm-spoke --lib` (9 passed)
- `yarn prettier --check programs/svm-spoke/V5_ADAPTER_SPEC.md programs/svm-spoke/fixtures/v5_adapter_v1.json test/evm/foundry/local/SvmSpokeV5Vectors.t.sol test/svm/SvmSpoke.V5Foundations.ts`
- `yarn ts-mocha -p ./tsconfig.json -t 1000000 test/svm/SvmSpoke.V5Foundations.ts` (3 passed)
- `yarn test-evm-foundry -- --match-contract SvmSpokeV5VectorsTest` (4 passed)

## Scope

This PR establishes Step 1 foundations only. It does not expose a live deposit or fill entrypoint.

---


programs/svm-spoke/src/v5.rs
```diff
@@ -0,0 +1,569 @@
...
+    pub authority: [u8; 20],
+    pub allow_output_amount: bool,
+    pub allow_exclusive_relayer: bool,
+}
+
+impl V5DepositModificationRules {
+    pub fn validate(&self) -> Result<()> {
+        let has_authority = self.authority != [0u8; 20];
+        let has_permission = self.allow_output_amount || self.allow_exclusive_relayer;
+        require!(has_authority == has_permission, V5Error::InvalidParamModificationRules);
```
droplet-rl, line 73: **Deliberate divergence from the EVM adapter — please document it here and in the spec.**

This rejects `authority == 0` with a permission flag set. `contracts-v5/src/adapters/AcrossDepositDelegateAdapter.sol` accepts exactly that shape:

```solidity
if (uint256(params.paramModificationRules) != 0) {      // whole word, not just the authority
    _applyParamModifications(...);
}
// ...and inside:
if (authority != address(0)) { /* verify signature */ }  // skipped entirely when authority == 0
if (allowAmountOut) { /* improvement-only bump, no signature required */ }
```

The EVM struct comment says "the signing authority **if one is required**. Zero disables JIT entirely" — zero *word*, not zero authority. So permissionless improvement-only `outputAmount` JIT is a valid EVM path shape, and this rejects it with `InvalidParamModificationRules`.

Failing closed is the right default and I'd keep the stricter rule. But `V5_ADAPTER_SPEC.md:54` states the SVM rule ("it never means permissionless modification") as if it were the shared rule, so a path builder reading the spec has no way to learn EVM differs. Since this PR is the freeze, please add an explicit "diverges from `AcrossDepositDelegateAdapter.sol`, which permits authority-less JIT" note — and confirm the integrator side never emits that shape for SVM.

programs/svm-spoke/src/v5.rs
```diff
@@ -0,0 +1,569 @@
...
+
+/// Borsh enum discriminants are frozen as Deposit=0 and Fill=1.
+#[derive(AnchorSerialize, AnchorDeserialize, Clone)]
+pub enum V5AdapterMode {
+    Deposit(AcrossDepositInput),
+    Fill(V5FillInput),
+}
+
+/// Literal uses the committed `input_amount`. Balance-relative mode resolves `bips` of the canonical Gateway input
+/// vault's live token amount, rounded down, and later enforces the committed amount as a floor.
```
droplet-rl: **This doc reads as execution-scoped; the account it resolves against is not.**

The EVM analog resolves `BalanceSub` against `address(this)` — under `ADAPTER_DELEGATECALL` that's the Executor, a per-execution context. Here "the canonical Gateway input vault" is `ATA(PDA(["vault_authority"], gateway), mint)`, which `solana-v5/AGENTS.md` describes as:

> Vaults are shared per mint rather than isolated per execution. Any residual balance or stale approval left between executions may be consumed by a later permissionless tape. Executions must therefore leave no residual vault balance; **the gateway does not yet enforce net-zero settlement.**

So `bips` of the *live* balance is only equivalent to the EVM semantics under an invariant the Gateway explicitly does not enforce. Combined with the spec's `["v5_source_delegate"]` allowance rule ("may grant any allowance at least the resolved amount, including `u64::MAX`"), `bips = 10_000` over a vault with residual balance sweeps whatever is sitting there.

The math here is correct and I'm not asking for a code change — but this is the most load-bearing unstated assumption in the frozen surface. Please cross-reference the Gateway's net-zero-settlement invariant in both this doc comment and `V5_ADAPTER_SPEC.md:85`, so Step 2 doesn't have to rediscover it.

programs/svm-spoke/src/v5.rs
```diff
@@ -0,0 +1,569 @@
...
+fn decode_strict<T: AnchorDeserialize>(data: &[u8]) -> Result<T> {
+    let mut remaining = data;
+    let decoded = T::deserialize(&mut remaining).map_err(|_| error!(V5Error::InvalidWireFormat))?;
+    require!(remaining.is_empty(), V5Error::InvalidWireFormat);
+    Ok(decoded)
+}
+
+pub fn decode_v5_adapter_input(data: &[u8]) -> Result<V5AdapterInput> {
+    let input: V5AdapterInput = decode_strict(data)?;
+    require_eq!(input.version, V5_ADAPTER_WIRE_VERSION, V5Error::UnsupportedVersion);
```
droplet-rl: **Version is checked after the whole body is deserialized, which inverts the error for the case this exists to handle.**

`decode_strict::<V5AdapterInput>` parses everything first, so a future v2 payload with a changed layout almost certainly fails inside Borsh and surfaces `InvalidWireFormat` — never reaching this `require_eq!`. `UnsupportedVersion` will realistically only fire for a v2 payload that happens to be layout-compatible with v1, which is the least interesting case.

For a versioned boundary whose job is graceful rejection of unknown versions, peel the leading byte first:

```rust
let (&version, body) = data.split_first().ok_or_else(|| error!(V5Error::InvalidWireFormat))?;
require_eq!(version, V5_ADAPTER_WIRE_VERSION, V5Error::UnsupportedVersion);
```

The on-wire bytes are unchanged, so this costs nothing at the freeze — it just makes the error honest.

droplet-rl (commented):
## Summary

Solid foundations PR. I reproduced every test claim locally and independently verified the frozen ABI constants against their upstream definitions rather than just against the fixture.

**Verified green:**
- `cargo test -p svm-spoke v5::tests --lib` — 7 passed
- `yarn ts-mocha -p ./tsconfig.json -t 1000000 test/svm/SvmSpoke.V5Foundations.ts` — 3 passed
- `yarn test-evm-foundry -- --match-contract SvmSpokeV5VectorsTest` — 4 passed

(`cargo +nightly fmt` and `cargo clippy` I could not run — neither component is installed in this environment.)

**Cross-repo conformance I checked by hand** against `across-protocol/solana-v5@programs/gateway`:

| Frozen constant | Upstream | Match |
|---|---|---|
| `GATEWAY_PROGRAM_ID` | `gateway/src/lib.rs` `declare_id!` | ✅ |
| `GATEWAY_ADAPTER_EXECUTE_V5_PREIMAGE` | `ADAPTER_EXECUTE_ACROSS_V5_PREIMAGE` | ✅ |
| `GATEWAY_DISPATCH_AUTHORITY_SEED` + `[seed, callee_program_id]` | `DISPATCH_AUTHORITY_SEED`, `execute_via_adapter` | ✅ |
| `GATEWAY_VAULT_AUTHORITY_SEED` / `["vault_authority"]` | `VAULT_AUTHORITY_SEED`, `execute.rs` | ✅ |
| `V5GatewayContext` field order/widths | `dispatch::CtxValues` | ✅ |
| dispatch layout `disc(8) ‖ ctx ‖ len(4) ‖ input ‖ len(4) ‖ jit` | `encode_dispatch_data` | ✅ |

And against `across-protocol/contracts-v5@src/adapters/AcrossDepositDelegateAdapter.sol`: `AcrossDepositParams` / `AcrossDepositInput` / `AcrossDepositJitParams` field order matches, the `paramModificationDigest` layout matches, the improvement-only + ignore-unpermitted semantics match, and the low-s / v∈{27,28} rules match what OZ `ECDSA.recover` enforces. `find_v5_account`'s search-by-derived-key contract also lines up with the Gateway's forwarding loop (dedups keys, preserves committed writability), so the design holds.

Nice work — the wire encoding is right, the strictness is right, and the fixture is genuinely useful.

## The one thing I'd like an answer on before this is "frozen"

`V5DepositModificationRules::validate()` rejects `authority == 0 && any permission set`. On EVM that is a **supported configuration**: `adapterDelegateExecuteAcrossV5` decodes JIT whenever the whole `paramModificationRules` word is nonzero, and `_applyParamModifications` only verifies a signature `if (authority != address(0))` — the struct comment even reads "the signing authority *if one is required*". So EVM supports permissionless improvement-only JIT; this SVM surface rejects it. It fails closed, so it isn't a vulnerability, but it means a cross-VM path builder can emit rules that are valid on EVM and revert on SVM. If intentional, please say so explicitly in `V5_ADAPTER_SPEC.md` and name the EVM behavior being deviated from. Details inline.

## Other themes

- **The Rust golden-vector tests never run in CI.** `pr.yml` runs `yarn lint-rust` but no `cargo test`. The TS and Solidity halves run; the Rust half — the primary guard on this surface — does not.
- **The load-bearing invariant behind `InputVaultBalance` is unstated.** The EVM analog resolves against a per-execution context; the SVM one resolves against a vault that `solana-v5/AGENTS.md` documents as *"shared per mint rather than isolated per execution ... the gateway does not yet enforce net-zero settlement."*
- **The three-language verification proves less than the PR description implies** — all three tests re-derive the same formulas from the same fixture.
- **CU on the hot path** — several `find_program_address` grinds that the in-repo Gateway adapters avoid by persisting bumps at init.

## Question not covered inline

EVM's `inputAmountParam` carries an `INPUT_PARAM_SET_CALL_VALUE` flag for native deposits. `V5InputAmountMode` has no analog, and this repo does ship a `nativeDeposit` path. Is native-SOL V5 deposit deliberately out of wire v1? Adding a Borsh variant later is additive, but it would sit awkwardly against a `version: u8 = 1` freeze — worth one line in the spec either way.

Not blocking — nothing here is callable yet — but the divergence and the CI gap are cheapest to fix now, before the behavior steps build on top.

programs/svm-spoke/src/v5.rs
```diff
@@ -0,0 +1,569 @@
...
+pub fn derive_v5_fill_payer(submitter: &Pubkey) -> (Pubkey, u8) {
+    Pubkey::find_program_address(&[V5_FILL_PAYER_SEED, submitter.as_ref()], &ID)
+}
+
+pub fn derive_fill_status(relay_hash: &[u8; 32]) -> (Pubkey, u8) {
+    Pubkey::find_program_address(&[FILL_STATUS_SEED, relay_hash], &ID)
+}
+
+pub fn require_gateway_dispatch_authority(account: &AccountInfo) -> Result<()> {
+    let (expected, _) = derive_gateway_dispatch_authority();
```
droplet-rl, line 325: **`find_program_address` on every dispatch, plus a misleading error.**

Two things:

1. This grinds the full derivation each call (~1.5k–15k CU depending on bump). The Gateway ships the callee-side check as `gateway::dispatch::is_dispatch_authority`, which takes a stored canonical bump and uses `create_program_address` — and its own in-repo adapters (`prefunded-adapter`, `cctp-mint-adapter`, `cctp-burn-adapter`) all persist `dispatch_authority_bump` in config at init for exactly this reason. Hand-rolling the check is fine and sanctioned here (`mock-executor`/`mock-planner` do the same, and the "no cross-repo Rust dependency" comment in `constants.rs` justifies it), but the bump caching is worth carrying over. Same applies to `derive_v5_source_delegate` / `derive_v5_fill_delegate` / `derive_v5_fill_payer`, which will all run per deposit/fill.

2. `V5Error::MissingAccount` ("Missing required Across V5 account!") also fires when the account *is* present but isn't a signer or has the wrong key. That's the more likely failure in practice and the message points debugging in the wrong direction. Worth a distinct variant.

programs/svm-spoke/Cargo.toml
```diff
@@ -22,4 +22,9 @@ test = []
 anchor-lang = { version = "0.31.1", features = ["init-if-needed","event-cpi"]}
 anchor-spl = "0.31.1"
 multicall-handler = { path = "../multicall-handler" }
-solana-security-txt = "1.1.1"
\ No newline at end of file
+solana-security-txt = "1.1.1"
+
+[dev-dependencies]
```
droplet-rl, line 27: **These dev-deps back 7 tests that CI will never run.**

`.github/workflows/pr.yml` has `lint-rust` (`cargo +nightly fmt --all -- --check && cargo clippy`) but no `cargo test` anywhere in the file. The TS vectors run via `anchor test` and the Solidity vectors via `test-evm`, so two of the three languages are covered — but the Rust half, which is the one that actually exercises `decode_v5_adapter_input`, `resolve_v5_deposit_modifications`, and the PDA derivations, is unguarded.

Given the whole point of this PR is a frozen surface protected by golden vectors, that's the gap most likely to let drift through. A step in the existing `lint-and-check-generated` job (it already installs a Rust toolchain and warms the cargo cache) would cover it:

```yaml
- name: Test rust
  run: cargo test -p svm-spoke --lib
```

test/evm/foundry/local/SvmSpokeV5Vectors.t.sol
```diff
@@ -0,0 +1,73 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.0;
+
+import { Test } from "forge-std/Test.sol";
+import { ECDSA } from "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
+
+/// @notice EVM-side conformance checks for the SVM V5 adapter's cross-VM hash and signature fixtures.
+contract SvmSpokeV5VectorsTest is Test {
```
droplet-rl, line 8: **Worth being precise about what this proves.**

This re-implements the *SVM* formulas in Solidity against the shared fixture. It doesn't call `contracts-v5`'s `AcrossDepositDelegateAdapter.paramModificationDigest`, and `testDepositIdVector` doesn't mirror this repo's own `SpokePool.getUnsafeDepositId` (`contracts/spoke-pools/SpokePool.sol:1289`) — which packs a 20-byte `msg.sender`, not a 32-byte program id. So a green run means "keccak and sha256 agree across three languages," not "the SVM surface matches the EVM adapter."

That's still worth having — it catches endianness and packing slips in the fixture. But the PR description's "golden vectors verified from Rust, TypeScript, and Solidity" reads like cross-VM conformance, and someone will later treat green CI as exactly that.

For what it's worth I did the comparison by hand and the digest layout **does** match EVM (`abi.encodePacked(domain, pathId, uint256 depositNonce, uint256 newOutputAmount, bytes32 newExclusiveRelayer)`); the `depositId` preimage necessarily differs because of the address widths. Two suggestions, either is fine: note in `V5_ADAPTER_SPEC.md` that these are self-consistency vectors rather than EVM conformance vectors, or record the corresponding EVM-side values in the fixture so the divergence is visible as data.

programs/svm-spoke/src/error.rs
```diff
@@ -80,6 +80,29 @@ pub enum SvmError {
     InvalidDelegatePda,
 }
 
+// Across V5 adapter specific errors.
+#[error_code]
```
droplet-rl: **Cheap moment to give this enum its own offset.**

Anchor defaults every `#[error_code]` enum to 6000, so `V5Error::InvalidWireFormat` == 6000 == `CommonError::InvalidQuoteTimestamp` == `SvmError::NotOwner` == `CallDataError::InvalidSelector`. Clients decoding by numeric code can't tell them apart, and the generated IDL will carry duplicate codes.

That's a pre-existing pattern in this file, so not something this PR broke — but `V5Error` is brand new with no clients depending on its codes yet, which makes it the one enum here that can still be fixed for free:

```rust
#[error_code(offset = 1000)]  // 7000..
pub enum V5Error {
```

Worth doing before Step 2 starts surfacing these errors to the Gateway and off-chain tooling.

programs/svm-spoke/src/constants.rs
```diff
@@ -9,6 +9,20 @@ pub const MESSAGE_TRANSMITTER_PROGRAM_ID: Pubkey = pubkey!("CCTPmbSD7gX1bxKPAmg7
 pub const MAX_EXCLUSIVITY_PERIOD_SECONDS: u32 = 31_536_000;
 
 pub const ZERO_DEPOSIT_ID: [u8; 32] = [0u8; 32];
+pub const BIPS_DENOMINATOR: u16 = 10_000;
+
+pub const V5_ADAPTER_WIRE_VERSION: u8 = 1;
+pub const V5_SOURCE_DELEGATE_SEED: &[u8] = b"v5_source_delegate";
+pub const V5_FILL_DELEGATE_SEED: &[u8] = b"v5_fill_delegate";
+pub const V5_FILL_PAYER_SEED: &[u8] = b"v5_fill_payer";
+pub const FILL_STATUS_SEED: &[u8] = b"fills";
```
droplet-rl, line 18: The constant is introduced but the three existing call sites still use the bare literal — `instructions/fill.rs:70`, `instructions/slow_fill.rs:36`, `instructions/slow_fill.rs:167`.

Right now there are two sources of truth for the same replay namespace, which is the opposite of what extracting the constant was for. Anchor's `seeds = [...]` accepts a const, so they can all take `FILL_STATUS_SEED` in this PR.

programs/svm-spoke/src/v5.rs
```diff
@@ -0,0 +1,569 @@
...
+        );
+        assert_eq!(digest, array(&fixture, "/jit/digest"));
+
+        let authority = array(&fixture, "/jit/authority");
+        let signature = array(&fixture, "/jit/signature");
+        verify_v5_authority(&authority, &digest, &signature).unwrap();
+        assert!(recover_v5_authority(&digest, &array(&fixture, "/jit/highSSignature")).is_err());
+        let mut invalid_v = signature;
+        invalid_v[64] = 0;
+        assert!(recover_v5_authority(&digest, &invalid_v).is_err());
```
droplet-rl, line 479: **Missing the negative test that matters most for the JIT boundary.**

Current coverage is high-s rejection and `v = 0` rejection — both malleability/encoding checks. What isn't covered is domain separation, which is the actual security property of the digest:

- a signature valid for `path_id` A rejected under `path_id` B
- a signature valid for `deposit_nonce` N rejected under N+1
- a well-formed signature from a *different* signer rejected against the committed authority

All three are a couple of lines each given `libsecp256k1` is already a dev-dep (the test at line 559 shows the pattern). The first is the one that would actually catch a regression if someone later reorders or drops a field in `v5_param_modification_digest` — a field-order swap would keep every existing test green.

programs/svm-spoke/src/v5.rs
```diff
@@ -0,0 +1,569 @@
...
+
+        deposit.modification_rules.allow_exclusive_relayer = false;
+        assert_eq!(
+            resolve_v5_deposit_modifications(&deposit, &jit, &gateway, &path_id)
+                .unwrap()
+                .1,
+            deposit.deposit_params.exclusive_relayer
+        );
+
+        deposit.modification_rules.allow_output_amount = true;
```
droplet-rl: This assignment is a no-op — the fixture already commits `allow_output_amount = true` (`wire.depositInput` ends `...0101`), and nothing between here and line 541 clears it.

Harmless, but it reads as "now enable output-amount modification," which makes the improvement-rule assertion below look like it's testing a state transition it isn't. Either drop the line, or set it to `false` first so the flip is real.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=c92722eabd148dcb8d2b4df2f9fb7393&trace_start_ts=1787832919&trace_end_ts=1787919619)

programs/svm-spoke/src/v5.rs
Reinis-FRP: Documented the intentional divergence in both `V5DepositModificationRules::validate()` and `V5_ADAPTER_SPEC.md`: zero authority disables JIT on SVM v1, while `AcrossDepositDelegateAdapter` permits authority-less JIT when a permission flag is set. The spec now states that SVM route builders must not emit that EVM-only shape. Step 1 does not include a live SVM route builder, so this is recorded as an explicit integration requirement rather than claiming an enforcement point that is not in this PR.

---

programs/svm-spoke/src/v5.rs
Reinis-FRP: Documented this invariant at both sites. The enum comment and spec now state that Gateway vaults are shared per mint, `InputVaultBalance` resolves against shared live state, the continuing tape must clear residual state, and Gateway does not currently enforce net-zero settlement. The spec also calls out stale approval explicitly.

---

programs/svm-spoke/src/v5.rs
Reinis-FRP: Fixed. The decoder now peels and validates the leading version byte before Borsh-decoding the mode body. Exact-error tests cover a full wrong-version payload and a one-byte future-version payload returning `UnsupportedVersion`, while empty input returns `InvalidWireFormat`.

---

programs/svm-spoke/src/v5.rs
Reinis-FRP: Added a dedicated `InvalidDispatchAuthority` error for the wrong-key/non-signer cases.

I have not added bump persistence in this foundations step: it exposes no callable V5 entrypoint and has no V5 config account in which to store the fixed dispatch bump. The submitter- and relay-derived PDAs also cannot share one cached bump. The derivation helpers continue to return their bumps so the behavior step can choose persisted or caller-provided bumps once the actual account topology and hot paths exist.

---

programs/svm-spoke/Cargo.toml
Reinis-FRP: Added `cargo test -p svm-spoke --lib` to the existing `lint-and-check-generated` job, immediately after Rust lint. The expanded local suite now passes 9 tests.

---

test/evm/foundry/local/SvmSpokeV5Vectors.t.sol
Reinis-FRP: Clarified this in `V5_ADAPTER_SPEC.md`: these are cross-language self-consistency vectors, not an invocation of the EVM adapter. The spec now records that the JIT digest layout matches `AcrossDepositDelegateAdapter`, while the deposit-ID preimage necessarily differs because SVM commits a 32-byte executor program ID and EVM commits a 20-byte caller address.

---

programs/svm-spoke/src/error.rs
Reinis-FRP: Added a dedicated V5 error range and a regression assertion that `InvalidWireFormat == 7000`. I used `#[error_code(offset = 7000)]`: in Anchor 0.31 an explicit offset replaces the default 6000 base rather than being added to it, so `offset = 1000` would generate code 1000.

---

programs/svm-spoke/src/constants.rs
Reinis-FRP: Fixed all three existing account constraints to use `FILL_STATUS_SEED`; there are no remaining `b"fills"` seed literals in those fill/slow-fill call sites.

---

programs/svm-spoke/src/v5.rs
Reinis-FRP: Added all three domain-separation negatives: the golden signature is rejected for a changed path ID, nonce + 1, and a different committed authority.

---

programs/svm-spoke/src/v5.rs
Reinis-FRP: Removed the no-op assignment so the improvement-rule assertion no longer suggests a permission transition that did not occur.

---

programs/svm-spoke/src/v5.rs
```diff
@@ -0,0 +1,612 @@
...
+    fn assert_error_name<T>(result: Result<T>, expected: &str) {
+        match result.err().unwrap() {
+            anchor_lang::error::Error::AnchorError(error) => assert_eq!(error.error_name, expected),
+            _ => panic!("expected Anchor error"),
+        }
+    }
+
+    #[test]
+    fn v5_errors_use_dedicated_range() {
+        assert_eq!(u32::from(V5Error::InvalidWireFormat), 7_000);
```
droplet-rl, line 377: Optional: this pins the base offset but not the variant ordering, and the ordering is the part that just moved.

`InvalidDispatchAuthority` was inserted between `MissingAccount` and `InvalidAccountMutability` in this very commit, which shifted the four variants after it by +1 (`InvalidAccountMutability` 7005→7006, ... `ParamModificationNotAnImprovement` 7008→7009). Harmless now — `V5Error` is unreleased and nothing decodes these yet — but once Step 2 ships and off-chain tooling starts matching on numeric codes, the same mid-enum insertion becomes a silent breaking change that this assertion would not catch.

Pinning both ends costs one line and turns any future reordering into a failing test rather than a client-side mystery:

```rust
assert_eq!(u32::from(V5Error::InvalidWireFormat), 7_000);
assert_eq!(u32::from(V5Error::ParamModificationNotAnImprovement), 7_009);
```

droplet-rl (approved):
## Re-review of `2414952b` — approving

All ten threads from the previous round are addressed: nine fixed, one deferred with a rationale I agree with. I re-verified the delta rather than taking the replies at face value.

**Reproduced:** `cargo test -p svm-spoke --lib` → **9 passed**, matching the PR description (7 prior v5 tests + `v5_errors_use_dedicated_range` + the pre-existing `test_id`).

### Verification of each fix

| Item | Status | How I checked |
|---|---|---|
| Permissionless-JIT divergence | ✅ | Documented at `validate()` and in the spec, with the explicit "route builders must not emit that EVM-only rule shape" requirement |
| Shared-vault invariant | ✅ | Both sites now state it; matches `solana-v5/AGENTS.md` verbatim, incl. stale approvals |
| Version-before-decode | ✅ | See below — wire-identical, and proven by the surviving round-trip assertion |
| `InvalidDispatchAuthority` | ✅ | Distinct variant; bump caching deferred (see below) |
| `cargo test` in CI | ✅ | Added to `lint-and-check-generated`, which already compiles the crate via `cargo clippy` and resolves `declare_program!` from the committed `idls/` — so it will work without downloading SVM artifacts |
| Vector-scope wording | ✅ | Spec now says "cross-language self-consistency vectors, not an invocation of the EVM adapter" and records the 32-byte-vs-20-byte deposit-ID divergence |
| Error offset | ✅ | `offset = 7000` + regression test |
| `FILL_STATUS_SEED` | ✅ | All three sites; IDL-safe, see below |
| Domain-separation tests | ✅ | path_id, nonce+1, and wrong-authority negatives all present |
| No-op test line | ✅ | Removed; `allow_output_amount` is still true from the fixture at that point, so the improvement assertion is unchanged |

**You were right and I was wrong on the error offset.** My suggestion of `#[error_code(offset = 1000)]  // 7000..` assumed the offset was added to the 6000 base. It replaces it — `anchor-attribute-error` substitutes the supplied value for `ERROR_CODE_OFFSET` outright. `offset = 7000` is correct, and the regression assertion is a good way to pin it.

**The new decoder is wire-identical.** Borsh lays `V5AdapterInput` out as `version: u8` immediately followed by `mode`, so peeling `data[0]` and decoding the remainder as `V5AdapterMode` consumes exactly the same bytes. The golden `assert_eq!(serialize(&input), input_bytes)` round-trip still passing is the proof — a framing slip would have broken it. Empty input still maps to `InvalidWireFormat`, and both the truncated and full future-version payloads now report `UnsupportedVersion`.

**The `FILL_STATUS_SEED` refactor does not perturb the IDL** — worth stating explicitly, since these are the only three `seeds = [CONST, ..]` sites in the entire program (every other seed in `svm-spoke` is a bare byte-string literal), so there was no in-repo precedent to lean on. In `anchor-syn-0.31.1/src/idl/accounts.rs::parse_seed`, a `syn::Expr::Path` matching neither an instruction arg nor an account field falls through to `IdlSeed::Const { value: AsRef::<[u8]>::as_ref(&FILL_STATUS_SEED).into() }`, evaluated at IDL-build time. That yields the same five bytes the previous `Expr::Lit` arm produced, so the generated IDL, the Codama clients, and any PDA derivation downstream are unchanged.

**Deferring the dispatch-authority bump caching is the right call.** There is no V5 config account in Step 1 to persist a bump into, and inventing one before the account topology exists would be speculative. The derivation helpers returning their bumps keeps the option open, and `derive_v5_fill_payer` / `derive_fill_status` genuinely cannot share a cached bump anyway. Worth revisiting when the behavior step wires up a hot path.

### Not re-run

The fixture, `SvmSpoke.V5Foundations.ts`, and `SvmSpokeV5Vectors.t.sol` are untouched in this delta, so their previously verified results (3 and 4 passing) still hold. I could not run `cargo +nightly fmt`, `cargo clippy`, or `prettier` — none of the three are installed in this environment — so those checks rest on CI.

One optional nit inline about the scope of the new error-code assertion. Nothing blocking. Nice turnaround.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=4d20c07d4f8e5dad622ebcc1fc2475a3&trace_start_ts=1787836389&trace_end_ts=1787923089)

programs/svm-spoke/src/v5.rs
Reinis-FRP: Fixed in `176cf8d4`. The regression test now pins both ends of the initial V5 error range: `InvalidWireFormat == 7000` and `ParamModificationNotAnImprovement == 7009`. The full `svm-spoke` library suite remains green at 9/9.

---

programs/svm-spoke/src/constants.rs
```diff
@@ -9,6 +9,20 @@ pub const MESSAGE_TRANSMITTER_PROGRAM_ID: Pubkey = pubkey!("CCTPmbSD7gX1bxKPAmg7
...
+
+// Mirrored from the Gateway dispatch ABI. Keep these local: Gateway and svm-spoke intentionally use different
+// Anchor versions and must not acquire a cross-repository Rust dependency.
+pub const GATEWAY_PROGRAM_ID: Pubkey = pubkey!("34trBszXuqhRjWaMxXWsunJNmyUsBvDNPxAwTzbPTm4p");
+pub const GATEWAY_DISPATCH_AUTHORITY_SEED: &[u8] = b"dispatch_authority";
+pub const GATEWAY_VAULT_AUTHORITY_SEED: &[u8] = b"vault_authority";
+pub const GATEWAY_ADAPTER_EXECUTE_V5_PREIMAGE: &[u8] = b"global:adapter_execute_across_v5";
```
fusmanii: lets do the same optimization that we did in solana-v5 where we computed the first eight sha256 bytes of this

programs/svm-spoke/src/v5.rs
```diff
@@ -0,0 +1,609 @@
...
+pub fn derive_v5_fill_payer(submitter: &Pubkey) -> (Pubkey, u8) {
+    Pubkey::find_program_address(&[V5_FILL_PAYER_SEED, submitter.as_ref()], &ID)
+}
+
+pub fn derive_fill_status(relay_hash: &[u8; 32]) -> (Pubkey, u8) {
+    Pubkey::find_program_address(&[FILL_STATUS_SEED, relay_hash], &ID)
+}
+
+pub fn require_gateway_dispatch_authority(account: &AccountInfo) -> Result<()> {
+    let (expected, _) = derive_gateway_dispatch_authority();
```
fusmanii: would we be able to import `is_dispatch_authority` from solana-v5 (once its public?)

programs/svm-spoke/src/v5.rs
```diff
@@ -0,0 +1,609 @@
...
+fn decode_strict<T: AnchorDeserialize>(data: &[u8]) -> Result<T> {
+    let mut remaining = data;
+    let decoded = T::deserialize(&mut remaining).map_err(|_| error!(V5Error::InvalidWireFormat))?;
+    require!(remaining.is_empty(), V5Error::InvalidWireFormat);
+    Ok(decoded)
+}
+
+pub fn decode_v5_adapter_input(data: &[u8]) -> Result<V5AdapterInput> {
+    let (&version, body) = data.split_first().ok_or_else(|| error!(V5Error::InvalidWireFormat))?;
+    require_eq!(version, V5_ADAPTER_WIRE_VERSION, V5Error::UnsupportedVersion);
```
fusmanii, line 132: Why do we need this version? do you see a case where we would want to support multiple versions?

programs/svm-spoke/src/v5.rs
```diff
@@ -0,0 +1,609 @@
...
+}
+
+/// Literal uses the committed `input_amount`. Balance-relative mode resolves `bips` of the canonical Gateway input
+/// vault's live token amount, rounded down, and later enforces the committed amount as a floor. Gateway vaults are
+/// shared per mint, not isolated per execution, so the continuing tape must leave no residual balance; Gateway does
+/// not currently enforce that net-zero settlement invariant.
+#[derive(AnchorSerialize, AnchorDeserialize, Clone, Copy)]
+pub enum V5InputAmountMode {
+    Literal,
+    InputVaultBalance { bips: u16 },
```
fusmanii, line 55: How would `InputVaultBalance` be used? is the idea to only pull in a portion of the `input_amount`? can't we just use balance substitution in the Gateway for this?

programs/svm-spoke/src/v5.rs
```diff
@@ -0,0 +1,609 @@
...
+    pub authority: [u8; 20],
+    pub allow_output_amount: bool,
+    pub allow_exclusive_relayer: bool,
+}
+
+impl V5DepositModificationRules {
+    pub fn validate(&self) -> Result<()> {
+        let has_authority = self.authority != [0u8; 20];
+        let has_permission = self.allow_output_amount || self.allow_exclusive_relayer;
+        // SVM v1 intentionally requires an authority for any JIT permission. The EVM adapter permits authority-less
```
fusmanii: Why don't we want to support authority-less permission bits in SVM?

fusmanii (commented):
Did a first pass, mostly questions around different decisions?

programs/svm-spoke/src/constants.rs
Reinis-FRP: Fixed in [c0b90db6](https://github.com/across-protocol/contracts/commit/c0b90db6). The adapter discriminator is now the precomputed eight-byte constant matching solana-v5; the production runtime SHA-256 is gone, with a test-only preimage assertion retained.

---

programs/svm-spoke/src/v5.rs
Reinis-FRP: Addressed in [48ebb01d](https://github.com/across-protocol/contracts/commit/48ebb01d). Rather than add a cross-repository dependency across incompatible Anchor versions, the four static PDA address/bump pairs are now pinned locally and seed-verified in tests. Dispatch authentication is a direct signer/key check; only the submitter- and relay-dependent PDAs remain dynamic.

---

programs/svm-spoke/src/v5.rs
Reinis-FRP: Included this so a future program upgrade can continue decoding already-committed inputs while introducing a new wire format

programs/svm-spoke/src/v5.rs
```diff
@@ -0,0 +1,614 @@
+//! Frozen wire, cryptographic, and PDA foundations for the Gateway-facing V5 adapter.
+//!
+//! This module deliberately contains no live deposit or fill entrypoint. The versioned types and helpers are the
+//! Step 1 compatibility boundary consumed by the later behavior steps.
```
fusmanii: Lets remove these comments that mention that these PRs are split into steps

programs/svm-spoke/src/v5.rs
Reinis-FRP: Agreed—there was no SVM-specific reason for this restriction. This was my oversight when translating the EVM implementation. Fixed in [c6b4dd72](https://github.com/across-protocol/contracts/commit/c6b4dd72be7ca520a9dfc7f01724da31e429c4b6): SVM now enables JIT when any part of the modification rules is nonzero, verifies a signature only for a nonzero authority, and otherwise applies the committed permission bits permissionlessly, matching EVM. Tests cover the authority-only case and both authority-less permission paths.

---

programs/svm-spoke/src/v5.rs
Reinis-FRP: `InputVaultBalance` is the typed SVM equivalent of the EVM adapter's embedded `BalanceSub`: it resolves bips of the live input-token vault balance while preserving the committed `inputAmount` as an implicit floor. Gateway `BalanceSub` could overwrite the encoded amount, but matching this behavior would also require a separate `BALANCE_REQ`, and that also permits an arbitrary mint and offset, whereas this mode is deliberately bound to `input_token`, semantically mirroring the EVM behavior.

programs/svm-spoke/src/v5.rs
```diff
@@ -0,0 +1,632 @@
...
+/// Verify and apply only the committed JIT permissions. Output-amount changes are improvement-only. When an authority
+/// is configured, all proposed values remain signature-bound even if unpermitted; otherwise the permitted changes are
+/// permissionless, matching the EVM adapter.
+pub fn resolve_v5_deposit_modifications(
+    input: &AcrossDepositInput,
+    jit: &AcrossDepositJitParams,
+    gateway_program_id: &Pubkey,
+    path_id: &[u8; 32],
+) -> Result<([u8; 32], Pubkey)> {
+    require!(input.modification_rules.jit_enabled(), V5Error::InvalidParamModificationRules);
```
fusmanii: Do we need to do this check here? There is no harm in letting jit data to be provided even if the modification flags are not set

programs/svm-spoke/src/v5.rs
```diff
@@ -0,0 +1,632 @@
...
+    let hashed = keccak::hash(public_key.to_bytes().as_ref()).to_bytes();
+    Ok(hashed[12..].try_into().unwrap())
+}
+
+pub fn verify_v5_authority(
+    expected_authority: &[u8; 20],
+    digest: &[u8; 32],
+    signature: &[u8; V5_SIGNATURE_LEN],
+) -> Result<()> {
+    require!(*expected_authority != [0u8; 20], V5Error::InvalidParamModificationRules);
```
fusmanii: Dont think we need this check either:
- `resolve_v5_deposit_modifications` below already checks to make sure `input.modification_rules.authority != [0u8; 20]`
- `secp256k1_recover` returns an error instead of zero key

I think in general we should try to reduce linting on contracts/programs as much as possible

programs/svm-spoke/src/v5.rs
```diff
@@ -0,0 +1,632 @@
...
+        path_id,
+        &nonce,
+        new_output_amount,
+        new_exclusive_relayer.as_ref(),
+    ])
+    .to_bytes()
+}
+
+/// Recover the 20-byte EVM authority from `r || s || v`. Only EVM-canonical v=27/28 and low-s signatures pass.
+pub fn recover_v5_authority(digest: &[u8; 32], signature: &[u8; V5_SIGNATURE_LEN]) -> Result<[u8; 20]> {
```
fusmanii: good stuff, what do you think about splitting this file into multiple files that each have different responsibilities? I am thinking maybe have a `v5/codec.rs`, `v5/jit.rs`, `v5/pda.rs`, `v5/tests.rs` all being in v5 folder. wdyt?

fusmanii (commented):
Looks good, couple points around the need for validation and also v5.rs file organization

programs/svm-spoke/src/v5.rs
```diff
@@ -0,0 +1,632 @@
...
+    0x73, 0x57, 0xa4, 0x50, 0x1d, 0xdf, 0xe9, 0x2f, 0x46, 0x68, 0x1b, 0x20, 0xa0,
+];
+
+/// Gateway-attested values prepended to every adapter call. Field order and widths mirror Gateway `CtxValues`.
+#[derive(AnchorSerialize, AnchorDeserialize, Clone, Copy)]
+pub struct V5GatewayContext {
+    pub step_id: [u8; 32],
+    pub path_id: [u8; 32],
+    pub submitter: Pubkey,
+}
```
grasphoper: This seems to be an external contract that Gateway is promising on the adapter call. In `solana-v5`, we're calling this `CtxValues`. Maybe we should change that name to something more explicit like `GatewayContextV1` and match that name here

I'm thinking about upgrade strategy for this. If we ever wanted to expand gateway context, it seems like `GatewayContextV2` + a new "call adapter branch" calling `adapter_execute_v2_across_v5` might be warranted.

programs/svm-spoke/src/v5.rs
```diff
@@ -0,0 +1,632 @@
...
+
+#[derive(AnchorSerialize, AnchorDeserialize, Clone)]
+pub struct V5AdapterInput {
+    pub version: u8,
+    pub mode: V5AdapterMode,
+}
```
grasphoper: Can we use something like
```
  #[derive(AnchorSerialize, AnchorDeserialize)]
  pub enum V5AdapterInput {
      DepositV1(DepositInputV1), // tag 0
      FillV1(FillInputV1),       // tag 1
  }
```
?

`enum` already implicitly "encodes" a tag for each variant on serialization (eg above, `DepositV1` carries tag 0)

Updating it would look like:
```
  #[derive(AnchorSerialize, AnchorDeserialize)]
  pub enum V5AdapterInput {
      DepositV1(DepositInputV1),
      FillV1(FillInputV1),      
      DepositV2(DepositInputV2),
  }
```

This also makes the shape of fill / deposit structs independent which I think is nice

programs/svm-spoke/src/v5.rs
```diff
@@ -0,0 +1,632 @@
...
+
+/// Destination acceptance bounds. Adapter mode requires `message` to be empty; retaining the field keeps the
+/// semantic V5FillInput shape explicit and makes a non-empty callback fail closed.
+#[derive(AnchorSerialize, AnchorDeserialize, Clone)]
+pub struct V5FillInput {
+    pub recipient: Pubkey,
+    pub output_token: Pubkey,
+    pub min_output_amount: u64,
+    pub message: Vec<u8>,
```
grasphoper: We reject `message` on fill-as-adapter always, I feel like we should remove it from here

grasphoper: Especially since we're going to retire the whole Across + Messsage flow anyway on Solana

programs/svm-spoke/src/v5.rs
```diff
@@ -0,0 +1,632 @@
...
+    pub repayment_chain_id: u64,
+    pub repayment_address: Pubkey,
+}
+
+fn decode_strict<T: AnchorDeserialize>(data: &[u8]) -> Result<T> {
+    let mut remaining = data;
+    let decoded = T::deserialize(&mut remaining).map_err(|_| error!(V5Error::InvalidWireFormat))?;
+    require!(remaining.is_empty(), V5Error::InvalidWireFormat);
+    Ok(decoded)
+}
```
grasphoper: Could we use `Borsh` here?

```
  fn decode_strict<T: AnchorDeserialize>(data: &[u8]) -> Result<T> {
      T::try_from_slice(data).map_err(|_| error!(V5Error::InvalidWireFormat))
  }
```

It also rejects trailing bytes

programs/svm-spoke/src/v5.rs
```diff
@@ -0,0 +1,632 @@
...
+
+pub fn decode_v5_adapter_input(data: &[u8]) -> Result<V5AdapterInput> {
+    let (&version, body) = data.split_first().ok_or_else(|| error!(V5Error::InvalidWireFormat))?;
+    require_eq!(version, V5_ADAPTER_WIRE_VERSION, V5Error::UnsupportedVersion);
+    let input = V5AdapterInput { version, mode: decode_strict(body)? };
+    match &input.mode {
+        V5AdapterMode::Deposit(deposit) => {
+            if let V5InputAmountMode::InputVaultBalance { bips } = deposit.input_amount_mode {
+                require!(bips <= BIPS_DENOMINATOR, V5Error::InvalidWireFormat);
```
grasphoper: We check `bips` both here and in amount resolution, perhaps we can do it in one place somewhere?

programs/svm-spoke/src/v5.rs
```diff
@@ -0,0 +1,632 @@
...
+}
+
+pub fn decode_v5_deposit_jit(deposit: &AcrossDepositInput, data: &[u8]) -> Result<Option<AcrossDepositJitParams>> {
+    if deposit.modification_rules.jit_enabled() {
+        Ok(Some(decode_strict(data)?))
+    } else {
+        require!(data.is_empty(), V5Error::InvalidWireFormat);
+        Ok(None)
+    }
+}
```
grasphoper: It seems like it's harmless to drop `.jit_enabled()` check from the contracts side. The deposit function can choose whether to decode jit data at all and will just drop the bytes if they're not needed in light of rules provided by input.

We then can also drop both these decoding functions and inline them as ~~ `decode_strict::<V5FillJit>(data)?;`

programs/svm-spoke/src/v5.rs
```diff
@@ -0,0 +1,632 @@
...
+    depositor: &Pubkey,
+    deposit_nonce: u64,
+) -> [u8; 32] {
+    let nonce = u64_to_evm_uint(deposit_nonce);
+    let synthetic_nonce = keccak::hashv(&[submitter.as_ref(), path_id, &nonce]).to_bytes();
+    keccak::hashv(&[executor_program_id.as_ref(), depositor.as_ref(), &synthetic_nonce]).to_bytes()
+}
+
+pub fn v5_param_modification_domain(gateway_program_id: &Pubkey) -> [u8; 32] {
+    let name_hash = keccak::hash(V5_PARAM_MODIFICATION_NAME).to_bytes();
```
grasphoper: could we make this hash a constant?

grasphoper (commented):
This PR makes sense to me overall. Left a few stylistic / struct composition comments

programs/svm-spoke/src/v5.rs
Reinis-FRP: Addressed in [aab5f329](https://github.com/across-protocol/contracts/commit/aab5f32999fa7997849ff4d3fb4f392faa06e5a5). Removed the standalone `V5_ADAPTER_WIRE_VERSION` and wrapper version field. `V5AdapterInput` now carries the action-specific schema version directly through `DepositV1(...)` and `FillV1(...)`; safe old variants can remain during an in-flight drain window, while unsafe variants can be rejected. Gateway-context changes are versioned separately through a new dispatch ABI and entrypoint.

---

programs/svm-spoke/src/v5.rs
Reinis-FRP: Addressed in aab5f329. Renamed the local context to `GatewayContextV1` and documented that input-variant versions do not cover the outer dispatch ABI: a context-layout change requires a new Gateway dispatch branch and adapter entrypoint. The comment describes the ABI without referring to the current upstream Rust type name, so a corresponding `solana-v5` rename will not make it stale.

---

programs/svm-spoke/src/v5.rs
Reinis-FRP: Addressed in aab5f329. Replaced the `version + mode` wrapper with `V5AdapterInput::{DepositV1(AcrossDepositInput), FillV1(V5FillInput)}`. The Borsh tags remain frozen at 0 and 1, future versions are append-only variants, and the regenerated Rust/TypeScript/Solidity vectors pass. Comments explicitly distinguish Across protocol V5 from SVM wire-schema V1.

---

programs/svm-spoke/src/v5.rs
Reinis-FRP: Addressed in ce29718e. Removed message from the SVM `V5FillInput` and regenerated the Rust and TypeScript golden wire fixture. The spec explicitly documents the intentional divergence from the EVM executor-mode shape because SVM adapter fills do not execute recipient callbacks.

---

programs/svm-spoke/src/v5.rs
Reinis-FRP: Addressed in c1a51c80. Removed the helper-side requirement that JIT be empty when the rules are zero. The behavior layer can now ignore jit_data in that case, matching EVM, and strictly decode it only when a rule is enabled.

---

programs/svm-spoke/src/v5.rs
Reinis-FRP: Addressed in c1a51c80. Removed the nonzero expected-authority check: the caller only invokes verification for a nonzero configured authority, while recovery failures or key mismatches return `InvalidParamModificationSignature`. The now-unused `InvalidParamModificationRules` error was removed as well.

---

programs/svm-spoke/src/v5.rs
Reinis-FRP: Addressed in c1a51c80. decode_strict now uses `T::try_from_slice`, preserving strict rejection of trailing bytes with less code.

---

programs/svm-spoke/src/v5.rs
Reinis-FRP: Addressed in c1a51c80. Removed BIPS validation from generic wire decoding. `resolve_v5_input_amount` is now the single authoritative semantic check, and the test confirms that syntactically valid input decodes before invalid BIPS are rejected at resolution.

---

programs/svm-spoke/src/v5.rs
Reinis-FRP: Addressed in c1a51c80. Removed both branch-specific JIT decoders and exposed the generic strict decoder crate-locally. On restack, the source-deposit implementation will decode only when the committed rules enable JIT and otherwise ignore the bytes; fills continue to decode `V5FillJit` strictly.

---

programs/svm-spoke/src/v5.rs
Reinis-FRP: Addressed in c1a51c80. Domain derivation now uses a precomputed `V5_PARAM_MODIFICATION_NAME_HASH`, with a test-only preimage assertion tying the constant to the documented string and golden fixture.

---

programs/svm-spoke/src/v5.rs
Reinis-FRP: Addressed in aab5f329. Removed the split-step wording from the module comments; the remaining documentation describes the compatibility and versioning boundaries without referring to PR rollout sequencing.

---

programs/svm-spoke/src/v5.rs
Reinis-FRP: Follow-up in c1a51c80: removing the now-unused InvalidParamModificationRules variant shifts the final initial V5 error to 7007. The regression test still pins both ends of the range, and there are no released consumers of these codes yet.

---

programs/svm-spoke/src/v5.rs
Reinis-FRP: Agreed—split this in 1465483a into `v5/codec.rs`, `v5/jit.rs`, `v5/pda.rs`, and `v5/tests.rs`, with `mod.rs` containing only module declarations. Callers use explicit responsibility-based paths; there are no broad compatibility re-exports. Wire layout, hashes/PDA derivations, and fixture coverage are unchanged.

---

programs/svm-spoke/V5_ADAPTER_SPEC.md
```diff
@@ -0,0 +1,130 @@
...
+
+`V5FillInput` contains `recipient[32] || output_token[32] || min_output_amount:u64_le`. Unlike the EVM executor-mode
+input, the SVM adapter wire intentionally omits a callback message because adapter fills do not execute recipient
+callbacks.
+
+`AcrossDepositInput` nests the canonical deposit fields under `deposit_params: AcrossDepositParams`, matching the EVM
+adapter's type boundary. Borsh serializes that fixed struct inline, so the nesting adds no bytes. All Rust fields
+serialize in declaration order. Integers use Borsh little-endian encoding. Pubkeys and `[u8; 32]` are raw 32-byte
+values. Vectors use a `u32_le` length. `input_amount_mode` is `Literal = 0` or
+`InputVaultBalance = 1 { bips: u16_le }`; `bips` must not exceed 10,000. The resolved SVM token amount is `u64`, while
```
droplet-rl: **The `bips` bound is documented here but no longer enforced here.**

This sentence sits in *Committed input and JIT wire*, so it reads as a decode-time constraint. After the consolidation in `c1a51c80` (Ihor's suggestion — correct), the only check is at `codec.rs:116`, inside `resolve_v5_input_amount`. `tests.rs:243-245` now explicitly asserts that `bips = 10001` **decodes successfully** and is rejected later at resolution.

So a path builder reading this section would expect `decode_v5_adapter_input` to reject an out-of-range `bips`, and it won't. Suggest dropping the clause here:

```suggestion
`InputVaultBalance = 1 { bips: u16_le }`. The resolved SVM token amount is `u64`, while
```

and adding it to the resolution paragraph below (around `:64-66`), phrased as resolution-time enforcement — something like: *"Amount resolution rejects `bips` greater than 10,000; the wire decoder does not."* That keeps the one-place-only property Ihor asked for while making the enforcement point explicit.

programs/svm-spoke/src/v5/codec.rs
```diff
@@ -0,0 +1,122 @@
...
+
+pub fn decode_v5_adapter_input(data: &[u8]) -> Result<V5AdapterInput> {
+    decode_strict(data)
+}
+
+pub fn resolve_v5_input_amount(mode: V5InputAmountMode, committed_amount: u64, vault_balance: u64) -> Result<u64> {
+    let amount = match mode {
+        V5InputAmountMode::Literal => committed_amount,
+        V5InputAmountMode::InputVaultBalance { bips } => {
+            require!(bips <= BIPS_DENOMINATOR, V5Error::InvalidWireFormat);
```
droplet-rl: **Prefer a dedicated semantic error over `InvalidWireFormat` here.**

By this point decoding has already succeeded — `InvalidWireFormat` ("Malformed Across V5 wire data!") is the one error message that is now guaranteed to be wrong when this fires. It's also the error the Gateway will surface, so it'll misdirect debugging.

Suggest a dedicated variant, e.g.:

```rust
#[msg("Across V5 input amount bips exceed the denominator!")]
InvalidInputAmountBips,
```

Two things worth getting right when adding it:

1. **Append it, don't insert it.** Anything inserted mid-enum shifts every later discriminant — the same silent-break risk that motivated the range test in the first place.
2. **`tests.rs:45` needs updating.** It currently pins `ParamModificationNotAnImprovement == 7_007` as the end of the range. Appending makes `InvalidInputAmountBips == 7_008`, so that assertion still passes but stops pinning the upper end. Move it to the new final variant to keep both ends pinned.

programs/svm-spoke/V5_ADAPTER_SPEC.md
```diff
@@ -0,0 +1,130 @@
...
+
+Unlike the EVM `inputAmountParam`, `DepositV1` has no set-call-value flag. Native SOL must first be wrapped by the
+ordinary Gateway `WRAP_SOL` command into its canonical WSOL vault; the deposit then consumes WSOL through the same
+token path as any SPL input. Direct lamport deposit from this adapter is outside `DepositV1`.
+
+Deposit JIT uses the EVM-aligned name `AcrossDepositJitParams`. It is decoded when the committed authority or either
+permission bit is nonzero and is the fixed 129 bytes
+`new_output_amount[32] || new_exclusive_relayer[32] || signature[65]`. A nonzero authority requires a valid signature;
+when authority is zero, enabled modifications are permissionless, matching the EVM `AcrossDepositDelegateAdapter`.
+When the authority and both permission bits are zero, `jit_data` is ignored rather than decoded. Fill mode always
```
droplet-rl: **These two sentences describe behavior this layer no longer implements.**

Both read as present-tense normative statements about the code in this PR:

- `:72-73` — "It is decoded when the committed authority or either permission bit is nonzero"
- `:76` — "When the authority and both permission bits are zero, `jit_data` is ignored rather than decoded"

After `c1a51c80` the foundations gate nothing: `decode_strict` is generic, and `jit_enabled()` (`codec.rs:48`) has **zero production call sites** — the only three references are in `tests.rs`. That's the right shape for a foundations PR and it matches EVM's `if (uint256(params.paramModificationRules) != 0)`, but as written the spec claims an enforcement point that doesn't exist yet, which is exactly the kind of thing the behavior step will read as already done.

Rephrasing as a handler requirement:

```suggestion
Deposit JIT uses the EVM-aligned name `AcrossDepositJitParams` and is the fixed 129 bytes
`new_output_amount[32] || new_exclusive_relayer[32] || signature[65]`. A nonzero authority requires a valid
signature; when authority is zero, enabled modifications are permissionless, matching the EVM
`AcrossDepositDelegateAdapter`. These foundations decode strictly and expose `jit_enabled()` but gate nothing
themselves: the deposit handler must decode `jit_data` only when the committed authority or either permission bit
is nonzero, and must ignore it entirely when all three are zero. Fill mode always
```

While you're in here, one optional addition: with `authority == 0` and `allow_exclusive_relayer` set, anyone can assign `exclusive_relayer` arbitrarily for the committed `exclusivity_parameter` window. That's EVM parity by design, not a bug — but the EVM side documents it (`AcrossDepositDelegateAdapter.sol:36` calls the field "arbitrary") and this spec doesn't. A sentence warning path builders off that shape when `exclusivity_parameter != 0` would carry the sharp edge across with the semantics.

programs/svm-spoke/V5_ADAPTER_SPEC.md
```diff
@@ -0,0 +1,130 @@
+# SVM SpokePool V5 adapter specification
+
+This document freezes the compatibility surface for the Gateway-facing `svm_spoke` V5 adapter. It intentionally
+describes foundations only: these schemas do not become callable until the source and destination behavior steps land.
```
droplet-rl: **Remaining rollout-step language.**

The module comments are clean after `aab5f329`, but two references to the PR sequencing survive here and at `:10`.

Line 4:

```suggestion
describes foundations only: these schemas do not become callable until the source and destination handlers land.
```

And `:10`, which currently reads "The frozen dispatch target for the later behavior steps is the single" — suggest:

> The frozen dispatch target is the single

Worth noting "step" is overloaded in this document: `step_id`, `dst_step_id`, and the Gateway execution tape's steps are all legitimate domain terms and should stay. It's only the rollout-sequencing sense that should go.

droplet-rl (commented):
Follow-up review of `1465483a` — the four documentation/ergonomics items agreed in Slack. All minor; none are blockers.

Context for anyone picking this up cold: I re-verified this revision independently rather than relying on the commit messages. `cargo test -p svm-spoke --lib` passes 10/10 locally and CI now runs it (`pr.yml:102`). I also re-derived the fixture's `syntheticNonce`, `depositId`, `domain`, and `digest` from scratch outside the repo, recovered the signer from the golden signature (`0xf39Fd6e5…`, matches the committed authority), confirmed `jit.highSSignature` is a genuine malleable counterpart (`s_low + s_high == N`, `v` flips `0x1c`→`0x1b`, same recovered address), checked `SECP256K1_HALF_ORDER == N/2`, and rebuilt the dispatch bytes from parts byte-for-byte. The JIT parity change in `c6b4dd72` was checked against the real `contracts-v5/src/adapters/AcrossDepositDelegateAdapter.sol` (`:99` gates on the whole rules word; the struct comment at `:36` documents `exclusiveRelayer` as deliberately *arbitrary*) — it matches. `V5_MAGIC_PREFIX` matches `V5Witness.sol` and `SpokePool.sol:189`, and is already covered by `SvmSpoke.Fill.ts:320`.

I withdrew a fifth item (suggesting `u64::try_from` instead of the `as u64` cast at `codec.rs:117`). Reinis pointed out the adjacent bound already proves the quotient fits, and the regression at `tests.rs:250` would catch its removal. I checked: with the bound deleted, `u64::MAX * 10001 / 10000` wraps to `1844674407370954`, which passes the `committed_amount = 0` floor and returns `Ok`, so the `is_err()` assertion fails. The test does pin the bound, and the fallible conversion would be redundant. Withdrawn.

programs/svm-spoke/V5_ADAPTER_SPEC.md
Reinis-FRP: Addressed in b6630544. Moved the `bips <= 10_000` constraint out of the wire-layout paragraph and documented that it is enforced during amount resolution, after syntactically valid Borsh decoding.

---

programs/svm-spoke/V5_ADAPTER_SPEC.md
Reinis-FRP: Addressed in b6630544. Removed the remaining rollout-step terminology while preserving protocol-domain terms such as `step_id` and `dst_step_id`. The spec now states directly that a callable surface requires handlers and names the frozen dispatch target without referring to PR sequencing.

---

programs/svm-spoke/V5_ADAPTER_SPEC.md
Reinis-FRP: Addressed in b6630544. Rephrased this as a handler obligation: the foundations expose strict decoding and `jit_enabled()` but do not gate JIT themselves; the deposit handler must decode or ignore `jit_data` from the committed rules. Also documented that zero authority plus `allow_exclusive_relayer` permits an arbitrary choice during the committed window and should be enabled by path builders only intentionally.

---

programs/svm-spoke/src/v5/codec.rs
Reinis-FRP: Addressed in b471bbd2. Added `V5Error::InvalidAmountBips` at the end of the V5 error range, matching the EVM `BalanceSubLib` name and resolution-time semantics. `resolve_v5_input_amount` now returns that error, the invalid-bips test asserts its exact name, and the range test pins the new upper end at 7008 without shifting existing variants.

---

droplet-rl (approved):
## Approving — `b6630544`

All four items from my previous review are correctly addressed, the frozen wire surface is untouched, and every CI gate passes locally.

**Verification at this head**

| Gate | Result |
|---|---|
| `cargo test -p svm-spoke --lib` | 10/10 pass |
| `cargo +nightly fmt --all -- --check` | exit 0 |
| `cargo clippy` | exit 0, **zero** warnings in `v5/` or `error.rs` |
| `generated/constants.json` regeneration | unaffected (networks/tokens only; no Rust error codes) |
| Spec line length vs `printWidth: 120` | max 119 |

**Per finding**

1. `V5_ADAPTER_SPEC.md` — the `bips` clause is out of the wire paragraph and restated at resolution as *"Amount resolution rejects `bips` greater than 10,000; the wire decoder does not."* Stating the negative explicitly is what closes the trap; a path builder can no longer read the wire section and expect decode-time rejection.
3. The JIT paragraph now reads as a handler requirement, and correctly says the foundations "gate nothing themselves". You also took the optional note — the permissionless `allow_exclusive_relayer` sharp edge is now documented with the `exclusivity_parameter` window called out, so the EVM parity decision carries its caveat forward.
4. `InvalidAmountBips` is **appended**, so `7000..7007` are unshifted and the new variant is `7008`; `tests.rs:45` moved to pin `7008`, keeping both ends of the range pinned. No checked-in IDL carries these codes (`idls/` holds only the external CCTP IDLs), so there's no generated-artifact drift.
5. Both rollout-step references are gone, and "handlers" is now used consistently with the #3 wording. `step_id` / `dst_step_id` / Gateway tape steps are correctly left alone.

**Two things worth recording**

You improved #4 past what I asked: the regression went from `.is_err()` to `assert_error_name(..., "InvalidAmountBips")`, which pins the specific error rather than merely *some* error. Because that reshaped the test that the declined `try_from` item depended on, I mutation-tested it — deleting the `bips` bound makes `resolve_v5_input_amount` return `Ok(1844674407370954)` (the wrapped `as u64` value), and the test now fails at `tests.rs:36` on the `err().unwrap()`. So the bound is still pinned, and the guarantee is strictly stronger than before. Your reasoning for declining that item holds at this head.

`fixtures/v5_adapter_v1.json`, `jit.rs`, `pda.rs`, and `constants.rs` are **byte-identical** to `1465483a`, and zero wire-definition lines were touched in `codec.rs` (the only change is the error name). The independent verification I did on the previous head therefore carries forward unchanged: re-derived `syntheticNonce` / `depositId` / `domain` / `digest` outside the repo, signer recovery from the golden signature, the malleable-counterpart check on `jit.highSSignature`, `SECP256K1_HALF_ORDER == N/2`, and the dispatch-byte reconstruction.

Nothing further from me. Note this PR is stacked on #1495 and targets `faisal/svm-spoke-v5`, so it needs that to land first.

grasphoper (approved):

PR #1536, opened 2026-08-28
feat(svm): add V5 source deposits

Reinis-FRP:
## Summary

- add the Gateway-authenticated `adapter_execute_across_v5` entrypoint with source `Deposit` enabled and `Fill` reserved fail-closed
- resolve literal or vault-balance input amounts, authenticate signed JIT changes, derive the V5 deposit ID, and reuse the shared deposit validation/transfer/event core
- resolve and validate dynamic mint, token-program, Gateway-vault, SpokePool-vault, and source-delegate accounts with static Anchor-equivalent checks
- add a test-only mock Gateway and source-deposit coverage for dynamic amounts, JIT domains, wrong accounts, insufficient allowance, malformed calls, events, and atomic rollback
- update the V5 adapter specification for the enabled source-deposit behavior

Part of ACP-184. This is Step 2 of the implementation plan and is stacked on #1535.

## Validation

- `git diff --check 5424fd77..HEAD`
- `cargo check -p svm-spoke`
- `cargo test -p svm-spoke --lib` (9 passed)
- direct `cargo build-sbf` for `svm-spoke`
- `yarn prettier --check programs/svm-spoke/V5_ADAPTER_SPEC.md test/svm/SvmSpoke.V5Source.ts`
- repository pre-commit hook

The focused validator suite could not be started locally because Anchor CLI 0.31.1 generates `--bind-address 0.0.0.0`, which the installed Solana validator 4.1.2 rejects. The SBF build itself passed; CI should run the validator test with the repository-supported toolchain.

---


programs/svm-spoke/src/instructions/v5_adapter.rs
```diff
@@ -0,0 +1,176 @@
...
+    let gateway_vault =
+        get_associated_token_address_with_program_id(&gateway_vault_authority, &input_token, &token_program_id);
+    let spoke_vault = get_associated_token_address_with_program_id(&state, &input_token, &token_program_id);
+    let (source_delegate, _) = derive_v5_source_delegate();
+    let gateway_vault_info = find_v5_account(remaining_accounts, &gateway_vault, true)?;
+    let spoke_vault_info = find_v5_account(remaining_accounts, &spoke_vault, true)?;
+    let source_delegate_info = find_v5_account(remaining_accounts, &source_delegate, false)?;
+
+    // Together with the canonical addresses above, these checks mirror the corresponding static mint and
+    // associated-token constraints. Delegate authorization and allowance are enforced by transfer_checked.
```
droplet-rl: **Blocking.** "Delegate authorization and allowance are enforced by `transfer_checked`" is true but leaves `require_v5_delegate_allowance` and `V5Error::InsufficientDelegateAllowance` (both added in #1535) as dead code, and it gives up a cheap invariant.

The concern is the *static* delegate. Everywhere else in this program the delegate PDA is `["delegate", derive_seed_hash(all deposit params)]`, so an approval is cryptographically bound to exactly one deposit's parameters. `V5_SOURCE_DELEGATE_SEED` is a single constant seed, so *any* allowance standing on a Gateway vault for this PDA is spendable by *any* `adapter_execute_across_v5` call — and `params.depositor` / `params.recipient` are unauthenticated wire fields.

Concretely: tape A approves 1,000,000 and spends 500,000, leaving 500,000 of allowance and 500,000 of balance on the shared per-mint vault. An attacker submits a new Gateway path with `Literal` mode, `input_amount = 500_000`, `depositor`/`recipient` = themselves. No `APPROVE` needed — the residual covers it. They get a valid `FundsDeposited` with the V5 magic prefix, funded by someone else's residual.

`V5_ADAPTER_SPEC.md` puts the burden on Gateway ("the continuing tape must leave no residual balance… Gateway does not currently enforce this"), but SpokePool can close it locally for ~free — `source` is already deserialized on the line below:

```rust
let source = load_token_account(gateway_vault_info, &token_program_id, &input_token, &gateway_vault_authority)?;
require_keys_eq!(
    source.delegate.ok_or_else(|| error!(V5Error::InsufficientDelegateAllowance))?,
    source_delegate,
    V5Error::InvalidTokenAccount
);
```

and then in `execute_v5_deposit`, after `resolve_v5_input_amount`:

```rust
require_v5_delegate_allowance(source.delegated_amount, input_amount)?;
```

That alone gets you the dedicated error and a non-brittle test. To actually kill the residual-allowance path you'd want exact consumption (`delegated_amount == input_amount`), which makes each Gateway `APPROVE` single-use from the SpokePool's side. Is there a reason the Gateway tape can't approve exactly the resolved amount? If exact-match is too strict for real tapes, please say so in the spec rather than asserting the residual is harmless.

(`load_v5_deposit_accounts` would need `input_amount` threaded in, or return `source.delegated_amount` alongside `source.amount`.)

programs/svm-spoke/src/instructions/v5_adapter.rs
```diff
@@ -0,0 +1,176 @@
...
+    if *token_program == anchor_spl::token::ID {
+        return Ok(());
+    }
+    let data = info.try_borrow_data()?;
+    let mint = StateWithExtensions::<SplMint>::unpack(&data).map_err(|_| error!(V5Error::InvalidTokenAccount))?;
+    let extensions = mint
+        .get_extension_types()
+        .map_err(|_| error!(V5Error::InvalidTokenAccount))?;
+    require!(
+        !extensions.contains(&ExtensionType::TransferFeeConfig) && !extensions.contains(&ExtensionType::TransferHook),
```
droplet-rl: **Blocking.** A two-entry denylist is the wrong polarity for a gate in a module whose whole stated design is fail-closed. Extensions that pass today but arguably shouldn't:

- `PermanentDelegate` — the mint's permanent delegate can move tokens straight back out of the SpokePool vault *after* the deposit event is emitted. That's the same class of loss as a transfer fee, but silent.
- `DefaultAccountState` (frozen) / `Pausable` — fail at transfer today, so no accounting break, but they can strand the SpokePool vault later.
- Anything Token-2022 adds after this ships is accepted by default. That's the part that rots.

Suggest inverting to an allowlist of extensions known to be inert for a vault-to-vault `transfer_checked` (metadata, metadata pointer, close authority, …) and rejecting the rest with `UnsupportedTokenExtension`. Same line count, and new extensions then fail closed instead of open.

I'll note the V3 `deposit` path is *looser* than this (it accepts transfer-fee mints), so this is already a net improvement — but since you're writing the gate anyway, an allowlist is strictly better and the spec section ("Transfer-fee mints are excluded… Transfer hooks remain disabled") reads as if it already is one.

test/svm/SvmSpoke.V5Source.ts
```diff
@@ -0,0 +1,351 @@
...
+  });
+
+  it("rejects wrong delegate, wrong vault accounts, and insufficient allowance", async () => {
+    const input = encodeDeposit(deposit, { literal: true });
+    await expectError(execute(input, Buffer.alloc(0), 1_000_000n, false, owner), "MissingAccount");
+    await expectError(
+      execute(input, Buffer.alloc(0), 1_000_000n, false, sourceDelegate, gatewayVault),
+      "MissingAccount"
+    );
+    await expectError(execute(input, Buffer.alloc(0), deposit.inputAmount - 1n), "insufficient funds");
```
droplet-rl: **Blocking (coverage).** Two things:

1. Asserting on `"insufficient funds"` couples the test to spl-token's log string. If you add the explicit `require_v5_delegate_allowance` call from my other comment, this becomes `"InsufficientDelegateAllowance"` and stops being brittle.
2. Branches added in this PR with zero validator coverage:
   - `UnsupportedTokenExtension` — brand-new error, brand-new `reject_unsupported_mint_extensions` path, never exercised. Needs a Token-2022 mint created with `TransferFeeConfig` (and ideally one with `TransferHook`) plus a plain Token-2022 mint proving the happy path still works. This is the gap I'd most want closed before merge — the function is currently unreachable-in-tests code guarding fund safety.
   - `DepositsArePaused` via the V5 entrypoint (the check moved from an Anchor constraint to a hand-rolled `require!`, so it's no longer structurally guaranteed).
   - `ResolvedInputAmountBelowCommitted` on-chain — unit-tested in `v5.rs`, but the wiring of `source.amount` into `resolve_v5_input_amount` is not.
   - `load_token_account`'s `account.owner == authority` check. Both current wrong-account cases short-circuit at `MissingAccount`, so the authority assertion never fires. Passing a same-mint ATA owned by an unrelated authority at the derived spoke-vault key isn't constructible — but a wrong-`state` ATA is, if you can reach a second initialized state.

Also worth a case for `bips` mode where the resolved amount exceeds the granted allowance, since that's the exact overshoot the shared-vault design makes possible.

programs/svm-spoke/src/instructions/v5_adapter.rs
```diff
@@ -0,0 +1,176 @@
...
+pub fn adapter_execute_across_v5<'info>(
+    ctx: Context<'_, '_, '_, 'info, AdapterExecuteAcrossV5<'info>>,
+    ctx_values: V5GatewayContext,
+    input: Vec<u8>,
+    jit_data: Vec<u8>,
+) -> Result<()> {
+    require_gateway_dispatch_authority(&ctx.accounts.dispatch_authority)?;
+    match decode_v5_adapter_input(&input)?.mode {
+        V5AdapterMode::Deposit(deposit) => execute_v5_deposit(ctx, ctx_values, deposit, &jit_data),
+        V5AdapterMode::Fill(_) => return err!(V5Error::UnsupportedMode),
```
droplet-rl: `return` is redundant — this `match` is the function's tail expression, and the sibling arm returns the value directly. `V5AdapterMode::Fill(_) => err!(V5Error::UnsupportedMode),` reads consistently with the `Deposit` arm.

programs/svm-spoke/src/instructions/v5_adapter.rs
```diff
@@ -0,0 +1,176 @@
...
+) -> Result<(DepositAccounts<'info>, TokenAccount)> {
+    let mint_info = find_v5_account(remaining_accounts, &input_token, false)?;
+    let token_program_id = *mint_info.owner;
+    // Mirror Interface<TokenInterface>: accept only a supported executable token program.
+    require!(
+        token_program_id == anchor_spl::token::ID || token_program_id == anchor_spl::token_2022::ID,
+        V5Error::InvalidTokenAccount
+    );
+    let token_program = find_v5_account(remaining_accounts, &token_program_id, false)?;
+    require!(token_program.executable, V5Error::InvalidTokenAccount);
```
droplet-rl: Two redundancies here, both cheap to drop (repo style leans hard on succinctness):

- `token_program.executable` — you already required `token_program_id` to be `anchor_spl::token::ID` or `token_2022::ID` on the line above. Both are executable by construction; the account at that address can't be anything else.
- `load_mint` re-checks `*info.owner == *token_program` (line 144), but `token_program_id` was *derived* from `mint_info.owner` on line 105. That comparison can never fail.

The `load_token_account` owner check is the one that earns its keep — keep that.

programs/svm-spoke/src/lib.rs
```diff
@@ -353,6 +354,17 @@ pub mod svm_spoke {
         Ok(utils::get_unsafe_deposit_id(signer, depositor, deposit_nonce))
     }
 
+    /// Executes one Gateway-authenticated Across V5 adapter branch. Wire version 1 enables source deposits; the
+    /// reserved Fill discriminant fails closed until its destination behavior is enabled.
```
droplet-rl, line 358: Every other handler in this `#[program]` block carries a `### Required Accounts:` / `### Parameters:` doc block. This one is the strongest candidate for it in the whole program: the accounts aren't in the `Accounts` struct at all, they're key-resolved out of `remaining_accounts`, so the doc comment is the only place an integrator can learn what to pass.

Worth listing: `dispatch_authority` (signer, Gateway PDA), `state`, then the remaining set — input mint, its token program, Gateway vault ATA (writable), pre-created SpokePool vault ATA (writable), `["v5_source_delegate"]`. The spec has this; the header doesn't, and `CLAUDE.md` asks for docs in the same change.

programs/svm-spoke/src/utils/transfer_utils.rs
```diff
@@ -11,19 +17,41 @@ pub fn transfer_from<'info>(
...
-        mint: mint.to_account_info(),
-        to: to.to_account_info(),
-        authority: delegate.to_account_info(),
-    };
-    let cpi_context = CpiContext::new_with_signer(token_program.to_account_info(), transfer_accounts, signer_seeds);
 
-    transfer_checked(cpi_context, amount, mint.decimals)
+    let bump_seed = [bump];
+    // Reuse the same offset after appending the bump, without allocating a seed vector.
+    let signer_seeds = [delegate_seeds[0], delegate_seeds[1], bump_seed.as_ref()];
```
droplet-rl, line 53: The right-alignment trick is correct — I traced both arms (`start=0` → `[b"delegate", hash, bump]`, `start=1` → `[seed, bump]`) and they match `find_program_address` — but it costs a fair bit of readability to avoid one small `Vec`, and the `[&[], *seed]` placeholder is the kind of thing that invites a wrong-index edit later.

More practically: Step 3 needs `derive_v5_fill_payer(submitter)`, which is a **two**-seed PDA with a runtime (non-`'static`) component. Neither `FunctionSeed(&'static [u8])` nor the fixed-2 array accommodates that, so this shape gets reworked one PR from now. Might be worth landing the general form here instead — e.g. take `seeds: &[&[u8]]` and build the signer array once — so Step 3 doesn't have to re-touch a fund-moving helper.

programs/svm-spoke/V5_ADAPTER_SPEC.md
```diff
@@ -51,8 +55,9 @@ cross-VM uint256 values remain 32-byte big-endian EVM words. The leading version
 is decoded, so any unsupported version reports `UnsupportedVersion` even when its body is not compatible with v1.
 
 Gateway token vaults are shared per mint rather than isolated per execution. `InputVaultBalance` therefore resolves
-against shared live state, and the continuing tape must leave no residual balance or stale approval that a later
-permissionless execution could consume. Gateway does not currently enforce this net-zero settlement invariant.
+against shared live state, and the continuing tape must leave no residual balance. Gateway does not currently enforce
+this net-zero settlement invariant. A residual delegate allowance is not privileged—the next permissionless
```
droplet-rl: > A residual delegate allowance is not privileged—the next permissionless `APPROVE` may replace it—but the SpokePool never delegates its own vault.

I'd push back on this. "May be replaced" isn't a safety property — nothing *requires* it to be replaced, and until it is, the residual is spendable by any Gateway path that routes into this adapter with attacker-chosen `depositor`/`recipient` (see my comment on `v5_adapter.rs`). Combined with the residual *balance* that the paragraph above concedes Gateway doesn't enforce away, the two together are a live path, not a theoretical one.

The second clause is accurate and worth keeping — the SpokePool vault is never delegated, so the exposure is bounded to whatever is sitting in the shared Gateway vault. But the framing should be "this depends on a Gateway invariant that is currently unenforced," not "not privileged." If you add the explicit allowance binding, this paragraph gets to make a much stronger claim instead.

programs/mock-gateway/src/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+    solana_program::{
+        instruction::{AccountMeta, Instruction},
+        program::invoke_signed,
+    },
+    InstructionData,
+};
+use anchor_spl::token_interface::{self, ApproveChecked, Mint, TokenAccount, TokenInterface};
+use svm_spoke::program::SvmSpoke;
+
+declare_id!("34trBszXuqhRjWaMxXWsunJNmyUsBvDNPxAwTzbPTm4p");
```
droplet-rl, line 12: Worth a second look at the blast radius of putting this under `programs/`, because two build scripts iterate `programs/*` unconditionally — not gated on `IS_TEST`:

- `scripts/svm/buildHelpers/buildSolanaVerify.sh` will now run a full `solana-verify build` for `mock_gateway` on `yarn build-verified`, i.e. a reproducible-build artifact for a mock, plus the CI time.
- `buildIdl.sh` + `generateSvmAssets.sh` copy **every** `target/idl/*.json` into `src/svm/assets/idl/` and auto-generate the barrel, so `mockGatewayIdl` and `MockGatewayAnchor` become public exports of the published `@across-protocol/contracts` package.

`programs/test` already does this, so it's a precedent — but this one declares `34trBszXuqhRjWaMxXWsunJNmyUsBvDNPxAwTzbPTm4p`, the **real** Gateway program ID. Shipping a mock IDL keyed to a live mainnet address is a meaningfully worse footgun than shipping `testIdl`: a consumer wiring `mockGatewayIdl` against that address gets an `execute_adapter` entrypoint that doesn't exist there. Suggest an exclusion list in `buildIdl.sh`/`generateSvmAssets.sh` (or drop the ID collision, though I assume it's load-bearing for the `dispatch_authority` derivation — which it is).

Since the validator suite didn't run locally, the IDL-generation job is also the most likely place this PR breaks CI: `anchor idl build --program-name mock_gateway` on a program crate that depends on another Anchor program crate is a known-fussy combination. Please confirm `upload-svm-artifacts` is green (it only runs on cache miss, and a new program should miss).

programs/mock-gateway/src/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+                    AccountMeta::new_readonly(ctx.accounts.state.key(), false),
+                    AccountMeta::new_readonly(ctx.accounts.event_authority.key(), false),
+                    AccountMeta::new_readonly(ctx.accounts.svm_spoke_program.key(), false),
+                    AccountMeta::new(ctx.accounts.gateway_vault.key(), false),
+                    AccountMeta::new(ctx.accounts.spoke_vault.key(), false),
+                    AccountMeta::new_readonly(ctx.accounts.mint.key(), false),
+                    AccountMeta::new_readonly(ctx.accounts.token_program.key(), false),
+                    AccountMeta::new_readonly(ctx.accounts.source_delegate.key(), false),
+                ],
+                data: svm_spoke::instruction::AdapterExecuteAcrossV5 {
```
droplet-rl, line 76: Nice — using `svm_spoke::instruction::AdapterExecuteAcrossV5 { .. }.data()` means the mock encodes via Anchor's own discriminator rather than hand-rolling it.

That does leave one thing unpinned, though: the mock proves the *Anchor* discriminator round-trips, and the `v5.rs` unit test proves `gateway_adapter_discriminator()` matches the golden fixture, but nothing asserts the two are the same value. They are today — I checked by adding this temporarily and it passes:

```rust
assert_eq!(
    svm_spoke::v5::gateway_adapter_discriminator(),
    svm_spoke::instruction::AdapterExecuteAcrossV5::DISCRIMINATOR,
);
```

The coupling is a string: `GATEWAY_ADAPTER_EXECUTE_V5_PREIMAGE = b"global:adapter_execute_across_v5"` has to track the Rust fn name. Rename `adapter_execute_across_v5` and the real Gateway silently dispatches to a discriminator this program no longer answers to — with the fixture test still green, because the fixture pins the constant, not the entrypoint. Now that the entrypoint exists, that two-line assertion closes the loop cheaply; `mock-gateway` is the natural home since it already depends on both.

droplet-rl (changes requested):
## Assessment

The structure here is good. Refactoring `_deposit` to take a `DepositAccounts` struct + `DepositId` enum instead of a concrete `Context<Deposit>` is the right way to share the validation/transfer/event core, and it preserves the existing ordering (validate → transfer → assign id). Resolving every remaining account by *derived key* rather than position (`find_v5_account`) is the correct instinct for a Gateway-dispatched entrypoint, and the mint/token-program/vault/authority checks do faithfully mirror the static Anchor constraints on `Deposit`. Fail-closed `Fill`, `paused_deposits`, and the JIT domain binding all look right.

Verified locally on `ea4f3770`:
- `cargo check -p svm-spoke -p mock-gateway` — clean
- `cargo test -p svm-spoke --lib` — 9 passed, as claimed
- I also confirmed the ABI concern is *not* a problem: `svm_spoke::v5::gateway_adapter_discriminator() == svm_spoke::instruction::AdapterExecuteAcrossV5::DISCRIMINATOR`. It holds today, but nothing pins it — see the inline note.

### Blocking

1. **The source-delegate allowance is never bound in-program.** `require_v5_delegate_allowance` and `V5Error::InsufficientDelegateAllowance` shipped in Step 1 and are now unreachable — the only enforcement is SPL's own `insufficient funds`. Because the source delegate is a *static* program-wide PDA (unlike the per-deposit `["delegate", seed_hash]` PDA everywhere else in this program), a residual allowance left on a shared Gateway vault by a partially-settled tape is directly spendable by the *next* permissionless Gateway path — with an attacker-chosen `depositor`/`recipient`, since neither is authenticated here. `source` is already deserialized, so binding `delegate`/`delegated_amount` costs almost nothing. Details inline.

2. **The Token-2022 gate is a two-entry denylist in a component whose stated design is fail-closed.** `PermanentDelegate` (mint admin can drain the SpokePool vault post-deposit) and any extension Token-2022 adds later pass silently.

3. **The new branches have no validator coverage.** `UnsupportedTokenExtension` — a brand-new error and code path — is never exercised; nor are `DepositsArePaused`, `ResolvedInputAmountBelowCommitted`, or `load_token_account`'s authority check on this path. The one allowance test asserts on an SPL log substring.

Since the validator suite didn't run locally, CI green is a hard prerequisite here — and note `buildIdl.sh`/`buildSolanaVerify.sh` both iterate `programs/*` unconditionally, so `mock-gateway` is now in the IDL-generation and verified-build paths too. That's the first thing to confirm.

### Non-blocking

Nits on redundant checks, the seed right-alignment trick, the missing `### Required Accounts:` doc block, and the published-artifact side effects of `mock-gateway` are inline. The `V5_ADAPTER_SPEC.md` claim that a residual allowance "is not privileged" is the one doc statement I'd push back on.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=c806e97835b9368c722fc31ffedc0572&trace_start_ts=1787923301&trace_end_ts=1788010001)

programs/svm-spoke/src/instructions/v5_adapter.rs
Reinis-FRP: Addressed in 669bd801. The loader now binds `source.delegate` to `v5_source_delegate`, and execution requires `delegated_amount >= input_amount` with the dedicated `InsufficientDelegateAllowance` error.

I intentionally did not require exact equality. The EVM adapter ultimately relies on ERC20 `transferFrom`, which accepts sufficient or maximum allowance while pulling exactly the requested amount; this path now preserves that behavior. Exact equality would reject valid maximum approvals without securing residual Gateway custody, because any balance left in the shared Gateway vault is already movable by a later committed Gateway `TRANSFER`. The spec now states that invariant explicitly. Validator coverage includes a 900k approval for a 750k pull, literal under-approval, and a bips-resolved allowance overshoot.

---

programs/mock-gateway/src/lib.rs
Reinis-FRP: Addressed in 669bd801. `mock-gateway` now asserts that `gateway_adapter_discriminator()` equals Anchor's `AdapterExecuteAcrossV5::DISCRIMINATOR`; `cargo test -p mock-gateway --lib` passes with both tests.

---

programs/svm-spoke/V5_ADAPTER_SPEC.md
Reinis-FRP: Addressed in 669bd801. The spec no longer calls residual allowance harmless: it documents the explicit delegate binding and sufficient-allowance check, explains why exact equality is intentionally not required for EVM `transferFrom` parity, and states that residual shared-vault balance remains a Gateway custody invariant because a later committed Gateway `TRANSFER` can move it. It also retains the distinction that the SpokePool vault itself is never delegated.

---

programs/svm-spoke/src/utils/transfer_utils.rs
Reinis-FRP: I kept this unchanged for now to avoid broadening the deposit-focused diff. The payer PDA is not a token-transfer delegate and will not use this helper; the next relevant token delegate is the static fill function seed, which still fits `FunctionSeed`. When fill is refactored in the follow-up step, both signed transfer call sites can be collapsed around the final shared shape instead of introducing an intermediate generalization here.

---

programs/svm-spoke/src/lib.rs
Reinis-FRP: Addressed in 669bd801. The entrypoint now documents all static accounts, the complete unordered remaining-account set and writability, PDA derivations, and each wire/JIT parameter.

---

programs/svm-spoke/src/instructions/v5_adapter.rs
Reinis-FRP: Addressed in 669bd801: removed both the redundant executable check and the mint-owner comparison while retaining the token-account owner, mint, and authority validation.

---

programs/svm-spoke/src/instructions/v5_adapter.rs
Reinis-FRP: Addressed in 669bd801: removed the redundant `return`, so the `Fill` arm is now the match tail expression.

---

test/svm/SvmSpoke.V5Source.ts
Reinis-FRP: Addressed in 669bd801. The allowance assertions now use `InsufficientDelegateAllowance`; coverage was added for bips overshoot, paused deposits, the committed dynamic-amount floor, transfer-fee and transfer-hook rejection, and a successful plain Token-2022 transfer. The existing over-approval case also proves the adapter pulls only the resolved amount.

I kept the token-account authority check as defense in depth. Because the account key must first equal the canonical ATA derived from that same authority, a supplied wrong-state or wrong-authority ATA necessarily has a different key and fails `MissingAccount`; reaching the authority branch with a live same-key account would require the owning PDA to authorize an authority mutation. I avoided adding an unsafe mock-only mutation just to manufacture that otherwise-unconstructible state.

---

programs/svm-spoke/src/instructions/v5_adapter.rs
Reinis-FRP: Addressed in 669bd801 by inverting the Token-2022 check to a fail-closed allowlist. Only mint-close authority and metadata/group pointer or data extensions are accepted; transfer fees, transfer hooks, permanent delegates, default account state, and unknown future extensions fail with `UnsupportedTokenExtension`. Unit coverage pins both accepted and rejected examples, and validator coverage exercises transfer-fee and transfer-hook rejection plus a plain Token-2022 happy path.

---

programs/mock-gateway/src/lib.rs
Reinis-FRP: Addressed in 669bd801. Production builds now exclude `mock_gateway` from verified artifacts, generated IDLs, generated TypeScript types, and published SVM assets; `IS_TEST=true` still includes it for validator execution, and stale mock artifacts are removed before production asset generation.

The fresh CI run on this head confirms both relevant paths: `Upload SVM artifacts` passed, including production SVM artifact generation/upload, and `Lint and Check Generated Files` passed with no generated-file drift.

---

programs/mock-gateway/src/lib.rs
```diff
@@ -0,0 +1,156 @@
...
+    #[msg("Forced failure after adapter execution")]
+    ForcedFailure,
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn adapter_discriminator_matches_gateway_abi() {
```
droplet-rl: This is the assertion I asked for and it passes (`cargo test -p mock-gateway --lib` → 2 passed). One problem with the placement, though: CI's "Test rust" step is

```yaml
- name: Test rust
  run: cargo test -p svm-spoke --lib
```

`-p svm-spoke` is package-scoped, so this test never runs in CI. The guard is inert exactly where it needs to be live — the failure mode it protects against (renaming `adapter_execute_across_v5` and silently desyncing `GATEWAY_ADAPTER_EXECUTE_V5_PREIMAGE` from the real Gateway's dispatch) would land green.

Cheapest fix is to move it into `svm-spoke`'s own test module, where CI already covers it. I verified this compiles and passes from inside `v5.rs::tests`:

```rust
#[test]
fn adapter_discriminator_matches_gateway_abi() {
    use anchor_lang::Discriminator;
    assert_eq!(
        gateway_adapter_discriminator(),
        crate::instruction::AdapterExecuteAcrossV5::DISCRIMINATOR,
    );
}
```

(`crate::instruction::` resolves — Anchor emits `pub mod instruction` at the crate root, beside the `#[program]` module.) That also drops the `anchor_lang::Discriminator` import from the mock entirely.

Alternative is widening CI to `cargo test --workspace --lib`, which would pick up `mock-gateway`'s `test_id` too and guard future mock-side tests — but that's a CI change for one assertion, so relocating seems better. Either way, worth doing before merge since it's the test that keeps the cross-repo ABI honest.

programs/svm-spoke/src/instructions/v5_adapter.rs
```diff
@@ -0,0 +1,204 @@
...
+    let gateway_vault_info = find_v5_account(remaining_accounts, &gateway_vault, true)?;
+    let spoke_vault_info = find_v5_account(remaining_accounts, &spoke_vault, true)?;
+    let source_delegate_info = find_v5_account(remaining_accounts, &source_delegate, false)?;
+
+    // Together with the canonical addresses above, these checks mirror the corresponding static mint and
+    // associated-token constraints.
+    let mint = load_mint(mint_info)?;
+    let source = load_token_account(gateway_vault_info, &token_program_id, &input_token, &gateway_vault_authority)?;
+    load_token_account(spoke_vault_info, &token_program_id, &input_token, &state)?;
+    require!(source.delegate == COption::Some(source_delegate), V5Error::InvalidTokenAccount);
```
droplet-rl, line 132: Good defensive addition, and the `COption` comparison is the correct way to express it (spl-token clears `delegated_amount` to 0 on `Revoke`, so there's no stale-amount path when `delegate` is `None`).

Worth knowing it's currently unreachable in tests: `mock_gateway::execute_adapter` always `approve_checked`s to `source_delegate` before the CPI, so `source.delegate` is always `Some(source_delegate)` by the time the adapter reads it — even at `approval_amount = 0`, which sets the delegate and trips the allowance check on the next line instead. The `delegate = owner` case in the wrong-accounts test short-circuits at `find_v5_account` with `MissingAccount`, well before this line.

Not worth contorting the mock for on its own, but if Step 3 adds a mock flag anyway (e.g. approve-to-arbitrary-delegate, or skip-approve so a prior standalone `Approve` instruction sets a different one), this check is a free extra case at that point. Purely defense-in-depth as things stand — the allowance check covers every practically reachable input.

programs/svm-spoke/src/instructions/v5_adapter.rs
```diff
@@ -0,0 +1,204 @@
...
+        .get_extension_types()
+        .map_err(|_| error!(V5Error::InvalidTokenAccount))?;
+    require!(extensions.iter().all(is_supported_v5_mint_extension), V5Error::UnsupportedTokenExtension);
+    Ok(())
+}
+
+fn is_supported_v5_mint_extension(extension: &ExtensionType) -> bool {
+    matches!(
+        extension,
+        ExtensionType::MintCloseAuthority
```
droplet-rl, line 181: Allowlist contents check out. Walking the Token-2022 mint extensions that are now rejected and weren't before: `PermanentDelegate` (mint admin could move tokens back out of the SpokePool vault after the event was emitted — the one I was most concerned about), `ScaledUiAmount` and `InterestBearingConfig` (both reinterpret amounts against the `transfer_checked` decimals assertion), `ConfidentialTransferMint`, `DefaultAccountState`, `Pausable`, `NonTransferable`. That's the right set.

`MintCloseAuthority` is fine to permit — `CloseAccount` on a mint requires zero supply, so it's only exercisable when there's nothing in any vault to lose.

One note for whoever revisits this: the allowlist covers *mint* extensions only, which is correct for what `get_extension_types` returns here. The account-side extensions that could matter on the two vaults (`CpiGuard` on the source blocking delegate-signed CPI transfers, `MemoTransfer` on the destination) all fail the transfer rather than corrupting accounting, so they're safe-by-default — but that's an implicit property, not something this function establishes. Might be worth a sentence in `V5_ADAPTER_SPEC.md` so a future reader doesn't assume this gate covers account extensions too.

droplet-rl (approved):
## Re-review of `669bd801` ("fix(svm): address V5 source review")

All three blocking items from my previous review are resolved, and one of them with a better answer than I proposed. Approving.

Verified locally on `669bd801` (base still `5424fd77`):
- `cargo test -p svm-spoke --lib` — 10 passed (up from 9; the new allowlist test is in the CI-covered crate)
- `cargo test -p mock-gateway --lib` — 2 passed
- `rustfmt.toml` sets only `reorder_imports`, so the split `use anchor_lang::{prelude::*}` / `use anchor_lang::solana_program::...` statements are already in canonical order — no `cargo +nightly fmt --check` failure
- Traced the new `IS_TEST` guards against `.github/workflows/pr.yml`: `test-svm-verified` sets `IS_TEST: true` on the "Create verified build" step, so `target/deploy/mock_gateway.so` still exists for `anchor test --skip-build`. `upload-svm-artifacts` does *not* set it, so the mock IDL is correctly excluded from published assets. The guards land on the right side of both paths.

### Item-by-item

**1. Delegate allowance (was blocking) — resolved, and I accept the pushback.** `require_v5_delegate_allowance(source.delegated_amount, input_amount)` plus the `source.delegate == Some(v5_source_delegate)` binding make the dedicated error reachable and the test now asserts on it instead of an spl-token log string.

You declined exact-consumption, and the spec rationale is the right call: *"Any residual Gateway-vault balance is already movable by a later committed Gateway `TRANSFER`; exact allowance would not replace that custody invariant."* That's correct and it's a better framing than mine — Gateway holds the vault authority, so residual balance is Gateway-custodied regardless of what the SpokePool asserts about the allowance. Requiring equality would close one route to a residual that remains reachable by several others, at the cost of breaking max-approval tapes. The net-zero invariant is genuinely Gateway's to enforce, and the spec now says so plainly instead of calling the residual "not privileged" — that was my actual ask.

**2. Token-2022 gate (was blocking) — resolved.** Inverting to `is_supported_v5_mint_extension` is exactly right, and the allowlist choices hold up: `ScaledUiAmount`, `InterestBearingConfig`, `ConfidentialTransferMint`, and `PermanentDelegate` all now fail closed, which the old two-entry denylist let through. `MintCloseAuthority` is safe to permit — close requires zero supply, so there's nothing to lose when it's exercisable. Nice touch adding the plain-Token-2022 happy-path assertion; that's what stops an allowlist from silently over-rejecting.

**3. Test coverage (was blocking) — resolved.** `UnsupportedTokenExtension` for both `TransferFeeConfig` and `TransferHook`, `DepositsArePaused`, `ResolvedInputAmountBelowCommitted`, and the bips-overshoot allowance case are all now exercised through the mock. Building the extended mints by hand with `getMintLen` is the right way to do it.

**4–7. Nits — all resolved.** Redundant `executable` / `load_mint` owner checks dropped (both were provably unreachable, since `token_program_id` derives from `mint_info.owner`), `return err!` cleaned up, the `### Required Accounts:` block added and accurate about order-independence, build scripts gated.

### One follow-up (non-blocking, one line)

The discriminator-pin test landed in `mock-gateway`, but CI's "Test rust" step runs `cargo test -p svm-spoke --lib` — package-scoped, so that assertion never executes in CI. Details and a verified one-line relocation inline.

I did not re-run the validator suite (no Anchor CLI here), so `test-svm-verified` remains the gate on the eight TS cases.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=d93d2d098c8050768284bbae8399f31b&trace_start_ts=1787925950&trace_end_ts=1788012650)

programs/mock-gateway/src/lib.rs
Reinis-FRP: Addressed in ef66daf2. The assertion now lives in `svm-spoke::v5::tests`, imports `Discriminator` there, and compares the computed Gateway discriminator with `crate::instruction::AdapterExecuteAcrossV5::DISCRIMINATOR`. The duplicate mock-side test was removed. `cargo test -p svm-spoke --lib` passes, so the guard now runs in the existing CI `Test rust` command.

---

programs/svm-spoke/src/instructions/v5_adapter.rs
Reinis-FRP: Addressed in ef66daf2. The spec now states explicitly that the allowlist covers mint extensions only and records the account-side behavior relevant to this path: source CPI guard and destination memo requirements cause the token transfer to fail rather than changing deposit accounting.

---

programs/svm-spoke/src/instructions/v5_adapter.rs
Reinis-FRP: Agreed. I kept this as defense in depth and did not add a mock-only flag solely to manufacture the otherwise unreachable branch. The current validator cases cover the practically reachable allowance failures; when the Step 3 mock gains another approval mode, I will add the wrong-active-delegate case there at essentially no additional mock complexity.

---

programs/mock-gateway/src/lib.rs
```diff
@@ -0,0 +1,143 @@
+use anchor_lang::{
```
fusmanii: is the goal to replace this mock with actual gateway once we make solana-v5 repo public?

programs/svm-spoke/src/instructions/v5_adapter.rs
```diff
@@ -0,0 +1,204 @@
...
+        params.fill_deadline,
+        params.exclusivity_parameter,
+        message,
+        DelegatePda::FunctionSeed(V5_SOURCE_DELEGATE_SEED),
+    )?;
+    emit_cpi!(event);
+    Ok(())
+}
+
+fn load_v5_deposit_accounts<'info>(
```
fusmanii, line 105: should this function live in `DepositAccounts` impl?

programs/svm-spoke/src/instructions/v5_adapter.rs
```diff
@@ -0,0 +1,204 @@
...
+    let (output_amount, exclusive_relayer) = match jit {
+        Some(jit) => resolve_v5_deposit_modifications(&deposit, &jit, &GATEWAY_PROGRAM_ID, &ctx_values.path_id)?,
+        None => (params.output_amount, params.exclusive_relayer),
+    };
+    let (accounts, source) =
+        load_v5_deposit_accounts(ctx.remaining_accounts, ctx.accounts.state.key(), params.input_token)?;
+    let input_amount = resolve_v5_input_amount(deposit.input_amount_mode, params.input_amount, source.amount)?;
+    // Match EVM transferFrom semantics: sufficient and max allowances are valid; the adapter pulls exactly
+    // input_amount.
+    require_v5_delegate_allowance(source.delegated_amount, input_amount)?;
```
fusmanii, line 70: is this mainly to have a nice error for the submitter? like if there isn't enough delegated then the transaction would still fail but with a convoluted error? if not enough is delegated and a transfer is attempted and the error is clear then I think we can remove this check here

programs/svm-spoke/src/instructions/v5_adapter.rs
```diff
@@ -0,0 +1,204 @@
...
+    let data = info.try_borrow_data()?;
+    let mint = StateWithExtensions::<SplMint>::unpack(&data).map_err(|_| error!(V5Error::InvalidTokenAccount))?;
+    let extensions = mint
+        .get_extension_types()
+        .map_err(|_| error!(V5Error::InvalidTokenAccount))?;
+    require!(extensions.iter().all(is_supported_v5_mint_extension), V5Error::UnsupportedTokenExtension);
+    Ok(())
+}
+
+fn is_supported_v5_mint_extension(extension: &ExtensionType) -> bool {
```
fusmanii, line 178: should we adopt the same set of extension exclusion list on the solana-v5 side as well? I remember talking about still allowing tokens with other extensions but if we are moving towards only allowing these extensions then it makes sense to stay consistent across the board

fusmanii (commented):
first pass

programs/svm-spoke/src/instructions/v5_adapter.rs
```diff
@@ -0,0 +1,204 @@
...
+    jit_data: &[u8],
+) -> Result<()> {
+    require!(!ctx.accounts.state.paused_deposits, CommonError::DepositsArePaused);
+
+    let jit = decode_v5_deposit_jit(&deposit, jit_data)?;
+    let params = &deposit.deposit_params;
+    let (output_amount, exclusive_relayer) = match jit {
+        Some(jit) => resolve_v5_deposit_modifications(&deposit, &jit, &GATEWAY_PROGRAM_ID, &ctx_values.path_id)?,
+        None => (params.output_amount, params.exclusive_relayer),
+    };
```
grasphoper, line 64: I think we should base this branching off of what the user specified in `AcrossDepositInput`, not on `jit`.

Take a look at the EVM side:

```
        if (uint256(params.paramModificationRules) != 0) {
            _applyParamModifications(
                deposit, abi.decode(jitData, (AcrossDepositJitParams)), pathId, params.paramModificationRules
            );
        }
```

If params are specified by the user, we expect jit to be a certain shape, otherwise we revert. This forces a submitter to supply an auction resolution for example

grasphoper: Ah I see. Technically we check that `jit` is nonzero in case of flags being set in `decode_v5_deposit_jit`. But this is a very non-trivial codepath. Decoding function should not be used for correctness checks like this IMO. A more straightforward way sould be:

```
if user asked (check `deposit` struct):
  resolve_v5_deposit_modifications(.., decode(jit)) else
  take from `deposit.deposit_params`
```

IMO the decode function should not concern itself with anything other than a strict decoding (see #1535 comments)

programs/svm-spoke/src/instructions/deposit.rs
```diff
@@ -67,8 +68,38 @@ pub struct Deposit<'info> {
...
 }
 
+pub struct DepositAccounts<'info> {
+    pub from: AccountInfo<'info>,
+    pub vault: AccountInfo<'info>,
+    pub delegate: AccountInfo<'info>,
+    pub mint: AccountInfo<'info>,
+    pub token_program: AccountInfo<'info>,
+    pub mint_decimals: u8,
+}
```
grasphoper, line 78: Do you think we can use `TransferChecked` inside this struct?

```
  pub struct DepositAccounts<'info> {
      pub transfer: TransferChecked<'info>,
      pub token_program: AccountInfo<'info>,
      pub mint_decimals: u8,
  }
```

programs/svm-spoke/src/instructions/v5_adapter.rs
```diff
@@ -0,0 +1,204 @@
...
+    let gateway_vault =
+        get_associated_token_address_with_program_id(&GATEWAY_VAULT_AUTHORITY, &input_token, &token_program_id);
+    let spoke_vault = get_associated_token_address_with_program_id(&state, &input_token, &token_program_id);
+    let gateway_vault_info = find_v5_account(remaining_accounts, &gateway_vault, true)?;
+    let spoke_vault_info = find_v5_account(remaining_accounts, &spoke_vault, true)?;
+    let source_delegate_info = find_v5_account(remaining_accounts, &V5_SOURCE_DELEGATE, false)?;
+
+    // Together with the canonical addresses above, these checks mirror the corresponding static mint and
+    // associated-token constraints.
+    let mint = load_mint(mint_info)?;
```
grasphoper, line 129: Should we deserialize mint once? `reject_unsupported_mint_extensions` wants it too, perhaps we can combine these?

programs/svm-spoke/src/instructions/v5_adapter.rs
```diff
@@ -0,0 +1,204 @@
...
+    let (accounts, source) =
+        load_v5_deposit_accounts(ctx.remaining_accounts, ctx.accounts.state.key(), params.input_token)?;
+    let input_amount = resolve_v5_input_amount(deposit.input_amount_mode, params.input_amount, source.amount)?;
+    // Match EVM transferFrom semantics: sufficient and max allowances are valid; the adapter pulls exactly
+    // input_amount.
+    require_v5_delegate_allowance(source.delegated_amount, input_amount)?;
+
+    let mut message = Vec::with_capacity(64);
+    message.extend_from_slice(&V5_MAGIC_PREFIX);
+    message.extend_from_slice(&deposit.dst_step_id);
```
grasphoper, line 74: What about

```
let message = [V5_MAGIC_PREFIX, deposit.dst_step_id].concat();
```

programs/svm-spoke/src/instructions/v5_adapter.rs
Reinis-FRP: Addressed by the Step 1 JIT-decoding revision in c1a51c80 and carried into this PR by the rebase in 6134b773. The current `execute_v5_deposit` branches directly on the committed `deposit.modification_rules.jit_enabled()` value: the enabled branch strictly decodes `jit_data` and applies the permitted modifications, while the disabled branch uses the committed deposit parameters unchanged. Missing or malformed JIT therefore still reverts whenever the committed rules require it.

`decode_v5_adapter_input` and `decode_strict` now only perform strict Borsh decoding; they no longer contain this correctness policy. GitHub marks the original line outdated because of the force-push.

---

programs/svm-spoke/src/instructions/v5_adapter.rs
Reinis-FRP: Addressed in d5a128cd. Replaced the separate `reject_unsupported_mint_extensions` and `load_mint` passes with `validate_v5_mint`, which borrows and deserializes the mint once via `StateWithExtensions::<SplMint>::unpack`—the same parser used by Anchor's token-interface mint account. For Token-2022 it validates the extension allowlist from that parsed state, and for both token programs it returns the base decimals.

This helper is used by the deposit loader now and can be reused by the V5 fill loader when Step 4 is restacked. `cargo test -p svm-spoke --lib` passes all 12 tests; the changed file passes rustfmt and `git diff --check`.

---

programs/svm-spoke/src/instructions/deposit.rs
Reinis-FRP: Addressed in 07fe127f. `DepositAccounts` now stores the `TransferChecked` account bundle directly alongside the token program and mint decimals. Both `From<&Deposit>` and the V5 remaining-account loader construct that same bundle, and `_deposit` passes it straight to `transfer_from_with_delegate` instead of unpacking and rebuilding it.

This keeps the same four `AccountInfo` conversions/clones and adds no runtime work. `cargo test -p svm-spoke --lib` passes all 12 tests; both changed files pass rustfmt and `git diff --check`.

---

programs/svm-spoke/src/instructions/v5_adapter.rs
Reinis-FRP: Changed in 8b285e4a. The V5 witness message now uses `[V5_MAGIC_PREFIX, deposit.dst_step_id].concat()`, directly expressing `prefix || stepId` while still allocating the resulting `Vec` once.

`cargo test -p svm-spoke --lib` passes all 12 tests; the changed file passes rustfmt and `git diff --check`.

---

programs/svm-spoke/src/instructions/v5_adapter.rs
Reinis-FRP: Addressed in 9ff3f632. `load_v5_deposit_accounts` is now the private associated constructor `DepositAccounts::load_v5`. This namespaces both deposit-account construction paths under `DepositAccounts`.

I kept the impl in `v5_adapter.rs` rather than `deposit.rs` because the loader is specifically a V5 adapter boundary: it understands Gateway remaining-account discovery, V5 authorities, ATA derivation, the Token-2022 extension policy, and `V5Error`. Moving those dependencies into `deposit.rs` would couple the context-agnostic shared deposit flow to Gateway/V5 concerns. Rust still allows the inherent impl to live here because the type is defined in the same crate.

It continues returning the already-decoded source `TokenAccount` alongside `Self`, avoiding a second deserialization for the balance and allowance checks. Validation behavior is unchanged. `cargo test -p svm-spoke --lib` passes all 12 tests; the changed file passes rustfmt and `git diff --check`.

---

programs/svm-spoke/src/instructions/v5_adapter.rs
Reinis-FRP: Fixed in `3ba2477a`. You are right that this preflight only changed the error surfaced to the submitter: SPL Token and Token-2022 `transfer_checked` already enforce `delegated_amount >= input_amount` atomically and return `InsufficientFunds` when the allowance is too small. The V5 source integration tests now assert that native error and unchanged balances after rollback.

I traced why the check had been added: it was responding to a residual-allowance concern. However, a `>=` precheck never eliminated residual allowance; only requiring exact allowance would do that, which would diverge from EVM `transferFrom` and reject maximum approvals. There is also no source-deposit self-transfer exception that needs the precheck: the source is the Gateway-vault-authority ATA while the destination is the distinct State-authority ATA, so the token program performs the normal debit and allowance update. Gateway-vault self-delivery on fills is separately recognized as in-place delivery and does not invoke a delegated transfer.

Validated with all 12 `svm-spoke` Rust unit tests and all 7 V5 source integration tests.

---

programs/svm-spoke/src/instructions/v5_adapter.rs
Reinis-FRP: Good point. I agree we should apply the same fail-closed principle to the relevant `solana-v5` component, but I would keep this allowlist in the V5 Spoke adapter and avoid making it a blanket Gateway restriction.

The SpokePool needs exact accounting and durable custody: transfer fees can make the amount received diverge from the deposited amount, while a permanent delegate can remove funds after the deposit event. In `solana-v5`, the corresponding gap is the PrefundedAdapter, which currently rejects only TransferHook. That adapter should separately move to a fail-closed allowlist because transfer fees can under-deliver its committed `min_amount`, and permanent-delegate/freeze/pause behavior can drain or strand stored credit. The generic Gateway has no single downstream accounting invariant, so extension support should remain owned by the consuming adapter.

---

programs/svm-spoke/src/instructions/v5_adapter.rs
```diff
@@ -0,0 +1,201 @@
...
+fn execute_v5_deposit<'info>(
+    ctx: Context<'_, '_, '_, 'info, AdapterExecuteAcrossV5<'info>>,
+    ctx_values: GatewayContextV1,
+    deposit: AcrossDepositInput,
+    jit_data: &[u8],
+) -> Result<()> {
+    require!(!ctx.accounts.state.paused_deposits, CommonError::DepositsArePaused);
+
+    let params = &deposit.deposit_params;
+    let (output_amount, exclusive_relayer) = if deposit.modification_rules.jit_enabled() {
```
grasphoper: nit: could we name `deposit.modification_rules.jit_enabled()` better?
Something like `modification_enabled()` or `_required()` or some other better name? I feel like `jit_enabled()` is a bit too mechanical

grasphoper (approved):

programs/svm-spoke/src/instructions/v5_adapter.rs
Reinis-FRP: Addressed. I renamed the predicate to `requires_jit()` in 6ae33115 and updated this entrypoint at its introduction in d374374a (current PR head 8604d216).

I checked the EVM adapter before choosing the name: it branches on `paramModificationRules != 0` and decodes `jitData` for any nonzero rules word, even if no permission flag ultimately changes a deposit field. `requires_jit()` therefore describes the predicate's actual contract more precisely than `modification_enabled()`. The V5 spec and foundation tests use the new name as well.

---

PR #1537, opened 2026-08-28
feat(svm): add V5 fill-status payer float

Reinis-FRP:
## Summary

- add a submitter-scoped V5 fill-status payer PDA that manually creates the standard fill-status PDA without requiring that PDA to be a transaction signer
- keep V5-tagged relays outside the slow-fill lifecycle: an existing `Filled` status returns `RelayFilled`, while any other program-owned fill status is invalid
- have the account helper return the validated payer PDA without serializing semantic fields, leaving Step 4's shared fill core to write `Filled`, `relayer`, and `fill_deadline` after token delivery
- bind V5 fill-status rent ownership in the existing `FillStatusAccount.relayer` slot without changing its account layout; legacy fills continue to store their signing relayer there
- make post-expiry `close_fill_pda` permissionless while keeping its recorded `relayer` as the immutable rent destination; only the V5 submitter may withdraw its payer float
- handle both empty and prefunded fill-status PDAs using the same create/top-up/allocate/assign sequence generated by Anchor `init_if_needed`, relying on the System Program to enforce whether the account can be allocated and assigned
- document the Step 4 safety contract: source the submitter from Gateway-attested context, derive the relay hash from the validated V5 relay, and complete semantic validation before account creation; the caller owns semantic status serialization
- generate test-feature IDLs after production artifacts in standard SVM test commands and verified-SVM CI, keeping test-only output confined to `target/`

The reserved V5 `Fill` adapter branch remains fail-closed for Step 4.

Part of ACP-184. This is Step 3 of the implementation plan and is stacked on #1536.

## Compatibility note

This deliberately relaxes the live `close_fill_pda` authorization rule for existing fill-status accounts. Old-IDL clients may continue supplying the recorded relayer signer meta and signature and remain compatible on-chain. Clients generated from the new IDL must stop explicitly passing that relayer to `.signers(...)`, because the account is no longer marked as signing. The account name is retained to avoid needless client churn, and the in-repository callers have been updated. The V5 adapter spec records the authorization change for release coordination.

## Validation

- `cargo check -p svm-spoke --features test`
- `cargo test -p svm-spoke --lib` (11 passed)
- `yarn lint-rust` (passed with existing warnings)
- focused Prettier checks for the workflow, V5 spec, and changed SVM tests
- target-only test-feature IDL generation for both `svm_spoke` and `mock_gateway`; confirmed `test_create_v5_fill_status` is present only in the test IDL while production package assets remain unchanged
- `yarn build-ts`
- focused validator coverage in `SvmSpoke.V5FillStatus.ts` (2 passed: empty and prefunded creation/reclaim, plus withdrawal authorization and amounts)
- repository pre-commit hook

---


programs/svm-spoke/src/instructions/v5_fill_status.rs
```diff
@@ -0,0 +1,155 @@
...
+pub fn create_v5_fill_status<'info>(
+    payer: &AccountInfo<'info>,
+    fill_status: &AccountInfo<'info>,
+    system_program_info: &AccountInfo<'info>,
+    submitter: &Pubkey,
+    relay_hash: &[u8; 32],
+    fill_deadline: u32,
+) -> Result<()> {
+    let (expected_payer, payer_bump) = Pubkey::find_program_address(&[V5_FILL_PAYER_SEED, submitter.as_ref()], &ID);
+    let (expected_fill_status, fill_status_bump) = Pubkey::find_program_address(&[FILL_STATUS_SEED, relay_hash], &ID);
```
droplet-rl: Both derivations already exist in `v5.rs` and are re-derived inline here:

```rust
pub fn derive_v5_fill_payer(submitter: &Pubkey) -> (Pubkey, u8)  // v5.rs:314
pub fn derive_fill_status(relay_hash: &[u8; 32]) -> (Pubkey, u8)  // v5.rs:318
```

That matters beyond DRY: `pda_domains_match_golden_fixture` (v5.rs:487) pins both of those helpers against the golden fixture, including `/pdas/fillPayer` and `/pdas/fillStatus`. This copy is outside that safety net, so a future seed-domain change would be caught in `v5.rs` and silently diverge here.

```rust
let (expected_payer, payer_bump) = derive_v5_fill_payer(submitter);
let (expected_fill_status, fill_status_bump) = derive_fill_status(relay_hash);
```

Separately, worth considering for Step 4: two `find_program_address` calls cost up to ~25k CU in what will be the fill hot path inside an atomic tape. Taking both bumps as arguments and using `create_program_address` is safe here — a non-canonical bump derives a different address and fails the `require_keys_eq!` below — and would recover most of that budget.

programs/svm-spoke/src/instructions/v5_fill_status.rs
```diff
@@ -0,0 +1,155 @@
...
+    require_keys_eq!(*payer.key, expected_payer, V5Error::InvalidFillPayer);
+    require_keys_eq!(*fill_status.key, expected_fill_status, V5Error::InvalidFillStatusAccount);
+    require_keys_eq!(*system_program_info.key, system_program::ID, V5Error::MissingAccount);
+    require!(system_program_info.executable, V5Error::MissingAccount);
+    require!(payer.is_writable && fill_status.is_writable, V5Error::InvalidAccountMutability);
+    require_keys_eq!(*payer.owner, system_program::ID, V5Error::InvalidFillPayer);
+    require!(payer.data_is_empty(), V5Error::InvalidFillPayer);
+    require_keys_eq!(*fill_status.owner, system_program::ID, V5Error::InvalidFillStatusAccount);
+    require!(fill_status.data_is_empty(), V5Error::InvalidFillStatusAccount);
```
droplet-rl: `3594622c` dropped the dedicated replay signal from just above these lines:

```rust
-    require_keys_neq!(*fill_status.owner, ID, CommonError::RelayFilled);
```

The replay is still rejected — the `owner == system_program` check catches it — but it now surfaces as `InvalidFillStatusAccount`, which is the same error as "wrong PDA" and "foreign-owned account." Those need different off-chain handling: `RelayFilled` means stop retrying and drop the relay, whereas a malformed-account error means retry or alert. Legacy `fill_relay` returns `CommonError::RelayFilled` for exactly this case (`fill.rs:121`), so V5 fills would report a different error for identical semantics.

The doc comment above still says "an existing program-owned fill status is rejected as a replay," which no longer matches the error the caller actually observes. Please restore the `require_keys_neq!` so the classification stays consistent with the legacy path — the new TS assertion on `InvalidFillStatusAccount` would need updating to `RelayFilled` alongside it.

programs/svm-spoke/src/instructions/v5_fill_status.rs
```diff
@@ -0,0 +1,155 @@
...
+        )?;
+        invoke_signed(
+            &system_instruction::assign(fill_status.key, &ID),
+            &[fill_status.clone(), system_program_info.clone()],
+            &[fill_status_seeds],
+        )?;
+    }
+
+    FillStatusAccount { status: FillStatus::Filled, relayer: expected_payer, fill_deadline }
+        .try_serialize(&mut &mut fill_status.try_borrow_mut_data()?[..])
```
droplet-rl, line 99: `3594622c` removed `write_v5_fill_status` and, with it, the only test asserting the layout invariant this PR is built on:

```rust
assert_eq!(V5_FILL_STATUS_SPACE, 8 + 1 + 32 + 4);
assert_eq!(FillStatusAccount::DISCRIMINATOR.len(), DISCRIMINATOR_SIZE);
```

Inlining the struct literal is fine, but that assertion is the guard for "preserving the legacy account layout" in the PR title. `V5_FILL_STATUS_SPACE` is computed from `INIT_SPACE`, so it tracks the struct automatically and never trips — but a field added to `FillStatus` or `FillStatusAccount` would silently change the on-chain layout and the rent math on both the legacy and V5 paths, with nothing failing. It costs nothing to keep as a standalone `#[test]` against the constant; please restore it.

programs/svm-spoke/src/instructions/v5_fill_status.rs
```diff
@@ -0,0 +1,155 @@
...
+            &system_instruction::allocate(fill_status.key, V5_FILL_STATUS_SPACE as u64),
+            &[fill_status.clone(), system_program_info.clone()],
+            &[fill_status_seeds],
+        )?;
+        invoke_signed(
+            &system_instruction::assign(fill_status.key, &ID),
+            &[fill_status.clone(), system_program_info.clone()],
+            &[fill_status_seeds],
+        )?;
+    }
```
droplet-rl, line 96: The prefunded branch — the transfer / allocate / assign sequence that the summary calls out as the "one-lamport account-creation denial of service" mitigation — is never executed by any test. `SvmSpoke.V5FillStatus.ts` funds `payer`, never `fillStatus`, so every test run takes the `current_lamports == 0` path and lines 64-82 are dead in CI.

I walked through the logic and believe it's correct, but this is the subtlest code in the PR (three chained `invoke_signed` calls with two different signer-seed sets) and it's the part a reader is least likely to verify by inspection. A one-line addition to the existing validator test — transfer 1 lamport to the fill-status PDA before calling `testCreateV5FillStatus`, then assert the account is created and the float is debited by `rentExemption(45) - 1` — would cover it and directly demonstrate the DoS resistance the summary claims.

The `InvalidFillPayer`, `InvalidAccountMutability`, and `MissingAccount` branches (lines 35-41) are likewise untested.

programs/svm-spoke/src/instructions/v5_fill_status.rs
```diff
@@ -0,0 +1,155 @@
...
+
+    pub system_program: Program<'info, System>,
+}
+
+/// Withdraws from the signing submitter's own fill-status rent float. `u64::MAX` drains the live balance.
+pub fn withdraw_v5_fill_payer(ctx: Context<WithdrawV5FillPayer>, amount: u64) -> Result<()> {
+    let submitter = ctx.accounts.submitter.key();
+    let balance = ctx.accounts.payer.lamports();
+    let amount = if amount == u64::MAX { balance } else { amount };
+    let seeds: &[&[u8]] = &[V5_FILL_PAYER_SEED, submitter.as_ref(), &[ctx.bumps.payer]];
```
droplet-rl, line 159: `13821579` removed the rent guard that used to sit here:

```rust
-    require_fill_payer_spend(balance, amount, Rent::get()?.minimum_balance(0))?;
```

and the matching spec sentence ("a nonzero remainder must be rent-exempt") was dropped from `V5_ADAPTER_SPEC.md` in this PR. The behaviour is still safe, but a partial withdrawal leaving a nonzero remainder below 890,880 lamports now fails at the runtime rent-state check with `InsufficientFundsForRent` — a bare transaction-level error with no program context — instead of the previous `FillPayerRemainderNotRentExempt`.

Note the new test never exercises this: it withdraws exactly `rentMinimum` from `rentMinimum * 2`, landing precisely on the rent-exempt floor, then drains via `u64::MAX`. Every value strictly between those two fails opaquely. Since this is a submitter-facing instruction and "withdraw most of my float" is the obvious usage, please either restore the guard or document the constraint in the spec and add a test pinning the failure mode.

Minor, same function: `amount == 0` is accepted and emits a `V5FillFloatWithdrawn` with `amount: 0`, which will show up as noise for indexers.

programs/svm-spoke/src/instructions/v5_fill_status.rs
```diff
@@ -0,0 +1,155 @@
...
+}
+
+/// Test-only entrypoint for exercising the PDA-signed account-creation path before Step 4 wires it into V5 Fill.
+#[cfg(feature = "test")]
+pub fn test_create_v5_fill_status(
+    ctx: Context<TestCreateV5FillStatus>,
+    submitter: Pubkey,
+    relay_hash: [u8; 32],
+    fill_deadline: u32,
+) -> Result<()> {
```
droplet-rl, line 124: `submitter` is an unauthenticated instruction argument, so anyone holding this entrypoint can spend an arbitrary submitter's float and permanently mark any relay hash as `Filled`, blocking a legitimate fill.

The gating is correct — `buildSolanaVerify.sh` only adds `--features test` under `IS_TEST=true`, and `publish.yml` builds it as a separate `svm-verified-test-binaries` artifact, distinct from the production one. But this is a step up from the existing `test`-gated code in `utils/testable_utils.rs`, which only mocks time; this is the first one that can burn another account's lamports, and the test binaries are published as a public release artifact.

Cheap hardening that costs nothing in coverage: make `submitter` a `Signer` in `TestCreateV5FillStatus` and drop the argument. `create_v5_fill_status` still receives no signature for the payer PDA, so the signerless property under test is fully preserved — the wrapper just can't be pointed at a third party's float if a test build is ever deployed somewhere it shouldn't be. The TS test would only need to retain the generated keypair and add `.signers([submitter])`.

Also worth a `# Safety`-style line on `create_v5_fill_status` stating the precondition Step 4 must honour: the caller is responsible for supplying a Gateway-attested `submitter`, since the helper authenticates the derivation but not the value.

programs/svm-spoke/src/instructions/fill.rs
```diff
@@ -201,8 +201,9 @@ fn unwrap_fill_relay_params(
 
 #[derive(Accounts)]
 pub struct CloseFillPda<'info> {
+    /// CHECK: The address constraint binds this non-signing account to the recorded rent recipient.
     #[account(mut, address = fill_status.relayer @ SvmError::NotRelayer)]
-    pub signer: Signer<'info>,
+    pub signer: UncheckedAccount<'info>,
```
droplet-rl, line 206: This relaxes an authorization check on an instruction that is live on mainnet today, and it's the one change here that reaches beyond V5.

I think it's safe: `close = signer` is still bound by `address = fill_status.relayer`, and every writer of that field records a real beneficiary — `fill_relay` the signing relayer, `request_slow_fill` the requester (`slow_fill.rs:75`), `create_v5_fill_status` the payer PDA. The deadline check still gates closure, `fill_relay` cannot re-fill past the deadline, so permissionless close opens no replay window. Rent cannot be redirected. I checked the in-repo callers (`SvmSpoke.Fill.ts:371`, `SvmSpoke.SlowFill.AcrossPlus.ts:482`, `scripts/svm/closeRelayerPdas.ts`) and they all still pass the relayer as a signer, which the runtime accepts as a redundant signature.

Two things I'd still want before this merges:

1. **Call it out in the PR description.** "Legacy `close_fill_pda` becomes permissionless" is a security-relevant change to a deployed program and shouldn't be discovered by reading the diff of a PR titled "add V5 fill-status payer float." It also warrants a note for whoever signs off on the next deploy.
2. **Confirm the IDL change downstream.** The `signer` account flips to `isSigner: false` in the published IDL. Clients built against the old IDL keep working, but anything that introspects the IDL to decide which keypairs a `closeFillPda` transaction needs — or that asserts on the account-meta shape — will see a different value. Worth a quick check against the relayer repo and a note in the release notes for `@across-protocol/contracts`.

Minor: the field is now named `signer` while explicitly not being one. Renaming would break the IDL account name, so keeping it is the right call, but the `/// CHECK` line and the updated `lib.rs` doc comment are now the only thing preventing a misreading — worth a brief inline note that the name is retained for IDL compatibility.

droplet-rl (changes requested):
## Assessment

I read the full diff against `reinis/acp-184-step-2-source-deposit`, built with `cargo check -p svm-spoke --features test` (clean, pre-existing warnings only), and ran the unit tests.

**The core mechanics look right.** I specifically checked and could not break:

- The account-creation sequence correctly mirrors Anchor 0.31.1's `generate_create_account`. The `current_lamports == 0` split is necessary — the System Program's `create_account` bails with `AccountAlreadyInUse` when `to.lamports() > 0` — so the one-lamport prefunding DoS is genuinely handled, and an attacker can only transfer lamports to the fill-status PDA (`allocate`/`assign` need the PDA signature), so there is no way to poison it into a non-empty state.
- `V5_FILL_STATUS_SPACE` is 45 bytes → rent-exempt minimum ≈ 1,204,080 lamports, comfortably above the 890,880 floor for the 0-data payer PDA, so permissionless close can never leave the float rent-paying.
- The front-run vector I went looking for is closed: `request_slow_fill` rejects V5-tagged messages (`slow_fill.rs:52`) and the relay hash binds the message, so nobody can squat the fill-status PDA ahead of a V5 fill.
- Float scoping is sound — `create_v5_fill_status` records the same PDA it spends from, and `WithdrawV5FillPayer`'s seed constraint binds withdrawal to the signing submitter (the `ConstraintSeeds` test confirms).

**What I'm asking for.** My concerns are not exploitable bugs; they're about what the last two commits quietly removed and what the PR body claims.

`13821579 refactor(svm): align fill payer with Anchor rent handling` and `3594622c refactor(svm): simplify V5 fill-status creation` deleted **every Rust unit test this PR added**, along with the `RelayFilled` replay error and the rent-remainder guard. Both are labelled as refactors, which is exactly where this kind of removal escapes review. Concretely, on HEAD:

```
$ cargo test -p svm-spoke --lib
test result: ok. 11 passed
```

11, not the 12 in the Validation section — and none of them are new. The only test-file change in the whole diff is the error-range assertion in `v5.rs`. So "add focused unit and validator coverage for signerless creation, replay rejection, …" is no longer accurate: unit coverage is zero, and the branchy creation helper is covered only by the two validator tests (happy path + replay). Please either restore the tests or correct the summary — right now the body over-states what was validated.

The individually actionable items are inline. The highest-value ones: restore the `RelayFilled` replay error (relayers need to distinguish "already filled, stop" from "malformed account, retry"), restore the layout-invariant assertion (it guards the exact property in the PR title), and cover the prefunded branch.

**One item needing explicit sign-off:** `CloseFillPda` relaxes `Signer` → `UncheckedAccount` on an instruction that is live on mainnet today. I believe it is safe — rent still lands on `fill_status.relayer`, and both legacy fills and slow-fill requests record a real beneficiary there — but it is an authorization relaxation plus an IDL signer-flag change to a deployed program, folded into a PR titled "add V5 fill-status payer float." It should be called out in the description and confirmed against downstream clients rather than discovered in the diff. Details inline.

The reserved `Fill` adapter branch is confirmed still fail-closed (`v5_adapter.rs:45` → `UnsupportedMode`), and the `test` feature is properly segregated from the production artifact in `buildSolanaVerify.sh` / `publish.yml`.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=ff13970c482a58da4a5047fac80c3d94&trace_start_ts=1788192180&trace_end_ts=1788278880)

programs/svm-spoke/src/instructions/v5_fill_status.rs
Reinis-FRP: Fixed in `ece3f332`: the helper now calls `derive_v5_fill_payer` and `derive_fill_status`, so both derivations remain under the golden-fixture coverage. I left passing bumps / switching to `create_program_address` for Step 4, when the hot-path account ABI is finalized and its CU tradeoff can be measured.

---

programs/svm-spoke/src/instructions/fill.rs
Reinis-FRP: Addressed in `ece3f332` and the PR description. The description and V5 adapter spec now explicitly call out that existing `close_fill_pda` accounts become permissionless after expiry, that the immutable recorded `relayer` remains the destination, and that regenerated IDLs flip only the signer requirement while retaining the account name.

I also checked downstream usage: the SDK helper currently accepts a `TransactionSigner` and passes it to the generated close instruction, while the in-repo scripts/tests still sign; those old-IDL callers remain valid because the signature is merely redundant. I found no direct close caller in the relayer repositories checked locally. The compatibility note is now visible for release/deploy signoff.

---

programs/svm-spoke/src/instructions/v5_fill_status.rs
Reinis-FRP: Fixed in `ece3f332`: `submitter` is now a `Signer` account in the test-only wrapper, the unauthenticated instruction argument is gone, and the TS caller signs with that keypair. The helper also has a `# Safety` precondition stating that Step 4 must source `submitter` from Gateway-attested context.

---

programs/svm-spoke/src/instructions/v5_fill_status.rs
Reinis-FRP: Addressed without restoring the custom guard. The program and V5 spec now explicitly document that partial withdrawals are subject to the runtime rent-state rule, keeping this path aligned with the System Program / Anchor behavior. `1dfa69f9` pins the failure atomically when a withdrawal would leave a nonzero rent-unsafe remainder; the same test covers an exact rent-floor remainder and the `u64::MAX` drain. I left zero withdrawals unchanged because the submitter can already emit events for its own float, so rejecting zero would not provide an integrity or anti-spam boundary.

---

programs/svm-spoke/src/instructions/v5_fill_status.rs
Reinis-FRP: Covered in `ece3f332`. The validator test now exercises both the zero-balance `create_account` branch and the prefunded transfer / allocate / assign branch, checks the payer delta, and reclaims both accounts. Under Agave 4.1.2, transferring only one lamport to the absent PDA is itself rejected by the runtime rent rule, so the test uses the smallest valid prefund (`minimum_balance(0)`) and still verifies that creation correctly tops it up. The PR description no longer overclaims coverage for the other validation branches.

---

programs/svm-spoke/src/instructions/v5_fill_status.rs
Reinis-FRP: I kept this test removed intentionally. The prior assertion only pinned the current type against a duplicated numeric size while round-tripping that same type; it did not prove compatibility with historical bytes. V5 now directly uses the shared `FillStatusAccount`, `INIT_SPACE`, discriminator, and serializer, so there is no separate V5 layout implementation to drift. A true legacy ABI invariant would be better added centrally as a golden-byte fixture rather than as a V5 helper test. I also removed the inaccurate unit-test claim from the PR description.

---

programs/svm-spoke/src/instructions/v5_fill_status.rs
Reinis-FRP: Fixed in `ece3f332`, including the slow-fill distinction. A program-owned account is now deserialized: `Filled` returns `CommonError::RelayFilled`, `RequestedSlowFill` is accepted and replaced as `ReplacedSlowFill`, and `Unfilled` is accepted as `FastFill`. The validator test creates a real requested slow fill and confirms that replacement succeeds without debiting the V5 payer float.

---

programs/svm-spoke/src/instructions/v5_fill_status.rs
```diff
@@ -0,0 +1,180 @@
...
+/// The caller must source `submitter` from Gateway-attested context; this helper validates accounts derived from that
+/// value but does not authenticate the value itself.
+#[allow(dead_code)] // Called when Step 4 enables the reserved Fill adapter branch.
+pub fn create_v5_fill_status<'info>(
+    payer: &AccountInfo<'info>,
+    fill_status: &AccountInfo<'info>,
+    system_program_info: &AccountInfo<'info>,
+    submitter: &Pubkey,
+    relay_hash: &[u8; 32],
+    fill_deadline: u32,
```
droplet-rl, line 39: `relay_hash` and `fill_deadline` arrive as independent parameters and nothing checks that the deadline is the one committed inside the hash. Every other writer of this field binds them: `fill_relay` and `request_slow_fill` both carry `is_relay_hash_valid(&relay_hash, &relay_data, &state)`, so the invariant "stored `fill_deadline` == the value committed in `relay_hash`" holds everywhere in the program today. This helper is the first place that can break it, and the new replacement path means it can now overwrite a deadline that `request_slow_fill` had already validated.

Why it matters downstream: the stored `fill_deadline` is the sole input to `close_fill_pda`'s expiry check, and that instruction is permissionless as of this PR. A caller passing a too-small deadline writes a fill status that anyone can close immediately, which removes the replay guard and reopens the relay to a second fill.

Relatedly, the helper performs no expiry check of its own (it has no `state` access, so it can't). Legacy `fill_relay` rejects `relay_data.fill_deadline < current_time`. Without that check in the Step 4 caller, an expired-then-closed V5 fill status can simply be re-created for the same relay hash.

Neither is reachable today — the helper is `#[allow(dead_code)]` and the only caller is test-gated — so this is purely about not losing the constraint between PRs. Two options:

- Cheapest: extend the `# Safety` block, which currently covers only `submitter`, to state that the caller must source `fill_deadline` from the same `RelayData` that produced `relay_hash`, and must reject deadlines already in the past.
- Sturdier: take `&RelayData` plus the chain ID and derive the hash internally, mirroring `is_relay_hash_valid`, so the binding is structural instead of contractual.

Either is fine by me; I'd just rather it be decided here than rediscovered in Step 4 review.

programs/svm-spoke/src/instructions/v5_fill_status.rs
```diff
@@ -0,0 +1,180 @@
...
+    require!(payer.data_is_empty(), V5Error::InvalidFillPayer);
+    let (fill_type, needs_init) = if fill_status.owner == &ID {
+        let data = fill_status.try_borrow_data()?;
+        let existing = FillStatusAccount::try_deserialize(&mut &data[..])
+            .map_err(|_| error!(V5Error::InvalidFillStatusAccount))?;
+        let fill_type = match existing.status {
+            FillStatus::Unfilled => FillType::FastFill,
+            FillStatus::RequestedSlowFill => FillType::ReplacedSlowFill,
+            FillStatus::Filled => return err!(CommonError::RelayFilled),
+        };
```
droplet-rl: Is the `RequestedSlowFill` arm reachable for a genuine V5 relay? As far as I can trace, no:

- A V5 relay's message is `V5_MAGIC_PREFIX || step_id`, so `is_v5_message` returns true.
- `request_slow_fill` rejects V5-tagged messages with `V5FillOnly` (`slow_fill.rs:52`).
- The relay hash commits to `message`, so nobody can request a slow fill under alternate relay data that hashes to a V5 relay's hash.

So no V5 relay hash can ever carry a `RequestedSlowFill` status. The new test reaches this branch only by constructing a non-V5 relay (`message: Buffer.alloc(0)`) and then invoking the V5 helper on it — a combination Step 4 would never produce, since the adapter only handles V5-tagged relays.

The spec already acknowledges this ("although new V5-tagged slow-fill requests are disabled"), so I assume it's deliberate symmetry that becomes live if V5 slow fills are ever enabled. Fine by me if so — just flagging that it's currently unreachable surface whose only coverage exercises a shape production can't produce, in case that wasn't the intent.

Same question for the `Unfilled` arm: `fill_relay` and `request_slow_fill` both set a terminal status in the same instruction that initializes the account, so a persisted `Unfilled` account shouldn't exist either. Harmless as a defensive default, just noting it.

programs/svm-spoke/src/instructions/v5_fill_status.rs
```diff
@@ -0,0 +1,180 @@
...
+use crate::{
+    constants::{DISCRIMINATOR_SIZE, FILL_STATUS_SEED, V5_FILL_PAYER_SEED},
+    error::{CommonError, V5Error},
+    event::{FillType, V5FillFloatWithdrawn},
+    state::{FillStatus, FillStatusAccount},
+    v5::{derive_fill_status, derive_v5_fill_payer},
+    ID,
+};
+
+pub const V5_FILL_STATUS_SPACE: usize = DISCRIMINATOR_SIZE + FillStatusAccount::INIT_SPACE;
```
droplet-rl, line 15: Nit, carried over and downgraded: the layout assertion deleted in `3594622c` is still gone.

```rust
assert_eq!(V5_FILL_STATUS_SPACE, 8 + 1 + 32 + 4);
```

Correcting the Validation section to "11 passed" resolves my actual complaint — the description no longer claims coverage that isn't there — so this is no longer a blocker. But the invariant itself is still unguarded: `V5_FILL_STATUS_SPACE` derives from `INIT_SPACE`, so it silently tracks any field added to `FillStatus` or `FillStatusAccount`, and a change there would shift the on-chain layout and rent math on both the legacy and V5 paths with nothing failing. Given "without changing its account layout" is a bullet in this PR's summary, a three-line `#[test]` pinning the constant seems worth keeping around.

droplet-rl (approved):
## Assessment

All seven points from my previous review are addressed, and the two follow-up commits are a genuine improvement rather than a minimal patch. Re-verified locally on `1dfa69f9`:

```
$ cargo check -p svm-spoke --features test    # clean, no new warnings
$ cargo test -p svm-spoke --lib
test result: ok. 11 passed
```

11 matches the corrected Validation section.

**Point-by-point:**

| Previous finding | Resolution |
|---|---|
| Duplicated PDA derivation | Now imports `derive_v5_fill_payer` / `derive_fill_status` from `v5.rs` — back under the golden-fixture test |
| Replay surfaced as `InvalidFillStatusAccount` | `CommonError::RelayFilled` restored (L56), TS asserts on it |
| Prefunded branch untested | Covered, and the assertions are stronger than I asked for — see below |
| Rent-remainder guard removed silently | Pinned by test (`"insufficient funds for rent"`) and documented in both the spec and the `lib.rs` doc comment |
| Test entrypoint took unauthenticated `submitter` | Now `Signer`, argument dropped |
| Missing caller precondition on `submitter` | `# Safety` section added |
| `close_fill_pda` relaxation undocumented | Dedicated Compatibility note in the description + recorded in the spec for release coordination |

The prefunded-path test is worth calling out: rather than just asserting the account exists, it tracks the float across both creation modes and both closes, landing on `initialFloat + prefundedLamports`. That arithmetic only balances if the donated lamports are correctly absorbed as a top-up rather than double-paid, which is exactly the property the branch exists to guarantee. Good test.

**On the new slow-fill replacement path.** I checked whether overwriting `relayer` and `fill_deadline` when replacing a `RequestedSlowFill` lets a V5 submitter capture rent that the slow-fill requester paid. It does — but legacy `fill_relay` already does precisely the same thing (`fill.rs:161-163`), so this faithfully mirrors existing semantics rather than introducing new behaviour. Documenting it as "preserve standard fill-status semantics" is accurate. `try_borrow_data` on L50 is scoped to the `if` block and released before the `try_borrow_mut_data` on L107, so no borrow conflict.

**Approving.** Nothing here blocks. Three non-blocking items inline; the first is the one I'd least like to see slip, because it lands in Step 4 rather than here — `relay_hash` and `fill_deadline` are independent parameters with nothing binding them, and this helper is the first place in the program where the stored `fill_deadline` can diverge from the value committed in the relay hash. Worth closing off in the `# Safety` contract now while it's fresh.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=92ff513841ff43004a7f793233b06070&trace_start_ts=1788202473&trace_end_ts=1788289173)

programs/svm-spoke/src/instructions/v5_fill_status.rs
Reinis-FRP: Fixed in [`b2474f10`](https://github.com/across-protocol/contracts/commit/b2474f10e81680c7e6bdab74b182e3e29d54327b). The helper's `# Safety` contract now requires the caller to derive `relay_hash` from the same validated V5 `RelayData` that supplies `fill_deadline`, and to reject an expired deadline before calling the helper. Step 4 should enforce those conditions structurally in its caller.

---

programs/svm-spoke/src/instructions/fill.rs
Reinis-FRP: Follow-up after exercising the regenerated IDL in CI: the on-chain compatibility conclusion stands, but the client wording needed one distinction. Old-IDL clients keep sending the signer meta/signature and remain compatible. Anchor clients generated from the new IDL must omit the explicit `.signers([relayer])`; otherwise Web3 rejects it locally as an unknown signer because no instruction account is marked signing.

Fixed in [`b2474f10`](https://github.com/across-protocol/contracts/commit/b2474f10e81680c7e6bdab74b182e3e29d54327b): the in-repository close callers now omit that explicit signer, verified-SVM CI regenerates the test-feature `svm_spoke` IDL, and the PR compatibility note documents both client cases.

---

programs/svm-spoke/src/instructions/v5_fill_status.rs
Reinis-FRP: Acknowledged. I am leaving the duplicate layout assertion out of this narrowly scoped change: the V5 path intentionally uses the shared `FillStatusAccount` type and its shared `INIT_SPACE`, so a V5-local assertion would not independently protect the legacy layout and could imply stronger coverage than it provides. If we add a golden byte-layout invariant, it should live centrally with the shared account definition and cover serialization, not only repeat the derived size expression.

---

programs/svm-spoke/src/instructions/v5_fill_status.rs
Reinis-FRP: @droplet-rl I checked this against the EVM V5 design before changing it. `contracts-v5/docs/SPOKE_V5_FILLS.md` explicitly quarantines V5-tagged deposits from slow-fill requests and executions, so `RequestedSlowFill` is not an intended V5 state; a persisted `Unfilled` state is likewise unreachable through the existing writers.

Fixed in [`b2474f10`](https://github.com/across-protocol/contracts/commit/b2474f10e81680c7e6bdab74b182e3e29d54327b): I removed the speculative replacement behavior and the synthetic non-V5 test. The helper now accepts only an uninitialized PDA, returns `RelayFilled` for `Filled`, and rejects either remaining program-owned enum state as `InvalidFillStatusAccount`. The spec now records the same V5 slow-fill quarantine. This also matches the intended Solana lite-chain direction; removing the general SVM slow-fill instruction remains outside this PR. Please recheck this piece.

---

droplet-rl: Rechecked `b2474f10`. The quarantine change is correct — and I confirmed it independently of the EVM doc rather than taking it on faith:

- `request_slow_fill` returns `V5FillOnly` for `is_v5_message` (`slow_fill.rs:52`), and `is_v5_message` matches on the first 32 bytes being `V5_MAGIC_PREFIX`.
- The relay hash commits to `message`, and `is_relay_hash_valid` binds the two, so nobody can request a slow fill under alternate relay data that hashes to a V5 relay's hash.

So neither `RequestedSlowFill` nor `Unfilled` is reachable for a V5-tagged relay, and rejecting both as `InvalidFillStatusAccount` is strictly more conservative than replacing them. Fail-closed on a state that can't occur is the right call for this program. Dropping `-> Result<FillType>` follows correctly, since the only reachable V5 outcome was `FastFill`. Verified on `b2474f10`: `cargo check -p svm-spoke --features test` clean, `cargo test -p svm-spoke --lib` 11 passed.

One thing the refactor dropped that I don't think was intentional. Flattening the `if/else` removed both validations that used to guard the non-program-owned path:

```rust
require_keys_eq!(*fill_status.owner, system_program::ID, V5Error::InvalidFillStatusAccount);
require!(fill_status.data_is_empty(), V5Error::InvalidFillStatusAccount);
```

Control now falls straight from the `owner == &ID` check into the creation sequence, so a `fill_status` owned by neither svm-spoke nor the System Program reaches `create_account`/`allocate` unvalidated.

Not exploitable, and I checked why: assigning or allocating that PDA requires its own signature, which only svm-spoke can produce via `invoke_signed`, so the owner is always svm-spoke or the System Program, and a system-owned one always has empty data. Even if it were reachable, the System Program's `allocate` returns `AccountAlreadyInUse` for a non-empty or non-system-owned account, so it fails safe. It's purely about keeping the validate-then-act shape and a program error instead of a bare System Program one:

```rust
if fill_status.owner == &ID {
    // existing replay / invalid-state handling
} else {
    require_keys_eq!(*fill_status.owner, system_program::ID, V5Error::InvalidFillStatusAccount);
    require!(fill_status.data_is_empty(), V5Error::InvalidFillStatusAccount);
}
```

Your call — I didn't push it, since you're mid-iteration on the branch and I'd rather not race your agent. Happy to if you want it.

Two smaller notes on the rest of the commit:

**CI change is correct and was actually needed.** The verified-test job deploys a binary built with `IS_TEST=true` but was regenerating only the `mock_gateway` IDL with `--features test`, so `target/types/svm_spoke.ts` wouldn't have carried `testCreateV5FillStatus` and the new test would have failed to build the instruction. Overwriting `target/idl/svm_spoke.json` is scoped to that job, so it doesn't leak into the published artifact.

**The `.signers([relayer])` removals.** I checked whether the regenerated IDL breaks the compatibility note, since Anchor builds the account meta as `isSigner: false` and the old client still passes the keypair. On the pinned `@solana/web3.js` 1.98.2 it does not — signing with a key that isn't a required signer is tolerated with a deprecation warning, not an error:

```
Transaction references a signature that is unnecessary, only the fee payer and
instruction signer accounts should sign a transaction. This behavior is
deprecated and will throw an error in the next major version release.
```

So the compat note holds as written. Worth knowing that web3.js states it will throw in the next major, and that removing the signer from both `SvmSpoke.Fill.ts` and `SvmSpoke.SlowFill.AcrossPlus.ts` leaves the repo with no coverage of the old-client path. Keeping `.signers([relayer])` at one of the two call sites and dropping it at the other would exercise both shapes for free.

Separately, the `# Safety` block now covers the `relay_hash`/`fill_deadline` binding and the expiry precondition — that fully addresses the point I flagged as the one I least wanted to see slip into Step 4. My approval stands.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=ce8ec27a096b24de0726218193333f3a&trace_start_ts=1788208177&trace_end_ts=1788294877)

test/svm/SvmSpoke.V5FillStatus.ts
```diff
@@ -0,0 +1,173 @@
...
+          fromPubkey: providerPayer.publicKey,
+          toPubkey: payer,
+          lamports: initialFloat,
+        })
+      ),
+      [providerPayer]
+    );
+
+    await program.methods
+      .testCreateV5FillStatus([...relayHash], fillDeadline)
```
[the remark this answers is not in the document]
Reinis-FRP: Fixed in [`b90ba45a`](https://github.com/across-protocol/contracts/commit/b90ba45ac4e88dca28a2eca87e19da802efaccf6). Both standard SVM test commands now generate normal production artifacts first, then call a shared `generate-svm-test-idls` helper before `anchor test`. Verified-SVM CI uses the same helper.

The helper writes only to `target/idl` and `target/types`; it never copies test-feature IDLs, mock-gateway artifacts, or generated clients into `src/svm/assets` or `src/svm/clients`. I verified that the production `svm_spoke` IDL excludes `test_create_v5_fill_status`, the target test IDL includes it, package assets remain unchanged, and `yarn build-ts` passes.

---

Reinis-FRP:
@droplet-rl please recheck this PR after the fill-status account helper responsibility was moved here from #1538. The helper now only creates or assigns account storage and returns the validated payer PDA; Step 4 owns the semantic `Filled` transition.

---

droplet-rl:
Rechecked at `c7fe69fe`. Verified locally: `cargo check -p svm-spoke` and `cargo check -p svm-spoke --features test` both clean, `cargo test -p svm-spoke --lib` 11 passed. `test/svm/` and `V5_ADAPTER_SPEC.md` are unchanged since `b2474f10`.

The split reads well, and returning `Ok(expected_payer)` is a nice touch — it hands the caller the already-validated float address instead of making Step 4 re-derive it, which also avoids a redundant PDA derivation in the fill hot path.

Two things the responsibility move introduces that I'd want pinned down before #1538 lands.

**The helper can now leave an unrecoverable account.** On success it returns with the PDA program-owned and its data all zeroed, including the 8-byte discriminator. If any Step 4 success path returns without serializing, the persisted result is a program-owned account with a zero discriminator, and that state is terminal in both directions:

- `close_fill_pda` takes `fill_status: Account<'info, FillStatusAccount>`, so the discriminator check fails with `AccountDiscriminatorMismatch` before `address = fill_status.relayer` can even be evaluated. The rent is stranded permanently.
- A retry hits `fill_status.owner == &ID`, `try_deserialize` fails, and returns `InvalidFillStatusAccount`. That relay hash becomes permanently unfillable.

Not attacker-reachable — it needs a bug on Step 4's success path, not hostile input — and before this refactor it was unreachable by construction, because the helper wrote a valid record itself. So it's new surface that depends entirely on caller discipline.

Worth noting the doc's "mirroring Anchor's exit behavior" is a little generous in that light: Anchor's exit is generated and runs unconditionally at instruction end, so you cannot forget it. Here it's a manual step with no compiler or runtime backstop. If Step 4 can be shaped so the write is the only path out — finalize immediately after the call, no early returns or `?` between them — that closes it. Otherwise it's worth an explicit invariant in the doc comment rather than the current framing.

**The validator coverage of finalization went tautological.** `test_create_v5_fill_status` now performs the write itself ("Mirror production finalization"), so these assertions no longer say anything about production code:

```ts
assert.hasAnyKeys(account.status, ["filled"]);
assert.equal(account.relayer.toBase58(), payer.toBase58());
assert.equal(account.fillDeadline, fillDeadline);
```

Line 61 was the one proving reclaim routes back to the float, and the close-to-`payer` chain later in that test rests on it. It now proves only that the test's own write agrees with itself. The rest of the test — signerless creation, prefunded top-up, replay rejection, rent accounting — still exercises the real helper and remains valuable. If #1538 factors the finalization into a helper both the test entrypoint and the Fill branch call, these regain their meaning for free; otherwise the two writes can drift silently.

Smaller items:

- The `# Safety` block traded "reject an expired `fill_deadline`" for "complete semantic validation". Defensible now that `fill_deadline` isn't a parameter, but the specific invariant I flagged earlier — the stored deadline must be the value committed in `relay_hash`, and must not already be expired — no longer has a home anywhere in this PR. It should land wherever the serialization now lives.
- The `else`-branch `require_keys_eq!(*fill_status.owner, system_program::ID, ...)` / `require!(fill_status.data_is_empty(), ...)` pair is still absent. Unreachable and fail-safe as I described last round, so purely your call; just noting it stayed open rather than assuming it was declined.
- `b90ba45a` is a better fix than what I suggested. Extracting `buildTestIdls.sh` and threading it through both `test-svm` and `test-svm-solana-verify` also repairs the local path, since `IS_TEST=true` only applied to `build-svm` and so `generate-svm-artifacts` was stripping the mock IDL, and it adds the `--out-ts target/types/mock_gateway.ts` the inline CI step had omitted.

Nothing here blocks this PR — both findings are about the contract #1538 has to honor, and the code in front of me is correct as written. My approval stands.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=ee2ddbf0c37965437947e8a809cc3167&trace_start_ts=1788779836&trace_end_ts=1788866536)

Reinis-FRP:
@droplet-rl could you please recheck the Step 3 side of the duplicate-PDA-derivation fix in [558d923e](https://github.com/across-protocol/contracts/commit/558d923ef28bbc7e9de7e918ea9014565ac591b0)?

It introduces a private-field `V5FillStatusPdas` bundle whose sole constructor derives both canonical keys and bumps. `create_v5_fill_status_account` now validates against that bundle and reuses its bumps for signing, so callers cannot supply unchecked raw bumps. I also added focused rejection coverage for noncanonical payer and fill-status accounts. Local validation passed: 11/11 Rust tests and 3/3 V5 fill-status validator tests.

I’ll rebase Step 4 and plumb this same bundle through its adapter account loader next, removing the duplicated hot-path derivations identified in [the PR #1538 thread](https://github.com/across-protocol/contracts/pull/1538#discussion_r3951472324).

---

droplet-rl:
Rechecked the Step 3 side of `558d923e`. The bundle shape is the right direction — a private-field type whose only constructor derives canonically is strictly better than accepting raw `u8` bumps. But it doesn't yet enforce the invariant it's meant to, and I found one regression it introduced that I've pushed a fix for.

**The bundle isn't bound to the seeds used for signing.** The helper now takes `submitter`, `relay_hash`, *and* `pdas`, with nothing tying them together. Validation reads from the bundle, but the signing seeds mix the loose parameters with the bundle's bumps:

```rust
let payer_seeds: &[&[u8]] = &[V5_FILL_PAYER_SEED, submitter.as_ref(), &[pdas.payer_bump]];
let fill_status_seeds: &[&[u8]] = &[FILL_STATUS_SEED, relay_hash, &[pdas.fill_status_bump]];
```

A caller that passes a bundle derived from different values than the `submitter`/`relay_hash` it also passes gets account keys checked against one set and signatures attempted with a hybrid of both. It fails closed — `invoke_signed` derives an address that doesn't match the account, so the System Program rejects it as an unauthorized signer — so this isn't exploitable. The problem is that "callers cannot supply unchecked raw bumps" is now true only in the narrow sense; they can still supply bumps that don't correspond to the seeds actually signed with. The previous code couldn't get this wrong, because everything came from one local derivation.

That matters specifically because of what you're about to do next. Threading the bundle through Step 4's adapter account loader means it will be constructed further from its use site, which is exactly where a stale or mismatched bundle becomes plausible. Making the bundle the single source of truth closes it — carry the seed material and drop the loose parameters, so `create_v5_fill_status_account(payer, fill_status, system_program_info, pdas)` has no way to disagree with itself:

```rust
pub struct V5FillStatusPdas<'a> {
    submitter: &'a Pubkey,
    relay_hash: &'a [u8; 32],
    payer: Pubkey,
    fill_status: Pubkey,
    payer_bump: u8,
    fill_status_bump: u8,
}
```

Borrowing rather than copying keeps this off the stack, which seems worth preferring here given `fill.rs:92` already documents having to move a constraint into the handler to avoid exceeding the maximum stack offset. Your call on the exact shape — the point is that one derivation should feed both the validation and the signing.

**Regression, fixed and pushed as `2f94cfa7`.** The production build started warning:

```
warning: associated function `derive` is never used
```

Confirmed new — `cargo check -p svm-spoke` gives 0 such warnings at `7de27b68` and 1 at `558d923e`. `derive` is only reached from the `test`-gated entrypoint, and it's the one of the three new helpers that didn't get an attribute: `create_v5_fill_status_account` has `#[allow(dead_code)]` and `write_v5_fill_status` has `#[cfg_attr(not(feature = "test"), allow(dead_code))]`. I applied the latter, since it matches `write_v5_fill_status`'s situation exactly and keeps the lint live under `--features test`. Worth noting why your validation missed it: `cargo check --features test` is clean because the test entrypoint calls it, and `yarn lint-rust` "passed with existing warnings" absorbed the new one. A production-config check would have caught it.

After the fix: `cargo check -p svm-spoke` 0 dead-code warnings, `--features test` clean, `cargo test -p svm-spoke --lib` 11 passed, `cargo +nightly fmt --all -- --check` clean.

Both findings from my last round landed well, and I want to flag that explicitly since they were the ones I cared most about:

- `9d74419b` extracting `write_v5_fill_status` fixes the tautological coverage. The validator assertions on `status`, `relayer`, and `fillDeadline` now exercise the same function Step 4 will call, so they're evidence about production again. The Safety block also now states the bricking invariant outright — "Every successful instruction path must then call `write_v5_fill_status` with the unexpired deadline committed in that `RelayData`; failed paths atomically roll back the zeroed intermediate account" — which both documents the hazard and restores the "unexpired deadline" wording that had gone missing.
- `7de27b68` renaming to `rent_recipient` and correcting the `FillStatusAccount.relayer` doc comment resolves the misnomer I raised back on `CloseFillPda`, without touching the layout.
- The new rejection test covers the `InvalidFillPayer` and `InvalidFillStatusAccount` address-validation branches I'd flagged as untested two rounds ago.

The bundle-binding item is a design cleanup rather than a defect, so it doesn't change my approval. It's just much cheaper to settle before the Step 4 plumbing than after.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=e1430d93b3338182926f8180fda5dfb9&trace_start_ts=1788857537&trace_end_ts=1788944237)

Reinis-FRP:
@droplet-rl thanks for catching the remaining split source of truth. Fixed in [0c46ac95](https://github.com/across-protocol/contracts/commit/0c46ac951d1f41bbc2ce5d5d68c076aa2148624f): `V5FillStatusPdas` now retains references to the authenticated submitter and validated relay hash, and `create_v5_fill_status_account` accepts only that bundle rather than separate loose seed arguments. Both account checks and signer seeds therefore use the same construction. Production check, 11/11 Rust tests, SBF build, and 3/3 focused validator tests pass. Could you please recheck Step 3 before I finish rebasing the Step 4 plumbing?

---

droplet-rl:
Rechecked `0c46ac95`. The split source of truth on the creation path is properly closed — the bundle owns the seed material, `create_v5_fill_status_account` takes only the bundle, and both the key checks and the signer seeds now read from it. I confirmed the type can't be built inconsistently: fields are private, `derive` is the only constructor, there's no `Default`/`Clone` escape hatch, and nothing outside the module references the type.

Verified locally at `0c46ac95` (my `2f94cfa7` is intact underneath it):

- `cargo check -p svm-spoke` — 0 errors, 0 dead-code warnings
- `cargo check -p svm-spoke --features test` — clean
- `cargo test -p svm-spoke --lib` — 11 passed
- `cargo +nightly fmt --all -- --check` — clean

**One seam of the same kind is still open, and it's the more dangerous one.** `write_v5_fill_status` still takes the rent recipient as a loose value:

```rust
pub fn write_v5_fill_status(fill_status: &AccountInfo<'_>, rent_recipient: Pubkey, fill_deadline: u32) -> Result<()>
```

It arrives hand-to-hand from `create_v5_fill_status_account`'s return. The bump mismatch you just fixed failed closed at `invoke_signed`, so the worst case was a rejected transaction. This one doesn't fail at all: `write_v5_fill_status` serializes whatever it's given, and `close_fill_pda` later credits whatever `fill_status.relayer` holds, with no validation in between. A wrong value here silently misdirects reclaimed rent. Much less likely to be hit than the bump case, but strictly worse if it is.

Taking the bundle completes the pattern and makes it unrepresentable:

```rust
pub fn write_v5_fill_status(fill_status: &AccountInfo<'_>, pdas: &V5FillStatusPdas<'_>, fill_deadline: u32) -> Result<()>
```

reading `pdas.payer()` internally, which also lets `create_v5_fill_status_account` return `Result<()>` and drops the `rent_recipient` threading from the test entrypoint. The one thing it costs is needing the bundle still in scope after token delivery — which it will be, given you're plumbing it through the loader anyway.

I deliberately didn't push this one. It changes the signature Step 4 is being rebased onto right now, and landing it mid-rebase would just hand you a conflict. Better folded in on your side, or deferred to the Step 4 PR if you'd rather not churn Step 3 again.

**Heads-up for the plumbing, since you asked before finishing it.** The borrow shape means `V5FillStatusPdas<'a>` cannot outlive the `Pubkey` and `[u8; 32]` it points at. If the adapter account loader computes the relay hash into a local and then returns the bundle, that won't compile — the borrow escapes. The relay hash needs to live in the enclosing scope, alongside the Gateway-attested submitter, rather than being derived inline in the loader. Worth arranging deliberately rather than discovering it as a borrow-checker fight, since the obvious factoring is the one that fails.

Approval stands. Nothing here blocks Step 3.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=bd381ff7d0a49540637aa61dab3f4831&trace_start_ts=1788858374&trace_end_ts=1788945074)

Reinis-FRP:
@droplet-rl addressed the remaining finalization seam in [d4e59080](https://github.com/across-protocol/contracts/commit/d4e59080a1d69ba418826ca5c7080cb537747ad7). `write_v5_fill_status` now takes the same canonical PDA bundle and reads the rent recipient from it; `create_v5_fill_status_account` returns `Result<()>`, so no loose payer value is threaded between creation and finalization. Production and test-feature checks, 11/11 Rust tests, SBF build, and 3/3 focused validator tests pass. Could you please give the finished Step 3 branch one final recheck?

---

droplet-rl:
Rechecked `d4e59080` and did a full-branch pass. The finalization seam is closed correctly — `write_v5_fill_status` reads the rent recipient from the bundle, `create_v5_fill_status_account` returns `Result<()>`, and no bare `Pubkey` is threaded between the two. The Safety block now mandates "the same bundle" for both calls, which is the right thing to have written down.

Verified at `d4e59080`:

- `cargo check -p svm-spoke` — 0 errors, 0 dead-code warnings
- `cargo check -p svm-spoke --features test` — same
- `cargo test -p svm-spoke --lib` — 11 passed
- `cargo clippy -p svm-spoke --lib --tests` — no errors; the `too_many_arguments` warnings are pre-existing in `admin.rs`, `deposit.rs`, and `lib.rs`, none in `v5_fill_status.rs`. Both helpers are now under the threshold, where the earlier six-argument shape was closer to it
- `cargo +nightly fmt --all -- --check` — clean

Scoping my sign-off honestly: everything above is mine. I have no Solana toolchain here, so the SBF build and the 3/3 validator results are yours and I haven't reproduced them — I've read `SvmSpoke.V5FillStatus.ts` and the entrypoint ABI is unchanged by these last three commits, so I'd expect them to hold, but that's inference rather than execution.

**One pair is still unbound, and then I'm done finding these.** `write_v5_fill_status` takes `fill_status: &AccountInfo` alongside `pdas`, with no check that they agree:

```rust
pub fn write_v5_fill_status(
    fill_status: &AccountInfo<'_>,
    pdas: &V5FillStatusPdas<'_>,
    fill_deadline: u32,
) -> Result<()>
```

In the mandated sequence this is redundant — `create_v5_fill_status_account` already validated that same pair a few lines earlier. The reason I'd still consider closing it is the failure mode if it's ever mis-wired: a non-program-owned account fails at the runtime's data-modification check, but a writable svm-spoke-owned account of sufficient size does not. Passing `state` would overwrite the first 45 bytes of the `State` PDA with a `FillStatusAccount` record. That's a different severity class from the two seams we just closed, both of which only ever cost a rejected transaction.

Genuinely a judgment call, and the argument against is real: it's one `require_keys_eq!` of CU in the fill hot path guarding a sequence the docs now spell out, and the repo's own style guidance leans against redundancy. I lean toward adding it because `write_v5_fill_status` is `pub` and Step 4 will call it from a different scope than the creation call, which is exactly where the pair can drift apart. But I didn't push it — you're closer to the Step 4 call-site shape than I am, and if the two calls end up adjacent, the check earns less. Fine to decline or defer to Step 4.

Step 3 looks finished to me. Worth noting how far it moved: the branch started with a replay path that reported the wrong error, no coverage of the prefunded branch, a silently-relaxed authorization rule on a live mainnet instruction, and unit tests that had been deleted by commits labelled as refactors while the description still claimed them. It ends with the replay classified as `RelayFilled`, the prefunded top-up proved by float arithmetic that only balances if the donation is absorbed correctly, the `close_fill_pda` relaxation documented for release coordination, and a single canonical derivation feeding validation, signing, and finalization. Approval stands.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=e3586e05f6193bc0bfb1cf55d2d46336&trace_start_ts=1788858965&trace_end_ts=1788945665)

Reinis-FRP:
@droplet-rl final-head correction: GitHub DCO rejected the unsigned one-line warning-fix commit, so I rebuilt it as signed commit [23586cc6](https://github.com/across-protocol/contracts/commit/23586cc6d58e95a879061bee5d5849b2bf2a0618) with you credited as co-author, then replayed the two reviewed changes. The Step 3 tree is byte-for-byte unchanged; its final head is [b9edf074](https://github.com/across-protocol/contracts/commit/b9edf07490ad78a62dddcce978f473d390025fbd). Please use that head for the requested final recheck.

---

droplet-rl:
Confirmed at `b9edf074`. The "byte-for-byte unchanged" claim holds, and I checked it against the tree objects rather than eyeballing a diff:

```
d4e59080 tree 9ddd8698ecac7b698628a04868b9ed5fc20de988
b9edf074 tree 9ddd8698ecac7b698628a04868b9ed5fc20de988
git diff d4e59080 b9edf074  ->  empty
```

Each replayed commit matches its original at every step, so the rebuild was faithful and not just equal at the tip:

```
2f94cfa7 -> 23586cc6   tree 7e8e888fcd44018ea45777a2d93e7d29c2f4b33d (both)
0c46ac95 -> 4dba9fc3   tree 585b279150a6c9e5d302808f34ec14ba334a60fe (both)
d4e59080 -> b9edf074   tree 9ddd8698ecac7b698628a04868b9ed5fc20de988 (both)
```

Also verified: all three carry `Signed-off-by`, `23586cc6` keeps the one-line `#[cfg_attr(not(feature = "test"), allow(dead_code))]` intact with the `Co-authored-by` credit, and my unsigned `2f94cfa7` is no longer reachable from any remote branch. Thanks for sorting the DCO out properly rather than dropping the fix.

I'm not re-running the build and test suite on this head, and I'd rather say that plainly than imply I did. Identical tree hashes are a stronger guarantee than a repeated run: the compiler inputs are cryptographically proven unchanged from the state I verified at `d4e59080`, so the prod and test-feature checks, 11/11 unit tests, clippy, and fmt results carry over unchanged. My working copy is clean at `b9edf074` and its tree matches what I checked.

For the record, my sign-off was never contingent on the last open item — the unbound `fill_status`/`pdas` pair in `write_v5_fill_status`. It's untouched here and remains your call, best judged against the Step 4 call-site shape.

Step 3 is good to land from my side.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=6353f8d2adcef4adedee7446d9b14d33&trace_start_ts=1788859558&trace_end_ts=1788946258)

Reinis-FRP:
@droplet-rl addressed the final account/bundle seam in [5a8a5350](https://github.com/across-protocol/contracts/commit/5a8a53500feb7e7cb88c412b39dccd3154532e51) without adding a duplicate runtime check. Creation now returns a private-field, `must_use` `PendingV5FillStatus` that owns the exact validated `AccountInfo` clone and canonical PDA bundle reference; its consuming `write_filled` method is the only finalization path. Step 4 keeps that handle across token delivery. Final checks pass: production/test builds, 11/11 Rust tests, SBF build, and 3/3 focused validator tests. Please recheck this final Step 3 head.

---

droplet-rl:
The typestate is a better answer than the runtime check I proposed — it closes the seam at zero CU instead of one comparison. Rechecked `5a8a5350`, and I found one regression, fixed and pushed as `11d8df8e` (signed off this time, so DCO shouldn't bounce it).

**The design.** Sorting the four failure modes by what the type system now guarantees:

- finalize with a mismatched account — structurally impossible; the handle owns the exact validated clone. This is the one I raised, and owning it beats re-checking it.
- finalize twice — impossible; `write_filled(self)` consumes.
- finalize without creating — impossible; private fields, no public constructor.
- create without finalizing — a `#[must_use]` lint. Worth being precise: that's a warning, not enforcement. `let _pending = create(...)?;` with no follow-up still compiles silently, as does `let _ = ...`. It catches the accidental drop, which was the realistic path to the bricked-account state, and Rust has no linear type to do better.

I also checked the two things your Step 4 plan leans on. Holding the handle across token delivery is safe because it carries an `AccountInfo` clone — an `Rc` clone, not a borrow guard — so it neither trips `AccountBorrowFailed` nor blocks the delivery CPI's own borrows. And the `try_borrow_data()` in the replay check sits in a block where every match arm returns, so no `Ref` is ever live on the fall-through path that reaches `write_filled`.

On the `&'a V5FillStatusPdas<'a>` shape: unifying those lifetimes looked like it might over-constrain a caller whose seed data outlives the bundle borrow, so rather than assume covariance saves it I compiled a reduced model of the Step 4 pattern — seed data outliving the bundle, handle held across an intervening call. It compiles. No constraint problem.

**The regression.** Production build picked up two dead-code warnings, 0 at `b9edf074` and 2 at `5a8a5350`:

```
warning: fields `fill_status` and `pdas` are never read
warning: method `write_filled` is never used
```

When `write_v5_fill_status` became `PendingV5FillStatus::write_filled`, its `#[cfg_attr(not(feature = "test"), allow(dead_code))]` didn't come along. Attributing the method alone clears both — making it a dead-code root means its field reads count as uses, so the struct warning goes too.

This is the same class and the same root cause as the one at `558d923e`, and it slipped through for the same reason: `cargo check --features test` is clean because the test entrypoint exercises the path, so the warning only appears in the production config. Given it has now happened twice while moving these helpers around, a `cargo check -p svm-spoke` with no features in your loop would catch it cheaply — that single command is what surfaced both.

After the fix, at `11d8df8e`: production check 0 warnings, `--features test` clean, 11/11 Rust tests, clippy no errors and nothing new in this file beyond the crate-wide `anchor-debug` cfg noise from `#[derive(Accounts)]`, `cargo +nightly fmt --check` clean. SBF build and the 3/3 validator tests remain yours — still no Solana toolchain here — though the entrypoint ABI is untouched by this commit.

Step 3 is good to land from my side.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=b2be2c75fe352cd2d2f69b3c39a77a2a&trace_start_ts=1788860274&trace_end_ts=1788946974)

programs/svm-spoke/src/instructions/v5_fill_status.rs
```diff
@@ -0,0 +1,205 @@
...
+pub fn create_v5_fill_status_account<'a, 'info>(
+    payer: &AccountInfo<'info>,
+    fill_status: &AccountInfo<'info>,
+    system_program_info: &AccountInfo<'info>,
+    pdas: &'a V5FillStatusPdas<'a>,
+) -> Result<PendingV5FillStatus<'a, 'info>> {
+    require_keys_eq!(*payer.key, pdas.payer(), V5Error::InvalidFillPayer);
+    require_keys_eq!(*fill_status.key, pdas.fill_status(), V5Error::InvalidFillStatusAccount);
+    require_keys_eq!(*system_program_info.key, system_program::ID, V5Error::MissingAccount);
+    require!(system_program_info.executable, V5Error::MissingAccount);
```
fusmanii: is it even possible that the system program can be unexecutable?

fusmanii (commented):
makes sense, just one comment

programs/svm-spoke/src/instructions/v5_fill_status.rs
Reinis-FRP: Agreed. Removed the redundant executable assertion in 4371e9d3. The preceding key check already pins this account to the canonical System Program address.

---

fusmanii (approved):

programs/svm-spoke/src/instructions/v5_fill_status.rs
```diff
@@ -0,0 +1,204 @@
...
+    error::{CommonError, V5Error},
+    event::V5FillFloatWithdrawn,
+    state::{FillStatus, FillStatusAccount},
+    v5::pda::{derive_fill_status, derive_v5_fill_payer},
+    ID,
+};
+
+pub const V5_FILL_STATUS_SPACE: usize = DISCRIMINATOR_SIZE + FillStatusAccount::INIT_SPACE;
+
+pub struct V5FillStatusPdas<'a> {
```
fusmanii, line 17: should this have `pdas` in the name given that `submitter` and `payer` are not PDAs?

Reinis-FRP: I lean toward keeping `V5FillStatusPdas` as both `payer` and `fill_status` are PDAs, while the other fields hold their seed inputs and cached bumps. `V5FillStatusPdaDerivations` might be technically more precise, but it feels unnecessarily long.

PR #1530, opened 2026-08-18
improve: add shared HyperCore primitives to HyperCoreLib

droplet-rl:
## Motivation

Review on [contracts-v5#203](https://github.com/across-protocol/contracts-v5/pull/203) (counterfactual HyperCore withdrawals) flagged that `CounterfactualHyperCoreWithdrawer` hand-rolls a set of HyperCore constants and calls that `HyperCoreLib` already owns the equivalents of — a second source of truth for precompile addresses and action headers, which is how they drift. Raised by @grasphoper in [three](https://github.com/across-protocol/contracts-v5/pull/203#discussion_r3801456992) [separate](https://github.com/across-protocol/contracts-v5/pull/203#discussion_r3801397574) [comments](https://github.com/across-protocol/contracts-v5/pull/203#discussion_r3801465303) and independently in [this one](https://github.com/across-protocol/contracts-v5/pull/203#discussion_r3793617876).

This PR upstreams them so the v5 leaf can drop its local copies. Opened separately because the lib sits behind the submodule boundary.

> **Retargeted** at v5#203's rewritten leaf. [`187e8d4e`](https://github.com/across-protocol/contracts-v5/commit/187e8d4ec2f4d5684cf9970a5affe39d83ee3a90) removed the rounding logic and [`933ddfc6`](https://github.com/across-protocol/contracts-v5/commit/933ddfc6) replaced the perp → spot → EVM two-step with a single `sendAsset` carrying a `sourceDex`, which changed what the leaf needs from here. See *Changed in the retarget* below.

## What's added

Purely additive — no existing function or constant changes behavior.

| Added | Replaces, in v5#203 |
| --- | --- |
| `WITHDRAWABLE_PRECOMPILE_ADDRESS` (`0x803`), `struct Withdrawable`, `withdrawable(address)` | the hand-rolled staticcall + bare `abi.decode(result, (uint64))` in `_perpWithdrawable` |
| `HYPE_SYSTEM_ADDRESS`, `HYPE_CORE_INDEX`, `HYPE_CORE_INDEX_TESTNET`, `hypeCoreIndex()`, `isHype(uint32)` | the local HYPE constants + the testnet/mainnet index ternary in `_bridgeTarget` |
[...]

PR #1541, opened 2026-09-02
fix: L04 Bridge Safety Check Can Revert Instead Of Returning False, Disabling HyperEVM Fallback On Fills

tbwebb22:
[HyperCoreLib.isCoreAmountSafeToBridge](https://github.com/across-protocol/contracts/blob/6e7a99497c051a45bd7e72615be3f68b244c070f/contracts/libraries/HyperCoreLib.sol#L413-L423) reports whether the asset bridge holds enough to cover a transfer, and callers use a false answer to pay the user on HyperEVM instead of bridging. This matters most on the Across fill path: HyperliquidDepositHandler is an [AcrossMessageHandler](https://github.com/across-protocol/contracts/blob/6e7a99497c051a45bd7e72615be3f68b244c070f/contracts/handlers/HyperliquidDepositHandler.sol#L22) whose [handleV3AcrossMessage](https://github.com/across-protocol/contracts/blob/6e7a99497c051a45bd7e72615be3f68b244c070f/contracts/handlers/HyperliquidDepositHandler.sol#L128-L133) is invoked by the SpokePool after it has already transferred the tokens to the handler, so something must be done with funds that have arrived. [_depositToHypercore](https://github.com/across-protocol/contracts/blob/6e7a99497c051a45bd7e72615be3f68b244c070f/contracts/handlers/HyperliquidDepositHandler.sol#L277-L289) runs the check before making any state changes or withdrawals and, on false, transfers the tokens to the user on HyperEVM and emits FallbackToHyperEVM, allowing the fill to succeed. [HyperCoreFlowExecutor](https://github.com/across-protocol/contracts/blob/6e7a99497c051a45bd7e72615be3f68b244c070f/contracts/periphery/mintburn/HyperCoreFlowExecutor.sol#L571) applies the same pattern through _fallbackHyperEVMFlow, as does [its swap path](https://github.com/across-protocol/contracts/blob/6e7a99497c051a45bd7e72615be3f68b244c070f/contracts/periphery/mintburn/HyperCoreFlowExecutor.sol#L717).

The check resolved the bridge address through toAssetBridgeAddress, plain arithmetic that could not fail, so it always returned true or false. [Pull request #1530](https://github.com/across-protocol/contracts/pull/1530) repoints [line 418](https://github.com/across-protocol/contracts/blob/6e7a99497c051a45bd7e72615be3f68b244c070f/contracts/libraries/HyperCoreLib.sol#L418) to [toSystemAddress](https://github.com/across-protocol/contracts/blob/6e7a99497c051a45bd7e72615be3f68b244c070f/contracts/libraries/HyperCoreLib.sol#L432-L441), which reverts when the core index exceeds uint32, when the token's Core entry has no linked HyperEVM contract, or when the tokenInfo precompile call fails. Reverting is correct where toSystemAddress resolves a destination that funds are about to be sent to, since sending to an unlinked token would strand them. However, inherited by a check whose purpose is to return an answer, it changes what happens on the fill path: for a token with no linked HyperEVM contract the handler now reverts, so the SpokePool fill reverts and the intent goes unfilled. Because the condition is permanent rather than transient, every relayer fails identically and the user waits for the origin-chain refund instead of receiving the tokens on HyperEVM.

The handler's other entry point, [depositToHypercore](https://github.com/across-protocol/contracts/blob/6e7a99497c051a45bd7e72615be3f68b244c070f/contracts/handlers/HyperliquidDepositHandler.sol#L110), performs its own safeTransferFrom, so a revert there simply unwinds and leaves the caller holding the tokens, and an unlinked token is a misconfiguration that arguably should fail loudly rather than silently divert users to HyperEVM.

Consider giving the check a non-reverting way to resolve the address, for example a variant returning a success flag alongside it, and having isCoreAmountSafeToBridge return false when resolution fails, so the fill path keeps its fallback; a token with no HyperEVM side is genuinely not safe to bridge. toSystemAddress should keep reverting for callers that resolve an address in order to send to it. If failing loudly on an unlinked token is intended, consider applying that only to depositToHypercore, where a revert unwinds the caller's own transfer and leaves them holding their tokens, while keeping the HyperEVM fallback on handleV3AcrossMessage, where a revert makes the SpokePool fill fail and the intent go unfilled.

### DIFF
--- a/contracts/libraries/HyperCoreLib.sol
+++ b/contracts/libraries/HyperCoreLib.sol
@@ -394,15 +394,6 @@ library HyperCoreLib {
     }
 
     /**
-     * @notice Whether `erc20CoreIndex` is native HYPE on the current chain.
-     * @param erc20CoreIndex The HyperCore index id of the token
-     * @return True if the index is native HYPE, false otherwise
-     */
-    function isHype(uint32 erc20CoreIndex) internal view returns (bool) {
-        return erc20CoreIndex == hypeCoreIndex();
-    }
-
-    /**
      * @notice Checks if an amount is safe to bridge from HyperEVM to HyperCore
      * @dev Verifies that the asset bridge has sufficient balance to cover the amount plus a buffer
      * @param erc20CoreIndex The HyperCore index id of the token
--- a/contracts/libraries/HyperCoreLib.sol
+++ b/contracts/libraries/HyperCoreLib.sol
@@ -415,8 +415,9 @@ library HyperCoreLib {
         uint64 coreAmount,
         uint64 coreBufferAmount
     ) internal view returns (bool) {
-        address bridgeAddress = toSystemAddress(erc20CoreIndex);
-        uint64 currentBridgeBalance = spotBalance(bridgeAddress, erc20CoreIndex);
+        // Deliberately not `toSystemAddress`: this is a predicate, and an unlinked token's system address simply
+        // holds nothing, so it evaluates to false instead of reverting a fill that has a HyperEVM fallback.
+        uint64 currentBridgeBalance = spotBalance(toSystemAddressNoRevert(erc20CoreIndex), erc20CoreIndex);
 
         // Return true if currentBridgeBalance >= coreAmount + coreBufferAmount
         return currentBridgeBalance >= coreAmount + coreBufferAmount;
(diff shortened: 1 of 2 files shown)


contracts/libraries/HyperCoreLib.sol
```diff
@@ -434,9 +435,24 @@ library HyperCoreLib {
         // linkage resolves via tokenInfo, declared `tokenInfo(uint32)` in canonical L1Read.sol — an id
         // beyond that domain (spotBalance's uint64 also carries encoded outcome asset ids) can't be resolved.
         if (erc20CoreIndex > type(uint32).max) revert TokenNotBridgeable(erc20CoreIndex);
-        uint32 index = uint32(erc20CoreIndex);
-        if (isHype(index)) return HYPE_SYSTEM_ADDRESS; // must precede the linkage check: HYPE has no evmContract
```
fusmanii, line 438: `isHype()` function is not longer being used, lets remove it

tbwebb22: good call, removed

grasphoper (commented):
Can a cleaner solution to this be returning an extra bool from `toSystemAddress`?

```
function toSystemAddress(uint64 erc20CoreIndex) internal view returns (address addr, bool success)
```

Then `isCoreAmountSafeToBridge` can return `false` if `toSystemAddress` returns false.

We can then have `toSystemAddressStrict` or something that will call `toSystemAddress` internally and revert if we want `revert` functionality in other places.

Or just make current `toSystemAddressNoRevert` return a `bool` too. Prevent duplication of functionality between `toSystemAddressNoRevert` and `toSystemAddress`

Before https://github.com/across-protocol/contracts/pull/1530 we never reverted on a similar call so instead of reverting, the callers could use a bool-returning function and just ignore the bool too 🤷

tbwebb22:
> Can a cleaner solution to this be returning an extra bool from `toSystemAddress`?
> 
> ```
> function toSystemAddress(uint64 erc20CoreIndex) internal view returns (address addr, bool success)
> ```
> 
> Then `isCoreAmountSafeToBridge` can return `false` if `toSystemAddress` returns false.
> 
> We can then have `toSystemAddressStrict` or something that will call `toSystemAddress` internally and revert if we want `revert` functionality in other places.
> 
> Or just make current `toSystemAddressNoRevert` return a `bool` too. Prevent duplication of functionality between `toSystemAddressNoRevert` and `toSystemAddress`
> 
> Before #1530 we never reverted on a similar call so instead of reverting, the callers could use a bool-returning function and just ignore the bool too 🤷

I like the bool suggestion. I went with that here https://github.com/across-protocol/contracts/commit/10ffc3df2ad212c5f2c05f673a823b7959928be8

Now naming is `toSystemAddress` for the reverting function, and `toSystemAddressIfBridgeable` for the bool returning function.

grasphoper (approved):
LGTM

On naming: there's a Rust ~convention we could borrow.

```
toSystemAddress() returns (address)
tryToSystemAddress() returns (address, bool)
```

We used it eg here https://github.com/across-protocol/sp1-helios/blob/b20cfa71dcec20052ee3c6c2d141a40e659bf8fe/zk-api/src/lib.rs#L27

PR #1544, opened 2026-09-08
test(svm): add real Gateway V5 conformance (ACP-184 Step 5)

Reinis-FRP:
## Scope

ACP-184 Step 5, stacked on #1538: real-Gateway integration tests, cross-VM fixtures, and docs. No production contract or packaged IDL/client changes.

- Pin Gateway/PrefundedAdapter to `solana-v5@457cf693` and run an isolated validator.
- Cover source funding, destination delivery, root binding/reuse, replay, payer lifecycle, malformed accounts, and rollback.
- Document unsafe short/aggregate delivery cases. The delivery policy is shared with EVM, not a new SVM-only security requirement.

## Follow-ups

- [ACB-637](https://linear.app/uma/issue/ACB-637): API builder port and applicable conformance tests; ACB-863 was consolidated here.
- [ENG-320](https://linear.app/uma/issue/ENG-320): SVM relayer integration checks, reusing EVM policy.
- Validate supported paths before SVM Across route enablement. Neither issue blanket-blocks ACP-184; no new generic validator or aggregation project is requested.
- Run `yarn test-svm-gateway` locally while `solana-v5` is private. The secret-dependent workflow has been removed; no cross-repository token or approval environment is needed.
- Ordinary CI runs the dependency-free Gateway hash vector and includes `test/svm-gateway/**` in the SVM path filter. It does not run the real-Gateway lane.
- Once `solana-v5` is public, add real-Gateway CI using a public checkout at the pinned commit. This is a temporary CI limitation, not a protocol blocker.

## Validation

At [1ea896d5](https://github.com/across-protocol/contracts/commit/1ea896d5c100389d21818bb9affebf36da5b3c58): real-Gateway suite 15/15; standalone TypeScript vector 1/1; Rust 12/12 (default) and 16/16 (`test` feature); Foundry vectors 5/5. Focused TypeScript, formatting, workflow-structure, and diff checks pass.

Review fixes cover setup-error cleanup/retry, attempted fill-event decoding from failed receipts, exact underdelivery and rejected-JIT assertions, IDL-derived program identity, explicit spec files, and both sibling-hash orders.

Limitations: the local compiler reports stack-size diagnostics in unchanged legacy fill handlers; this V5 lane does not certify them. The Step 4 base also has three unrelated EVM setup failures ([CI run](https://github.com/across-protocol/contracts/actions/runs/34218616113)). Verified-build checks remain separate.


droplet-rl (commented):
## Summary

Reviewed at `2d4155ab` against base `reinis/acp-184-step-4-destination-fill` (13 files, +1211/-0). This is genuinely test/docs/CI-only: the `programs/svm-spoke/src/v5.rs` addition sits inside the `#[cfg(test)] mod tests` block at line 341, `package.json` only gains a script, and no packaged IDL/client is touched. Scope matches the description.

The suite is strong. What I like in particular:

- It derives destination relays from **actual origin `FundsDeposited` events** rather than hand-built `RelayData`, so the deposit-ID derivation, the `V5_MAGIC_PREFIX || stepId` witness, and `get_relay_hash` are exercised end-to-end against the real Gateway instead of being restated in the test.
- The account-authentication cases are meaningful, not tautological. `relayJit(relay, wrong)` / `relayJit(relay, status, wrong)` fail with `MissingAccount` because `load_v5_fill_accounts` re-derives `V5FillStatusPdas` and the real PDA is *not in the forwarded set* — which is precisely the property that makes the outer remaining-account pool safe. Adding `wrong` to the pool via `extra` and still getting a rejection is the right way to pin "pool membership does not equal forwarding."
- Deliberately keeping the unsafe primitives (short consumption, two fills over one balance, floor-of-committed-minima with larger JIT outputs) as *passing counterexamples* rather than pretending the program enforces delivery is the honest framing, and the spec/README text around it is accurate w.r.t. `v5_adapter.rs:196` (`source.amount >= output_amount` is an assertion, not a debit).
- Good separation from the mock lane: `InsufficientVaultBalance`, allowance, pause, Token-2022 extension and exclusivity cases already live in `test/svm/SvmSpoke.V5Fill.ts` / `.V5Source.ts`, so this lane doesn't duplicate them.

### Independent verification of the new fixture

I re-derived `programs/svm-spoke/fixtures/v5_gateway_path.json` from scratch (pure-Python Keccak-256, no repo code):

- `pathId`, `siblingPathId`, `stepRoot`, `witness` — all four match.
- `executor` base58-encodes to `34trBszXuqhRjWaMxXWsunJNmyUsBvDNPxAwTzbPTm4p`, i.e. `constants::GATEWAY_PROGRAM_ID`. Since `reference.ts::pathId` hardcodes `GATEWAY.toBuffer()`, the TS assertion transitively pins that too.
- `V5_MAGIC_PREFIX` = `keccak256("AcrossV5MessagePrefix.V1")`, matching the comment at `constants.rs:27`.
- The `message` blob decodes byte-for-byte as the Borsh tape `[BALANCE_REQ(mint, 500000), TRANSFER(mint, recipient, u64::MAX, 10000bps)]` — op vector `02000000 0011`, then two length-prefixed inputs of `0x28` and `0x4a` bytes.

The fixture is self-consistent and correctly cross-checked in all three languages.

### Main thing I'd like addressed

**Nothing here runs in CI, and one piece of it easily could.** The lane is `workflow_dispatch`-only and blocked on an environment/secret that doesn't exist yet (correctly disclosed in the PR body). But `PathVectors.ts` needs no validator and no private dependency — it's pure hashing — while its Rust and Solidity counterparts *do* run on every PR. That leaves `reference.ts`, the encoder most likely to be reused downstream by ACB-637/ENG-320, as the only unpinned side of the vector. Note also that `pr.yml`'s `paths-filter` matches `test/svm/**`, which does **not** match `test/svm-gateway/**`, so a future PR touching only this directory triggers no SVM job at all. Details inline.

Everything else below is non-blocking: CI hygiene and a few assertions that could be tightened. Not requesting changes — this is test infrastructure with no production surface, and the substance is sound.

*Caveat on validation:* I could not execute the lane (private `solana-v5` + Solana toolchain unavailable here), so I'm relying on your reported 14/14 · 16/16 · 5/5. My review is static analysis plus the independent fixture re-derivation above.

test/svm-gateway/PathVectors.ts
```diff
@@ -0,0 +1,24 @@
+import { assert } from "chai";
+import { PublicKey } from "@solana/web3.js";
+import fixture from "../../programs/svm-spoke/fixtures/v5_gateway_path.json";
+import { GATEWAY_COMMIT, PREFIX, floor, pair, pathId, tape, transfer } from "./reference";
+
+describe("Gateway path cross-VM vectors", () => {
+  it("pins Borsh tape, EVM path hashing, sorted sibling root and V5 witness", () => {
```
droplet-rl, line 7: This spec has no validator or private-dependency requirement — it is pure hashing over a checked-in fixture — yet it only runs inside `yarn test-svm-gateway`, which is `workflow_dispatch`-only and blocked on the not-yet-configured `svm-gateway-integration` environment.

Meanwhile the Rust half (`v5.rs::gateway_path_root_and_witness_match_cross_vm_fixture`, via `cargo test -p svm-spoke --lib`) and the Solidity half (`SvmSpokeV5Vectors.t.sol`, via `yarn test-evm-foundry`) both run on every PR. So the one implementation that *isn't* pinned by ordinary CI is `reference.ts` — the encoder downstream consumers in ACB-637/ENG-320 are most likely to copy.

Worth moving this file somewhere the default suite picks it up (or adding a tiny mocha invocation to `lint-and-check-generated`) so all three encoders stay locked together.

Separately: `pr.yml`'s `dorny/paths-filter` lists `test/svm/**`, which does not glob-match `test/svm-gateway/**`. A future PR touching only this directory triggers no SVM job.

.github/workflows/svm-gateway.yml
```diff
@@ -0,0 +1,48 @@
...
+          token: ${{ secrets.SVM_GATEWAY_READ_TOKEN }}
+          persist-credentials: false
+      - uses: actions/setup-node@v4
+        with:
+          node-version: "22.18"
+          cache: yarn
+      - uses: dtolnay/rust-toolchain@stable
+      - uses: solana-developers/github-actions/setup-all@v0.2.5
+        with:
+          anchor_version: 0.31.1
```
droplet-rl: Toolchain versions are hardcoded here (`anchor_version: 0.31.1`, `solana_version: 4.1.2`, and `node-version: "22.18"` on line 31), but the repo already has `.github/actions/setup-solana-anchor`, which calls `solana-developers/github-actions/extract-versions@v0.2.5` to derive them from the repo itself, and `pr.yml` centralises Node as `env.NODE_VERSION`.

Reusing the composite action would keep this lane from silently drifting when `anchor-lang` is bumped in `programs/svm-spoke/Cargo.toml`. The drift is quiet, too — a version mismatch here surfaces as a confusing build failure in a manually-dispatched job rather than a clear signal.

(The pinned Gateway commit is duplicated across `reference.ts`, the fixture, this file's `ref:`, and the spec link, but `testRealGateway.ts:35` hard-fails on mismatch, so that one is safe.)

.github/workflows/svm-gateway.yml
```diff
@@ -0,0 +1,48 @@
...
+          solana_version: 4.1.2
+      - name: Install Gateway Anchor CLI
+        run: |
+          curl -sSfL --retry 3 https://github.com/otter-sec/anchor/releases/download/v1.1.2/anchor-1.1.2-x86_64-unknown-linux-gnu -o target/anchor-gateway
```
droplet-rl: This fetches an executable over the network and runs it with no integrity check. The URL is version-pinned, but GitHub release assets are mutable, and this job is one where `secrets.SVM_GATEWAY_READ_TOKEN` has just been used to pull a private repository.

The rest of the workflow's security posture is deliberate and good — `permissions: contents: read`, `persist-credentials: false` on both checkouts, token scoped only to the checkout step so it is absent from `yarn install` and the test run, environment gating with required reviewers. A `sha256sum -c` on this download would round it out.

`dtolnay/rust-toolchain@stable` on line 33 is also a moving target; `pr.yml` pins `actions-rs/toolchain@v1` with an explicit toolchain.

test/svm-gateway/RealGateway.ts
```diff
@@ -0,0 +1,629 @@
...
+      opts.proof ?? [],
+      opts.jit ?? relays.map((r) => relayJit(r)),
+      opts.funds ?? []
+    );
+    const accountDiscriminator = createHash("sha256").update("account:ExecuteParamsAccount").digest().subarray(0, 8);
+    const digest = createHash("sha256")
+      .update(Buffer.concat([accountDiscriminator, bytes]))
+      .digest();
+    const buffer = pda(GATEWAY, Buffer.from("execute_params"), owner.toBuffer(), digest);
+    await send(
```
droplet-rl: `initialize_execute_params` and the fragment writes sit *outside* the `try`, so only the `execute` call itself is covered by the `finally` cleanup on line 229.

If any of these setup sends throws (RPC hiccup, a fragment write failing), the params PDA is left allocated. Because the PDA seed is a digest over `bytes`, a later `execute()` with byte-identical params then fails inside `initialize_execute_params` with an "account already in use" error — and since `expectFailure` only substring-matches, a negative test could report a pass-shaped failure for entirely the wrong reason, or a positive test could fail confusingly.

Moving lines 176-192 inside the `try` makes the existing `finally` cover them at no cost.

test/svm-gateway/RealGateway.ts
```diff
@@ -0,0 +1,629 @@
...
+  it("rolls back recorded fills, emitted event effects, tokens and rent after a downstream failure", async () => {
+    const dst = path([...canonicalInPlace(fillCommand(), mint, amount, recipient), floor(mint, 1n)]);
+    const relay = await origin(pathId(dst));
+    const userBefore = (await getAccount(connection, userAta)).amount;
+    const payerBefore = await connection.getBalance(fillPayer);
+    const result = await execute(dst, { relays: [relay], funds: [funding(userAta, mint, amount)], failed: true });
+    assert.include(
+      result.logs.join("\n"),
+      `Program ${spoke.programId} success`,
+      "fill completed before the later failure"
```
droplet-rl: This asserts `Program <spoke> success` in the failed transaction's logs, which shows the fill CPI completed before the later `BalanceRequirementNotMet`. But the hazard the spec and README actually warn about is narrower:

> Failed transaction logs may contain attempted fill events; consumers must check transaction success before accepting them.

A generic "program succeeded" line isn't that. What pins the warning is a `FilledRelay` event genuinely appearing in the logs of a transaction that `meta.err` marks failed. Asserting the `emit_cpi!` log (the event-authority self-invoke, or the base64 program-data line) would make this test the actual evidence for the documented consumer requirement, rather than a proxy for it.

The rollback assertions on lines 531-535 are thorough and correct as-is.

test/svm-gateway/RealGateway.ts
```diff
@@ -0,0 +1,629 @@
...
+        floor(mint, amount),
+        ...(consumed ? [transfer(mint, recipient, consumed)] : []),
+      ]);
+      const relay = await origin(pathId(dst));
+      await fund();
+      const before = (await getAccount(connection, recipientAta)).amount;
+      await execute(dst, { relays: [relay] });
+      await filled(relay);
+      const delivered = (await getAccount(connection, recipientAta)).amount - before;
+      assert.throws(() => assertAggregateDelivery([amount], delivered), "aggregate underdelivery");
```
droplet-rl, line 625: The only assertion on the outcome routes through `assert.throws` on a helper defined in this same PR, so what's really pinned is "delivered < amount" — the test would still pass if the primitive started delivering some *different* short amount.

Since `consumed` is known per iteration, asserting the exact value is strictly stronger and costs one line:

```ts
assert.equal(delivered, consumed);
```

The test on line 626 already does exactly this (`assert.equal(delivered, amount * 2n)` alongside the `assert.throws`), so this is just bringing the two counterexamples into line with each other.

test/svm-gateway/RealGateway.ts
```diff
@@ -0,0 +1,629 @@
...
+      assert.isNull(await connection.getAccountInfo(status(relay)));
+    }
+    const dst = destination();
+    const relay = await origin(pathId(dst));
+    const wrong = Keypair.generate().publicKey;
+    await fund();
+    for (const jit of [relayJit(relay, wrong), relayJit(relay, status(relay), wrong)]) {
+      await expectFailure(execute(dst, { relays: [relay], jit: [jit], extra: [writable(wrong)] }), "MissingAccount");
+    }
+    await expectFailure(execute(dst, { relays: [relay], jit: [Buffer.alloc(31)] }), "InjectedAccountsMismatch");
```
droplet-rl, line 596: Minor asymmetry: the invalid-meta loop above asserts `assert.isNull(await connection.getAccountInfo(status(relay)))` after each rejection (line 547), but these three JIT cases only assert that the send throws.

Atomicity makes it true regardless, but the whole point of this lane is pinning that no fill state survives a rejected execution — and `await fund()` on line 552 means the vault is live here, so there is something to leak. Worth the same one-line check after the loop.

scripts/svm/testRealGateway.ts
```diff
@@ -0,0 +1,162 @@
...
+      "--upgradeable-program",
+      GATEWAY.toBase58(),
+      path.join(checkout, "target/deploy/gateway.so"),
+      wallet.publicKey.toBase58(),
+      "--upgradeable-program",
+      PREFUNDED.toBase58(),
+      path.join(checkout, "target/deploy/prefunded_adapter.so"),
+      wallet.publicKey.toBase58(),
+      "--upgradeable-program",
+      "DLv3NggMiSaef97YCkew5xKUHDh13tVGZ7tydt3ZeAru",
```
droplet-rl: The svm-spoke program ID is a bare string literal here while `GATEWAY` and `PREFUNDED` come from `reference.ts` two lines up and down. If `declare_id!` in `lib.rs:17` or `Anchor.toml` ever changes, this silently deploys the freshly-built `.so` at a stale address and the failure shows up as an opaque Anchor error deep in the suite.

`target/idl/svm_spoke.json` is generated on line 56-68, immediately above — its `address` field can be read here, or export the ID from `reference.ts` alongside the other two.

scripts/svm/testRealGateway.ts
```diff
@@ -0,0 +1,162 @@
...
+      } catch {
+        /* RPC starts after genesis setup. */
+      }
+      await new Promise((resolve) => setTimeout(resolve, 500));
+    }
+    if (!ready) throw new Error(`Validator did not start; see ${logPath}`);
+    console.log(`Real Gateway ${GATEWAY_COMMIT}; validator logs: ${logPath}`);
+    const tests = spawnSync(
+      "yarn",
+      ["ts-mocha", "--bail", "-p", "tsconfig.json", "-t", "1000000", "test/svm-gateway/*.ts"],
```
droplet-rl: The glob `test/svm-gateway/*.ts` also matches `reference.ts` and `provider.ts`, so mocha loads both as spec files.

Harmless today — `reference.ts` is side-effect-free and `provider.ts`'s `setProvider` is idempotent — but it means any future import-time side effect in `reference.ts` runs at an unintended point in the mocha lifecycle, and file ordering (`PathVectors`, `RealGateway`, `provider`, `reference` under case-sensitive sort) becomes load-bearing without anyone noticing.

An explicit file list, or a `*.spec.ts` naming convention for the two real specs, removes the coupling.

test/evm/foundry/local/SvmSpokeV5Vectors.t.sol
```diff
@@ -58,6 +58,25 @@ contract SvmSpokeV5VectorsTest is Test {
...
+        bytes32 messageHash = keccak256(vm.parseJsonBytes(pathFixture, ".message"));
+        uint256 chainId = vm.parseJsonUint(pathFixture, ".chainId");
+        bytes32 executor = vm.parseJsonBytes32(pathFixture, ".executor");
+        bytes32 a = keccak256(abi.encode(chainId, vm.parseJsonBytes32(pathFixture, ".salt"), executor, messageHash));
+        bytes32 b = keccak256(
+            abi.encode(chainId, vm.parseJsonBytes32(pathFixture, ".siblingSalt"), executor, messageHash)
+        );
+        assertEq(a, vm.parseJsonBytes32(pathFixture, ".pathId"));
+        assertEq(b, vm.parseJsonBytes32(pathFixture, ".siblingPathId"));
+        bytes32 root = a < b ? keccak256(abi.encodePacked(a, b)) : keccak256(abi.encodePacked(b, a));
```
droplet-rl: I confirmed by re-deriving the fixture that `a < b` holds, so the `: keccak256(abi.encodePacked(b, a))` branch is dead here — and identically at `v5.rs:405-409`.

The TypeScript vector already covers this properly: `PathVectors.ts:20-21` asserts both `pair(a, b)` and `pair(b, a)` equal `stepRoot`. Rust and Solidity could pin the same commutativity for one extra `assertEq` each, so all three sorted-pair implementations are fully exercised rather than two of them being half-tested.

Also a nit on line 75: the V5 prefix is a bare literal, whereas the Rust vector references `crate::constants::V5_MAGIC_PREFIX` — meaning the Rust test would catch a drift in the production constant and this one wouldn't. Reading it from the SpokePool's own constant (if reachable from this test) would make the Solidity side an equally real cross-check. For the record, I verified the literal equals `keccak256("AcrossV5MessagePrefix.V1")`.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=1e6dbec590130c5690bac43d0cd52e18&trace_start_ts=1788868908&trace_end_ts=1788955608)

test/svm-gateway/PathVectors.ts
Reinis-FRP: Fixed in [1ea896d5](https://github.com/across-protocol/contracts/commit/1ea896d5c100389d21818bb9affebf36da5b3c58). Ordinary CI now runs the dependency-free `yarn test-svm-gateway-vectors`, and the SVM path filter includes `test/svm-gateway/**`. The actual private-Gateway lane remains local-only.

---

.github/workflows/svm-gateway.yml
Reinis-FRP: Removed in [294d5cd9](https://github.com/across-protocol/contracts/commit/294d5cd91e6ba608e7aad2d45331c0393de32dd9). The entire private-dependency workflow is gone. Real-Gateway CI is deferred until publication; no toolchain setup remains in that workflow.

---

.github/workflows/svm-gateway.yml
Reinis-FRP: Removed in [294d5cd9](https://github.com/across-protocol/contracts/commit/294d5cd91e6ba608e7aad2d45331c0393de32dd9). The workflow, executable download, and cross-repository credential requirement are gone. Real-Gateway testing stays local until publication.

---

test/svm-gateway/RealGateway.ts
Reinis-FRP: Fixed in [1ea896d5](https://github.com/across-protocol/contracts/commit/1ea896d5c100389d21818bb9affebf36da5b3c58). Initialization and fragment writes are now inside the cleanup `try/finally`. A regression injects confirmation errors after each setup operation and proves byte-identical execution can retry.

---

test/svm-gateway/RealGateway.ts
Reinis-FRP: Fixed in [1ea896d5](https://github.com/across-protocol/contracts/commit/1ea896d5c100389d21818bb9affebf36da5b3c58). The rollback test now decodes the failed receipt's inner CPI instructions, asserts exactly one attempted `FilledRelay`, and matches its deposit ID. The receipt must still have `meta.err`; balance, status, and rent rollback checks remain.

---

test/svm-gateway/RealGateway.ts
Reinis-FRP: Fixed in [1ea896d5](https://github.com/across-protocol/contracts/commit/1ea896d5c100389d21818bb9affebf36da5b3c58). Both counterexamples now assert `delivered === consumed` before checking aggregate underdelivery.

---

test/svm-gateway/RealGateway.ts
Reinis-FRP: Fixed in [1ea896d5](https://github.com/across-protocol/contracts/commit/1ea896d5c100389d21818bb9affebf36da5b3c58). Fill-status absence is checked after each wrong-account JIT rejection and after malformed JIT rejection.

---

scripts/svm/testRealGateway.ts
Reinis-FRP: Fixed in [1ea896d5](https://github.com/across-protocol/contracts/commit/1ea896d5c100389d21818bb9affebf36da5b3c58). The validator now reads the SpokePool address from the freshly generated test IDL.

---

scripts/svm/testRealGateway.ts
Reinis-FRP: Fixed in [1ea896d5](https://github.com/across-protocol/contracts/commit/1ea896d5c100389d21818bb9affebf36da5b3c58). Mocha now receives the two explicit spec files; helpers are loaded only through their imports.

---

test/evm/foundry/local/SvmSpokeV5Vectors.t.sol
Reinis-FRP: Fixed in [1ea896d5](https://github.com/across-protocol/contracts/commit/1ea896d5c100389d21818bb9affebf36da5b3c58). Rust and Solidity now check both sibling orders; Solidity uses OpenZeppelin's commutative hash helper. This checkout has no exposed production EVM V5 prefix, so the witness derives it from `keccak256("AcrossV5MessagePrefix.V1")` instead of a hex literal.

---

test/svm-gateway/README.md
```diff
@@ -0,0 +1,73 @@
...
+`SVM_GATEWAY_CHECKOUT` may point to an existing clean checkout at the exact pin to avoid another clone. Builds still
+run; arbitrary prebuilt binaries are not accepted as conformance evidence. Dependency downloads and local validator
+ports require network permission in sandboxed environments.
+
+While `solana-v5` is private, run this lane locally with existing repository access; do not add a cross-repository
+credential to `contracts` CI. Ordinary PR checks run `yarn test-svm-gateway-vectors` for the dependency-free hash
+fixture, but do not run the real-Gateway lane. Changes in `test/svm-gateway` also trigger the ordinary SVM tests.
+
+After `solana-v5` is public, add CI coverage using a public checkout at the immutable `GATEWAY_COMMIT`, without a
+cross-repository token or approval environment. This is a temporary CI limitation, not a protocol blocker.
```
droplet-rl, line 29: Worth capturing the specifics here so they survive to the follow-up, since the deleted workflow resolved two of my earlier comments by removal rather than by fix. When this CI comes back:

- Prefer the existing `.github/actions/setup-solana-anchor` composite over hardcoding `anchor_version` / `solana_version` / `node-version`. It calls `extract-versions` to derive them from the repo, so the lane won't silently drift when `anchor-lang` is bumped in `programs/svm-spoke/Cargo.toml`. `pr.yml` also centralises Node as `env.NODE_VERSION`.
- If the Gateway's Anchor CLI (otter-sec 1.1.2) still has to be fetched as a release binary, add a `sha256sum -c`. Release assets are mutable even when the URL is version-pinned.

Neither applies today — purely a note for whoever writes the public-checkout workflow.

test/svm-gateway/RealGateway.ts
```diff
@@ -0,0 +1,662 @@
...
+        assert.isNotNull(receipt!.meta!.err, "must be a failed transaction, never an accepted fill event");
+        return {
+          signature,
+          logs: receipt!.meta!.logMessages ?? [],
+          attemptedEvents: processEventFromTx(receipt!, [spoke]),
+        };
+      }
+      const signature = await send(ComputeBudgetProgram.setComputeUnitLimit({ units: 1_400_000 }), instruction);
+      assert.isNull(await connection.getAccountInfo(buffer), "successful execution closes parameter buffer");
+      return { signature, logs: [] as string[], attemptedEvents: [] };
```
droplet-rl: Optional nit: `attemptedEvents: []` (and `logs: []`) on the success path are stubs that exist only to keep the return shape uniform, but the name doesn't signal that. A future test author who calls `execute(...)` on a successful path and checks `result.attemptedEvents` gets an empty array and could reasonably read that as "no events were emitted" — when in fact successful-path events are read via `readEventsUntilFound`, as `origin()` and the first test do.

Either naming it for its actual scope or returning it only under `opts.failed` would remove the ambiguity. Not worth holding the PR for.

droplet-rl (approved):
## Approving

Re-reviewed the delta `2d4155ab..1176c013` (3 commits, 10 files, +109/−108) and re-checked the full PR against base. All ten comments from my previous review are addressed, several more thoroughly than I asked. LGTM.

### Fixes verified

| Previous comment | Resolution |
|---|---|
| `PathVectors.ts` never runs in CI | New `test-svm-gateway-vectors` script wired into `lint-and-check-generated` (`pr.yml:104`) |
| `paths-filter` misses `test/svm-gateway/**` | Added at `pr.yml:147` |
| Hardcoded toolchain versions | Workflow deleted |
| Unverified binary download | Workflow deleted |
| Params-buffer leak on setup failure | Setup moved inside `try` — **plus** a new regression test |
| `FilledRelay` event in failed tx not asserted | Now decoded from the failed receipt |
| `assert.throws` on a local helper | `assert.equal(delivered, consumed)` added |
| Missing status-PDA assertions after JIT rejections | Added at both sites |
| Hardcoded spoke program ID | Read from `target/idl/svm_spoke.json` |
| Glob picked up `reference.ts`/`provider.ts` | Explicit spec file list |

### Two fixes that went beyond the ask

**The failed-receipt event decoding (`RealGateway.ts:558-560`) is now the real thing.** Rather than grepping logs, it calls `processEventFromTx(receipt, [spoke])` — which walks `meta.innerInstructions` and decodes the `emit_cpi!` self-invoke — then asserts exactly one attempted `filledRelay` and matches its `depositId` against the relay. That is precisely the hazard `V5_ADAPTER_SPEC.md` warns about ("Failed transaction logs may contain attempted fill events"), and the test is now the evidence for it rather than a proxy. I confirmed the signature at `solanaProgramUtils.ts:55` accepts a `VersionedTransactionResponse`, which is what `getTransaction(..., {maxSupportedTransactionVersion: 0})` returns.

**The new retry test (`RealGateway.ts:405-427`) is self-verifying**, which is the part I like. It patches `provider.sendAndConfirm` to let the setup instruction land on-chain and *then* throw, so the buffer genuinely exists when `finally` runs. Because it wraps `expectFailure`, a patch that silently failed to match would surface as `expected injected setup confirmation failure` rather than a false pass. The close instruction correctly escapes the injection (its discriminator doesn't match `failAfter`), so cleanup still runs inside the patched window, and `path([floor(mint, 0n)])` encodes to 177 bytes — one fragment — so both loop branches are reachable. Restoration is in a `finally`.

### Checks I ran

- **OZ `Hashes.commutativeKeccak256`** — `@openzeppelin/contracts` is pinned at `5.5.0` and `Hashes.sol` landed in 5.1.0, so the new import resolves. Swapping the hand-rolled ternary for the canonical helper also documents that the Gateway's sorted-pair rule *is* the standard commutative hash, and deriving the prefix via `keccak256("AcrossV5MessagePrefix.V1")` instead of the literal means the Solidity vector now catches constant drift the way the Rust one already did. Both orders asserted in Rust (`v5.rs:405-412`) and Solidity.
- **The new CI step works on a fresh checkout.** `lint-and-check-generated` does not download SVM artifacts, so I traced the import graph: `PathVectors.ts` → `reference.ts` → `src/types/svm.ts`, which imports only `@coral-xyz/anchor`, `@solana/web3.js` and `ethers`. No `target/` or generated-client dependency, and `resolveJsonModule` is already set. It will run green.
- **No dangling references.** Grepped the tree for `SVM_GATEWAY_READ_TOKEN`, `svm-gateway-integration`, `svm-gateway.yml` and `linear.app` — all clean. Removing the private Linear links from a public repo is the right call; the replacement prose keeps the policy without the unresolvable pointers.
- **`v5_gateway_path.json` and `reference.ts` are byte-identical to what I reviewed last round**, so my independent Keccak re-derivation of `pathId` / `siblingPathId` / `stepRoot` / `witness` still holds.

### Carrying forward

The toolchain-pinning and binary-integrity concerns I raised were resolved by deleting the workflow rather than by fixing it. They'll be live again when `solana-v5` goes public and CI returns — flagged inline on the README so they aren't lost with the file.

*Same caveat as last time:* I could not execute the real-Gateway lane here (private `solana-v5`, no Solana toolchain), so the reported 15/15 · 1/1 · 12/12 · 16/16 · 5/5 is taken from the PR description. My verification is static analysis plus the independent fixture derivation.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=0f4c2e414dd4991a652f7c4f1c1f4469&trace_start_ts=1788870890&trace_end_ts=1788957590)

test/svm-gateway/README.md
Reinis-FRP: Documented in [935b2709](https://github.com/across-protocol/contracts/commit/935b27096c045070e70bd14b5c7326a908bf32fd): reuse the shared toolchain setup and Node version, and verify any downloaded Gateway Anchor binary against a reviewed, pinned SHA-256 checksum before execution. CI remains deferred until publication.

---

test/svm-gateway/RealGateway.ts
Reinis-FRP: Fixed in [935b2709](https://github.com/across-protocol/contracts/commit/935b27096c045070e70bd14b5c7326a908bf32fd). Success returns only `signature`; logs and attempted events exist only under `failedReceipt` for a confirmed failed transaction. Updated both success-path and rollback assertions. Full Gateway suite: 15/15; focused typecheck and formatting pass.

---

PR #1545, opened 2026-09-08
fix(ci): pin Foundry and HyperEVM test fork

Reinis-FRP:
Two independent issues make master's EVM CI fail: the unpinned Foundry installation moved from v1.7.1 to v1.8.1 and triggered Solidity 0.8.30's `Tag too large for reserved space` compiler error; HyperEVM simulator setup also races the public RPC's latest block and fails with `invalid block height`.

Pin both CI build and test installations to v1.7.1, the version used by [master's last successful build](https://github.com/across-protocol/contracts/actions/runs/32507876728). Include the build action in the artifact cache key so changing the toolchain pin invalidates cached artifacts, and document how to match CI locally.

Pin the simulator's EVM fork to block 44,686,920, whose availability was checked against the public RPC. Its explicit HyperCore precompile reads remain live. This changes only the test fixture; production contract sources and compiler configuration remain unchanged.

Validation: formatting and diff checks pass, and all 45 tests across the three affected simulator suites pass with the pinned fork. The Foundry pin restored successful EVM artifact generation in CI; the same isolated v1.8.1 build fails with dynamic test linking enabled and passes with it disabled. All checks now pass on commit 2081ee9786599278928db2d07a53aacb29a26cc5 in [CI run 34253659880](https://github.com/across-protocol/contracts/actions/runs/34253659880). The TypeScript job required one diagnostic rerun after GitHub's internal artifact ListArtifacts API returned an intermediary 403; both artifact downloads and compilation then passed without code or permission changes.

### DIFF
--- a/.github/actions/cache-evm-artifacts/action.yml
+++ b/.github/actions/cache-evm-artifacts/action.yml
@@ -23,7 +23,7 @@ runs:
         # The job that generates the artifacts is responsible for archiving them to the cache tarball. This avoids any
         # conflicts with other caching actions that might have cleaned some of cached contents.
         path: evm-artifacts.tar
-        key: evm-artifacts-${{ runner.os }}-node-${{ steps.resolved-node.outputs.version }}-${{ hashFiles('yarn.lock', 'foundry.toml', 'contracts/**/*.sol') }}
+        key: evm-artifacts-${{ runner.os }}-node-${{ steps.resolved-node.outputs.version }}-${{ hashFiles('yarn.lock', 'foundry.toml', 'contracts/**/*.sol', '.github/actions/generate-evm-artifacts/action.yml') }}
     - name: Unpack restored EVM artifacts
       if: steps.evm-artifacts-cache.outputs.cache-hit == 'true'
       shell: bash
--- a/.github/actions/generate-evm-artifacts/action.yml
+++ b/.github/actions/generate-evm-artifacts/action.yml
@@ -18,6 +18,8 @@ runs:
       run: yarn install --frozen-lockfile
     - name: Install Foundry
       uses: foundry-rs/foundry-toolchain@v1
+      with:
+        version: v1.7.1
     - name: Install forge dependencies
       shell: bash
       run: forge install
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -205,6 +205,8 @@ jobs:
         run: yarn install --frozen-lockfile --ignore-optional
       - name: Install Foundry
         uses: foundry-rs/foundry-toolchain@v1
+        with:
+          version: v1.7.1
       - name: Install forge dependencies
         run: forge install
       - name: Inspect storage layouts
--- a/README.md
+++ b/README.md
@@ -35,6 +35,9 @@ yarn build # Will build all code. Compile solidity & rust (local toolchain), gen
 yarn build-verified # Will build all code. Compile solidity & rust (verified docker build), generate ts outputs
 ```
 
+EVM CI builds and tests use Foundry **v1.7.1**. Run `foundryup --install v1.7.1` to match CI locally.
+Keep the version pins in `.github/actions/generate-evm-artifacts/action.yml` and `.github/workflows/pr.yml` in sync.
+
 ## Test
 
 ```shell
--- a/README.md
+++ b/README.md
@@ -50,6 +50,9 @@ yarn test-svm # Only test SVM code (local toolchain build)
 yarn test-svm-solana-verify # Only test SVM code (verified docker build)
 ```
 
+HyperEVM simulator tests fork EVM state at block **44,686,920** to avoid races when the public RPC reports a new
+block before all backends can serve it. The simulator's explicit HyperCore precompile calls still use live data.
+
 ## Lint
 
 ```shell
--- a/test/evm/foundry/local/external/hyper-evm-lib/test/BaseSimulatorTest.sol
+++ b/test/evm/foundry/local/external/hyper-evm-lib/test/BaseSimulatorTest.sol
@@ -33,7 +33,9 @@ abstract contract BaseSimulatorTest is Test {
 
     function setUp() public virtual {
         string memory hyperliquidRpc = "https://rpc.hyperliquid.xyz/evm";
-        vm.createSelectFork(hyperliquidRpc);
+        // Pin EVM state to avoid latest-block races across the public RPC's backends.
+        // The simulator's explicit HyperCore precompile calls still read live data.
+        vm.createSelectFork(hyperliquidRpc, 44_686_920);
 
         hyperCore = CoreSimulatorLib.init();
 


test/evm/foundry/local/external/hyper-evm-lib/test/BaseSimulatorTest.sol
```diff
@@ -33,7 +33,9 @@ abstract contract BaseSimulatorTest is Test {
 
     function setUp() public virtual {
         string memory hyperliquidRpc = "https://rpc.hyperliquid.xyz/evm";
-        vm.createSelectFork(hyperliquidRpc);
+        // Pin EVM state to avoid latest-block races across the public RPC's backends.
+        // The simulator's explicit HyperCore precompile calls still read live data.
+        vm.createSelectFork(hyperliquidRpc, 44_686_920);
```
droplet-rl, line 38: The pin fixes the reported race, but the second comment line undersells how much nondeterminism is left, and it's worth being precise since this is the file people will read when the suite next flakes.

`CoreSimulatorLib.init()` etches `PrecompileSim` over `0x0800`–`0x0810`. That sim serves 10 precompiles from simulated state; the rest fall through to `PrecompileSim.sol:94`, which is a `vm.rpc("eth_call", ... "latest")` — not the fork block. Same in `RealL1Read.sol:128`. So the genuinely-live set is ORACLE_PX `0x0807`, SPOT_PX `0x0808`, L1_BLOCK_NUMBER `0x0809`, PERP_ASSET_INFO `0x080A`, SPOT_INFO `0x080B`, TOKEN_SUPPLY `0x080D`, BBO `0x080E`, plus `readSpotPx`/`readMarkPx` when uncached (`CoreView.sol:23,31`).

Net effect: EVM state frozen at 2026-08-31 21:09:40 UTC, Core reads at live head — a skew that widens ~1 day/day (~7.8 days as of this review). Production code under test reads this way (`HyperCoreLib.sol:338` `spotPx`, `:349` `tokenInfo`).

I checked the actual exposure and it's low, which is why this isn't a change request: every deadline in the three suites is relative to `block.timestamp` (so the pin is safe there, and arguably *more* deterministic now), and `HyperCoreFlowExecutor.t.sol` sets token info explicitly via `callSetCoreTokenInfo` (line 77) and forces bridge balance (line 108) instead of trusting live values.

The residual risk is a token relink or a new listing after the pinned block: live `tokenInfo.evmContract` would return an address with no code at block 44,686,920, and `HyperCoreFlowExecutor.sol:320` asserts `evmContract != address(0) && weiDecimals != 0`. That failure would be genuinely confusing to debug.

Worth knowing: `RealL1Read.sol:131` has `bool useArchivedBlockNumber = false;` — a disabled hook that exists to make exactly this coherent by routing reads at `block.number`. Enabling it (or leaving a note on why it stays off) would close the loop. Minimum ask: reword line 37 to name which reads stay live, so the next reader isn't misled into thinking the suite is fully pinned.

README.md
```diff
@@ -35,6 +35,9 @@ yarn build # Will build all code. Compile solidity & rust (local toolchain), gen
 yarn build-verified # Will build all code. Compile solidity & rust (verified docker build), generate ts outputs
 ```
 
+EVM CI builds and tests use Foundry **v1.7.1**. Run `foundryup --install v1.7.1` to match CI locally.
+Keep the version pins in `.github/actions/generate-evm-artifacts/action.yml` and `.github/workflows/pr.yml` in sync.
```
droplet-rl, line 39: "Keep the version pins ... in sync" is a manual-sync requirement, and this repo already has the pattern that removes it: `pr.yml` declares `env: NODE_VERSION: 22.18` once and threads it through as a `node_version` input to the composite actions.

Same shape works here:

```yaml
env:
  NODE_VERSION: 22.18
  FOUNDRY_VERSION: v1.7.1
```

plus a `foundry_version` input on `generate-evm-artifacts` (an explicit input rather than relying on `env` reaching into the composite action's `with:`, matching how `node_version` is already handled). One literal, no sync note needed, and it lets the cache key reference the version directly — see my comment on `cache-evm-artifacts/action.yml`.

Also: this documents *what* is pinned but not *why*. Add the reason — "1.8.1+ fails solc 0.8.30 with `Tag too large for reserved space` under `via_ir`" — otherwise the next person to helpfully bump the pin rediscovers it the hard way.

.github/actions/cache-evm-artifacts/action.yml
```diff
@@ -23,7 +23,7 @@ runs:
         # The job that generates the artifacts is responsible for archiving them to the cache tarball. This avoids any
         # conflicts with other caching actions that might have cleaned some of cached contents.
         path: evm-artifacts.tar
-        key: evm-artifacts-${{ runner.os }}-node-${{ steps.resolved-node.outputs.version }}-${{ hashFiles('yarn.lock', 'foundry.toml', 'contracts/**/*.sol') }}
+        key: evm-artifacts-${{ runner.os }}-node-${{ steps.resolved-node.outputs.version }}-${{ hashFiles('yarn.lock', 'foundry.toml', 'contracts/**/*.sol', '.github/actions/generate-evm-artifacts/action.yml') }}
```
droplet-rl, line 26: Right instinct, and it's correct as-is — but hashing the whole action file is a coarse proxy for "the toolchain changed." Any unrelated edit (a comment, a step reorder, adding a step) invalidates every cached EVM artifact for a full rebuild.

If you take the `FOUNDRY_VERSION` suggestion from the README comment, this becomes both tighter and self-documenting:

```yaml
key: evm-artifacts-${{ runner.os }}-node-${{ steps.resolved-node.outputs.version }}-foundry-${{ inputs.foundry_version }}-${{ hashFiles('yarn.lock', 'foundry.toml', 'contracts/**/*.sol') }}
```

Invalidates on exactly the thing you care about, and a reader can tell which toolchain a cache entry belongs to from the key alone.

.github/actions/generate-evm-artifacts/action.yml
```diff
@@ -18,6 +18,8 @@ runs:
       run: yarn install --frozen-lockfile
     - name: Install Foundry
       uses: foundry-rs/foundry-toolchain@v1
+      with:
+        version: v1.7.1
```
droplet-rl, line 22: The pin itself is the right call — unpinned toolchains in CI are how you get this class of surprise.

Two follow-ups:

1. **Track getting off 1.7.1.** You already established that `dynamic_test_linking = false` makes 1.8.1 pass, so a path forward exists. Without a linked issue, the pin is indistinguishable from a permanent decision and the repo quietly ages out of the toolchain. A TODO with an issue link here would do it.
2. **`foundry-toolchain@v1` is itself a floating major tag.** Given this PR is specifically about toolchain reproducibility, the action that *interprets* `version:` still moves underneath you. Pinning to a full tag or SHA would be consistent with the intent (and closes the usual third-party-action supply-chain gap, since it runs with the workflow token). `actions/checkout@v3` / `setup-node@v3` have the same issue but are pre-existing and out of scope here.

.github/workflows/pr.yml
```diff
@@ -205,6 +205,8 @@ jobs:
         run: yarn install --frozen-lockfile --ignore-optional
       - name: Install Foundry
         uses: foundry-rs/foundry-toolchain@v1
+        with:
+          version: v1.7.1
```
droplet-rl, line 209: Confirmed this is the second and last `foundry-toolchain` site in the repo — I grepped, and there's no third one left unpinned. Good.

This pin is also load-bearing in a way that isn't obvious: `test-evm` runs `FOUNDRY_PROFILE=local-test forge test`, which compiles into `out-local` from scratch rather than reusing the downloaded `out`/`cache-foundry` artifacts. So the test job compiles the tree itself and would hit the same compiler error independently — pinning both sites is necessary, not belt-and-braces.

The flip side, which supports keeping the cache key scoped to the build action only (as you've done): this pin does *not* need to feed the artifact cache key, since nothing this job compiles is cached.

droplet-rl (approved):
Approving. Both root causes are correctly diagnosed, the fixes are minimal and correctly scoped, and no production source or compiler configuration changes. Comments below are follow-ups, none blocking.

### Verified independently

I confirmed both failure modes from CI logs rather than taking the description on faith:

- **`Tag too large`** — run [34230528573](https://github.com/across-protocol/contracts/actions/runs/34230528573) (`Upload EVM artifacts`): `Internal compiler error: ... Assembly exception for bytecode: Tag too large for reserved space.` Reproduced against a wide set of branches on 2026-09-08, consistent with a toolchain-level break rather than anything branch-specific. Note `master`'s `Checks` workflow hasn't actually run since 2026-08-21 (the last pushes were Dependabot-only), so the break surfaces on PRs, not on a master run.
- **`invalid block height`** — run [34250627445](https://github.com/across-protocol/contracts/actions/runs/34250627445), at commit `0d42c719` (Foundry pin applied, fork pin not yet): `[FAIL: vm.createSelectFork: ... error code -32603: invalid block height: 45373614] setUp()`. 695 passed, 1 failed. Only `HyperCoreFlowExecutor` tripped, which fits an intermittent race on fork instantiation — putting the fix on the shared base is the right call.

Other things I checked:

- `foundry-rs/foundry-toolchain` appears in exactly two places in the repo, and both are now pinned — no third install site left floating.
- The cache-key change is correct **and sufficient**: cached artifacts come from the default-profile `forge build` (`out`, `cache-foundry`), while `test-evm` compiles into `out-local` fresh under the `local-test` profile. So only the build action's pin needs to feed the key; the `pr.yml` pin doesn't.
- `foundryup --install v1.7.1` is valid (`-i, --install <VERSION>`), so the README instruction works as written.
- **The pin won't rot from pruning.** This was my main worry with a hard-coded block, so I probed it: `rpc.hyperliquid.xyz/evm` served `eth_getStorageAt` / `eth_getCode` / `eth_call` at 20M blocks back. It's effectively an archive node. Recording it here so it doesn't get re-litigated later.
- Block 44,686,920 is `2026-08-31 21:09:40 UTC`; head at review time was 45,374,639 (`2026-09-08 17:04 UTC`).
- I could not reproduce the compiler error locally with 1.8.1 under `FOUNDRY_PROFILE=local-test forge build` — it succeeded with warnings only. That is *consistent* with the CI evidence rather than contradicting it: the failing step is the default profile, which differs in `revert_strings = "strip"`, and that setting changes bytecode layout, which is exactly what an assembly tag-space error is sensitive to. Worth knowing when someone tries to reproduce this by hand.

### Main substantive note

The pin freezes EVM state but Core reads still track live head, so these suites remain partly nondeterministic — with a skew that grows about a day per day. Detail in the inline comment on `BaseSimulatorTest.sol`; low exposure for the three suites as written, so it's a comment and not a change request.

One description nit: "all 45 tests across the three affected simulator suites" — run [34253659880](https://github.com/across-protocol/contracts/actions/runs/34253659880) shows **53** (2 + 8 + 8 + 35; `SponsoredDirectFlow.t.sol` declares two test contracts). All pass, so the conclusion holds; just the count is off.

droplet-rl:
  • Reproduce v1.8.1 build failure
  • forge 1.8.1 build for Tag-too-large error

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=161c228d430663d85bff841d2092be70&trace_start_ts=1788886718&trace_end_ts=1788973418)

PR #1549, opened 2026-09-09
build: isolate Foundry with mise locally and in CI

Reinis-FRP:
Use repository-scoped Foundry installations so local commands and CI use the same versions without overwriting the shared `~/.foundry/bin` tools used by other repositories.

- Pin standard Foundry **1.7.1** in `mise.toml` and the optional **foundry-zksync v0.1.9** fork in `mise.zksync.toml`.
- Select the appropriate installation explicitly for build, test, script, and verification commands. Repo-owned shell scripts load the pin automatically, including when an incompatible Foundry is already on PATH. Shell activation is unnecessary.
- Use the same mise setup for CI artifact generation, EVM tests, and publishing, and include the standard Foundry configuration and `foundry.lock` in the artifact cache key.
- Exclude test/script directories from zkSync builds, including ordinary `.sol` helpers that use EVM-only features such as `type(T).runtimeCode`.
- Read selected versions through mise and report installation hints for missing executables.
- Document setup and add smoke tests for installation, version selection, compilation, and isolation from another repository’s tools. Run the platform matrix when toolchain/compiler configuration, dependencies, CI, or tooling scripts change.

The compiler settings remain unchanged: solc **0.8.30** and zksolc **1.5.15**.

Validation: local isolation tests and the full zkSync target build pass. CI covers Linux, Apple Silicon macOS, and Intel macOS; the zkSync checks run on Linux and Apple Silicon because the pinned fork has no Intel macOS binary. Formatting and lint checks pass.

To try on macOS:

```sh
brew install mise
gh pr checkout 1549
yarn install --frozen-lockfile
mise trust mise.toml
yarn pin-foundry
yarn test-foundry-toolchain
```

For zkSync on Apple Silicon macOS or Linux:

```sh
mise trust mise.zksync.toml
yarn pin-foundry-zksync
yarn test-foundry-toolchain --zksync
yarn forge-build-zksync
```

Use `yarn foundry forge ...` for direct standard Foundry commands and `yarn foundry-zksync forge ...` for the fork.


package.json
```diff
@@ -42,20 +45,24 @@
     "prepublishOnly": "yarn build-ts && yarn export-idl && yarn export-evm-artifacts",
     "build": "yarn build-evm-foundry && yarn build-svm && yarn generate-svm-artifacts && yarn build-ts",
     "build-verified": "yarn build-evm-foundry && yarn build-svm-solana-verify && yarn generate-svm-artifacts && yarn build-ts",
-    "pin-foundry": "sh scripts/foundry-version.sh pin",
-    "build-evm-foundry": "forge build",
+    "pin-foundry": "mise install foundry && yarn check-foundry",
+    "check-foundry": "mise exec -- node scripts/checkFoundryVersion.js",
+    "test-foundry-toolchain": "node scripts/testFoundryToolchain.js",
+    "foundry": "mise exec --",
+    "check-storage-layouts": "mise exec -- bash scripts/checkStorageLayout.sh",
```
droplet-rl: This is the right pattern, but it was applied to only one of the scripts that shells out to Foundry. Still calling bare `forge`/`cast`:

- `scripts/diffStorageLayout.sh:12` — `forge inspect ... storagelayout`. Direct sibling of the script you just wrapped here, and storage layouts are exactly the compiler-version-sensitive output this PR exists to pin.
- `scripts/evm-contract-sizes.sh:15` — `forge build --sizes --skip test`. The `evm-contract-sizes` entry in package.json got `mise exec`, but this standalone twin didn't, so the two now disagree on toolchain.
- `script/mintburn/checkSponsoredPeripheryProdReadiness.sh:1000` — `require_cmd cast`. This is the sharper one: the README now lists mise as the only Foundry requirement, so a fresh contributor has no `cast` on `PATH` at all and this exits with "missing command" rather than working.
- `scripts/verifyBytecode.sh`, `scripts/verifyBytecodeEraVM.sh` — bare `cast` throughout. Low risk in practice (`cast keccak`/`chain-id`/`tx` are version-stable), but same `require`-style failure for a mise-only setup. `verifyBytecodeEraVM.sh:86` also *prints* `FOUNDRY_PROFILE=zksync forge script ...` as a suggested next command, which is now the wrong invocation — line 12 of the same file already says `yarn forge-script-zksync`.

I verified `mise exec` sets PATH for the entire child process tree, so the cheapest fix is a yarn wrapper per script (as you did here), or a two-line self-re-exec guard at the top of each script:

```sh
[ -n "${MISE_SHELL:-}" ] || command -v forge >/dev/null || exec mise exec -- "$0" "$@"
```

The README does say "Bare `forge` still uses your shell's selected tool", so this is a documented limitation rather than a surprise — but for scripts the repo itself owns I'd rather close the gap than document it.

package.json
```diff
@@ -17,12 +17,15 @@
...
+    "pin-foundry-zksync": "mise -E zksync install github:matter-labs/foundry-zksync && yarn check-foundry-zksync",
+    "check-foundry-zksync": "mise -E zksync exec -- node scripts/checkFoundryVersion.js --zksync",
+    "foundry-zksync": "mise -E zksync exec --",
     "preforge-script-zksync": "yarn -s check-foundry-zksync",
-    "forge-script-zksync": " FOUNDRY_PROFILE=zksync forge script --zksync --skip test --skip '**/contracts/test/**' --suppress-errors sendtransfer",
+    "forge-script-zksync": "FOUNDRY_PROFILE=zksync mise -E zksync exec -- forge script --zksync --skip '**/test/**' --suppress-errors sendtransfer",
     "preforge-build-zksync": "yarn -s check-foundry-zksync",
-    "forge-build-zksync": "FOUNDRY_PROFILE=zksync forge build --zksync --suppress-errors sendtransfer --skip test --skip script --skip '**/contracts/test/**'",
-    "forge-verify-zksync": "FOUNDRY_PROFILE=zksync forge verify-contract --zksync",
+    "forge-build-zksync": "FOUNDRY_PROFILE=zksync mise -E zksync exec -- forge build --zksync --suppress-errors sendtransfer --skip '**/test/**' --skip '**/script/**'",
```
droplet-rl: Worth knowing what was traded away here. `--skip test` / `--skip script` aren't globs — forge special-cases those two strings into `SkipBuildFilter::Tests`/`Scripts`, which match `*.t.sol` / `*.s.sol` **by filename, anywhere in the tree**. Replacing them with directory globs is a net win (it's what catches the non-`.t.sol` helpers under `test/evm/foundry/`), but it does drop the filename-based coverage: a `.t.sol` added outside a `test/` directory would now get pulled into the zkSync build.

I checked — nothing currently lives outside `test/`/`script/`, so this is latent, not broken. The filters are additive, so keeping both costs nothing:

```
--skip test --skip script --skip '**/test/**' --skip '**/script/**'
```

Same for `forge-script-zksync` on line 24 (keep `--skip test` alongside `--skip '**/test/**'`; obviously not `--skip script` there).

package.json
```diff
@@ -17,12 +17,15 @@
   "types": "dist/index.d.ts",
   "main": "dist/index.js",
   "scripts": {
-    "check-foundry-zksync": "forge --version | grep -q foundry-zksync-v0.1.9 || (echo 'zksync builds require foundry-zksync v0.1.9 (current: '\"$(forge --version | head -1)\"'). Install: https://foundry-book.zksync.io/getting-started/installation' && exit 1)",
+    "pin-foundry-zksync": "mise -E zksync install github:matter-labs/foundry-zksync && yarn check-foundry-zksync",
```
droplet-rl: The tool id is now spelled out in two places — here and in `mise.zksync.toml`. Since `mise.zksync.toml` disables `foundry`, the fork is the *only* tool in that env, so a bare `mise -E zksync install` installs exactly the same thing and can't drift from the config. I confirmed `mise -E zksync ls` lists only `github:matter-labs/foundry-zksync`.

scripts/checkFoundryVersion.js
```diff
@@ -0,0 +1,15 @@
+const assert = require("node:assert/strict");
+const { execFileSync } = require("node:child_process");
+const { readFileSync } = require("node:fs");
+const { join } = require("node:path");
+
+const zksync = process.argv.includes("--zksync");
+const config = zksync ? "mise.zksync.toml" : "mise.toml";
+const pin = readFileSync(join(__dirname, "..", config), "utf8").match(/^version = "([^"]+)"$/m)?.[1];
```
droplet-rl: Regex-parsing the TOML takes the *first* `version = "..."` line in the file, so this silently reads the wrong tool's pin the moment either config grows a second entry — and it won't see the inline form (`[tools]` + `foundry = "1.7.1"`) at all. Since mise is already a hard dependency, it can answer the question itself:

```js
const pin = execFileSync("mise", [...(zksync ? ["-E", "zksync"] : []), "current", tool], { encoding: "utf8" }).trim();
```

I checked both shapes on mise 2026.9.3: `mise current foundry` → `1.7.1`, `mise -E zksync current` → `github:matter-labs/foundry-zksync 0.1.9`, and `mise ls --json` gives `requested_version` plus the originating config path if you want to keep asserting *which* file the pin came from.

Minor, same file: when a binary isn't on PATH, `execFileSync` throws a raw `ENOENT` rather than the nice "run yarn pin-foundry" hint on line 13. Worth catching.

droplet-rl (commented):
The approach is sound and the mechanics check out. I reproduced the whole flow locally on Linux in a container that happened to already have an ambient `~/.foundry/bin` at forge **1.8.1** — i.e. exactly the version AGENTS.md warns breaks the solc 0.8.30 build — and confirmed:

- `mise install foundry` → `forge/cast/anvil/chisel` all report `1.7.1`; `checkFoundryVersion.js` passes.
- `mise -E zksync install github:matter-labs/foundry-zksync` resolves `foundry_zksync_v0.1.9_linux_amd64.tar.gz`; `forge --version` → `1.3.5-foundry-zksync-v0.1.9`. `disable_tools = ["foundry"]` works — `mise -E zksync ls` shows only the fork, and `mise ls` (default) still shows `foundry 1.7.1` active.
- `yarn test-foundry-toolchain --zksync` passes end-to-end (~10s warm), including the EraVM compile and the `--help` selection checks. I verified yarn 1.22 does *not* swallow `--help`/`--version`, so those assertions are real rather than vacuous.
- `mise exec` mutates PATH for the whole child process tree, so FFI-spawned tools (e.g. `@openzeppelin/foundry-upgrades`) get the pinned binaries too. `mise exec` also auto-installs a missing pin, so `yarn build-evm-foundry` works without `yarn pin-foundry` first.
- `jdx/mise-action@c2a8761` is genuinely `v4.3.0`, and `install_args` / `add_shims_to_path` / `export_path` / `env` are all real inputs on that ref. It sets `MISE_TRUSTED_CONFIG_PATHS=$PWD` and `MISE_YES=1`, so the `mise trust mise.zksync.toml` step in CI is redundant (harmless) — it's only needed locally.
- No Intel macOS asset exists on the `foundry-zksync-v0.1.9` release, so gating `--zksync` off `macos-15-intel` is correct.
- The `--skip` change is a real fix, not just a reshuffle: `test/evm/foundry/` holds plenty of non-`.t.sol` helpers (`eraVM_EIP7702.sol`, `HubPoolTestBase.sol`, the vendored `hyper-evm-lib` tree) that `--skip test` never excluded.

One thing I'd like resolved before merge, and a handful of optional nits, inline. The main gap: this moves Foundry off `PATH`, but several repo-owned shell scripts still shell out to bare `forge`/`cast`, so they now either silently use an unpinned toolchain or hard-fail for a contributor who follows the new README and installs only mise.

Note CI was still mid-flight when I looked — everything green except **EVM Tests**, which was `in_progress` (the 12-minute EVM artifact job passed). Worth confirming that one before merge.

.github/actions/setup-foundry/action.yml
```diff
@@ -0,0 +1,15 @@
+name: Set up Foundry
+description: Install the repository's Foundry version without activating it globally
+runs:
+  using: composite
+  steps:
+    - uses: jdx/mise-action@c2a87611a18de5b3828c5652fe268e992400cb5c # v4.3.0
+      with:
+        version: 2026.9.3
+        install_args: foundry
```
droplet-rl: `install_args: foundry` is currently a no-op (foundry is the only tool in `mise.toml`) but it turns into a silent footgun later: add a second tool to `mise.toml` and CI keeps installing only foundry, with no error to point at. Dropping the input installs everything the config asks for, which is the same thing today and the right thing tomorrow.

Also worth a line in the `description`: this action assumes the repo is already checked out **and** Node is set up, because of the `node scripts/checkFoundryVersion.js` step on line 15. All three current call sites satisfy that, but it's an invisible ordering constraint for the next caller.

.github/actions/cache-evm-artifacts/action.yml
```diff
@@ -23,7 +23,7 @@ runs:
         # The job that generates the artifacts is responsible for archiving them to the cache tarball. This avoids any
         # conflicts with other caching actions that might have cleaned some of cached contents.
         path: evm-artifacts.tar
-        key: evm-artifacts-${{ runner.os }}-node-${{ steps.resolved-node.outputs.version }}-${{ hashFiles('yarn.lock', 'foundry.toml', '.foundry-version', 'contracts/**/*.sol') }}
+        key: evm-artifacts-${{ runner.os }}-node-${{ steps.resolved-node.outputs.version }}-${{ hashFiles('yarn.lock', 'foundry.toml', 'mise.toml', 'contracts/**/*.sol') }}
```
droplet-rl: Swapping `.foundry-version` → `mise.toml` is the correct translation. While you're in here: `foundry.lock` isn't in the hash, and it pins the `lib/` deps that `forge install` resolves (`forge-std` v1.11.0, `murky`, `sp1-contracts`) — all of which compile into `out/`. Bumping a submodule rev today reuses stale artifacts. Pre-existing, not introduced here, but this is the one line that would fix it.

README.md
```diff
@@ -68,23 +90,33 @@ yarn lint-fix
...
+commands select it with `mise -E zksync exec` and set `FOUNDRY_PROFILE=zksync`. Use `yarn forge-script-zksync ...`
+to run deployment scripts, `yarn forge-verify-zksync ...` for verification, or `yarn foundry-zksync forge ...`
+for direct fork commands (set the Foundry profile/flags yourself for direct commands).
+The zkSync build excludes test and script directories, including ordinary `.sol` helpers that use EVM-only features.
+
+Standard Foundry and other repositories keep their own tools; no `foundryup-zksync` or shell activation is needed.
+The independent compiler pins remain in `foundry.toml`: `solc 0.8.30` and `zksolc 1.5.15`.
+Run `yarn test-foundry-toolchain --zksync` to test both pins, compile a small EraVM contract, and check isolation.
+CI runs this on Linux and Apple Silicon macOS. The pinned fork has no Intel macOS release binary; use a Linux
+environment for zkSync work on an Intel Mac. Standard Foundry remains supported on Intel macOS.
```
droplet-rl, line 124: Removing the "Alternatively, the `yarn forge-script-zksync` command can be used to deploy the contract." sentence orphaned the ```shell block right below this paragraph — it now reads as a dangling example hanging off a sentence about Intel macOS. Either fold it into the `yarn forge-script-zksync ...` mention two paragraphs up or give it a lead-in line.

.github/workflows/pr.yml
```diff
@@ -203,20 +203,38 @@ jobs:
...
       - name: Test evm-foundry
         run: yarn test-evm-foundry
+
+  test-foundry-toolchain:
+    name: Foundry toolchain (${{ matrix.os }})
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-15, macos-15-intel]
```
droplet-rl, line 248: Optional. This is a toolchain-plumbing smoke test that only changes behavior when `mise.toml` / `mise.zksync.toml` / `scripts/*FoundryToolchain*` / this action change, but it runs a 3-OS matrix on every PR — and macOS runners bill at 10×. From this PR's own run (43s + 1m13s on the two macOS legs) that's ~20 billable minutes per PR forever.

The repo already has the pattern for this: `test-svm-verified` uses `dorny/paths-filter`. A `paths:` filter or a filter step would keep the coverage where it matters. Genuinely fine to leave as-is if you'd rather have it always-on for the next few weeks while the setup settles.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=0f84d79fea71f1d2cc473a53f1ec25af&trace_start_ts=1788947208&trace_end_ts=1789033908)

package.json
Reinis-FRP: Fixed in [ada6bcc5](https://github.com/across-protocol/contracts/commit/ada6bcc5dad07f8557e4c8db296a93a81080bbfb). The shell entry points now source a shared mise PATH helper, so an ambient Foundry cannot win. The regression test checks pin selection from another working directory and preserves arguments/stdin. The EraVM command hint also uses the Yarn wrapper.

---

package.json
Reinis-FRP: Fixed in [ada6bcc5](https://github.com/across-protocol/contracts/commit/ada6bcc5dad07f8557e4c8db296a93a81080bbfb). Kept the filename aliases alongside the directory globs. The zkSync smoke test now includes EVM-only `.t.sol` and `.s.sol` files outside the test/script directories.

---

package.json
Reinis-FRP: Fixed in [ada6bcc5](https://github.com/across-protocol/contracts/commit/ada6bcc5dad07f8557e4c8db296a93a81080bbfb). Changed this to `mise -E zksync install`, so the environment configuration defines which tools are installed.

---

scripts/checkFoundryVersion.js
Reinis-FRP: Fixed in [ada6bcc5](https://github.com/across-protocol/contracts/commit/ada6bcc5dad07f8557e4c8db296a93a81080bbfb). The checker now queries `mise current` for the named tool. Added regressions for inline TOML with another tool's version first, and for a missing executable returning the installation hint.

---

.github/actions/setup-foundry/action.yml
Reinis-FRP: Fixed in [ada6bcc5](https://github.com/across-protocol/contracts/commit/ada6bcc5dad07f8557e4c8db296a93a81080bbfb). Removed `install_args` and documented the checkout/Node prerequisites in the action description. Also removed the redundant CI trust step.

---

.github/actions/cache-evm-artifacts/action.yml
Reinis-FRP: Fixed in [ada6bcc5](https://github.com/across-protocol/contracts/commit/ada6bcc5dad07f8557e4c8db296a93a81080bbfb). Added `foundry.lock` to the EVM artifact cache key so changes to the pinned Foundry dependencies invalidate the cache.

---

README.md
Reinis-FRP: Fixed in [ada6bcc5](https://github.com/across-protocol/contracts/commit/ada6bcc5dad07f8557e4c8db296a93a81080bbfb). Added an explicit lead-in identifying the command as a deployment and verification example.

---

.github/workflows/pr.yml
Reinis-FRP: Fixed in [ada6bcc5](https://github.com/across-protocol/contracts/commit/ada6bcc5dad07f8557e4c8db296a93a81080bbfb). Added an Ubuntu paths-filter job before the matrix. The platform tests run for toolchain/compiler configuration, dependency locks, CI configuration, and shell/tooling script changes; contract-only changes skip the matrix.

---

scripts/testFoundryToolchain.js
```diff
@@ -0,0 +1,154 @@
...
+      join(repo, "scripts/setupFoundryEnv.sh"),
+      join(repo, "scripts/checkFoundryVersion.js"),
+      otherRepo,
+      "argument with spaces",
+      process.execPath,
+    ],
+    otherRepo,
+    "preserved input\n"
+  );
+  run("bash", [join(repo, "script/mintburn/checkSponsoredPeripheryProdReadiness.sh"), "--help"], otherRepo);
```
droplet-rl, line 59: This assertion doesn't cover what it appears to. `--help` is handled at `checkSponsoredPeripheryProdReadiness.sh:988-990` (`-h | --help)` → `usage; exit 0`), which is **eleven lines before** the `require_cmd cast` at line 1001. So the script exits before it ever probes for `cast`, and this only proves that sourcing `setupFoundryEnv.sh` from an unrelated cwd doesn't break startup.

The behavior itself is fine — I confirmed separately that sourcing the helper puts the pinned `cast` ahead of a planted foreign one, so `require_cmd cast` does pass now. It's the coverage that's thin, which matters because this is the regression the line is standing in for.

Cheapest way to actually reach the gate is to assert on the helper's effect rather than on the script's `--help`:

```js
run("bash", ["-euc", 'source "$1"; command -v cast; command -v forge', "bash", join(repo, "scripts/setupFoundryEnv.sh")], otherRepo);
```

which fails loudly if either binary is missing from the sourced `PATH`. Keep the `--help` call too if you want the smoke coverage — it's cheap.

scripts/setupFoundryEnv.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+# Source before invoking Foundry. Preserve the caller's cwd, arguments, and stdin.
+_across_foundry_root="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+PATH="$(mise -C "$_across_foundry_root" exec -- sh -c 'printf "%s" "$PATH"')" || exit $?
```
droplet-rl: Three small things on this line, all cosmetic-to-minor:

1. **`exit` in a sourced file kills the caller's shell.** I confirmed that `source scripts/setupFoundryEnv.sh` in an interactive shell with mise unavailable terminates the shell outright. Nobody's documented workflow does this (the README points at `yarn foundry ...`), but the filename and the "Source before invoking Foundry" comment invite it. `return` is the correct construct for a sourced file; the usual portable idiom is `return 1 2>/dev/null || exit 1`.

2. **No install hint.** The failure output is just `mise: command not found` (exit 127). Correct and fail-fast — I verified it never falls through to the ambient forge, which is the property that matters — but a guard would point somewhere useful:
   ```bash
   command -v mise >/dev/null || { echo "mise not found; see Requirements in README.md" >&2; return 1 2>/dev/null || exit 1; }
   ```

3. **`PATH` is assigned before the status is checked**, so on failure it is briefly clobbered to the empty string. Unreachable in practice — `exit` fires immediately, and I couldn't reach a state where the caller continues with an empty `PATH` even with `source ... || true` — so this is belt-and-braces only. Landing it in a temp var and validating before `export` sidesteps it and reads more clearly.

Also `_across_foundry_root` stays set in the caller's shell after sourcing (verified). Namespaced, so harmless, but an `unset` on the last line would keep it tidy.

script/mintburn/checkSponsoredPeripheryProdReadiness.sh
```diff
@@ -156,7 +157,7 @@ build_cast_send() {
 	local chain_id="$3"
 	shift 3
 
-	local cmd="cast send $address \"$signature\""
+	local cmd="yarn foundry cast send $address \"$signature\""
```
droplet-rl, line 160: Right call updating this copy-paste hint alongside the one in `verifyBytecodeEraVM.sh:87`. One sibling got missed with the same problem: `script/mintburn/cctp/setUpTokens.sh` lines 41, 44 and 47 still `echo` bare `cast call ...` / `cast send ...` as commands for the operator to paste. Same failure mode you just fixed here — a contributor set up per the new README has no `cast` on `PATH`, so the printed instruction doesn't run.

That file has no executable `forge`/`cast` calls of its own (I checked — the three hits are all inside `echo`), so it correctly does *not* need to source `setupFoundryEnv.sh`. Just the printed strings need the `yarn foundry ` prefix.

.github/workflows/pr.yml
```diff
@@ -203,20 +203,59 @@ jobs:
...
+              - 'package.json'
+              - 'yarn.lock'
+              - '.github/**'
+              - 'scripts/**'
+              - 'script/**/*.sh'
+
+  test-foundry-toolchain:
+    name: Foundry toolchain (${{ matrix.os }})
+    needs: foundry-toolchain-changes
+    if: needs.foundry-toolchain-changes.outputs.changed == 'true'
```
droplet-rl, line 243: The path filter does the job — the gate job costs 7s and the matrix legs are 22s/43s/1m1s, so this is cheap. Two notes:

1. **Skipped jobs and required checks.** If `Foundry toolchain (ubuntu-latest|macos-15|macos-15-intel)` are configured as required status checks in branch protection, gating them behind `if:` makes them report `skipped` on unrelated PRs. GitHub generally treats a skipped job's check run as non-blocking, but the semantics are configuration-dependent and worth confirming before this lands on `master` — I couldn't read the protection rules (404, no permission) so I can't check for you.

2. **The gate could be a step rather than a job.** `test-svm-verified` already does the filter inline (`dorny/paths-filter` at line 134) and conditions the *step*, which avoids both the extra runner and the skipped-check question. Following that precedent would make the two mechanisms consistent. Entirely optional — the separate job reads more clearly for a matrix, which is presumably why you did it this way.

Filter coverage itself looks right: `mise*.toml` catches both configs, and `script/**/*.sh` picks up the mintburn script that now sources the helper.

package.json
droplet-rl: Now that `checkStorageLayout.sh` sources `setupFoundryEnv.sh` itself, this wrapper double-wraps — `mise exec` sets up the env, then the script asks mise for the `PATH` a second time. Harmless (~2 mise resolutions, tens of ms) and arguably good defence in depth, but `bash scripts/checkStorageLayout.sh` is now sufficient and would make this entry consistent with how the other self-loading scripts are invoked.

droplet-rl (approved):
All eight items from my last pass are addressed, and I re-validated the new code rather than taking the description's word for it. This time the container came up with **forge 1.8.1 on the default `PATH`** at `~/.foundry/bin/forge` — the exact version AGENTS.md says breaks the solc 0.8.30 build — so the hostile case was live rather than simulated.

What I verified on `ada6bcc5`:

- **`setupFoundryEnv.sh` actually wins over a foreign toolchain.** Planted a fake `cast` (exit 73) ahead of everything on `PATH`: before sourcing it resolved to the fake, after sourcing it resolved to `~/.local/share/mise/installs/foundry/1.7.1/cast` reporting `1.7.1`. The `require_cmd cast` gate in the mintburn script now passes on a mise-only machine.
- **Failure mode is fail-fast, not fail-open.** With mise absent the sourcing scripts abort with 127 and never silently fall through to an ambient forge — checked bare, under `set -euo pipefail`, and via `checkStorageLayout.sh` directly.
- **The restored `--skip test` / `--skip script` filters do real work.** The zkSync leg of the smoke test now plants `src/EvmOnly.t.sol` and `src/EvmOnly.s.sol` containing `type(T).runtimeCode` inside `src/`, and the build logs `Compiling 1 files with zksolc` — so the filename filters excluded them. That would have failed under the previous directory-glob-only version, which is exactly the latent gap I flagged.
- **`checkFoundryVersion.js` via `mise current` is correct.** `mise -C <root> current foundry` → `1.7.1`; `mise -E zksync current github:matter-labs/foundry-zksync` → `0.1.9`, and it reports the *configured* version even before install, which is the right semantic for a pin check. The ENOENT path produces both `spawnSync forge ENOENT` and the `Run yarn pin-foundry` hint at exit 1.
- **`mise -E zksync install` installs only the fork** — `disable_tools` keeps standard Foundry out, so dropping the explicit tool id is safe.
- `yarn test-foundry-toolchain --zksync` passes cold from an empty mise store in 13s. Full CI is now green including **EVM Tests (7m27s)**, the one check that was still in flight last time.

Nice touch on the `test "$PWD" -ef "$3"` in `ada6bcc5` — string-comparing cwd would have been flaky on macOS via the `/tmp` → `/private/tmp` symlink.

Approving. Four nits inline, none blocking; the only one I'd actually bother with is the test assertion that short-circuits before the thing it looks like it's testing.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=c52619465c9c784a9cae0a0a738a1167&trace_start_ts=1788950405&trace_end_ts=1789037105)

scripts/testFoundryToolchain.js
Reinis-FRP: Fixed in [2b80742d](https://github.com/across-protocol/contracts/commit/2b80742d609221c6627c8649b5353e5f0769008c). The sourced-shell probe now explicitly checks `command -v cast` and `command -v forge`, then verifies the actual pinned versions. The `--help` call is labelled as startup-only coverage.

---

scripts/setupFoundryEnv.sh
Reinis-FRP: Fixed in [2b80742d](https://github.com/across-protocol/contracts/commit/2b80742d609221c6627c8649b5353e5f0769008c). The helper now returns failure without changing PATH, reports an install hint when mise is missing, and cleans up its local state. Repo entry points explicitly stop on failure. Regression tests cover missing/failing mise, caller-shell survival, PATH preservation, and the storage script stopping before any ambient forge runs.

---

script/mintburn/checkSponsoredPeripheryProdReadiness.sh
Reinis-FRP: Fixed in [2b80742d](https://github.com/across-protocol/contracts/commit/2b80742d609221c6627c8649b5353e5f0769008c). All three printed commands in `setUpTokens.sh` now use `yarn foundry cast`; the script itself still needs no Foundry environment setup.

---

.github/workflows/pr.yml
Reinis-FRP: Checked the available `master` metadata: `required_status_checks.contexts` and `checks` are empty, and `/rules/branches/master` returns `[]`. Keeping the separate Ubuntu gate avoids allocating the macOS matrix for unrelated PRs; a step-level filter would still start those runners. No branch rules were changed.

---

package.json
Reinis-FRP: Fixed in [2b80742d](https://github.com/across-protocol/contracts/commit/2b80742d609221c6627c8649b5353e5f0769008c). Removed the outer `mise exec`; the Yarn command now calls `bash scripts/checkStorageLayout.sh`, which loads the pinned tools itself.

---

scripts/setupFoundryEnv.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+# Source before invoking Foundry. Preserve the caller's cwd, arguments, and stdin.
+_across_setup_foundry_env() {
+    local repo_root pinned_path
+    command -v mise >/dev/null || {
+        echo "mise not found; see Requirements in README.md" >&2
+        return 1
+    }
+    repo_root="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)" || return 1
+    pinned_path="$(mise -C "$repo_root" exec -- sh -c 'printf "%s" "$PATH"')" || return 1
```
droplet-rl, line 14: One silent-failure path survives, and it's worth a one-line guard because it defeats the whole point of the helper.

`mise -C <dir> exec` **exits 0 when `<dir>` has no mise config** — it just returns the ambient `PATH` unchanged. So if `repo_root` ever computes wrong, the helper reports success while handing back an unpinned toolchain. I hit this by sourcing the helper through a symlink:

```
$ ln -sf /workspace/contracts/scripts/setupFoundryEnv.sh /tmp/linked.sh
$ cd /tmp && bash -c 'source /tmp/linked.sh && echo "helper rc=0"; forge --version'
helper rc=0
forge Version: 1.8.1        # <-- the ambient forge this PR exists to avoid
```

`${BASH_SOURCE[0]}` is the unresolved path, so `dirname .. ` became `/`, and `mise -C /` succeeded with no config. I checked the five realistic invocation styles (absolute, relative from repo root, relative from `scripts/`, `../`-relative from another cwd, and via the real callers) — **all correct**; only the symlinked-helper case misresolves. So this is unlikely, not reachable through any documented workflow, and I'd have skipped it entirely if the consequence were an error rather than a silently wrong compiler.

Symlink resolution isn't worth it (no portable `readlink -f` across macOS versions). Asserting the root is the root is, and it catches every misresolution rather than just this one:

```bash
repo_root="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)" || return 1
[ -f "$repo_root/mise.toml" ] || {
    echo "mise.toml not found under $repo_root; run this from a repository checkout" >&2
    return 1
}
```

I confirmed the guard fails loudly for the symlink case (`computed repo_root=[/]`) and passes for the real root.

Unrelated, purely cosmetic: `unset -f` is duplicated on lines 15 and 17. Hoisting it above the `if` would read slightly better, but the current form is correct.

scripts/exportEvmArtifacts.sh
```diff
@@ -13,7 +13,7 @@ if ! command -v jq >/dev/null 2>&1; then
 fi
 
 if [ ! -d "$OUT_DIR" ]; then
-  echo "Error: $OUT_DIR not found. Run 'forge build' first." >&2
+  echo "Error: $OUT_DIR not found. Run 'yarn build-evm-foundry' first." >&2
```
droplet-rl, line 16: Good catch updating this hint. One stale sibling in the same class you've been sweeping: `scripts/verifyPrDeployments.sh:15` still documents its prerequisite as `#   - Contracts built (forge build)`. Comment-only, no executable path, so it correctly doesn't source the helper — just the wording is now inconsistent with `yarn build-evm-foundry` here. Genuinely trivial; fold it in only if you're touching the file anyway.

droplet-rl (dismissed):
All four nits from the last pass are fixed, and I re-tested the rewritten helper rather than eyeballing it. The container again came up with **forge 1.8.1 on the default `PATH`**, so every check below ran against a live incompatible toolchain.

Verified on `2b80742d`:

- **The `return 1 2>/dev/null || exit 1` idiom is correct in both modes.** Sourced → returns 1 and the caller's `|| exit $?` fires; executed directly (`bash scripts/setupFoundryEnv.sh`) → exits 1 with the message. And sourcing it interactively with mise missing no longer kills the shell — `source ... || echo handled` prints `handled rc=1` and the shell survives. That was my main concern last round.
- **`PATH` is genuinely untouched on failure.** Reproduced both new fixture cases by hand: mise absent → `mise not found; see Requirements in README.md`, `PATH` byte-identical, function unset; mise present but exiting 42 → same, and the garbage stdout never reaches `PATH` because it lands in the `local`. `checkStorageLayout.sh` under that broken mise exits 1 having printed nothing — it does *not* fall through to the planted ambient `forge`.
- **The strengthened probe now shows its work.** `command -v cast; command -v forge` in the sourced-shell test prints `…/installs/foundry/1.7.1/{cast,forge}` in the log even with the fake bin dir first on `PATH`, so the assertion is no longer inferential. `assert.doesNotMatch(stopped.stdout, /Comparing storage layout/)` is also non-vacuous — that string is `checkStorageLayout.sh:20`, the first thing its loop prints.
- **Cold start is clean.** From an empty mise store with no `yarn pin-foundry`, sourcing the helper auto-installs and yields a correct `PATH` — mise's progress output goes to stderr, so it can't corrupt the captured value. I specifically went looking for that failure mode; it isn't there.
- **Dropping the `mise exec` wrapper on `check-storage-layouts` is safe**, and CI confirms it: the `Inspect storage layouts` step ran the unwrapped `yarn check-storage-layouts` with `MISE_TRUSTED_CONFIG_PATHS` set by mise-action, and **EVM Tests passed** (I waited for it — all 11 checks are now green).
- **Coverage is complete.** All six sourcing sites carry `|| exit $?`; the only remaining `forge`/`cast` mentions in shell scripts are comments or `echo`'d hints, and `setUpTokens.sh` now prefixes all three with `yarn foundry`.

Approving. One low-severity find inline — a silent wrong-toolchain path I hadn't probed before — plus one trivial doc leftover. Neither blocks.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=f049f8edc83bd48f92a7f1552e4520e9&trace_start_ts=1788951471&trace_end_ts=1789038171)

scripts/setupFoundryEnv.sh
Reinis-FRP: Fixed in [4b2ba98d](https://github.com/across-protocol/contracts/commit/4b2ba98dfe644223113a352a5e331bb770c163d5). Added the repository mise.toml guard before invoking mise. The regression test sources the helper through a symlink and verifies it fails with a clear error while preserving PATH and the caller's shell.

---

scripts/exportEvmArtifacts.sh
Reinis-FRP: Fixed in [4b2ba98d](https://github.com/across-protocol/contracts/commit/4b2ba98dfe644223113a352a5e331bb770c163d5). Updated verifyPrDeployments.sh's build prerequisite to `yarn build-evm-foundry`.

---

PR #1551, opened 2026-09-09
feat(svm): retire callbacks, verify V5 swaps, and separate error ranges

Reinis-FRP:
Retires legacy SVM fill callbacks and assigns distinct Spoke runtime error ranges. Destination actions move to the V5 Gateway tape. Legacy `fill_relay` now accepts only an empty message and explicitly rejects callback payloads. Plain legacy fills and the V5 relay witness remain supported. Removes Spoke's legacy callback execution, its `invoke_handler` helper, and its dependency on `multicall-handler`. Updates fixtures and the legacy devnet script, and documents the migration. The standalone MulticallHandler program and package exports remain available.

Adds a pinned Raydium CPMM source fixture to the real-Gateway lane. It exercises an Across fill followed by delegate approval, a real swap CPI, an output floor, and full output delivery to the committed recipient. Tests cover reserve movement, replay rejection, and actual failed-transaction rollback for swap slippage and an unmet output floor. The fixture reads on-chain time from the Clock sysvar. This validates pinned source with local liquidity, not equivalence to a deployed mainnet binary or arbitrary swap routes.

The migration docs apply the same full-balance semantics and delivery policy as EVM V5, and describe deployment sequencing for old routes and in-flight deposits. Runtime errors now use distinct ranges: Common 6000, SVM 7000, CallData 8000, V5 9000. Compatibility tests pin each range boundary, including `LegacyFillMessageUnsupported = 7019`. Common codes are preserved; SVM/CCTP codes are renumbered and V5 is new in this release. The [complete runtime-code migration table](https://github.com/across-protocol/contracts/blob/0ef4e852647f2332ef13d0999f0ce890775d282f/programs/svm-spoke/ERROR_CODES.md) lists all 62 current variants and eight removed callback errors against a pinned master baseline, distinguishes new V5 errors from existing errors, and explains historical decoding. Deployment sequencing requires inspecting numeric-code consumers and updating affected maps before the upgrade, including non-callback paths. The existing incomplete generated Anchor IDL error table remains a separate limitation.

Stacked on #1550 (`reinis/acp-221-remove-slow-fills`).

Validation:

- Real-Gateway suite: 18 passing with the review fixes to the Clock wait, receipt diagnostics, and account assertions, before the final error-range renumbering. Covers delivery/replay, failed-swap rollback, and failed-output-floor rollback.
- Documentation follow-up: all 70 rows (62 current and eight removed variants) checked against the Rust enums and pinned master baseline; local links, formatting, and normal commit hooks passed.
- Rust library tests after error-range renumbering: 21 passing with the test feature; 17 passing without it, including error-code compatibility assertions.
- Full ordinary SVM suite: [145 passing in verified-build CI](https://github.com/across-protocol/contracts/actions/runs/34618323308/job/103326466421) at `88181f55`, including the reduced callback fixture. Callback tests cover inline and buffered parameter loading, with malformed and well-formed payloads rejected identically before parsing; V5-prefixed message rejection is covered separately.
- Focused TypeScript checks include the changed test files. Rust formatting, Prettier, migration-script exit/message checks, and `git diff --check` passed. Production IDL/client regeneration was validated separately from test-only IDLs.

The EVM artifact job currently fails inside solc 0.8.30 with `Tag too large for reserved space`; [the base PR's build has the same failure](https://github.com/across-protocol/contracts/actions/runs/34591005930/job/103236050809). Route cutover still requires API and relayer support for the replacement destination path.

---


programs/svm-spoke/src/error.rs
```diff
@@ -82,6 +82,8 @@ pub enum SvmError {
     InvalidDelegatePda,
     #[msg("Inconsistent optional parameters!")]
     InconsistentOptionalParameters,
+    #[msg("Legacy fill messages are unsupported; use a V5 Gateway destination path!")]
+    LegacyFillMessageUnsupported,
```
droplet-rl, line 86: The spec tells consumers to key off the **numeric code** because Anchor omits `SvmError` from the generated IDL error table. That makes 6019 a load-bearing part of the external contract, but two things weaken it:

1. **Nothing pins it.** There's already precedent for pinning in `v5.rs::v5_errors_use_dedicated_range`. Inserting a variant anywhere above this one silently shifts the code that integrators are matching on, with no test failure. Worth adding:

```rust
#[test]
fn legacy_rejection_code_is_pinned() {
    assert_eq!(u32::from(SvmError::LegacyFillMessageUnsupported), 6_019);
}
```

2. **The code space is shared.** `CommonError`, `SvmError`, and `CallDataError` all take Anchor's default 6000 offset, so codes already collide across enums (`SvmError::InvalidMint` and `CommonError::InvalidMerkleProof` are both 6007). I probed the current values: `CommonError::V5FillOnly` is 6016, so `CommonError` is exactly three variants away from minting its own 6019 and making the documented identifier ambiguous.

Giving `SvmError` an explicit offset would fix the root cause but is breaking for every existing `SvmError` consumer, so I'm not suggesting that here. Pinning the code plus a sentence in the spec noting that 6019 is only unique within `SvmError` — and that the log name is the authoritative discriminator — seems like the right trade.

programs/svm-spoke/V5_ADAPTER_SPEC.md
```diff
@@ -197,3 +197,32 @@ authenticate all JIT variants of an aggregate production route. The canonical re
...
+also undoes Anchor account initialization and preserves buffered parameters. V5-tagged messages still fail with
+`V5FillOnly` on this entrypoint. Slow-fill selectors have already been retired, including their callback path.
+Source deposits retain their message field for other destination chains.
+
+Anchor 0.31.1's existing multi-enum IDL error generation omits `SvmError`, including this new rejection. Consumers
+can identify it by its runtime log name or numeric code; generated error-name tables alone are insufficient.
+
+V5 keeps two distinct fields: the relay witness is exactly `V5_MAGIC_PREFIX || stepId`, while
+`V5FillInput.message` must be empty. The replacement destination flow is a single in-place Across fill followed by
+Gateway `APPROVE(inputMint, executor_authority, full balance)`, `CALL(swap)`, a committed output-mint `BALANCE_REQ`,
```
droplet-rl, line 220: This prescribes `APPROVE(inputMint, executor_authority, full balance)` plus (per the README) a `BalanceSub` that "patches the entire live input balance into the swap." I confirmed both are literally full-balance in the encoders: `reference.ts` `approve` emits `u64(0xffff_ffff_ffff_ffff)` with `u16(10000)`, and `swapFixture.ts` builds `inputMint || u16(10000) || u32(8)`.

That's sound in the fixture because `vault` holds exactly `amount` after a single JIT fill. But `vault` is the **shared** `pda(GATEWAY, "vault_authority")` ATA, and this section reads as the normative replacement flow for integrators. A tape that delegates the full balance and then substitutes 100% of the live balance into `amount_in` will sweep whatever else happens to be sitting in that vault at execution time into *this* tape's swap, and then `TRANSFER(all output)` hands the proceeds to *this* recipient.

The README already spells out the analogous hazard for the transfer path — "a fixed `min X` floor and full drain protect only `X`" and "two distinct fills observing one unchanged balance" — so the concept is established; the new APPROVE/BalanceSub pair just isn't covered by it yet. The fixture can't surface this either, since it uses an exclusive vault with exactly one fill, so the docs are the only guard.

Suggest extending the existing hazard paragraph to state that full-balance `APPROVE`/`BalanceSub` is sound only when the tape is the sole consumer of that vault balance for the execution window — the same "cover every recorded fill" invariant that already governs `TRANSFER`.

programs/svm-spoke/V5_ADAPTER_SPEC.md
```diff
@@ -197,3 +197,32 @@ authenticate all JIT variants of an aggregate production route. The canonical re
...
+The vault owner signs only Gateway token operations; the swap receives its distinct SPL delegate as signer.
+A failed swap or output floor reverts the entire execution, including the fill and payer rent.
+
+The [real-Gateway suite](../../test/svm-gateway/README.md) pins and exercises a concrete compatible swap program.
+It covers one fill and one committed swap, not arbitrary JIT routes or aggregate fills. Production builders must
+bind the complete destination outcome to every matching relay, validate venue instruction/version compatibility,
+and retain any auction winner and bid requirements outside a JIT swap window. Separate relayer/swapper executions
+can use prefunded chaining; they are distinct from this atomic fill-and-swap fixture.
+
+Roll out callback rejection together with replacement route-building and relayer execution support. A legacy
```
droplet-rl: "Roll out ... together with replacement route-building and relayer execution support" covers the *forward* path but not the in-flight window, and that gap is sharper than it looks because #1550 already removed slow fills.

Concretely, for an Across+ SVM deposit that is in flight when this deploys: `fill_relay` now reverts, so no relayer can fill it; there is no slow-fill fallback left; so it can only resolve by expiring and being refunded on the origin chain (consistent with `SLOW_FILL_RETIREMENT.md:18`). Funds are safe and the failure mode is the correct fail-closed one — but it's a multi-hour stuck deposit for the user, and it's silent from this repo's perspective.

`SLOW_FILL_RETIREMENT.md` handles exactly this class of problem well, with an explicit "Before deploying the combined upgrade, verify the active dataworker/configuration excludes ... and reconcile ..." checklist plus the honest caveat that "the compatibility tests do not establish that the live in-flight window is empty." This section would benefit from the parallel checklist: disable Across+ SVM routes in the API first, then let the in-flight window drain past `fillDeadline`, and only then deploy the rejection. Otherwise the ordering has to be rediscovered at deploy time.

test/svm/utils.ts
```diff
@@ -1,4 +1,4 @@
-import { BN, Program, workspace } from "@coral-xyz/anchor";
+import { BN, Program } from "@coral-xyz/anchor";
```
droplet-rl: `Program` is now unused after `testAcrossPlusMessage` was removed — the only remaining reference in this file is this import line. Same for `AccountMeta` on line 9.

Neither will be caught by CI: `lint-js` is `prettier --list-different` (no unused-import rule), and `tsconfig.json` `include` is only `./index.ts`, `utils/utils.ts`, `src/types`, so `test/` is never type-checked. Worth dropping both by hand.

droplet-rl (commented):
Reviewed the diff against `reinis/acp-221-remove-slow-fills` on a fresh checkout of `34e21ab`.

**The on-chain change is small, correct, and fail-closed.** `validate_legacy_fill_message` runs before `_fill`, so rejection precedes token delivery, the status transition, and event emission; the V5 check is ordered first so a witness still yields `V5FillOnly` rather than the new error. I verified there are no dangling references to `invoke_handler`, `AcrossPlusMessage`, or `AcrossPlusError` anywhere in the tree, and `V5FillInput.message` emptiness *is* enforced on-chain (`v5.rs:149`), so the spec's claim there is backed by code rather than convention.

**What I reproduced locally:**
- `cargo test -p svm-spoke --lib --features test` → 20 passing, matching the PR description.
- `cargo build -p svm-spoke --features test` → no new unused/dead-code warnings (`is_v5_message` is still reachable via `validate_legacy_fill_message`).
- Error numbering, by probing `u32::from(..)` directly: `LegacyFillMessageUnsupported` = **6019** ✅, `CommonError::V5FillOnly` = 6016, `SvmError::InconsistentOptionalParameters` = 6018. The `6019` in the spec is accurate.
- SVM spoke IDLs/clients are gitignored (`.gitignore:56-57`), so there is correctly no IDL diff to review here.

The swap fixture is carefully built — the 236-byte `AmmConfig` layout and the `trade_fee_rate` offset of 12 are both right, and `create_pool_fee = 0` is what lets the zero-amount WSOL fee account work. `beforeEach` mints a fresh input mint per test, so each of the three swap cases gets its own pool and the reserve assertions are properly isolated. The rollback assertions (`attemptedEvents` containing exactly one `filledRelay`, plus a full `deepEqual` over ten watched accounts) are the right shape for proving atomicity. I also checked that the new `sendTransaction` log-unwrapping wrapper doesn't break the existing injected-setup-failure test — the `if (receipt?.meta?.err)` guard correctly falls through when the tx actually succeeded.

**Three items I'd want addressed before this deploys** (all documentation/regression-coverage, not program bugs — hence COMMENT rather than REQUEST_CHANGES):

1. `6019` is advertised to consumers as the identifier but isn't pinned by any test, and it shares Anchor's default 6000 offset with `CommonError` — three variants of headroom. See inline on `error.rs`.
2. The spec promotes a full-balance `APPROVE` + 100%-bps `BalanceSub` on the **shared** Gateway vault to "the replacement destination flow" without restating the cross-fill exposure the README already documents for `TRANSFER`. The fixture structurally cannot catch this. See inline on the spec.
3. No in-flight drain precondition. With slow fills already gone in #1550, a callback-bearing SVM deposit that is in flight when this ships becomes permanently unfillable and can only resolve via origin-chain expiry refund. `SLOW_FILL_RETIREMENT.md` set a good precedent with its explicit pre-deploy checklist; this change needs the parallel one.

**One validation question:** the description reads "Ordinary SVM suite: 144 passing with one callback-fixture transaction-size failure. After reducing that fixture, the focused callback-rejection suite passed all four cases." That phrasing suggests the *full* ordinary suite wasn't re-run to green after the fixture reduction — only the focused subset. Could you confirm the whole suite is green on `34e21ab`? Related: because rejection happens on `!message.is_empty()` before any parsing, the `validCallback`/`malformed` axis exercises an identical branch, so the real coverage is 2 cases (inline vs buffered), not 4.

Remaining inline notes are minor: two now-unused imports, a markdown line that breaks the file's own wrap convention, a duplicated account fetch, a truthy-zero-signature guard, and a `getBlockTime` non-null assertion that can throw on a skipped slot.

Nice work on the migration docs — the "must never be reported as successfully filled with its requested action silently omitted" framing is exactly the right invariant, and keeping the MulticallHandler program and its exports intact while retiring only the Spoke's execution path is the right scoping call.

test/svm-gateway/README.md
```diff
@@ -2,7 +2,7 @@
 
 Run `yarn test-svm-gateway`. It builds Gateway and PrefundedAdapter from the immutable `GATEWAY_COMMIT` in
 `reference.ts`, builds this checkout's SpokePool with `--features test`, generates only the target test IDL, and starts
-an isolated validator with all three real programs loaded as upgradeable programs. It does not clone mainnet state.
+an isolated validator with Gateway, PrefundedAdapter, SpokePool and Raydium CPMM loaded as upgradeable programs. It does not clone mainnet state.
```
droplet-rl: This line is now 145 characters. Every other prose line in this file wraps at 120, and the repo's `.prettierrc` sets `printWidth: 120` — it just doesn't reflow markdown prose (default `proseWrap: "preserve"`), and `lint-js` only globs `**/*.js **/*.ts`, so nothing flags it. I checked the base: lines 37 and 75 were already over, but this one is newly introduced and is ordinary prose that wraps cleanly.

(The Raydium URL line further down is also long, but that's an unavoidable single token.)

test/svm-gateway/RealGateway.ts
```diff
@@ -404,6 +432,108 @@ describe("SVM V5 with the pinned real Gateway", () => {
...
+      } else {
+        const { signature } = await execute(dst, opts);
+        await filled(relay);
+        const delivered = (await getAccount(connection, fixture.recipientAta)).amount;
+        assert.isTrue(delivered >= minimum);
+        assert.equal((await getAccount(connection, fixture.poolInput)).amount, 1_000_000_000n + amount);
+        assert.equal((await getAccount(connection, fixture.poolOutput)).amount, 1_000_000_000n - delivered);
+        assert.equal((await getAccount(connection, vault)).amount, 0n);
+        assert.equal((await getAccount(connection, fixture.outputVault)).amount, 0n);
+        const inputAccount = await getAccount(connection, vault);
```
droplet-rl, line 520: `vault` is fetched twice — line 519 already reads it to assert `amount === 0n`, and this re-fetches the same account for the owner/delegate assertions. Hoisting one read covers all three:

```ts
const inputAccount = await getAccount(connection, vault);
assert.equal(inputAccount.amount, 0n);
assert.isTrue(inputAccount.owner.equals(vaultAuthority));
assert.isNull(inputAccount.delegate, "full swap consumes the exact delegate allowance");
```

Also reads more clearly as one coherent set of claims about a single observed state, which matters here since the point is that the sweep left the account fully drained *and* de-delegated.

test/svm-gateway/RealGateway.ts
```diff
@@ -86,8 +91,23 @@ describe("SVM V5 with the pinned real Gateway", () => {
...
   const ix = (programId: PublicKey, name: string, data: Buffer, keys: AccountMeta[]) =>
     new TransactionInstruction({ programId, keys, data: Buffer.concat([discriminator(name), data]) });
-  const send = (...instructions: TransactionInstruction[]) =>
-    provider.sendAndConfirm(new Transaction().add(...instructions));
+  async function sendTransaction(tx: Transaction | VersionedTransaction) {
+    try {
+      return await provider.sendAndConfirm(tx);
+    } catch (error) {
+      // Older Anchor clients can lose the logs when wrapping a failed receipt.
+      const signature = tx instanceof VersionedTransaction ? tx.signatures[0] : tx.signature;
```
droplet-rl, line 99: For an unsigned `VersionedTransaction`, `signatures[0]` is a 64-byte zero-filled `Uint8Array`, which is truthy — so the `if (signature)` guard below passes and you bs58-encode all zeros into a lookup that can never hit. Harmless (`receipt` comes back null and the original error is rethrown) but it costs an RPC round-trip on every versioned failure.

A zero check makes the guard mean what it reads as:

```ts
const signature = tx instanceof VersionedTransaction ? tx.signatures[0] : tx.signature;
if (signature?.some((byte) => byte !== 0)) {
```

The legacy branch is already fine — `tx.signature` is `null` until signed.

test/svm-gateway/swapFixture.ts
```diff
@@ -0,0 +1,183 @@
...
+      new TransactionInstruction({
+        programId: SWAP_PROGRAM,
+        keys,
+        data: Buffer.concat([discriminator("initialize"), u64(liquidity), u64(liquidity), u64(0)]),
+      })
+    )
+  );
+  // initialize sets open_time to clock + 1. Wait for on-chain time, not wall time.
+  const initializedSlot = await connection.getSlot();
+  const initializedTime = (await connection.getBlockTime(initializedSlot))!;
```
droplet-rl: The non-null assertion here is unguarded, while the equivalent call inside the loop below defends with `?? 0`. `getBlockTime` errors on a skipped slot rather than returning null, so if the slot picked up by `getSlot()` was skipped this throws out of fixture setup and takes all three swap cases with it.

Since `getSlot()` is called after `sendAndConfirm` returns, the slot isn't guaranteed to be the initialize tx's slot anyway. Retrying a couple of slots back, or reusing the confirmed signature's slot from `sendAndConfirm`, would remove the flake. Low probability on a fresh localnet, but the failure mode is a confusing setup-time crash rather than a test failure.

test/svm/SvmSpoke.Fill.AcrossPlus.ts
```diff
@@ -167,377 +142,64 @@ describe("svm_spoke.fill.across_plus", () => {
...
-        recipientAccounts.push(recipientATA);
-
-        // Construct ix to transfer tokens from handler to the recipient.
-        const transferInstruction = createTransferCheckedInstruction(
+  for (const buffered of [false, true]) {
+    for (const validCallback of [false, true]) {
+      it(`rejects ${validCallback ? "encoded callback" : "malformed message"} with ${buffered ? "buffered" : "inline"} params atomically`, async () => {
+        // A valid callback with a destination SOL transfer fits inline without a lookup table.
+        const coder = new MulticallHandlerCoder([], finalRecipient);
+        const message = validCallback
```
droplet-rl, line 150: Worth being precise about what this matrix covers. `validate_legacy_fill_message` rejects on `!message.is_empty()` before anything inspects the bytes, so a well-formed `AcrossPlusMessageCoder` payload and `Buffer.from([1])` traverse an identical branch and produce an identical error — the message contents are never parsed. The real axis is `buffered` (inline vs. buffer-account params), which does exercise genuinely different Anchor account handling and is a good case to have.

No change needed to the code; the assertions are solid (no handler invocation, no `filledRelay` event, full state rollback). But the PR description and README both frame this as four distinct callback cases, which oversells it — worth wording as "inline and buffered params, with both malformed and well-formed payloads rejected identically" so a future reader doesn't assume there's payload-shape coverage here that could regress.

One genuinely useful addition, if you want a fourth distinct case: an integration assertion that a V5-prefixed message on this entrypoint yields `V5FillOnly` and not `LegacyFillMessageUnsupported`. `SvmSpoke.Fill.ts:314` covers it, so this is only a note that the ordering guarantee lives there rather than here.

scripts/svm/fakeFillWithRandomDistribution.ts
```diff
@@ -1,241 +1,6 @@
...
-  );
-
-  console.log("Transaction signature:", txSignature);
-}
-
-// Run the fillRelayToRandom function
-fillRelayToRandom();
+// Retained as an explicit migration notice for existing devnet callers.
+// Destination actions now belong in the V5 Gateway tape; see test/svm-gateway/README.md.
+throw new Error(
```
droplet-rl: A module-level `throw` surfaces to the user as an unhandled-rejection stack trace, which buries the migration message it exists to deliver. `console.error(...)` + `process.exit(1)` reads better for a notice whose entire job is to be read.

Alternatively, just delete the file and drop the `Anchor.toml:57` entry — a file whose whole body is a throw is an odd artifact to carry, and `git log` preserves the history for anyone looking for the old script. Either is fine; the current form is the least good of the three.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=5864d2bed780df53c220ee9ba1e21968&trace_start_ts=1788975574&trace_end_ts=1789062274)

programs/svm-spoke/src/error.rs
Reinis-FRP: Addressed in [88181f55](https://github.com/across-protocol/contracts/commit/88181f55172ca5e31452f488666c19733cc2a94d). Assigned distinct runtime ranges: Common 6000, SVM 7000, CallData 8000, V5 9000. Compatibility tests pin each range boundary, including `LegacyFillMessageUnsupported == 7019`. Common codes are preserved; SVM/CCTP codes are renumbered and V5 is new in this release. The spec separately documents the existing incomplete generated IDL error table.

---

programs/svm-spoke/V5_ADAPTER_SPEC.md
Reinis-FRP: Addressed in [88181f55](https://github.com/across-protocol/contracts/commit/88181f55172ca5e31452f488666c19733cc2a94d). Clarified that full-balance APPROVE, BalanceSub, and terminal transfer follow EVM V5 semantics and the same delivery policy. Every recorded fill must have proportional or authenticated aggregate delivery; the fixture covers the single-fill template.

---

programs/svm-spoke/V5_ADAPTER_SPEC.md
Reinis-FRP: Addressed in [88181f55](https://github.com/across-protocol/contracts/commit/88181f55172ca5e31452f488666c19733cc2a94d). Added deployment sequencing: disable old routes, reconcile finalized deposits/fills and deadlines across origins, account for direct deposits and indexing/finality lag, and drain or explicitly migrate remaining obligations before upgrading. The docs state there is no slow-fill fallback and that local tests do not establish an empty live window.

---

test/svm/utils.ts
Reinis-FRP: Addressed in [88181f55](https://github.com/across-protocol/contracts/commit/88181f55172ca5e31452f488666c19733cc2a94d). Removed both unused imports (`Program` and `AccountMeta`). The focused TypeScript check includes the changed test files.

---

test/svm-gateway/README.md
Reinis-FRP: Addressed in [88181f55](https://github.com/across-protocol/contracts/commit/88181f55172ca5e31452f488666c19733cc2a94d). Wrapped the new introductory prose line.

---

test/svm-gateway/RealGateway.ts
Reinis-FRP: Addressed in [88181f55](https://github.com/across-protocol/contracts/commit/88181f55172ca5e31452f488666c19733cc2a94d). The amount, owner, and delegate assertions now use one fetched input account.

---

test/svm-gateway/RealGateway.ts
Reinis-FRP: Addressed in [88181f55](https://github.com/across-protocol/contracts/commit/88181f55172ca5e31452f488666c19733cc2a94d). The receipt lookup now requires a nonzero signature byte, so an unsigned versioned transaction preserves the original error without querying an all-zero signature.

---

test/svm-gateway/swapFixture.ts
Reinis-FRP: Addressed in [88181f55](https://github.com/across-protocol/contracts/commit/88181f55172ca5e31452f488666c19733cc2a94d). The fixture now reads `unix_timestamp` from the Clock sysvar, matching the program's time source and avoiding block-time lookups on skipped slots. The bounded wait remains in place.

---

test/svm/SvmSpoke.Fill.AcrossPlus.ts
Reinis-FRP: Addressed in [88181f55](https://github.com/across-protocol/contracts/commit/88181f55172ca5e31452f488666c19733cc2a94d). Clarified the README and PR description: inline and buffered parameters are the two loading paths; malformed and well-formed payloads are rejected identically before parsing. The existing V5FillOnly ordering test remains in SvmSpoke.Fill.ts.

---

scripts/svm/fakeFillWithRandomDistribution.ts
Reinis-FRP: Addressed in [88181f55](https://github.com/across-protocol/contracts/commit/88181f55172ca5e31452f488666c19733cc2a94d). The script now prints the migration notice and exits with status 1. Verified the exit code and that stderr contains no stack trace.

---

programs/svm-spoke/src/error.rs
```diff
@@ -42,7 +42,7 @@ pub enum CommonError {
 }
 
 // SVM specific errors.
-#[error_code]
+#[error_code(offset = 7000)]
```
droplet-rl, line 45: This is the renumbering I flagged as breaking-and-out-of-scope last round. To be clear about where I land: the change is **technically correct** and fixes a genuine latent bug — the cross-enum collision meant 6007 was simultaneously `SvmError::InvalidMint`, `CommonError::InvalidMerkleProof`, and `CallDataError::InvalidUint32`, so numeric disambiguation was impossible. Explicit offsets on all four enums (including `CommonError`, which you didn't have to touch) is the right end state.

What I want on the record is that this mutates observable ABI on a mainnet-deployed program (`DLv3NggMiSaef97YCkew5xKUHDh13tVGZ7tydt3ZeAru`). I checked the blast radius and it's bounded — `RelayFilled`/`NotExclusiveRelayer`/`ExpiredFillDeadline`/`FillsArePaused` all stay put, and the moved codes (`InvalidRelayHash` 6001→7001, `InvalidMint` 6007→7007, `InvalidDelegatePda` 6017→7017) are misconfiguration conditions rather than relayer retry signals. Every in-repo test matches on the log-derived name, so CI can't catch a regression here either way.

The residual risk is that this degrades *silently*: a consumer holding `6001 → InvalidRelayHash` now reads a genuine `InvalidFillDeadline` as `InvalidRelayHash`, and stops recognizing real `InvalidRelayHash` entirely. No exception, no log line, just a wrong label on a diagnostic path.

Two options, either fine by me:
- **Keep it here**, and add the old→new mapping table plus a coordination step to deployment sequencing (see my other two comments).
- **Split it out** into its own PR so it's reviewed on its own merits with the off-chain repos looped in, and land the callback retirement first.

What I'd rather avoid is it shipping as an unremarked line item under a title about callback retirement.

programs/svm-spoke/V5_ADAPTER_SPEC.md
```diff
@@ -197,3 +197,55 @@ authenticate all JIT variants of an aggregate production route. The canonical re
...
+`LegacyFillMessageUnsupported` (7019) before token delivery, the filled-status transition, or event emission; rollback
+also undoes Anchor account initialization and preserves buffered parameters. V5-tagged messages still fail with
+`V5FillOnly` on this entrypoint. Slow-fill selectors have already been retired, including their callback path.
+Source deposits retain their message field for other destination chains.
+
+Runtime error ranges are distinct: `CommonError` starts at 6000, `SvmError` at 7000, `CallDataError` at 8000, and
+`V5Error` at 9000. This preserves Common error numbers and renumbers SVM/CCTP errors from their overlapping legacy
+range; V5 errors are new in this release. Compatibility tests pin each range's first and last codes.
+Anchor 0.31.1's existing multi-enum IDL error generation remains incomplete and omits `SvmError`, including this
+new rejection. Consumers should use runtime log names or an explicit runtime-code mapping; generated error-name
```
droplet-rl: "Consumers should use runtime log names or **an explicit runtime-code mapping**" points at an artifact that doesn't exist anywhere in the repo. For a renumbering, that table is the single most useful thing you can hand an integrator, and it's cheap to write since the values are already pinned in `custom_error_ranges_are_stable`.

Suggest adding it right here, at minimum for the codes that moved:

| name | enum | old | new |
|---|---|---|---|
| `InvalidRelayHash` | `SvmError` | 6001 | 7001 |
| `InvalidMint` | `SvmError` | 6007 | 7007 |
| `InvalidDelegatePda` | `SvmError` | 6017 | 7017 |
| `LegacyFillMessageUnsupported` | `SvmError` | — | 7019 (new) |
| `InvalidSelector` | `CallDataError` | 6000 | 8000 |
| `InvalidWireFormat` | `V5Error` | 7000 | 9000 |

It's also worth stating explicitly that **`CommonError` codes are unchanged**, since that's the fact that makes the migration low-risk for relayers and it's currently only implied by "preserves Common error numbers." A reader scanning for "did my error code move" will want that answered in one line.

The surrounding paragraph is otherwise accurate — I confirmed that distinct ranges genuinely don't fix the generated IDL table, and that the runtime log name remains the only complete discriminator.

programs/svm-spoke/V5_ADAPTER_SPEC.md
```diff
@@ -197,3 +197,55 @@ authenticate all JIT variants of an aggregate production route. The canonical re
...
+
+Before deploying callback rejection:
+
+1. Disable routes that create callback-bearing SVM deposits in API/builders and coordinate relayer cutover to the
+   replacement V5 path. A route flag alone does not prevent direct deposits through permissionless entrypoints.
+2. Reconcile finalized deposits and successful fills across supported origin chains. Allow outstanding callback
+   deposits to fill before their deadlines, or wait past the remaining `fillDeadline` values and verify origin-chain
+   expiry-refund handling. Include finality/indexing lag and confirm no new callback deposits entered the window.
+3. Verify no unexpired callback-bearing obligations remain before deploying, or handle them through a separately
+   reviewed migration procedure. Verify replacement route-building and relayer execution support before enablement.
```
droplet-rl, line 246: This checklist is thorough for the callback cutover but silent on the error renumbering, which has its own independent coordination requirement — and unlike the callback change, it affects consumers on paths that have nothing to do with callbacks or V5.

Suggest a fourth step along the lines of:

> 4. Update off-chain consumers that map `svm_spoke` custom error numbers before upgrading. `SvmError` moves from the 6000 range to 7000 and `CallDataError` to 8000; `CommonError` numbers are unchanged. Consumers keying on runtime log names need no change. Because a stale numeric map silently mis-labels a preserved `CommonError` code rather than failing, audit for hardcoded 6xxx values rather than waiting for an error.

The "audit rather than wait for an error" framing is the part that matters — steps 1–3 all describe conditions you can *observe* (deposits drained, deadlines passed), whereas this one has no observable signal and has to be found by inspection. Worth calling out so it doesn't get deferred as a post-deploy cleanup.

droplet-rl (commented):
Re-reviewed at `88181f55`. Both branches were force-updated, so I reset and re-diffed against the new base `f7f705af` — the `constants.rs` / `v5_adapter.rs` / `v5.rs` deltas between the old and new first commit are base-branch (#1550) churn, not this PR.

**All ten items from the previous review are addressed, several better than I suggested.** The Clock-sysvar read is the right fix rather than the retry-on-skipped-slot workaround I proposed — I verified the layout (`unix_timestamp` is `i64` at offset 32 of a 40-byte account, and `initializedTime` is sampled *after* `sendAndConfirm`, so the wait is strictly conservative against `open_time`). The zero-signature guard, the hoisted `vault` fetch, the two unused imports, the wrapped README line, the `console.error`/`exit(1)`, and the corrected coverage wording for the callback matrix are all done. The new deployment-sequencing section is genuinely good — "a route flag alone does not prevent direct deposits through permissionless entrypoints" is a sharper observation than what I raised, and step 2's finality/indexing-lag caveat is the detail that usually gets missed.

**Verified locally at this commit:**
- `cargo test -p svm-spoke --lib --features test` → **21 passing**; without the feature → **17 passing**. Both match the description exactly.
- `custom_error_ranges_are_stable` pins first *and last* code per enum. That's stronger than the single assertion I asked for: because the last variant's code shifts if anything is inserted anywhere above it, a mid-enum insertion now fails the test. Appending still passes, which is the correct intent.
- No stale references to the old numbering anywhere in `programs/`, `test/`, or `src/`.
- All four enums now carry explicit offsets, including `CommonError` — better than leaving one on the implicit default.

---

**One thing went further than I asked, and I want to flag it rather than wave it through.**

I wrote last time that giving `SvmError` an explicit offset "would fix the root cause but is breaking for every existing `SvmError` consumer, so I'm not suggesting that here." The renumbering landed anyway: `SvmError` 6000→7000, `CallDataError` 6000→8000, `V5Error` 7000→9000. `SvmSpoke` is live on Solana mainnet (`DLv3NggMiSaef97YCkew5xKUHDh13tVGZ7tydt3ZeAru`, per `broadcast/deployed-addresses.md:553`), so these codes are observable production ABI.

I probed the actual blast radius before deciding how hard to push, and it's narrower than my first instinct:

| code | before | after |
|---|---|---|
| `RelayFilled` | 6004 | 6004 ✅ |
| `NotExclusiveRelayer` | 6002 | 6002 ✅ |
| `ExpiredFillDeadline` | 6006 | 6006 ✅ |
| `FillsArePaused` | 6012 | 6012 ✅ |
| `InvalidRelayHash` | 6001 | **7001** |
| `InvalidMint` | 6007 | **7007** |
| `InvalidDelegatePda` | 6017 | **7017** |

Every code a relayer actually branches on for retry decisions lives in `CommonError` and is preserved. The renumbered ones are misconfiguration/programming conditions. Combined with the fact that numeric matching was *already* ambiguous pre-change (6007 was simultaneously `InvalidMint`, `InvalidMerkleProof`, and `InvalidUint32`), no consumer could have been reliably depending on the old numbers. I also confirmed every in-repo TS test matches on `error.error.errorCode.code` — the name Anchor parses out of the runtime log, not the number — so nothing here breaks.

So I'm not blocking on it. But the failure mode is silent mis-attribution rather than a loud break (a consumer holding `6001 → InvalidRelayHash` will now quietly read `InvalidFillDeadline` as `InvalidRelayHash` and never recognize a real one), and two cheap additions would close it — see the inline comments. Worth an explicit team decision on whether this belongs in a callback-retirement PR at all, or as its own change with its own coordination; right now it gets reviewed as a footnote to an unrelated headline.

Everything else is in good shape. Happy to approve once the mapping table and the renumbering coordination step are in, or if you'd rather split the renumbering out.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=459f82de7c0a7b567d2fda89673fbacc&trace_start_ts=1789141765&trace_end_ts=1789228465)

programs/svm-spoke/src/error.rs
Reinis-FRP: Addressed in [354cdca0](https://github.com/across-protocol/contracts/commit/354cdca073764c26d06df9b711090db8299eaaa8). Keeping the renumbering in this PR as explicit migration scope; the title and description now call it out. Added the complete old/new mapping and pre-upgrade consumer coordination. Existing Common codes remain unchanged; V5 errors are new relative to master, with earlier undeployed assignments documented separately.

---

programs/svm-spoke/V5_ADAPTER_SPEC.md
Reinis-FRP: Addressed in [354cdca0](https://github.com/across-protocol/contracts/commit/354cdca073764c26d06df9b711090db8299eaaa8). Added [ERROR_CODES.md](https://github.com/across-protocol/contracts/blob/354cdca073764c26d06df9b711090db8299eaaa8/programs/svm-spoke/ERROR_CODES.md), covering all 62 current variants against a pinned master baseline. It explicitly states that existing Common codes are unchanged, marks new errors, records the earlier pre-release V5 range, and distinguishes historical decoding. All entries were checked against the Rust source.

---

programs/svm-spoke/V5_ADAPTER_SPEC.md
Reinis-FRP: Addressed in [354cdca0](https://github.com/across-protocol/contracts/commit/354cdca073764c26d06df9b711090db8299eaaa8). Added deployment step 4: inspect hardcoded numeric-code consumers and update affected maps before upgrading, including non-callback paths. It explicitly covers silent mislabeling, unchanged Common codes, log-name consumers, final V5 numbering, and historical decoding. The requirement is inspection and coordination before deployment, rather than waiting for an observable failure.

---

programs/svm-spoke/ERROR_CODES.md
```diff
@@ -0,0 +1,101 @@
...
+`SvmError` moves from 6000–6018 to 7000–7018, and `CallDataError` moves from 6000–6006 to 8000–8006.
+The callback rejection and V5 errors are new relative to
+[`master` at `75d968e4`](https://github.com/across-protocol/contracts/blob/75d968e4e86c37ef12c01277345ea8ed2f550901/programs/svm-spoke/src/error.rs),
+the comparison used in the tables below. A dash means the error did not exist in that baseline.
+
+Earlier undeployed revisions of this PR used `V5Error` 7000–7017; those codes are now 9000–9017, with the same
+variant order. They also used 6019 for the new `LegacyFillMessageUnsupported`, now 7019. These earlier assignments
+were pre-release values.
+
+The tables cover every current variant in [error.rs](src/error.rs). Use the mapping for the program version being
```
droplet-rl: "The tables cover every current variant in [error.rs](src/error.rs)" is literally true — I verified all 62 — but scoping to *current* variants leaves out the one family a numeric-map audit most needs: the errors this PR **deletes**.

`AcrossPlusError` exists on the `75d968e4` baseline at 6000–6007 with 8 variants, and this PR removes the enum entirely along with `invoke_handler`:

| Error | Baseline | This release |
|---|---|---|
| `MessageDidNotDeserialize` | 6000 | removed |
| `InvalidMessageKeyLength` | 6001 | removed |
| `InvalidReadOnlyKeyLength` | 6002 | removed |
| `InvalidMessageHandler` | 6003 | removed |
| `InvalidMessageAccountKey` | 6004 | removed |
| `NotReadOnlyMessageAccountKey` | 6005 | removed |
| `NotWritableMessageAccountKey` | 6006 | removed |
| `MissingValueRecipientKey` | 6007 | removed |

Why this specific gap bites rather than being a cosmetic omission: those numbers don't go quiet after the upgrade. 6000–6007 is now occupied by `CommonError` `InvalidQuoteTimestamp`…`InvalidMerkleProof`. So a consumer holding `6004 → InvalidMessageAccountKey` will, post-upgrade, silently relabel a genuine **`RelayFilled`** — arguably the most operationally consequential code a relayer reads, since it's the "stop retrying" signal. That is precisely the silent-mislabel failure mode step 4 of the deployment sequence warns about, and it's the one case the table currently can't help them diagnose.

It also interacts with the historical-decoding guidance two lines below. These codes *were* emitted on mainnet while `invoke_handler` was live, so real historical transactions carry them — but a reader following "do not relabel them using the new table" has no way to look up what 6003 meant back then, because the names appear nowhere in this document.

Suggest a `## Removed (AcrossPlusError)` section with the rows above, plus a line noting that these codes are never emitted again and that the same numbers are now live `CommonError` values. Worth landing before the step-4 coordination goes out to consumers.

droplet-rl (approved):
Re-reviewed at `354cdca0`. Code is byte-identical to `88181f55` (`git diff --stat 88181f55 HEAD` touches only the two markdown files), which I already verified at 21 tests with `--features test` / 17 without. This round is purely the docs response, so I focused on whether the new migration table is actually *correct* — a wrong mapping table is worse than no table.

**I verified all 62 rows programmatically rather than spot-checking.** Parsed the enum variants and offsets out of both `master` at the pinned `75d968e4` and current `HEAD`, computed every `u32::from(...)`, parsed the four markdown tables, and cross-checked in both directions:

- Every **"This release"** value matches current `error.rs`. ✅
- Every **"Before release"** value matches master at `75d968e4`. ✅
- Row count is exactly 62, matching the description, with **no missing and no extra rows** versus the code. ✅
- Every `—` is correct: master `CommonError` has 16 variants (no `V5FillOnly`), master `SvmError` has 19 (no `LegacyFillMessageUnsupported`), and `V5Error` genuinely does not exist on master. ✅

I also confirmed the baseline choice is sound: `75d968e4` is the *current* head of `origin/master` (zero commits between them), so this is a live baseline rather than an arbitrary historical pin. And the honesty about the pre-release churn in lines 9–11 — that earlier undeployed revisions of this PR used `V5Error` 7000–7017 and 6019 for the callback rejection — is the right call; that's exactly the kind of detail that confuses someone bisecting a branch later.

**Everything from the last round is addressed:**
- The mapping table exists and is accurate (verified above).
- Deployment sequencing step 4 landed, and it kept the framing that mattered most: *"Audit numeric maps by inspection: a stale 6xxx mapping can silently mislabel a preserved Common error, so waiting for an observable failure is insufficient."* That's the distinction between this step and steps 1–3, and it's stated explicitly.
- "Existing `CommonError` codes are unchanged" is now a one-line answer at the top of both documents.
- The scoping concern is resolved by retitling the PR to name the error-range work. Folding it in with the coordination docs and the boundary-pinning test is a reasonable resolution — my objection was to it shipping unremarked, not to it shipping.

Also checked, all clean: all four tables are prettier-canonical (uniform row widths per section), every relative link resolves (`src/error.rs`, `ERROR_CODES.md`, and the `#deployment-sequencing` anchor against the actual `### Deployment sequencing` heading at line 236), and the only >120-char line is a single unbreakable GitHub permalink.

**Approving.** One follow-up in the inline comment: `AcrossPlusError` is deleted by this PR and doesn't appear in the migration table, which leaves a gap for exactly the consumer the document targets. It's an 8-row addition and doesn't need to gate the merge, but I'd want it in before the deploy coordination in step 4 goes out, since that step tells people to audit their 6xxx maps and this is the one family of 6xxx codes the table can't help them with.

Nice work across all three rounds — the deployment-sequencing section and this migration table are both stronger than what I asked for.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=7ac1f2a8d34636ca6b6da985ff9bb48f&trace_start_ts=1789143294&trace_end_ts=1789229994)

programs/svm-spoke/ERROR_CODES.md
Reinis-FRP: Addressed in [0ef4e852](https://github.com/across-protocol/contracts/commit/0ef4e852647f2332ef13d0999f0ce890775d282f). Added all eight removed `AcrossPlusError` variants with their baseline codes and historical-decoding guidance. The docs explicitly state that those numbers remain in use by `CommonError` and that stale callback labels must be removed from active maps. Verified all 70 rows against the baseline and current Rust source.

---

PR #1552, opened 2026-09-09
chore: remove PermissionSplitterProxy

tbwebb22:


contracts/hub-pool/PermissionSplitterProxy.sol
```diff
@@ -1,128 +0,0 @@
...
-import "contracts/external/uma/core/contracts/common/implementation/MultiCaller.sol";
-import "@openzeppelin/contracts-v4/access/AccessControl.sol";
-
-/**
- * @notice This contract is designed to own an Ownable "target" contract and gate access to specific
- * function selectors based on specific roles. Practically, this contract should own the HubPool contract.
- * All ownable function calls to target should be sent through this contract's fallback function.
- * @custom:security-contact bugs@across.to
- */
-contract PermissionSplitterProxy is AccessControl, MultiCaller {
```
droplet-rl, line 13: Confirmed safe to delete. The mainnet instance at `0x0Bf07B2e415F02711fFBB32491f8ec9e5489B2e7` is inert:

- It does not own the `HubPool` — `owner()` is the DAO Safe `0xB5247…`, and an `eth_call` of `setPaused(bool)` with `from` set to this proxy reverts `Ownable: caller is not the owner`.
- `eth_getLogs` from genesis returns only 5 events, all in the deploy block 19084679 (2024-01-25). It has never been used since deployment day.
- `DEFAULT_ADMIN_ROLE` was granted to the deployer and then renounced, with no grant to a replacement admin, so `__setTarget` and `__setRoleForSelector` can never be called again.

No governance path depends on this.

script/hub-pool/DeployPermissionSplitterProxy.s.sol
```diff
@@ -1,46 +0,0 @@
...
-        // Grant anyone with the pause role the ability to call setPaused.
-        permissionSplitter.__setRoleForSelector(PAUSE_SELECTOR, PAUSE_ROLE);
-        // Revoke the deployer's default admin role.
-        permissionSplitter.renounceRole(DEFAULT_ADMIN_ROLE, deployer);
-
-        // Sanity check.
-        assertTrue(permissionSplitter.hasRole(PAUSE_ROLE, defaultAdmin));
-        assertFalse(permissionSplitter.hasRole(PAUSE_ROLE, deployer));
-        assertFalse(permissionSplitter.hasRole(DEFAULT_ADMIN_ROLE, deployer));
-        assertTrue(permissionSplitter.hasRole(DEFAULT_ADMIN_ROLE, defaultAdmin));
```
droplet-rl, line 44: Incidental finding while verifying the deployment, for the record: this final assertion never actually held on-chain. The deploy tx emitted `RoleGranted(PAUSE_ROLE, 0xB5247…)` and `RoleRevoked(DEFAULT_ADMIN_ROLE, deployer)` but **no** `RoleGranted(DEFAULT_ADMIN_ROLE, defaultAdmin)` — `grantRole(PAUSE_ROLE, …)` was called, but nothing ever gave `defaultAdmin` the admin role. Post-broadcast `assert*` failures in a `forge script` don't revert the broadcast, so it went unnoticed.

Nothing to fix here since the file is being deleted — it just corroborates that the deployment was abandoned rather than adopted.

broadcast/deployed-addresses.md
```diff
@@ -43,7 +43,6 @@ This file contains the latest deployed smart contract addresses from the broadca
 | Optimism_Adapter                       | [0x3562e309C6C79626E5F0Cf746FB5Bf4f6b8EebE5](https://etherscan.io/address/0x3562e309C6C79626E5F0Cf746FB5Bf4f6b8EebE5) |
 | PermissionedMulticallHandler           | [0x64a43393866DBA0044879979fAa7AD3d000622e9](https://etherscan.io/address/0x64a43393866DBA0044879979fAa7AD3d000622e9) |
 | PermissionedMulticallHandler_CCTP_USDC | [0xDd52f8134f85f3979fbA24387Ce0CEC05937259E](https://etherscan.io/address/0xDd52f8134f85f3979fbA24387Ce0CEC05937259E) |
-| PermissionSplitterProxy                | [0x0Bf07B2e415F02711fFBB32491f8ec9e5489B2e7](https://etherscan.io/address/0x0Bf07B2e415F02711fFBB32491f8ec9e5489B2e7) |
```
droplet-rl, line 46: Worth a conscious decision rather than a change request: `broadcast/deployed-addresses.{json,md}` are generated from `broadcast/`, so deleting the run artifacts is the only way to drop this row — but the contract still exists on mainnet and this was its only in-repo record.

Given it's bricked and owns nothing, removing it arguably *helps* (nobody will mistake it for live infrastructure), and git history preserves the address. `deployments/legacy-addresses.json` isn't the right home either — that's scoped to legacy Hardhat deployments. Fine to leave as-is; just flagging that the provenance is intentionally being dropped.

test/evm/foundry/local/MulticallHandler.t.sol
```diff
@@ -5,9 +5,6 @@ import { Test } from "forge-std/Test.sol";
 import { MulticallHandler } from "../../../../contracts/handlers/MulticallHandler.sol";
 import { IERC20 } from "@openzeppelin/contracts-v4/token/ERC20/IERC20.sol";
 
-// Run this test to verify PermissionSplitter behavior when changing ownership of the HubPool
```
droplet-rl, line 8: Good catch — this comment was copy-pasted from the PermissionSplitter fork test and was already wrong for this file. `MulticallHandlerTest` lives under `test/evm/foundry/local` and runs under the `local-test` profile in CI with no fork URL, so the "run this as a fork test" instruction was actively misleading. Removing it is correct.

droplet-rl (approved):
## Approve — the removal is complete, and I confirmed on-chain that the deployed proxy is inert

The main risk in a PR like this is deleting the source for something still load-bearing on mainnet. It isn't. Details below.

### On-chain state of `0x0Bf07B2e415F02711fFBB32491f8ec9e5489B2e7`

- `HubPool.owner()` → `0xB524735356985D2f267FA010D681f061DfF03715` (the DAO Safe, per `script/mintburn/prod-readiness-multisigs.json`), **not** the proxy. Ownership was never moved to it.
- `eth_call HubPool.setPaused(true)` with `from` = the proxy reverts `Ownable: caller is not the owner`, so the `PAUSE_ROLE` wiring is non-functional.
- `eth_getLogs` from genesis → **5 events total, all in the deploy block 19084679 (2024-01-25)**. Never touched since.
- Those events show `DEFAULT_ADMIN_ROLE` granted to the deployer and then renounced, with **no grant to the intended admin**. `__setTarget` / `__setRoleForSelector` are therefore permanently uncallable — the contract is bricked as well as unused.

So this is genuinely dead code. Removing it is the right call.

### Repo-side verification

| Check | Result |
|---|---|
| Dangling references | None. Only remaining hits are inside historical `deployments/**/solcInputs/*.json` verification blobs — immutable artifacts, correctly left alone. |
| Orphaned deps | `MultiCaller` (9 other consumers) and `HubPoolTestBase` (18 other consumers) both still used. Nothing left stranded. |
| CI `lint-and-check-generated` | Ran `yarn extract-addresses` + prettier on this branch → **zero diff** against the committed `deployed-addresses.{json,md}`. The generated-file gate will pass. |
| `FOUNDRY_PROFILE=local-test forge test` | 646 passed / 5 failed. All 5 failures are fork-RPC/environment (`vm.createSelectFork` against `rpc.hyperliquid.xyz`, ERC-6492 permit2 tests) and unrelated to this change. |
| Fork tests | Not run in CI (`local-test` profile is `test/evm/foundry/local` only), so deleting `fork/PermissionSplitter.t.sol` changes no gate. |

One note on my environment so it isn't mistaken for a finding: `forge build` (default profile) hit a solc ICE — `Assembly exception for bytecode: Tag too large for reserved space`. I reproduced it **identically on `master`** in a clean worktree, so it is pre-existing and caused by my local forge 1.8.1 vs the repo's pinned 1.7.1 (`.foundry-version`). Not a PR concern.

### Non-blocking

1. **Empty PR description.** For a chore that deletes a mainnet-deployed contract and its deployment provenance, a one-liner ("unused, never owned the HubPool, superseded by direct Safe ownership") would save the next reader the archaeology I just did.
2. **Published-package semantics.** `files` includes `/contracts/**/*.sol` and `exportEvmArtifacts.sh` ships the ABI, so this is technically a breaking removal for anyone importing `@across-protocol/contracts/contracts/hub-pool/PermissionSplitterProxy.sol` or that ABI. Org-wide code search found zero consumers outside this repo and its private mirror (`UMAprotocol/across-contracts-private`), so the practical risk is nil — just worth a thought at the next version bump.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=9d7531e0f1d6cb3fdd16ed63400e98d9&trace_start_ts=1788994608&trace_end_ts=1789081308)

grasphoper (approved):

PR #1466, opened 2026-06-12
chore: deploy to Arc

tbwebb22:

PR #1554, opened 2026-09-11
chore: deploy MulticallHandler and AcrossEventEmitter to Arc

droplet-rl:
Deploys the two remaining zero-dependency periphery contracts to Arc (5042) and records them in the generated address files.

| Contract | Address | Block |
| --- | --- | --- |
| `MulticallHandler` | `0xA07480456C4EbaD7626e4FdF4a180709E238547b` | 20298820 |
| `AcrossEventEmitter` | `0x446Fe2Be6a8D90fA89F30f3d0A686aAFfe621237` | 20298840 |

Both via CREATE2 from the canonical deployer, using the existing scripts unchanged (salts `0x12345678` and `0x1234`). Deployed bytecode at both addresses matches the locally compiled artifacts byte for byte.

Prerequisites confirmed live on Arc before deploying: CREATE2 factory, Permit2 and Multicall3 all at their canonical addresses.

### Not included

- **`SpokePoolVerifier`** — it is a passthrough for native-token deposits only ("this problem only exists for native tokens"). Arc s native asset is USDC while the bridgeable USDC is a 6-decimal ERC20, so the contract has no coherent use there. There is also no deploy script for it in the repo; it is only instantiated in tests.
- **`SpokePoolPeriphery`** — its script skips any chain without a `permit2` entry in `generated/constants.json`, and 5042 has none. Permit2 itself is already live on Arc at the canonical address, so this is a constants-package addition, not a deployment blocker. Left for a follow-up.

### Notes

- Both contracts are **unverified** on the explorer: `foundry.toml` has no `[etherscan]` entry for Arc.
- `MulticallHandler` does not land on the address used by mainnet/Arbitrum/Base, since that one was built from different initcode. Consumers resolve it per chain via `getDeployedAddress`, which the regenerated files now cover.
- These are additive and stateless: nothing routes through them until config references them.


droplet-rl:
Codex's review finished at 11:09:15Z and came back **clean** — no review body, no inline comments, no findings on `6c09137`. So there's nothing to address, and I've made no code changes.

Since a deploy PR's risk lives in the artifacts rather than the source, I spent the round independently re-verifying them instead. All of it passed:

**Addresses are genuine, not transcribed**
- Both `contractAddress` values re-derive exactly from `keccak256(0xff ++ factory ++ salt ++ keccak256(initcode))` using the initcode committed in each `run-latest.json`, via the canonical factory `0x4e59b448...`:
  - `MulticallHandler` → `0xA07480456C4EbaD7626e4FdF4a180709E238547b`
  - `AcrossEventEmitter` → `0x446Fe2Be6a8D90fA89F30f3d0A686aAFfe621237`
- Salts in the recorded calldata (`0x12345678`, `0x1234`) match the unchanged deploy scripts; `chainId` `0x13b2` = 5042; both receipts are `status: 0x1`; nonces `0x6f`/`0x70` are sequential; recorded block numbers and tx hashes agree between the receipts and `deployed-addresses.json`.

**Deployed code is this repo's source**
- Locally compiled creation bytecode is byte-identical to the deployed initcode for both contracts — 2935 B and 274 B respectively, *including* the trailing metadata hash, so same solc version and settings.

**Generated files are reproducible, not hand-edited**
- Re-running `ExtractDeployedFoundryAddresses.ts` reproduces `deployed-addresses.json` byte-for-byte.
- `deployed-addresses.md` reproduces identically modulo table column padding (the generator emits unpadded tables; the committed file is prettier-formatted — expected).
- No stale generated artifact: `DeployedAddresses.sol` reads `broadcast/deployed-addresses.json` at runtime via `vm.readFile`, so `getDeployedAddress` picks up the new Arc entries with no regeneration needed.

**PR body claims spot-checked and accurate**
- `foundry.toml` `[etherscan]` has only `ethereum` and `ink` — the unverified-on-explorer note is correct.
- `L2_ADDRESS_MAP.5042` has only the two CCTP v2 entries and no `permit2`, so the `SpokePoolPeriphery` deferral is correctly diagnosed.
- The `explorer.arc.io` links match `PUBLIC_NETWORKS.5042.blockExplorer` in `generated/constants.json`.
- On the `MulticallHandler` address not matching mainnet/Arbitrum/Base: that divergence is already the norm, not a new wart — the contract has 14 distinct addresses across 38 chains today, and `AcrossEventEmitter` is unique on every chain it's deployed to.

**One thing I could not check:** I have no reachable Arc RPC from this environment (`rpc.arc.network` isn't egress-permitted and 5042 isn't in the read connector's network map), so I verified local artifact == the initcode in the committed receipt, but did not independently re-`eth_getCode` the two addresses on Arc. Worth one manual `cast code` against each before merge if you want that leg closed, along with the PR's CREATE2-factory/Permit2/Multicall3 liveness claims.

No further iteration needed from my side. If you do want the on-chain leg confirmed or the `SpokePoolPeriphery` follow-up picked up, assign me to the PR and I'll take it from there.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=2ce8d0b700fa99ceaefca105fb91e2ab&trace_start_ts=1789124435&trace_end_ts=1789211135)

pxrl:
Remember to bump the patch version for this.

droplet-rl:
Done — bumped `5.0.27` → `5.0.28` in `78e0fafb`.

Checked the next free number rather than assuming: `origin/master` is still on `5.0.27` and no `v5.0.28` tag exists, so `.28` is correct. Matches the convention in comparable deploy PRs (#1466 Arc, #1462 Tron, #1507 RobinHood all bump patch in the deploying PR).

A few things I confirmed while in there:

- **Version isn't duplicated anywhere** — `5.0.27` appears only in `package.json`, so this is the whole change. One file, one line.
- **No `src/consts.ts` gap.** #1507 also touched `src/consts.ts`, but that file only carries per-chain `permit2` entries, not handler/emitter addresses — these two contracts resolve through `deployed-addresses.json`, which this PR already updates. (Incidentally that same `permit2` map is what's missing for 5042, which is exactly the `SpokePoolPeriphery` blocker described in the PR body.)
- **Master moved** to `fc0348f1` (#1529, #1530) since the branch point, but neither commit touches `package.json` or `broadcast/`, so no conflict and no rebase needed.

Two notes on process, since they affected how I did this:

1. **I did not bypass the pre-commit hook.** It initially failed with `husky-run not found` — the container had been recycled and `node_modules` was gone. I reinstalled with `--frozen-lockfile` and committed normally so the hook ran for real. It passed end to end.
2. **The hook's regeneration was a clean no-op**, which is a useful signal: it re-ran `generate-constants-json` and `extract-addresses` and `git add`ed the results, and the resulting diff against HEAD was empty. So `generated/constants.json` and `broadcast/deployed-addresses.{json,md}` are exactly reproducible from the committed broadcast artifacts — the repo's own tooling independently confirming what I checked by hand last round. I committed with an explicit `-- package.json` pathspec, and the final tree is clean with the commit touching that one file only.

This also explains the MD formatting detail from my earlier review: the hook runs `prettier --write` on `deployed-addresses.md` after generation, which is where the table column padding comes from.

Still outstanding from last round and unchanged by this: I have no reachable Arc RPC here, so the on-chain `eth_getCode` leg on the two addresses is still unverified by me — worth a manual `cast code` before merge.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=cc1291eb75e5354758984bc22006da1a&trace_start_ts=1789553437&trace_end_ts=1789640137)

pxrl (approved):

droplet-rl:
Thanks for the review — it's already merged as `9391027c`, so nothing further from me. The approval carried no inline comments and an empty body, so there was no feedback to action. Recording final state for the thread:

All five files landed and `package.json` on master reads `5.0.28`. Six of seven checks passed.

**One thing I got wrong: the red DCO check was mine.** My version-bump commit `78e0fafb` had a `Co-Authored-By` trailer but no `Signed-off-by`, which is what failed the check. I should have caught it — the deploy commit I was appending to already carried `Signed-off-by: droplet-rl`, and I'd even seen a sign-off trailer while reading the version-bump history of #1507. I matched the wrong part of the convention.

No action needed now: your squash-merge collapsed both commits into one and the resulting master commit carries a valid `Signed-off-by: droplet-rl` line, so nothing propagated into master's history. I've noted the sign-off requirement for future commits here rather than treating the green squash as the check passing — it went red on the PR and only the squash saved it.

Two items that stay open, neither blocking:

- **On-chain `eth_getCode` was never verified by me.** I retried just now against `rpc.arc.network` and `rpc.arc.io`; both return empty from this environment, and 5042 isn't in my read connector's network map, so it's genuinely unreachable rather than something I skipped. Everything I asserted about the bytecode was derived from the committed receipts — the CREATE2 addresses re-derive from the recorded initcode, and that initcode matches locally compiled artifacts byte for byte including metadata. What that does *not* cover is whether the code currently live at those two addresses on Arc matches. A single `cast code` against each would close it. Same applies to the PR body's CREATE2-factory/Permit2/Multicall3 liveness claims.
- **`SpokePoolPeriphery` on Arc** still needs a `permit2` entry for 5042 in the constants package before its deploy script will target the chain. Permit2 is already live at the canonical address, so it's a constants addition, not a deployment problem.

Happy to pick up the `SpokePoolPeriphery` follow-up if you want it — just assign me.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=a65b555485cdbba9f92e83b5ff2ac307&trace_start_ts=1789554682&trace_end_ts=1789641382)

PR #1525, opened 2026-08-14
fix: Pass gateway arg to Ethereum_SpokePool in Tron periphery test

droplet-rl:
## Problem

`master` has failed to compile since #1502 merged. The only failing check is **Upload EVM artifacts** (`forge build`); `Build TypeScript` and `EVM Tests` then skip because they consume its artifacts. Every PR opened since inherits the failure — e.g. #1523.

```
Error (6160): Wrong argument count for function call: 3 arguments given but expected 4.
  --> test/evm/foundry/local/Tron_SpokePoolPeriphery.t.sol:59:35
```

## Cause

A semantic merge conflict — neither PR was wrong in isolation:

- #1481 (`feat: v5 + Spoke`, merged 2026-08-05) added a fourth `_gateway` parameter to the `Ethereum_SpokePool` constructor.
- #1502 (`improve: add Tron variants for SpokePoolPeriphery and SwapProxy`, merged 2026-08-13) added `Tron_SpokePoolPeriphery.t.sol`, which constructs `Ethereum_SpokePool` with three arguments.

#1502 branched from master at `4d67daed` (2026-07-30), six days before #1481 landed, and was never rebased — its head commit does not contain `b032205a`. Its own CI was fully green against that stale base, so nothing flagged the incompatibility until the merge landed on `master`.

Timeline: `db91f503` green → `50263367` (#1502) red → still red at `ef3feca3`.

## Fix

Pass `address(0)` for the gateway, matching the three other `new Ethereum_SpokePool(...)` call sites in the test suite (`Ethereum_SpokePool.t.sol`, `SpokePoolVerifier.t.sol`, `MultiCallerUpgradeable.t.sol`). One line; no production code touched.

## Verification

- Reproduced the exact error on clean `master` @ `ef3feca3`.
- `forge build` after the fix: compiles clean, zero errors.
- `FOUNDRY_PROFILE=local-test forge test --match-contract Tron_SpokePoolPeriphery`: **4 passed, 0 failed**.
- Prettier-clean under the repo's `.prettierrc`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1474, opened 2026-06-23
chore: Robinhood deployment

tbwebb22:

PR #1557, opened 2026-09-16
chore: deploy SpokePoolPeriphery and SpokePoolVerifier to Arc

droplet-rl:
Adds the Arc (5042) `permit2` entry to `L2_ADDRESS_MAP` and deploys the two remaining periphery contracts there.

| Contract | Address | Block |
| --- | --- | --- |
| `SpokePoolPeriphery` | `0xE791A2669bef779Ff7A4A9CF789F8Ee2CA20a32c` | 21152202 |
| `SpokePoolVerifier` | `0x3Fb9cED51E968594C87963a371Ed90c39519f65A` | 21155012 |

Permit2 is live on Arc at the canonical address, confirmed by recomputing its EIP-712 domain separator for chain 5042 and matching `DOMAIN_SEPARATOR()`. The periphery deploy script skips any chain with no `permit2` entry in `generated/constants.json`, hence the constants change.

### SpokePoolPeriphery — note on the address

This does not land on `0x97CCDBea4632140639aD5eA9b944aa034eb15fD4`, the address used on the other EVM chains. The salt (`0x1236`), CREATE2 factory and constructor args are all unchanged; the initcode differs because the compiled `SpokePoolPeriphery` has grown 741 bytes since those deployments, via #1502 (which also changed `SwapProxy`, whose initcode the constructor embeds) and #1529. Consumers resolve the address per chain via `getDeployedAddress`, as they already do for Arc's `MulticallHandler`.

#1529 also changed the signed-payload hash, so an Arc signing flow needs the post-#1529 shape rather than the one the other chains currently expect.

### SpokePoolVerifier — deployed at the canonical address

This contract only works if it sits at the same address everywhere: it makes a native-token deposit aimed at the wrong chain revert instead of paying into a codeless address. It therefore had to land on `0x3Fb9cED51E968594C87963a371Ed90c39519f65A`.

Adds `script/periphery/DeploySpokePoolVerifier.s.sol`, as the repo had no way to deploy this contract after the move to Foundry. It deploys the creationCode recorded in `deployments/mainnet/SpokePoolVerifier.json` verbatim rather than `type(SpokePoolVerifier).creationCode` — the fleet's instances were compiled with solc 0.8.19 under hardhat-deploy, and recompiling the (unchanged) source with the current toolchain yields different initcode and so a different address. The script asserts the predicted address matches the canonical one and refuses to deploy otherwise. Salt `0x1234`, as used originally.

Verification: deployed bytecode on Arc is byte-identical to both the recorded artifact and the live instance on Base, and simulating a native deposit that names Base's SpokePool reverts with `InvalidSpokePool`.

Recorded in `deployments/legacy-addresses.json` rather than via a broadcast receipt — forge cannot name a contract whose initcode matches no local artifact, so the receipt registered as `Unknown`. This matches how every other chain records this contract.

### Not included

- **`SwapProxy`** — not deployed on any chain today.

### Docs

No separate doc update proposed: the non-obvious reasoning (canonical address requirement, why the recorded bytecode is used) lives in the new script's header comment, next to the code it governs.


droplet-rl:
Codex finished with **no findings** — nothing in the review to address.

When this trigger fired, the review was still `🔄 Running` (the comment was a status placeholder with zero feedback — no review comments, no review body). Rather than burn the single allowed automated round replying to an empty comment, I waited for it to land. It completed at 11:52:29Z and reacted 👍 on the PR body, which is its documented "all reviews finished with no findings" signal.

### What I did change: DCO

Not from Codex — CI was flagging `DCO` as `action_required`, a real merge blocker. Both commits were missing `Signed-off-by`, which this repo enforces (droplet-rl signed off on #1530 and #1525).

Fixed with `git rebase HEAD~2 --signoff`. The sign-off trailer matches the commit author exactly (`droplet-rl <284132418+droplet-rl@users.noreply.github.com>`), so DCO is now ✅ **success**.

**Heads-up: this rewrote history** on the branch — `96fc266` → `0b5eec5`. Re-pull if you have it checked out. Content is byte-identical; the tree hash is `804909b6f9378ae98dcd101f0c025c8f65bad956` before and after, so only commit messages changed. I used `--force-with-lease` pinned to `96fc266`.

### Also checked while waiting

- `SpokePoolPeriphery` sorting before `SpokePool` in `deployed-addresses.json` looked like a hand-edit, but it isn't: 40 of 42 chains are unsorted there. Only the `.md` is sorted (generator sorts by `contractName`), and the `.md` entry is correctly placed. Non-issue.
- Only `run-latest.json` in the broadcast dir matches all 30 other periphery chains.
- `foundry.toml` has no `[etherscan]` Arc entry, confirming the "unverified on explorer" note.
- `Lint and Check Generated Files` passed pre-rebase, so `generated/constants.json` and the address files match a fresh generator run.

I did **not** re-verify the on-chain claims in the PR body (permit2 domain separator, bytecode-vs-artifact, `wrappedNativeToken() == address(0)`) — the evm connector has no Arc network and `NODE_URL_5042` isn't set here. Those remain as originally verified.

### Status

CI is re-running on `0b5eec5`; the tree is identical to the one that already passed, so I'd expect the same results. Still `REVIEW_REQUIRED` — needs a human approval, and I won't self-merge.

That was the one automated Codex round, so further activity from it won't re-trigger me. Since it came back clean there's nothing outstanding, but if you want another pass after any changes, assign me to the PR — that raises the automated round budget.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=74e914c90cd90f50e4fc7d39e0033810&trace_start_ts=1789559122&trace_end_ts=1789645822)

deployments/legacy-addresses.json
```diff
@@ -316,5 +316,8 @@
   "808813": {
     "SpokePool": { "address": "0x3baD7AD0728f9917d1Bf08af5782dCbD516cDd96", "blockNumber": 15393172 },
     "MulticallHandler": { "address": "0xAC537C12fE8f544D712d71ED4376a502EEa944d7", "blockNumber": 15392366 }
+  },
+  "5042": {
+    "SpokePoolVerifier": { "address": "0x3Fb9cED51E968594C87963a371Ed90c39519f65A", "blockNumber": 21155012 }
   }
```
pxrl, line 322: Why is SpokePoolVerifier here but not SpokePool, MulticallHandler, etc ?

droplet-rl: Because this contract has no other place to live — it is recorded this way on all 22 chains it is deployed to, and there is no `SpokePoolVerifier` broadcast receipt anywhere in the repo.

The asymmetry you are pointing at is real, but it splits on *how the contract is built*, not on which chain it is:

- `SpokePool`, `MulticallHandler`, `SpokePoolPeriphery` are compiled from current sources, so `forge script` produces a named receipt under `broadcast/<Script>.s.sol/5042/run-latest.json`. That is the source of truth and they are not in this file (Arc had no entry here at all before this commit).
- `SpokePoolVerifier` is deployed from the initcode recorded in `deployments/mainnet/SpokePoolVerifier.json`, not `type(SpokePoolVerifier).creationCode`. The fleet was compiled with solc 0.8.19 under hardhat-deploy; recompiling the unchanged source with the current toolchain yields different initcode and therefore a different CREATE2 address. Since the entire value of this contract is that it sits at one canonical address on every chain (so a native deposit aimed at the wrong chain reverts instead of paying into a codeless address), it has to be the old initcode. Forge cannot name a contract whose initcode matches no local artifact, so the receipt would have registered as `Unknown`.

So the entry is not a fallback — it is the only mechanism that records this contract, and the address matches the other 21 chains.

Worth noting it is still resolvable: `yarn extract-addresses` merges this file with the broadcast receipts (legacy first, receipts winning on conflict) into `broadcast/deployed-addresses.json`, which is what `DeployedAddresses.sol` reads — so `getDeployedAddress("SpokePoolVerifier", 5042)` works. The CI step that regenerates `deployed-addresses` and then fails on uncommitted changes is what proves the generated entry really derives from this file.

Where I think your question lands a hit is the filename. `legacy-addresses.json` says "old Hardhat stuff", but in practice it means "addresses with no usable Foundry receipt", and it has taken new entries as recently as #1474 (June 2026). That was undocumented, which is presumably why the entry looked misfiled. I have written it down in `deployments/README.md` in cbefb47, including the merge order and an explicit note that contracts built from current sources must not be added here.

Happy to move it if you would rather this file stay frozen as a pure Hardhat artifact — but the alternatives are a hand-written receipt with a fake tx or a second addresses file, and both seem worse than one documented exception.

droplet-rl:
Thanks — replied inline on the `legacy-addresses.json` thread ([discussion_r4026036214](https://github.com/across-protocol/contracts/pull/1557#discussion_r4026036214)).

Short version: `SpokePoolVerifier` is in that file because it's the only place the contract is ever recorded — all 22 chains it's deployed to, and there is no `SpokePoolVerifier` broadcast receipt anywhere in the repo. The split isn't Arc-specific, it's about how the contract is built. `SpokePool` / `MulticallHandler` / `SpokePoolPeriphery` compile from current sources and get named Foundry receipts, so they're recorded there instead (Arc had no entry in the legacy file at all before this). The verifier has to be deployed from the recorded 0.8.19 initcode to keep its canonical cross-chain CREATE2 address, and forge can't name a contract whose initcode matches no local artifact, so its receipt would have recorded as `Unknown`.

It does still resolve through `getDeployedAddress()` — `yarn extract-addresses` merges the legacy file with the receipts (legacy first, receipts winning on conflict) into `broadcast/deployed-addresses.json`, which is what `DeployedAddresses.sol` reads.

### Doc fix

The part of your question I think is a genuine hit is the filename. `legacy-addresses.json` reads as "old Hardhat stuff", but in practice it means "addresses with no usable Foundry receipt" — and it's taken new entries as recently as #1474 (June 2026). That was nowhere in writing, which is probably why the entry looked misfiled.

`cbefb47` documents it in `deployments/README.md`: what the file is actually for, the merge order, and an explicit note that contracts built from current sources must not be added here. Doc-only, no behavior change. DCO is green.

If you'd rather keep that file frozen as a pure Hardhat artifact, say so and I'll move the entry — but the alternatives I can see are a hand-written receipt with a fabricated tx hash, or a third addresses file, and both seem worse than one documented exception.

### Note on the branch

Heads-up for your own context: the branch was rebased and extended between my last pass and this review — it's now `SpokePoolPeriphery` **and** `SpokePoolVerifier`, rebased onto master post-#1554, hence the retitle. I synced to `26511456` before committing rather than pushing on top of my stale checkout, so nothing was clobbered. Sign-offs from the earlier DCO fix survived the rebase.

Still `REVIEW_REQUIRED` — needs your approval, I won't self-merge.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=4cfbf1513c4e69cc84f2f7e88fccdc6e&trace_start_ts=1789561487&trace_end_ts=1789648187)

pxrl (approved):

droplet-rl:
Thanks for the approval — nothing outstanding on my side.

Taking the approval-without-follow-up on the [`legacy-addresses.json` thread](https://github.com/across-protocol/contracts/pull/1557#discussion_r4026036214) as you being satisfied with the explanation and declining the offer to relocate the entry, so `SpokePoolVerifier` stays where it is, now documented in `deployments/README.md` (`cbefb47`).

State: all 7 checks green (incl. DCO), `reviewDecision: APPROVED`, head `cbefb47`.

Ready to merge whenever you want to press the button — I won't self-merge.

droplet-rl:
🔎 [View trace](https://ui.honeycomb.io/risklabs/environments/droplet-prod/datasets/droplet-runner/trace?trace_id=6946c4beb77c94ce8b99eeab615242bb&trace_start_ts=1789564104&trace_end_ts=1789650804)

PR #1559, opened 2026-09-17
feat: migrate Solana_Adapter to CCTP V2 (EVM side)

grasphoper:
EVM-only slice of the Solana CCTP V2 migration, split out of #1548 so the EVM and SVM changes can be audited separately. The Solana program, TS client and operational script changes stay in #1548, which is stacked on this branch.

- Solana_Adapter sends admin messages through IMessageTransmitterV2 with destinationCaller = bytes32(0) and a configurable minimum finality threshold (finalized, 2000). The Solana spoke pool only implements the finalized message handler.
- Add IMessageTransmitterV2.sendMessage to CCTPInterfaces and a MockCCTPMessageTransmitterV2 mock; existing V2 transmitter mocks gain a no-op sendMessage.
- Solana_Adapter Foundry tests migrated to the V2 mocks, with constructor and invalid-target coverage.
- Add DeploySolanaAdapter Foundry script plus cctpV2MessageTransmitter L1 constants (src/consts.ts -> generated/constants.json -> Constants.sol) and a CCTP_FINALITY_THRESHOLD_FINALIZED constant.

PR #1555, opened 2026-09-11
chore: update Solana-related TS code to cctp v2

grasphoper:
Typescript part of ACP-226

PR #1561, opened 2026-09-17
feat: CCTP v2 for Solana Spoke

grasphoper:


Reinis-FRP (approved):
Approved. Verified that #1561 and #1559 together preserve the exact changes from #1548, including #1555, with the SVM changes applied over the V5 epic base.

Only caveat: please add a PR body describing the SVM/script scope, link #1559 as the EVM counterpart, and note that these replace #1548 and require coordinated deployment despite being audited separately.

---

